PR #8, opened 2017-05-09
Added docstrings and neaten

willingc:
@yuvipanda As I'm trying to understand the source, I've added some docstrings or edited existing ones based on my understanding. Please correct since bound to have errors.


yuvipanda:
Thank you :) I'll clean them up as I change them :)

PR #14, opened 2017-05-16
[WIP] first flow diagram + docs

choldgraf:
Dummy version of the flow diagram is here:

Will iterate on this...LMK your original thoughts


yuvipanda:
there's a cycle of sorts that's missing here. 

Build waits for s2i (or dockerfile - they should be at same level) builder to complete, then the build pushes the image to gcr.io (or 'docker registry'). *then* it returns this status to builderhub, which redirects you to a  JupyterHub. 

Not sure what the 'jhub' box around gcr.io is supposed to mean?

yuvipanda:
also are you tied to the gradients? :)

choldgraf:
> also are you tied to the gradients? :)

No not at all - this is just copy/pasted boilerplate stuff from the MNE repo. I just haven't had time to change the gradients yet :)

Will incorporate your comments into another version!

choldgraf:
@yuvipanda see latest version...feel free to edit your comment above w/ updated comments :)

choldgraf:
hey @yuvipanda, latest version causes the diagram to auto-build when you do `make html`. I also added a little sphinx website we can point people to. One day I think it should be similar to `Zero to JupyterHub`.

I'm gonna merge this but lemme know if you want me to make other modifications and I'm happy to do so. I also pushed a version of the docs to the upstream `gh-pages` branch, so we should be able to point people to it :)

PR #31, opened 2017-05-24
Add a stable URL for launching binders!

yuvipanda:
- Change the provider interface to do more things separately.
  This lets us re-use it in two places

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -142,9 +142,9 @@ class BinderHub(Application):
         }
 
         self.tornado_app = tornado.web.Application([
-            (r"/build/([a-z0-9]+)/([^?]+)", BuildHandler),
+            (r"/build/([^/]+)/(.+)", BuildHandler),
             (r"/redirect", RedirectHandler),
-            (r"/v2/([a-z0-9]+)/([^?]+)", ParameterizedMainHandler),
+            (r"/v2/([^/]+)/(.+)", ParameterizedMainHandler),
             (r'/', MainHandler)
         ], **self.tornado_settings)
 
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -12,7 +12,7 @@ from traitlets.config import Application
 
 from .builder import BuildHandler
 from .redirect import RedirectHandler
-from .main import MainHandler
+from .main import MainHandler, ParameterizedMainHandler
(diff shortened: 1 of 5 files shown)


binderhub/app.py
```diff
@@ -144,7 +144,8 @@ def initialize(self, *args, **kwargs):
         self.tornado_app = tornado.web.Application([
             (r"/build/([a-z0-9]+)/([^?]+)", BuildHandler),
             (r"/redirect", RedirectHandler),
-            (r"/", MainHandler)
+            (r"/v2/([a-z0-9]+)/([^?]+)", ParameterizedMainHandler),
```
minrk: We can greedily match any url part, so that ParameterizedMain gets called and can handle a more informative "no such provider" error. Plus, this pattern only applies to path, so there's no need to split on `?`, e.g. `v2/([^/])/(.*)`

binderhub/main.py
```diff
@@ -3,15 +3,20 @@
...
-        filepath = self.get_argument('filepath', '', True).replace('|', '#')
-        submit = self.get_argument('submit', False, True)
+        self.render("index.html", url=None, ref='master', filepath=None, submit=False)
+
+
+class ParameterizedMainHandler(web.RequestHandler):
+    def get(self, provider_prefix, spec):
+        providers = self.settings['repo_providers']
+        if provider_prefix not in self.settings['repo_providers']:
+            raise Exception('wat')
```
minrk: 404: No such provider...

minrk:
```python
raise web.HTTPError(404, "No such provider: %s" % provider_prefix)
```

binderhub/repoproviders.py
```diff
@@ -53,24 +57,29 @@ class GitHubRepoProvider(RepoProvider):
...
-    def resolve_spec(self, spec):
-        """
-        Parses a URL spec and returns a git URL + ref to check out
-        """
-        # We want spec to be user/repo/branch
-        spec_parts = spec.split('/')
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        spec_parts = self.spec.split('/')
         if len(spec_parts) != 3:
```
choldgraf: `"provided spec {}".format(spec)`?

PR #34, opened 2017-05-26
adding setup information

choldgraf:
@yuvipanda this is a first pass at some more in-depth documentation. There's some incorrect information in there because I just eyeballed the text w/o internet while I was on the plane, but let me know what changes you'd like to see and I'll throw 'em in there!

Rendered docs are here: http://predictablynoisy.com/binderhub/


choldgraf:
(I also added in a logo in the top left so that you can get back to the home page wherever you are)

willingc:
@choldgraf Overall, looks good. Do we have permission to use the binder logo? I am assuming we do. Do we need to provide attribution?

We should go ahead and host on RTD as well if that has not yet been done.

Happy to merge once the logo bit is clarified.

choldgraf:
that's a v. good point @willingc ! Let me follow up with Jeremy to see if he's OK with this. In the meantime, perhaps we can just use the "jupyterhub" logo, and replace the words with "binderhub"? WDYT?

doc/setup.rst
```diff
@@ -0,0 +1,51 @@
...
+up on their platforms as well...we should do this in z2jh and link here.>>>.
+
+Set up a Kubernetes cluster
+---------------------------
+This is also platform-dependent, though it is relatively straightforward
+to set up kubernetes on most cloud computing providers. For instructions
+on how to do this, see the `Zero to JupyterHub <https://zero-to-jupyterhub.readthedocs.io/en/latest/create-k8s-cluster.html>`_
+tutorial on setting up Kubernetes.
+
+Set up and configure JupyterHub
```
yuvipanda: I helped @minrk set these up at PyCon - we can either take these sections out until he (or me) have time to write it out or wait till we update it before merging. I prefer the former since there is a lot of good stuff in here without these two thing.s

choldgraf: could you be specific about which sections you're talking about?

yuvipanda: This and the section below it :)

choldgraf: gotcha - yep that sounds good to me. I can just strike them from this PR in the next push

willingc:
Let's do this for now...

1. Remove the sections suggested by @yuvipanda 
2. Remove the logo temporarily
3. Merge 

We will likely get approval to use the Binder logo and can add back in a quick PR at that point. Let's get the rest of the content out for others :smile:

choldgraf:
Well the main reason I wanted the logo is just so that it's easy for people to get back to the home page once they're on a child page. I couldn't find a super easy way to do that w/o some kinda logo to click on :) is there a way to include just a text logo w/ sphinx? I'll look into that before the next push.

willingc:
@choldgraf I think that depends on the theme and whether hosted on RTD or on a static site. On RTD, most themes allow clicking on the doc title to return to the doc home page. There is a jupyter logo with no wording in the jupyter design repo. For the moment I would just use that as is without text.

choldgraf:
Cool, will go with that

choldgraf:
ok changes made + sections removed!

willingc:
Thanks @choldgraf!

PR #43, opened 2017-05-31
Fix some UI text typos and edit wording

### LINKED ISSUE #38: UI - Add indicator/message that file path is optional
`File to Launch` is optional.

screen shot 2017-05-30 at 12 02 23 pm

willingc:
Closes #38

### DIFF
--- a/binderhub/index.html
+++ b/binderhub/index.html
@@ -11,7 +11,7 @@
         <link href="/static/vendor/xterm.js-2.6.0/xterm.css" rel="stylesheet" />
         <link href="/static/index.css" rel="stylesheet" />
         <!-- FIXME: vendor this -->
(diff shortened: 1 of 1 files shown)


binderhub/index.html
```diff
@@ -29,18 +29,19 @@
                     </div>
 
                     <form id="build-form" class="form jumbotron">
-                        <h4>Build & Launch a Repository</h4>
+                        <h4>Build and Launch a Binder</h4>
```
yuvipanda: Ideally I'd like to keep this to just Repository for now, rather than 'Binder'. Both to keep it consistent with current website (which Jeremy Freeman asked us) and to avoid introducing new terminology.

willingc: Sure no problem

PR #47, opened 2017-05-31
rtd in readme

choldgraf:


willingc:
@choldgraf Do you mind if I close this in favor of #48 which has both the doctr and RTD? It might be nice to keep both around for the moment to see if images work the same on both with graphviz.

PR #55, opened 2017-06-02
[WIP] Refactor JavaScript to suck less!

yuvipanda:
- Create an old school JS 'class' that can build and launch
  an image
- Fix some of the UX issues that came up:
  - Show log when build fails
  - Don't show waiting when we know the build already exists
  - Clear out the progress bar if we start a new build
- Write comment about the state transitions that are valid. Should
  enforce it at some point!A
- Separate build and runtime parameters slightly better than before!

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -79,16 +79,14 @@ class BinderHub(Application):
         config=True
     )
 
-    hub_redirect_url_template = Unicode(
+    hub_login_url = Unicode(
         None,
         allow_none=True,
         help="""
-        Template used to generate the URL to redirect user to after building.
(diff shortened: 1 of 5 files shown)


yuvipanda:
I need to add support back for file path detection - maybe move that logic to the server side actually (on JupyterHub), rather than the client.

yuvipanda:
I'm going to do the following:

1. In the tmpauthenticator derivative, I'll just stash all of the options from query params into user.spawner.user_options
2. In pre_start, change the spawner options based on the items from user_options

This should be more generalizable than the current solution

yuvipanda:
Merged!

PR #80, opened 2017-08-07
Add docstrings and autodoc traitlets

willingc:
This PR adds some docstrings and uses autodoc to document modules, classes, and configurable traits. Fixed a few Python style items. [Rendered docs](http://test-binderhub.readthedocs.io/en/doc-comment/)


README.rst
```diff
@@ -1,7 +1,7 @@
 `BinderHub`_
 ============
 
-.. image:: https://readthedocs.org/projects/binderhub/badge/?version=latest
+.. image:: http://readthedocs.org/projects/binderhub/badge/?version=latest
```
minrk: Shouldn't this stay https?

willingc: It can. This is a leftover of trying to work around an error in the latest Sphinx that warns for non-local images. I just cut and pasted the RTD sample code. Going to change it back to https. Good catch.

PR #81, opened 2017-08-09
[WIP] throw error if URL isn't right

choldgraf:
This is an attempt at raising an error if the build parameters somebody gives isn't correct. I suspect that I'm missing something here, as I couldn't get this error to pop up on my local build. Then again I'm pretty sure that my local build is broken in a more fundamental way than this error because I can't get minikube to work.

Let me know if this is generally the right direction though.


yuvipanda:
Superseeded by #91

Issue #86, opened 2017-08-15
Documentation improvements


choldgraf:
- [ ] the [DNS page](https://binderhub.readthedocs.io/en/latest/setup-dns.html) could be removed and instructions placed in [setup binderhub](https://binderhub.readthedocs.io/en/latest/setup-binderhub.html)
- [x] the page on [tearing down a binder deployment](https://binderhub.readthedocs.io/en/latest/turn-off.html) should be filled out.
- [x] the [contributor guide](https://binderhub.readthedocs.io/en/latest/contributor-guide.html#binderhub) isn't really a contributor guide...it's just the readme. We should either flesh this out, or remove the content and link to the appropriate section in the readme IMO
- [x] the [API page](https://binderhub.readthedocs.io/en/latest/api/api-index.html) doesn't seem to be rendering properly.

choldgraf:
a few to-do items @yuvipanda @willingc ... any thoughts / new items welcome!

willingc:
Resolved API page build. One setting in RTD config for setup.py install wasn't checked.

choldgraf:
@yuvipanda on this page: https://binderhub.readthedocs.io/en/latest/setup-binderhub.html

we have the wording

"Now find the IP of the nginx-ingress-controller:

`kubectl --namespace=support get svc support-nginx-ingress-controller`

And make DNS records for both binder and jupyterhub that point to that IP."

What do we mean by "make DNS records for both Binder and JupyterHub point to that IP"?

PR #89, opened 2017-08-18
Add some detail to tear down doc

willingc:
Addresses #86 

Refers to Zero to K documentation for shrinking a cluster and deleting a cluster.

### DIFF
--- a/doc/turn-off.rst
+++ b/doc/turn-off.rst
@@ -11,16 +11,21 @@ Contracting the size of your cluster
 
 If you would like to shrink the size of your cluster, refer to the
 `Expanding and contracting the size of your cluster <https://zero-to-jupyterhub.readthedocs.io/en/latest/extending-jupyterhub.html#expanding-and-contracting-the-size-of-your-cluster>`_
-section of the Zero to Jupyter documentation. Resizing the cluster to zero nodes
-could be used if you wish to temporarily reduce the cluster (and save costs)
-without deleting the cluster.
+section of the `Zero to JupyterHub`_ documentation. Resizing the cluster to
+zero nodes could be used if you wish to temporarily reduce the cluster (and
+save costs) without deleting the cluster.
 
 Deleting the cluster
 --------------------
 
 To delete a Binder cluster, follow the instructions in the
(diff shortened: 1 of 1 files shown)


doc/turn-off.rst
```diff
@@ -1,9 +1,26 @@
...
 to remember to delete each of these containers or else they will continue
 to exist (and cost money!).
 
-<<<TODO>>>
+Contracting the size of your cluster
+------------------------------------
+
+If you would like to shrink the size of your cluster, refer to the
+`Expanding and contracting the size of your cluster <https://zero-to-jupyterhub.readthedocs.io/en/latest/extending-jupyterhub.html#expanding-and-contracting-the-size-of-your-cluster>`_
+section of the Zero to Jupyter documentation. Resizing the cluster to zero nodes
```
choldgraf: This sounds reasonable to me, but I'd like to hear @yuvipanda 's opinion on this too...we're not unintentionally asking people do to something they'd regret later by doing this, are we?

doc/turn-off.rst
```diff
@@ -1,9 +1,26 @@
...
+could be used if you wish to temporarily reduce the cluster (and save costs)
+without deleting the cluster.
+
+Deleting the cluster
+--------------------
+
+To delete a Binder cluster, follow the instructions in the
+`Turning Off JupyterHub and Computational Resources <https://zero-to-jupyterhub.readthedocs.io/en/latest/turn-off.html>`_
+section of the Zero to Jupyter documentation.
```
choldgraf: ~Jupyter~ JupyterHub. Also we might as well hyperlink those words too :-)

doc/turn-off.rst
```diff
@@ -1,9 +1,26 @@
...
+
+Double check your cloud provider account to make sure all resources have been
```
choldgraf: maybe make this a note, and add a sentence that semi-frequently resources don't get deleted for whatever reason so this is not a small-probability kind of statement :-)

choldgraf:
looks good! a couple comments then I'm +1

doc/turn-off.rst
minrk: Yes, resizing to zero is safe. It doesn't delete the master node, which is managed separately by the cloud provider, and doesn't delete volumes or deployments.

choldgraf:
looks great, thanks! :-)

PR #96, opened 2017-08-30
trim trailing and leading / on repo

minrk:
if users input:

    https://github.com/ipython/ipython/

the trailing '/' would cause errors


choldgraf:
LGTM

PR #98, opened 2017-09-06
update setup.py

minrk:
- author is Project Jupyter Contributors
- point to jupyter mailing list
- require Python 3.5 (allows async/await!)


choldgraf:
LGTM

PR #100, opened 2017-09-06
use threadpool and tornado queue for messaging

minrk:
- threadpool limits concurrent builds and thus DOS by spawning runaway threads
- tornado Queue removes sleep when waiting for events

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -1,11 +1,9 @@
 """
 Contains build of a docker image from a git repository.
 """
-import json
-import threading
 
-from kubernetes import client, config, watch
-
+from kubernetes import client, watch
+from tornado.ioloop import IOLoop
 
 class Build:
     """Represents a build of a git repository into a docker image.
@@ -57,7 +55,7 @@ class Build:
 
     def progress(self, kind, obj):
         """Put the current action item into the queue for execution."""
-        self.q.put_nowait({'kind': kind, 'payload': obj})
+        IOLoop.instance().add_callback(self.q.put, {'kind': kind, 'payload': obj})
 
     def submit(self):
         """Submit a image spec to openshift's s2i and wait for completion """
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -4,12 +4,12 @@ Handlers for working with version control services (i.e. GitHub) for builds.
 
 import hashlib
 import json
-from queue import Queue, Empty
 import threading
 
 import docker
 from kubernetes import client, config
 from tornado import web, gen
+from tornado.queues import Queue
 from tornado.iostream import StreamClosedError
 from tornado.log import app_log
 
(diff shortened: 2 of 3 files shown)


choldgraf:
Could you give a short description of what this adds? Or if @yuvipanda thinks the code looks good I'm cool w/ it too...I probably don't have the skillset to properly review this :-/

binderhub/build.py
```diff
@@ -57,7 +55,7 @@ def get_cmd(self):
 
     def progress(self, kind, obj):
         """Put the current action item into the queue for execution."""
-        self.q.put_nowait({'kind': kind, 'payload': obj})
+        IOLoop.instance().add_callback(self.q.put, {'kind': kind, 'payload': obj})
```
yuvipanda: I'm trying to wrap my head around this. This looks like we're wanting to call on the main thread a call to put. However, put can block, and put_nowait is just put with block=False set. It's also safe to put things into queue from multiple threads. Why this change?

binderhub/builder.py
```diff
@@ -152,55 +152,57 @@ def get(self, provider_prefix, spec):
...
-
-        while True:
-            # FIXME: use Futures
-            try:
-                progress = q.get_nowait()
-            except Empty:
-                yield gen.sleep(0.5)
-                continue
+        while not done:
+            progress = yield q.get()
```
yuvipanda: Does this work? I don't see anything about generator / async support in https://docs.python.org/3/library/queue.html

binderhub/build.py
minrk: This is not a stdlib `queue.Queue`, it's a tornado `queues.Queue`.

`progress` is called from a thread and `q.put` must be called from the main thread. The only threadsafe way to call these things is to place them on the main thread via `.add_callback()`.

binderhub/builder.py
minrk: This is not a stdlib `queue.Queue`, it's a tornado `queues.Queue`, which is like a Queue except it's async (hence `yield`).

minrk:
> Could you give a short description of what this adds?

This uses a ThreadPool to limit concurrent builds and remove the spawning of new threads for every build. As a result, there is now a limit on the number of concurrent builds. It also switches to using the tornado async Queue instead of the stdlib Queue for communication. tornado Queues are async, which removes the need for a `gen.sleep` because we can just wait for messages on the Queue with `yield q.get()`.

choldgraf:
Sounds good - if we hit the limit of concurrent builds, what happens? Does it just hang until a spot is opened up? Or error?

minrk:
It'll wait

binderhub/build.py
yuvipanda: Aaah, I totally missed that change! This all makes sense now.

PR #110, opened 2017-09-15
launch via Hub API

### LINKED ISSUE #103: Launch jupyter-notebook instead of jupyterhub-singleuser
One of the tensions with adopting jupyterhub for binder is that the authentication mechanism used for jupyterhub requires jupyterhub to be installed in the container and to be a compatible version. This is a bit frustrating because the authentication is mostly irrelevant to binder.

jupyterhub-singleuser does two things right now, on top of `jupyter notebook`:

- authenticate with the Hub
- provide links back to the hub

The links back to the hub are actively unhelpful in Binder because they won't work. Authenticating with the Hub is also unnecessary for Binder containers.

If, instead, we launched `jupyter notebook`, we would get rid of all of the problems dealing with matching hub and container versions. This matches the behavior of old binder and makes it much more resilient to version changes (jupyterhub doesn't need to be installed at all).

We could even keep authentication if we require `notebook>=4.3` and launch with `--NotebookApp.token=...`, letting the notebook server authenticate with a token instead of deferring to jupyterhub.

minrk:
- use null authenticator to disable all login to the Hub itself
- create users via POST hub/api/users/:user
- spawn servers via POST hub/api/users/:user/server with image and token in user_options body
- spawns jupyter-notebook with token access (requires notebook 4.3), not jupyterhub-singleuser (closes #103)

<del>This *should* work with jupyterhub 0.7, but I've been testing with 0.8 via https://github.com/jupyterhub/helm-chart/pull/79</del> This now requires v0.5.x jupyterhub chart for the service tokens, but it isn't technically sensitive to the Hub version.

Questions:

- [x] where should I put the BinderSpawner bit? Publish a package or leave it in the extraConfig like this? **extraConfig for now**
- [x] Binder itself needs an API token to pass to the Hub. **passed via binder-secret to binder and hub-secret to hub**
- [x] It would be much nicer if I could write a regular `jupyterhub_config.py` Python config file, rather than writing Python in yaml. Is there a way to 'include' a local file with helm? **helm doesn't support it yet**

cc @willingc

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -117,13 +117,13 @@ class BinderHub(Application):
         help="""
         The base URL of the JupyterHub instance where users will run.
 
-        Must end in '/'
         e.g. https://hub.mybinder.org/
         """,
         config=True,
     )
     @validate('hub_url')
     def _add_slash(self, proposal):
+        """trait validator to ensure hub_url ends with a trailing slash"""
         if proposal.value is not None and not proposal.value.endswith('/'):
             return proposal.value + '/'
         return proposal.value
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -9,7 +9,7 @@ import tornado.ioloop
 import tornado.options
 import tornado.log
 import tornado.web
-from traitlets import Unicode, Integer, Bool, Dict
+from traitlets import Unicode, Integer, Bool, Dict, validate
 from traitlets.config import Application
 
 from .base import Custom404
@@ -109,17 +109,24 @@ class BinderHub(Application):
         config=True
     )
 
-    hub_login_url = Unicode(
-        None,
-        allow_none=True,
-        help="""
-        The hub login URL to redirect the user to after image is built.
-
-        It'll get all runtime parameters as query parameters, and it is the responsibility
-        of the hub to launch this properly
-        """,
-        config=True
+    hub_api_token = Unicode(
+        help="""API token for talking to the JupyterHub API""",
+        config=True,
     )
+    hub_url = Unicode(
+        help="""
+        The base URL of the JupyterHub instance where users will run.
+
+        Must end in '/'
+        e.g. https://hub.mybinder.org/
+        """,
+        config=True,
+    )
+    @validate('hub_url')
+    def _add_slash(self, proposal):
+        if proposal.value is not None and not proposal.value.endswith('/'):
+            return proposal.value + '/'
+        return proposal.value
 
     build_namespace = Unicode(
         'default',
@@ -169,7 +176,8 @@ class BinderHub(Application):
             "static_path": os.path.join(os.path.dirname(__file__), "static"),
             "github_auth_token": self.github_auth_token,
             "debug": self.debug,
-            'hub_login_url': self.hub_login_url,
+            'hub_url': self.hub_url,
+            'hub_api_token': self.hub_api_token,
             "build_namespace": self.build_namespace,
             "builder_image_spec": self.builder_image_spec,
             'repo_providers': self.repo_providers,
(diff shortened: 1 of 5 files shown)


binderhub/app.py
```diff
@@ -109,17 +109,24 @@ class BinderHub(Application):
...
-        It'll get all runtime parameters as query parameters, and it is the responsibility
-        of the hub to launch this properly
+        Must end in '/'
+        e.g. https://hub.mybinder.org/
         """,
-        config=True
+        config=True,
     )
+    @validate('hub_url')
+    def _add_slash(self, proposal):
```
willingc: @minrk `proposal` is a bit vague. A docstring would help clarify.

binderhub/redirect.py
```diff
@@ -1,16 +1,91 @@
...
+from urllib.parse import quote
+
+from tornado.log import app_log
+from tornado import web
+from tornado.httpclient import AsyncHTTPClient, HTTPRequest, HTTPError
+from tornado.httputil import url_concat
 
 from .base import BaseHandler
 
+# Add a random lowercase alphanumeric suffix to usernames to avoid collisions
```
willingc: Perhaps:
```
# Add a random lowercase alphanumeric suffix to usernames to avoid collisions
# Set characters which may be used to generate a suffix
SUFFIX_CHARS...
# Set length of suffix. Number of combinations = SUFFIX_CHARS**SUFFIX_LENGTH = 36**8 ~= 2**41
SUFFIX_LENGTH...
```

minikube-config.yaml
```diff
@@ -3,46 +3,54 @@
...
+            raise web.HTTPError(400, "image required")
+
+          self.singleuser_image_spec = self.user_options['image']
+          return super().start()
+
+    # launch jupyter notebook instead of jupyterhub-singleuser
+    c.JupyterHub.spawner_class = BinderSpawner
+
+    # disable all XSS protections:
+    # FIXME: Make this more secure before allowing any authenticated state into binder
```
willingc: Let's make sure that this gets fixed :wink:

binderhub/app.py
```diff
@@ -109,17 +109,24 @@ class BinderHub(Application):
...
+        config=True,
+    )
+    hub_url = Unicode(
         help="""
-        The hub login URL to redirect the user to after image is built.
+        The base URL of the JupyterHub instance where users will run.
 
-        It'll get all runtime parameters as query parameters, and it is the responsibility
-        of the hub to launch this properly
+        Must end in '/'
```
choldgraf: Why not just do a quick `if not hub_url.endswith('/'): hub_url += '/'`?

yuvipanda:
This is absolutely awesome, since it takes away the whole jupyterhub version problem!

My general personal rule is to leave it in extraConfig until it reaches a critical size (for me it's about 50 lines) and then it moves to a package. The package can just live in a separate dir in the binderhub repo maybe? I want to move the binderhub chart to the binderhub repo too - the only reason it is here is because we weren't publishing a chart after each commit, but now we are!

Unfortunately we can not include a file yet. I should write up a feature request for helm.

Generating token should just be manually supplied as part of binderhub chart's values.yaml I think.

minrk:
What should we do about updating helm? Should I do the updates in the 0.5.x branch of helm-chart? Or is now the time to move the binder helm chart over here?

binderhub/app.py
minrk: That's what the `@validate` below does. Updated to clarify.

binderhub/app.py
minrk: docstring added. This is a traitlets validator, which always have this signature.

yuvipanda:
@minrk I think now is the time to move the binderhub helm chart over here!

minrk:
@yuvipanda Sounds good! I'll do that in a separate PR form this one, I think.

This one seems to be good to go.

Issue #112, opened 2017-09-18
Improve setup instructions to avoid confusion


choldgraf:
I'm going to dump a few things that need improving here, per @henchc 's attempt at setting up binderhub...feel free to append chris!

https://binderhub.readthedocs.io/en/latest/setup-binderhub.html

- Make it more clear that you need to have 2 different DNS addresses point to the IP we're setting up

henchc:
"Now find the IP of the nginx-ingress-controller:" should just add "EXTERNAL-IP" for clarification

henchc:
"<the json file from service account>"

this is more difficult than copying and pasting, you have to make sure the JSON can be parsed to yaml, I just made it all one line

choldgraf:
We should also make it really clear that you *need* to have a DNS set up, you can't use IP addresses

minrk:
Closed by #113

Issue #111, opened 2017-09-18
config.yaml valid yaml


henchc:
```
registry:
  # Note this is project *ID*, not just project
  #
  # `<prefix>` can be any string and will be appended to image names
  # e.g., `dev` and `prod`.
  prefix:  gcr.io/<google-project-id>/<prefix>

hub:
  # This is the DNS that you've configured. E.g. `beta.mybinder.org`
  # Note that there is `http://` here
  url: http://<dns-entry-for-jupyterhub>

ingress:
  enabled: true
    # But no `http://` here
    host: <dns-entry-for-binder>

jupyterhub:
  ingress:
    enabled: true
    # But no `http://` here
    host: <dns-entry-for-jupyterhub>
```

the comment and host of the ingress block need to be moved back to be valid yaml

Chris edit:
The problem here is that the yaml we told people to put into their config files was invalid - there was an extra tab where there shouldn't be. This should be fixed once #113 is merged

minrk:
Closed by #113

PR #113, opened 2017-09-19
improve instructions

choldgraf:
@henchc and I took a shot at setting up a binderhub and some things were confusing...this PR addresses a few points of confusion we had. Should still add some more to it...I'll remove the WIP when it's ready.

Addresses #112 and #111

### DIFF
--- a/doc/setup-binderhub.rst
+++ b/doc/setup-binderhub.rst
@@ -1,40 +1,46 @@
 Set up BinderHub
 ================
 
+.. note::
+
+   If you wish to use encryption with ``letsencrypt``, BinderHub requires
+   that you configure the DNS of a web address to point to the BinderHub IP
+   address. This means that you need to have control over a URL such as
+   ``www.mydomain.com``.
+
 Download the Helm Chart and deployment repo for BinderHub
 ---------------------------------------------------------
 
-::
+BinderHub uses Helm Charts to set up the applications we'll use in our Binder
+deployment. If you're curious about what Helm Charts are and how they're
+used here, see the `Zero to JupyterHub guide
+<https://zero-to-jupyterhub.readthedocs.io/en/latest/tools.html#helm>`_.
 
-    git clone https://github.com/jupyterhub/helm-chart
-    git clone https://github.com/jupyterhub/binderhub-deploy
+First grab the latest helm chart for BinderHub and install it.::
 
-
-Install the "support" applications
-----------------------------------
-
-::
-
-    cd binderhub-deploy/support
-    helm dep up
-    cd ..
-    cd ..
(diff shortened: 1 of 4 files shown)


choldgraf:
some more edits added in there. when @minrk is done making changes to the binderhub app, let's take another pass at the instructions to make sure they're still correct. In the meantime we can either merge this and open a new PR, or wrap that work up into this PR

doc/setup-binderhub.rst
```diff
@@ -1,40 +1,56 @@
 Set up BinderHub
 ================
 
+.. note::
+
+   Currently BinderHub requires that you configure the DNS of a web
+   address to point to the BinderHub IP address. This means that you need to
+   have control over a URL such as ``www.mydomain.com``.
```
minrk: Specifically, it's the use of `binderhub-deploy/support` to setup letsencrypt that has this requirement. If these docs update to only cover using the binderhub helm-chart directly (resolving #114), then this requirement goes away as well.

But presumably we also want to cover how to enable LEGO as well, at least at some point.

choldgraf:
@minrk now that those big binderhub PRs are merged, should I change anything in this PR to update instructions?

minrk:
Yes, I think so.

There user-facing changes, I think:

- helm-chart and binderhub-deploy repos don't need to be cloned. They can be fetched with helm:

        helm repo add jupyterhub https://jupyterhub.github.io/helm-chart

- installing the binderhub chart should now be similar to jupyterhub itself in zero-to-jupyterhub:

        helm install --name binder jupyterhub/binderhub --version v0.1.0-464c4f5

- in addition to the existing `proxy.secretToken` and `hub.cookieSecret`, there is one more required secret: `hub.services.binder.apiToken`. It can be generated in the same way as the others.

This could be a good opportunity to scrub the `support` instructions, which confuse things a bit, I think, but that can be another PR.

doc/setup-binderhub.rst
choldgraf: now that binderhub can be installed with helm, can we edit this note to say this is only necessary w/ letsencrypt?

choldgraf:
OK I'm gonna update this PR so we can merge now that we've a buncha new features in binderhub. I'll try adding in the comments you gave above Min, and will ping here if something doesn't work. Maybe this is something @yuvipanda and I can do together.

Anything else I should change per the new additions?

doc/setup-binderhub.rst
minrk: yes

minrk:
This is a big improvement, so I'll merge now despite the WIP. There's always future PRs!

PR #116, opened 2017-09-20
import binderhub helm-chart

minrk:
binderhub helm-chart is now included here

initial import was done with

    git subtree add helm-chart github.com/jupyterhub/helm-chart v0.5.x --squash

so history is not preserved except by reference.

The main question when doing this is how does the helm chart version related to the binderhub version. Are they the same? Are they versioned separately? If separately, how do we handle that wrt git tags, etc.

- [x] update jupyterhub dependency url when https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/195 is merged
- [x] Where should this chart be published? To the same `jupyterhub/helm-chart` gh-pages repo that jupyterhub is published to?

includes #110

### DIFF
--- a/helm-chart/build.py
+++ b/helm-chart/build.py
@@ -4,6 +4,8 @@ import subprocess
 import argparse
 import yaml
 
+NAME = 'binderhub'
+
 def last_git_modified(path):
     return subprocess.check_output([
         'git',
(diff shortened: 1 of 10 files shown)


choldgraf:
I'm gonna defer to yuvi to confirm this is a good spot for the chart

re: publish...do you mean for documentation? If so, I'd imagine it'd be in the binderhub RTD, no?

minrk:
Ah, no, I mean uploading the chart itself to serve as a helm repository. We are publishing the jupyterhub helm-chart to [helm-chart@gh-pages](https://github.com/jupyterhub/helm-chart/tree/gh-pages) so that people can add `https://jupyterhub.github.io/helm-chart` as a helm repository. I don't think there's a strong reason we need to use a different location. We don't need to do this yet for binder since the chart's less mature, but we should think about it at least. It's not a blocker for this PR.

yuvipanda:
@minrk yeah, I think my current preference is to continue using the helm-chart repo as our published source of helm charts, jupyterhub and otherwise. GitHub doesn't do redirects in github pages, and we already have people using that URL. So it sounds like a good place to keep these.

helm-chart/build.py
```diff
@@ -0,0 +1,117 @@
...
+        env=dict(os.environ, GIT_SSH_COMMAND='ssh -i travis')
+    )
+    subprocess.check_call(['git', 'checkout', 'gh-pages'], cwd='gh-pages')
+    subprocess.check_call([
+        'helm', 'package', 'jupyterhub',
+        '--destination', 'gh-pages/'
+    ])
+    subprocess.check_call([
+        'helm', 'repo', 'index', '.',
+        '--url', 'https://jupyterhub.github.io/binderhub'
```
yuvipanda: Can you change this so this pushes to helm-chart still?

yuvipanda:
Works locally! \o/

PR #117, opened 2017-09-21
Make travis build and publish images + chart

yuvipanda:


yuvipanda:
Let's merge and see if it works!

PR #130, opened 2017-09-27
Wait for server to start before sending users to it

yuvipanda:
Fixes 403 errors on multi node kubernetes clusters

### DIFF
--- a/binderhub/redirect.py
+++ b/binderhub/redirect.py
@@ -9,7 +9,7 @@ import uuid
 from urllib.parse import quote
 
 from tornado.log import app_log
-from tornado import web
+from tornado import web, gen
 from tornado.httpclient import AsyncHTTPClient, HTTPRequest, HTTPError
 from tornado.httputil import url_concat
 
@@ -76,7 +76,7 @@ class RedirectHandler(BaseHandler):
         # start server
         app_log.info("Starting server for user %s with image %s", username, image)
         try:
-            await self.api_request(
+            resp = await self.api_request(
                 'users/%s/server' % username,
                 method='POST',
                 body=json.dumps({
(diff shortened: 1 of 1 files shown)


binderhub/redirect.py
```diff
@@ -76,14 +76,33 @@ def username_from_image(self, image):
...
                 method='POST',
                 body=json.dumps({
                     'token': token,
                     'image': image,
                 }).encode('utf8'),
             )
+            if resp.code == 202:
+                # Server hasn't actually started yet
+                # We wait for it!
+                for i in range(10):
```
choldgraf: `ii` :-)

binderhub/redirect.py
```diff
@@ -76,14 +76,33 @@ def username_from_image(self, image):
...
+                        method='GET',
+                    )
+
+                    body = json.loads(resp.body.decode('utf-8'))
+                    if body['server']:
+                        break
+                    # FIXME: make this configurable
+                    # FIXME: Measure how long it takes for servers to start
+                    # and tune this appropriately
+                    await gen.sleep(1.4 ** i)
```
choldgraf: why 1.4? is 14 some kind of special dev-ops maximum wait time?  +1 to being able to set this in the helmchart

binderhub/redirect.py
```diff
@@ -76,14 +76,33 @@ def username_from_image(self, image):
...
+            resp = await self.api_request(
                 'users/%s/server' % username,
                 method='POST',
                 body=json.dumps({
                     'token': token,
                     'image': image,
                 }).encode('utf8'),
             )
+            if resp.code == 202:
+                # Server hasn't actually started yet
```
willingc:
```
# Server has been requested.
# It hasn't actually started yet. That takes a small amount of time.
# We wait for it!
```

binderhub/redirect.py
```diff
@@ -76,14 +76,33 @@ def username_from_image(self, image):
...
+
+                    body = json.loads(resp.body.decode('utf-8'))
+                    if body['server']:
+                        break
+                    # FIXME: make this configurable
+                    # FIXME: Measure how long it takes for servers to start
+                    # and tune this appropriately
+                    await gen.sleep(1.4 ** i)
+                else:
+                    raise web.HTTPError(500, "Image %s for user %s took too long to launch" % (image, username))
```
choldgraf: the chart config could be to set a max wait time, and then this could just split that time up into logarithmically-sized chunks (so earlier chunks are shorter than later ones)

binderhub/redirect.py
yuvipanda: I played with values between 1 and 2 and found 1.4 to provide total wait times that weren't too large but also not too small. Agree it needs to be settable in helm chart.

binderhub/redirect.py
yuvipanda: It is! We have a utility function in jupyterhub to do that for us (along with some jitter). Might copy and use that here.

willingc: +1 to using the jitter utility.

PR #135, opened 2017-09-28
launch directly from build page

### LINKED ISSUE #133: Launch server & wait for completion as part of eventstream
Right now, we wait for the event stream to finish, and then redirect the user to a separate 'redirect' url that does the launching of the server and waiting for it to start & then redirecting.

Instead, we should do that as part of the eventstream itself.  It should just return the url + token at the end. Then the mybinder.org frontend code should be able to read that and construct appropriate URL to redirect to.

This way, other third party code can easily get the token + URL and use it as it sees fit.

minrk:
- replaces `RedirectHandler` with a `Launcher` object
- Launcher.launch returns `{url:, token:}` dict
- This info is published via event stream with `ready` state
- Frontend builds URL with token and filepath and triggers redirect
- When there's an error in build, the event stream sends a `phase: failed` event because eventstream cannot handle normal error codes.

Potential questions:

- What should be the public-api build URL?
- <del>Launch errors are now poorly handled because eventstream can't have error codes. I'll investigate dealing with this.</del> 

While in the builder page, a few more things:

- use async/await
- centralize `StreamClosed` handling in emit to avoid duplicating it everywhere

closes #133


minrk:
I added an example for consuming the event-stream with requests and launching the final URL with `webbrowser.open`

yuvipanda:
@minrk this is great! I've one primary question / suggestion:

Can we make the Launcher be aware of the Build object? We needed to use only image name when we were doing this with URL redirects, but since that's no longer the case Launcher can be passed an actual Build object. This way we can actually have reliable user names from images without having to guess, and is also cleaner I feel. In fact, we could just make the whole launcher be just a 'launch' object on Build. What do you think?

yuvipanda:
There also seems to be a merge conflict.

minrk:
> Can we make the Launcher be aware of the Build object?

👍

minrk:
I opened #142 as an alternative. It's the same as this one, but leaving the Launcher separate from the Builder. I think I like it a bit better, since they don't seem to have much to do with each other.

yuvipanda:
You are right, launcher can be invoked without having a build at all, so I'll merge #142.

yuvipanda:
@ellisonbg this has been deployed! Note that the contents of the response aren't set in stone yet, but should be not too many breaking changes!

PR #148, opened 2017-10-02
build.py: include python package changes when setting image, chart versions

### LINKED ISSUE #143: Have build.py set commit hash of image to be built
Right now, we are manually bumping the hash we are building in the Dockerfile. This is quite tedious. We should have build.py do it, but only rebuild it if the python package has actually changed. If it hasn't, we shouldn't spend time building the docker image unnecessarily.

minrk:
- path_changed, last_git_changed accept lists of paths
- include Python package files in git consideration for both image
- specify BINDERHUB_VERSION in image via `--build-arg` rather than manual changes
- fix bug where chart version would get `0.1.0-hash-hash-hash-hash` after building multiple times

closes #143

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -8,7 +8,7 @@ rbac:
 
 image:
   name: jupyterhub/k8s-binderhub
-  tag: 8022f73
+  tag: local
 
 repo2dockerImage: jupyter/repo2docker:v0.4.1
 
--- a/helm-chart/build.py
+++ b/helm-chart/build.py
@@ -59,10 +59,11 @@ def build_chart():
     with open(os.path.join(CHARTPATH, 'Chart.yaml')) as f:
         chart = yaml.load(f)
 
-    chart['version'] = chart['version'] + '-' + version
+    chart['version'] = chart['version'].split('-')[0] + '-' + version
 
     with open(os.path.join(CHARTPATH, 'Chart.yaml'), 'w') as f:
         yaml.dump(chart, f)
+
 
 def publish_pages():
     version = last_git_modified('.')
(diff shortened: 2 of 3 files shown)


yuvipanda:
Wouldn't it be simpler to just move the python package to a dir? This feels like it'll cause issues when we do something that we expect to cause a rebuild (like adding a MANIFEST.in file maybe) and it doesn't..

yuvipanda:
(going to merge this as is first)

yuvipanda:
I tried to test this with https://travis-ci.org/jupyterhub/binderhub/builds/282433344 but it didn't bump the binderhub version as it should have.

yuvipanda:
Fixed in https://github.com/jupyterhub/binderhub/pull/151

minrk:
> Wouldn't it be simpler to just move the python package to a dir?

I don't think we can do that if we want this to be a public Python package. `setup.py` and all other build-related files have to be in the top-level of the repo for things like `pip install <url>` to work (it's not technically impossible, `pip install binder-url#subdirectory=py-pkg-dir` can work, but it's a convention that we shouldn't deviate from without a *big* reason).

> like adding a MANIFEST.in file maybe

MANIFEST.in doesn't affect installation, so we *shouldn't* include that. but there are other examples, e.g. adding additional scripts to the build process. I'm not so concerned about this, though.

Alternately, if we *don't* want this to be a true publish-to-pypi, pip-installable package, we can move the whole package into the `image` directory, and then this gets a lot simpler because it becomes

```dockerfile
ADD binderhub binderhub
RUN pip install ./binderhub
```

 i.e. binderhub should *only* be installed via helm, not anything else.

Ultimately, though, it seems that maybe the helm-chart doesn't quite fit in the same repo as the Python package. Presumably, when we reach a degree of stability, the chart will start referencing release versions, rather than git revisions. Or is that not the case?

yuvipanda:
ah, I didn't know about the pip limitation. In that case this seems ideal, tyvm for thinking this through :)

PR #159, opened 2017-10-04
small updates to UI

choldgraf:
for when @yuvipanda shows me how to deploy updates to staging

these address some of the points in #132

### DIFF
--- a/binderhub/static/index.js
+++ b/binderhub/static/index.js
@@ -121,7 +121,6 @@ function rstBadge(url) {
   return '.. image:: ' + BADGE_URL + ' :target: ' + url
 }
 
-
 $(function(){
     var failed = false;
     var logsVisible = false;
@@ -153,6 +152,14 @@ $(function(){
       $('#badge-snippet')
         .show()
         .text(rstBadge(url));
+      return false;
+    });
+
+    $('#basic-url-toggle').on('click', function () {
+      var url = updateUrl();
+      $('#badge-snippet')
+        .show()
+        .text(url);
       return false;
     });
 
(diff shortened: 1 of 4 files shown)


binderhub/templates/index.html
```diff
@@ -14,6 +14,7 @@
 {% block main %}
 <div id="main" class="container">
   <div class="row">
+    <h3 style="margin-top: -60px;text-align: center;">(beta)</h3>
```
minrk: Can we avoid inline styles? The bootstrap css class "text-center" applies the text-align bit. CSS can be used for the margin adjustment.

choldgraf: Sounds good - I'll change this yuvi when runs me through deploying changes via PRs today

binderhub/static/index.js
```diff
@@ -121,6 +121,11 @@ function rstBadge(url) {
   return '.. image:: ' + BADGE_URL + ' :target: ' + url
 }
 
+function basicUrl(url) {
```
yuvipanda: I think we should just use the constant directly rather than adding a wrapper method.

binderhub/templates/index.html
yuvipanda: +1 on what @minrk said

binderhub/static/index.js
yuvipanda: Unless this function was supposed to do more?

binderhub/templates/index.html
```diff
@@ -14,6 +14,7 @@
 {% block main %}
 <div id="main" class="container">
   <div class="row">
+    <h3 class="text-center" style="margin-top: -60px;">(beta)</h3>
```
yuvipanda: you should add another id or class here, and add css rules for that in index.css :)

choldgraf: fineeeee :-)

PR #165, opened 2017-10-05
add GitHub auth support

minrk:
- [x] support client_id, client_secret in github provider
- [x] better error when rate limit exceeded
- [x] support env in helm chart

Once this lands, we can add the secrets for https://github.com/jupyterhub/mybinder.org-deploy/issues/58

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -62,6 +62,27 @@ spec:
             secretKeyRef:
               name: binder-secret
               key: "binder.hub-token"
+        {{ if .Values.github.clientId -}}
+        - name: GITHUB_CLIENT_ID
+          valueFrom:
+            secretKeyRef:
+              name: binder-secret
+              key: "github.client-id"
+        {{- end }}
+        {{ if .Values.github.clientSecret -}}
+        - name: GITHUB_CLIENT_SECRET
+          valueFrom:
+            secretKeyRef:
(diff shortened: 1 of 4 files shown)


helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -62,6 +62,27 @@ spec:
             secretKeyRef:
               name: binder-secret
               key: "binder.hub-token"
+        {{ if .Values.github.clientId -}}
```
yuvipanda: While this is ok for now, I think in the long run we should load the secret as a volume and read these from binderhub_config.py. Just  to keep the deployment definition clean.

yuvipanda:
@minrk I'm curious about possible merge strategies here. How coupled should merges here to deployment be? In past lives, I've found it very important that merges in repos like this are immediately deployed - this makes sure that we don't end up with a large number of changes that are undeployed, potentially breaking on the next person who does a deploy. If we do that, then a possible CR workflow is: 

1. someone else does an 'approve'
2. person who did the PR merges
3. person who did the PR immediately deploys / rolls back as appropriate.

what do you think?

minrk:
After chatting a bit: I think greenkeeper-style automatic PRs to staging when PRs land here or repo2docker would be pretty nice to keep the size of deploys small, but stopping short of automatic deployment of PRs from here to staging for now.

minrk:
rate limit prometheus PR coming soon!

PR #167, opened 2017-10-05
Add logging and prometheus gauge for github rate limit

minrk:
So we should have better tracking of when this is being consumed.

### DIFF
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -19,7 +19,7 @@ from tornado.httputil import url_concat
 from traitlets import Dict, Unicode, default
 from traitlets.config import LoggingConfigurable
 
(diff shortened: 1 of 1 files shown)


binderhub/repoproviders.py
```diff
@@ -5,18 +5,21 @@
...
+from prometheus_client import Gauge
+
 from tornado import gen
 from tornado.httpclient import AsyncHTTPClient, HTTPError
 from tornado.httputil import url_concat
 
 from traitlets import Dict, Unicode, default
 from traitlets.config import LoggingConfigurable
 
+GITHUB_RATE_LIMIT = Gauge('binderhub_github_rate_limit', 'GitHub rate limit remaining')
```
yuvipanda: I think this should be named 'binderhub_github_rate_limit_remaining' so it's clear it's the remaining requests. Shouldn't rely solely on the descripton for that info.

PR #169, opened 2017-10-06
adding URL to docs

choldgraf:
adds a URL for the binder user documentation on the front page, so it now looks like:

image


yuvipanda:
I'll deploy this along with #170 shortly

PR #180, opened 2017-10-10
Restore cross-origin access

### LINKED ISSUE #177: CORS is not enabled for BinderHub API
I'd like to use BinderHub's API to start kernels from another webpage.

When trying to start a notebook server using the BinderHub API, the browser will error because the Access-Control-Allow-Origin HTTP header is missing.

screenshot 2017-10-05 19 48 03

I believe this should be a simple implementation of [`set_default_headers`](https://stackoverflow.com/a/35259440) in the `base.py:BaseHandler` class but I'm still trying to get binderhub set up locally so I can't test it myself.

In addition, the current release of `notebook` (5.1.0) is missing a header that allows starting kernels from another domain. I fixed it in https://github.com/jupyter/notebook/pull/2876 but is there something we can do to get the fix in for the current deployment of Binder?

Happy to talk more about this (in person works too).

Related to #13.

minrk:
Still leave out disabling of xsrf, csp that was here before.

Since we are now using token auth, XSRF + token auth allows CORS to still be safe as only sites with access to tokens will be allowed access to notebook servers. Only disabling CORS and auth together makes things truly unsafe.

closes #177

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -59,6 +59,7 @@ jupyterhub:
                 '--port=%i' % self.port,
                 '--NotebookApp.base_url=%s' % self.server.base_url,
                 '--NotebookApp.token=%s' % self.user_options['token'],
+                '--NotebookApp.allow_origin=*',
             ] + self.args
 
         def start(self):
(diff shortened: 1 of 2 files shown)


minrk:
Thanks. Rebased.

yuvipanda:
Thanks @minrk! Can you deploy this when you get the time?

minrk:
Will deploy when #182 lands, which adds a missing piece to this.

PR #181, opened 2017-10-11
Fix typo in localtest sample binderhub_config.py

yuvipanda:
This makes it work again for people who have never installed
kubernetes tools on their local machine


yuvipanda:
I discovered this yesterday while helping @ian-r-rose get binderhub set up locally to test.

PR #182, opened 2017-10-11
Make it possible to enable CORS

### LINKED ISSUE #177: CORS is not enabled for BinderHub API

minrk:
enables access to the eventsource endpoint from other websites, not just scripts.

closes #177

Potential questions:

- Do we want to enable this by default (as this PR does)? Or should it be more locked down by default and opt-in on the mybinder.org deployment? (**answer: not enabled by default**)

cc @SamLau95

CORS is not enabled by default. To enable CORS from any origin in `values.yaml`:

```yaml
cors: &cors
  allowOrigin: '*'

jupyterhub:
  extraConfigMap:
    cors: *cors
```

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -39,10 +39,12 @@ github:
 # once in top-level cors.allowOrigin,
 # and again in jupyterhub.hub.extraConfigMap.cors.allowOrigin
 
-# &cors anchor allows us to remove redundancy in this values.yaml,
-# but it does not extend to user values.yaml.
-# They must still set cors separately for binderhub and notebooks.
-# However, they can use the same anchor pattern in their own values.yaml as well.
+# Using YAML anchors, `&cors` for the first appearance and `*cors` for subsequent
+# appearances, allows us to remove redundancy in this (BinderHub) `values.yaml`.
+# The anchors do not extend beyond this file.
+# As such, users must set `cors` separately for their own user `values.yaml` for notebooks.
+# `cors` will be set separately in the user `values.yaml` and binderhub`values.yaml`.
+# The same anchor pattern (`&cors`, `*cors`) can be used in the user `values.yaml`.
 
 cors: &cors
   allowOrigin:
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -25,7 +25,7 @@ service:
   type: LoadBalancer
   labels: {}
   annotations:
-    prometheus.io/scrape: "true"
+    prometheus.io/scrape: 'true'
   nodePort:
 
 github:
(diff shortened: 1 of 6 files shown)


yuvipanda:
@minrk yeah, I think the default should have this off and we should enable this for mybinder.org

Otherwise lgtm!

minrk:
Makes sense. I will do that in the morning:

1. switch default to off
2. expose it in helm chart
3. update mybinder.org-deploy to enable it

minrk:
It would be ideal if we can make the origin setting the same for notebook servers and binderhub itself. I'll see if that's doable, and do the simple thing first, if not.

minrk:
Okay, this should be set. The simplest way to enable CORS:

```yaml
cors: &cors
  allowOrigin: '*'

jupyterhub:
  extraConfigMap:
    cors: *cors
```

I was looking up info about yaml anchors to try to avoid setting this in more than one place and I couldn't find it. In a given `values.yaml`, anchors can be used to avoid setting *values* twice, but the anchor link must be established in each `values.yaml` (i.e. user values.yaml doesn't inherit anchors from the chart's values.yaml). The only way I think we could get this to be truly a single location is to put it in `Values.global`, but that would require teaching the hub chart about something that belongs in the Hub configmap coming from Values.global.

SamLau95:
Thanks @minrk ! I won't be able to immediately make API calls to beta.mybinder.org after this PR is merged; I'll have to wait until you guys update the deployment, is that right?

minrk:
@SamLau95 yes, but that is usually very soon after the PR lands with our current workflow. I'll let you know when it's deployed on beta.

helm-chart/binderhub/values.yaml
```diff
@@ -35,6 +35,18 @@ github:
   # ...or access_token
   accessToken:
 
+# have to set cors.allowOrigin twice:
+# once in top-level cors.allowOrigin,
+# and again in jupyterhub.hub.extraConfigMap.cors.allowOrigin
+
+# &cors anchor allows us to remove redundancy in this values.yaml,
```
willingc: Perhaps:

```
# Using YAML anchors, `&cors` for the first appearance and `*cors` for subsequent
# appearances, allows us to remove redundancy in this (BinderHub) `values.yaml`.
# The anchors do not extend beyond this file.
# As such, users must set `cors` separately for their own user `values.yaml` for notebooks. 
# `cors` will be set separately in the user `values.yaml` and binderhub`values.yaml`.
# The same anchor pattern (`&cors`, `*cors`) can be used in the user `values.yaml`.
```

willingc: @minrk A bit more context above on YAML anchors will make the pattern more understandable.

minrk: Thanks! added your comment.

willingc:
Thanks @minrk. Nice addition about anchors not extending beyond the file.

Issue #164, opened 2017-10-05
BinderHub fails with internal server error if repo2docker produces non JSON output


yuvipanda:
Due to a broken docker image, repo2docker produced the following stdout output:

```
/ # jupyter-repo2docker https://github.com/yuvipanda/example-requirements --json-logs
Traceback (most recent call last):
  File "/usr/local/bin/jupyter-repo2docker", line 11, in <module>
    load_entry_point('jupyter-repo2docker==0.4.1', 'console_scripts', 'jupyter-repo2docker')()
  File "/usr/local/lib/python3.6/site-packages/repo2docker/__main__.py", line 6, in main
    f.start()
  File "/usr/local/lib/python3.6/site-packages/repo2docker/app.py", line 309, in start
    checkout_path
  File "/usr/local/lib/python3.6/site-packages/repo2docker/app.py", line 95, in fetch
    capture=self.json_logs):
  File "/usr/local/lib/python3.6/site-packages/repo2docker/utils.py", line 12, in execute_cmd
    proc = subprocess.Popen(cmd, **kwargs)
  File "/usr/local/lib/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/usr/local/lib/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'git': 'git'
```

BinderHub then tried to parse this as JSON, choked and produced an 'internal server error' that isn't very useful.

This is what caused https://github.com/jupyterhub/mybinder.org-deploy/pull/52 to fail

yuvipanda:
I'm still unsure why / how it worked on staging tho.

choldgraf:
I thought that this broken r2d image had been fixed,  but I'm getting the same server error when trying to build this repo:

https://github.com/binder-examples/latex

Note that this doesn't happen for all repos (it built a repo with *just* requirements.txt fine)

choldgraf:
ah after trying it locally with r2d I see it is in fact the same problem...per https://github.com/jupyter/repo2docker/pull/85 we're raising an error if postBuild isn't executable, but we still don't have the pieces in place to grab this information and display it to the user

choldgraf:
Now I'm getting this strange error, it detects `postBuild` and then tries to run it, but when it does it can't seem to find the file

[...]

PR #186, opened 2017-10-17
allow phase to be undefined in events

minrk:
enables log messages that are not state changes

right now, if a message is published with no phase, an error is raised calling toLowerCase() on undefined.

I'm not 100% sure this is right, but exploring #164 I found a number of solutions where I don't really have a meaningful value of 'phase' to use when I'm wrapping up unhandled messages.


yuvipanda:
@minrk can you do a deployment when you get the chance? tyvm for fixing these!

minrk:
Deploying to staging now https://github.com/jupyterhub/mybinder.org-deploy/pull/86

PR #188, opened 2017-10-17
Bump timeout a bit

yuvipanda:
Needs reworking, but is better than the super short timeout
we currently have


binderhub/launcher.py
```diff
@@ -111,7 +111,7 @@ def username_from_repo(self, repo):
             if resp.code == 202:
                 # Server hasn't actually started yet
                 # We wait for it!
-                for i in range(10):
+                for i in range(32):
```
minrk: Because of the exponential, this is 32 hours. Maybe `min(1.4**i, 10)`?

PR #197, opened 2017-10-19
use google cloud shell

aculich:
GCS already has gcloud cli tool installed, so suggest that to people as their first experience setting up binderhub— don't suggest they try to install the gcloud cli tool locally on their laptop (which is more suitable for a more advanced user).

### DIFF
--- a/doc/create-cloud-resources.rst
+++ b/doc/create-cloud-resources.rst
@@ -37,10 +37,7 @@ connect your credit card or other payment method to your google cloud account.
    commands to Google Cloud and lets you do things like create and delete
    clusters.
 
-   - Go to the `gcloud downloads page <https://cloud.google.com/sdk/downloads>`_
-     to **download and install the gcloud SDK**.
-   - See the `gcloud documentation <https://cloud.google.com/sdk/>`_ for
-     more information on the gcloud SDK.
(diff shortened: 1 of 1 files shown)


doc/create-cloud-resources.rst
```diff
@@ -37,10 +37,7 @@ connect your credit card or other payment method to your google cloud account.
    commands to Google Cloud and lets you do things like create and delete
    clusters.
 
-   - Go to the `gcloud downloads page <https://cloud.google.com/sdk/downloads>`_
-     to **download and install the gcloud SDK**.
-   - See the `gcloud documentation <https://cloud.google.com/sdk/>`_ for
-     more information on the gcloud SDK.
+   - Use Google Cloud Shell which will give you access to the ``gcloud`` command-line tool.
```
willingc: Thanks @aculich. I agree that this approach makes sense and I would add a second sentence here on where to find on the nav bar for Google Cloud.

I would put the removed text in a directive for advanced users directly below.

Line 36 needs to be tweaked a bit. In fact, it may be better to break out line 41 as it's own step (new Step 4).

choldgraf:
pinging @aculich :-) a few small tweaks and then I think we're good to go on this

willingc:
Going to go ahead and merge this and then do a follow up PR. Thanks @aculich :cookie:

PR #212, opened 2017-10-19
Make `master` a placeholder.

Carreau:
The `master` form already has `master` as a placeholder; this (should)
make full use of this; by replacing empty branches by master

### DIFF
--- a/binderhub/static/index.js
+++ b/binderhub/static/index.js
@@ -11,6 +11,8 @@
   pushing -> failed
 */
 
+"use strict";
+
 function Image(provider, spec) {
     this.provider = provider;
     this.spec = spec;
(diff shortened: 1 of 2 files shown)


yuvipanda:
Hmm, I feel like we explicitly should not do this, since that will affect
API semantics for non-ui use cases.

yuvipanda:
I take that back! This is actually totally fine and an improvement! We
should merge and deploy this.

Carreau:
I'll attempt to do that later!

binderhub/static/index.js
```diff
@@ -102,7 +104,7 @@ function updateUrl() {
   repo = repo.replace(/^(https?:\/\/)?github.com\//, '');
   // trim trailing or leading '/' on repo
   repo = repo.replace(/(^\/)|(\/?$)/g, '');
-  var ref = $('#ref').val().trim();
+  var ref = $('#ref').val().trim() || 'master';
```
minrk: empty strings aren't falsy in javascript, so `|| master` will never run. We'd have to add

```javascript
if (ref.length === 0) ref = 'master';
```

to get this behavior.

Carreau: `||` seem to work for me.... but happy to change. 
screen shot 2017-10-20 at 11 20 56

Carreau: [According to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators), empty-string is casted to false.

minrk: Whoa, not sure where I'm getting my info. Sorry for misleading!

PR #215, opened 2017-10-20
Remove jupyter_alabaster_theme

Carreau:
It is not maintained anymore and have  a number of style and navigation
issue. Reverting to RTD is (IMHO) better, in particular because of the
Prev/Next button support; which alabaster does not support even if it
looks lighter ; which I think is important.

-- 

cc @willingc (which I think was pro alabaster), and @minrk for his thoughts as if this get approved I'll send identical PRs to other repositories like maybe JupyterHub.

Thoughts ?

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -101,7 +101,7 @@ todo_include_todos = False
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'jupyter_alabaster_theme'
+html_theme = 'alabaster'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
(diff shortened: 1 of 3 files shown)


choldgraf:
I really like the theme we use w/ z2jh...is that what you mean by the RTD theme?

If the team is into this idea then so am I!

willingc:
To clarify z2jh uses the standard alabaster theme with a few modifications in the static directory (sidebar template and css tweaks), I am in favor of using alabaster for all jupyterhub/binderhub repos. I would like to work with @Carreau and others to standardize any custom templates or css that we wish to use. Overall, I think alabaster will better suit our needs. That said, I think we can work on fonts and colors to better reflect the Jupyter design standard. @minrk thoughts?

Carreau:
> @Carreau looks good. You missed line 105 and changing that to alabaster.

Done. Thanks !

minrk:
Just catching up. I trust you folks on these!

PR #217, opened 2017-10-21
Fix reference in doc and remove unneeded doc file

willingc:
- Fix reference deleted in error #209 
- Remove redirect doc and its listing in the api index toctree as `redirect.py` no longer exists


yuvipanda:
tyvm, @willingc!

PR #218, opened 2017-10-21
add markdown config to docs

### LINKED ISSUE #210: Suggestion: add markdown rendering for readthedocs
See https://github.com/dib-lab/sourmash/blob/master/doc/conf.py#L295 for an example of using in this in our project.

You will also need to install the `recommonmark` package.

ctb:
Fixes #210.


doc/conf.py
```diff
@@ -171,3 +171,14 @@
      author, 'BinderHub', 'One line description of project.',
      'Miscellaneous'),
 ]
+
```
willingc: Please move this higher up in the file. Ideally around line 58.

willingc:
Thanks @ctb. Let's keep this consistent in location in conf.py as we have in other Jupyter docs. See jupyterhub for example if needed.

setup.py
```diff
@@ -9,7 +9,8 @@
         'traitlets',
         'docker',
         'jinja2',
-        'prometheus_client'
+        'prometheus_client',
+        'recommonmark'
```
yuvipanda: also should we have a different requirements.txt inside the doc folder? binderhub doesn't require recommonmark to run, so am not sure if this belongs here.

What's the general convention for packages needed only for generating docs?

willingc: I typically put packages needed for docs only in a requirements-doc.txt or environment.yml within the docs directory. See jupyterhub repo for more guidance.

choldgraf:
I'm +1 on this as well once @yuvipanda + @willingc comments are addressed

ctb:
ready for review again, @willingc!

yuvipanda:
LGTM, but I'll wait for @willingc to be back before merging!

willingc:
I'll be back on Thursday :-)

We may want to play around with this (https://github.com/miyakogi/m2r) at some point too.

betatim:
> We may want to play around with this (https://github.com/miyakogi/m2r) at some point too.

I've used pandoc (I think) to convert markdown READMEs to rst. Mostly works but almost always it needed some hand crafting afterwards. It didn't generate syntax errors but the output wasn't quite the same as the original markdown output.

willingc:
@betatim Agree that pandoc isn’t an elegant solution. In my quick scan the other day m2r takes into more account the details of rst, sphinx, docutils.

willingc:
@yuvipanda Thanks for the ping. Made one small change and merged.

PR #220, opened 2017-10-23
fixing disappearing dropdown

choldgraf:
This was bugging the crap outta me so I figured I'd give a shot at actually fixing this. I googled around a bit and it seems like this is a related problem

https://stackoverflow.com/questions/16050743/bootstrap-dropdown-disappears-on-input-select

This PR implements the solution suggested there, which seemed to do the trick for me locally. Can somebody take a look and let me know if this will have unforeseen consequences? Because I am a total noob w/ javascript...

### DIFF
--- a/binderhub/static/index.js
+++ b/binderhub/static/index.js
@@ -143,6 +143,11 @@ $(function(){
       $('#badge-snippet').hide();
     });
 
+    // prevent badge text dropdown menu from disappearing on click
+    $('.dropdown-menu#badge-text li').click(function(e) {
+          e.stopPropagation();
+    });
+
     $('#markdown-badge-toggle').on('click', function () {
       var url = updateUrl();
       $('#badge-snippet')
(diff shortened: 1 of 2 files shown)


binderhub/static/index.js
```diff
@@ -143,6 +143,11 @@ $(function(){
       $('#badge-snippet').hide();
     });
 
+    // prevent dropdown menu from disappearing on click
+    $('.dropdown-menu li').click(function(e) {
```
yuvipanda: This is now generic to all dropdown menus. Currently we only have one, but if we have another in the future it'll be very confusing.

We should:
1. Add an id to the dropdown here in the HTML
2. Use that in the selector
3. Explicitly specify which dropdown we're talking about in comments

choldgraf:
latest push should have those fixes...LMK if that looks OK

yuvipanda:
Thanks @choldgraf! I've merged! Would you like to do a deploy?

choldgraf:
yes!
--

yuvipanda:
go for it :)

choldgraf:
done!

betatim:
🍻 Excellent!

PR #222, opened 2017-10-25
[WIP] Auto update deployment on successful build.

Carreau:
Push to a 3rd party repo, we would still need to make the PR manually,
but next step is to get a token that can do the PR for us.

### DIFF
--- a/helm-chart/build.py
+++ b/helm-chart/build.py
@@ -23,6 +23,9 @@ IMAGE_FILES = [SETUP_PY, PYPKGPATH, IMAGE_PATH]
 # namely, all image files plus the chart itself
 CHART_FILES = IMAGE_FILES + [CHARTPATH]
 
+HELM_CHART_DEPLOY_KEY_NAME = 'travis'
+MYBINDER_DEPLOY_KEY_NAME = 'unknown'
+
 def last_git_modified(paths):
     """Return the short hash of the last commit on one or more paths"""
     if isinstance(paths, str):
@@ -82,10 +85,13 @@ def build_chart():
     with open(os.path.join(CHARTPATH, 'Chart.yaml')) as f:
         chart = yaml.load(f)
 
+    raw_version = chart['version']
+
     chart['version'] = chart['version'].split('-')[0] + '-' + version
 
     with open(os.path.join(CHARTPATH, 'Chart.yaml'), 'w') as f:
         yaml.dump(chart, f)
+    return raw_version
 
 
 def publish_pages():
@@ -93,7 +99,7 @@ def publish_pages():
     subprocess.check_call([
         'git', 'clone', '--no-checkout',
         'git@github.com:jupyterhub/helm-chart', 'gh-pages'],
-        env=dict(os.environ, GIT_SSH_COMMAND='ssh -i travis')
+        env=dict(os.environ, GIT_SSH_COMMAND=f'ssh -i {HELM_CHART_DEPLOY_KEY_NAME}')
     )
     subprocess.check_call(['git', 'checkout', 'gh-pages'], cwd='gh-pages')
     subprocess.check_call(['helm', 'repo', 'add', 'jupyterhub', 'https://jupyterhub.github.io/helm-chart/'])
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -26,6 +26,9 @@ before_install:
   if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
     openssl aes-256-cbc -K $encrypted_d8355cc3d845_key -iv $encrypted_d8355cc3d845_iv -in travis.enc -out travis -d
     chmod 0400 travis
+    ## deploy to MeeseeksDev Org's repo
+    openssl aes-256-cbc -K $encrypted_e0a1711769d1_key -iv $encrypted_e0a1711769d1_iv -in deploy_key.enc -out deploy_key -d
```
yuvipanda: Unfortunately this won't work. You can have only one encrypted file per travis repo, which totally sucks...

yuvipanda:
Much excite!

Carreau:
Sorry added the deploy key to the wrong repo. I need to retry. Are you sure you can only get one encrypted file/repo ?

yuvipanda:
will be very happy if we can add more than one, but couldn't. in http://github.com/berkeley-dsep-infra/datahub/ I use a combination of git-crypt and a tarball...

Carreau:
Actually yes, you are right, can only encrypt one file... I'll rethink it. I hope I have not void the previous file by sending this PR.

Carreau:
@yuvipanda; question
Am I right that right now the secure variable are setting: `DOCKER_USERNAME`, `DOCKER_PASSWORD`,  as well as 1 deploy-key for https://github.com/jupyterhub/helm-chart ?

Carreau:
Ok, let me know what you think about that ? It should not change how things works currently but provide a script to regenerate the ssh-keys , tar them and add the to a `secret.tar.enc` file.

To which org/repo should we send the automatic update ? Likely not jupyterhub/binderhub directly, I guess. I have a few orgs we can use; I don't know if you can have 2 forks/org, in which case we can push a a `jupyterhub/mybiner-autopush` ?

helm-chart/build.py
```diff
@@ -115,7 +121,38 @@ def publish_pages():
...
+    with open(os.path.join('staging', 'config', 'common.yaml')) as f:
+        values = yaml.load(f)
+
+    values['version'] = version
+
+    with open(os.path.join('staging', 'config', 'common.yaml')) as f:
+        yaml.dump(values, f)
+
+
+    subprocess.check_call(['git', 'add', '.'], cwd='staging')
```
yuvipanda: git add . is very dangerous always, no? especially with possible secrets around. I'd rather we explicitly only added the two files we have changed.

tools/regenerate_secrets.py
```diff
@@ -0,0 +1,54 @@
+"""
```
yuvipanda: I think we should just use git-crypt instead of this.

helm-chart/build.py
```diff
@@ -115,7 +121,38 @@ def publish_pages():
...
+    commit = last_git_modified('.')
+    subprocess.check_call([
+        'git', 'clone',
+        'https://github.com/jupyterhub/mybinder.org-deploy', 'staging'],
+        env=dict(os.environ, GIT_SSH_COMMAND='ssh -i {MYBINDER_DEPLOY_KEY_NAME}')
+    )
+    subprocess.check_call(['git', 'checkout', 'staging'], cwd='staging')
+    subprocess.check_call(['git', 'remote', 'add', 'meeseeks', 'ssh://git@github.com/MeeseeksBox/mybinder.org-deploy'])
+
+    with open(os.path.join('staging', 'config', 'common.yaml')) as f:
```
yuvipanda: when I moved projects yesterday, the location for the binderhub version has also changed. I haven't updated README yet, but the version is now in mybinder/requirements.yaml. Apologies for changing this out from under you!

Carreau:
I'll try to understand git-crypt and tell you.

Carreau:
Hum, I check ed GitHub API; you can't make cross-repositories PRs using GitHub integrations; and I'm not really a fan of using a Personal Access Token as it would get access to everything the person who created the Token have access to.

So either we directly push to a separate branch on Jupyter/BinderHub ; or I don't know.

yuvipanda:
What about using a deploy key with push access? That is what I do in other
repos in the past. You can give a specific (new) ssh key push access to a
specific branch...

Carreau:
The push is not a problem; it's opening the PR which is. I suppose we do not want something to automatically push on master by mistake; or credentials that can push on master to be accidentally leaked. 

What I can do is 
  - A) push on `jupyterhub/mybinder.org-deploy:a-random-branch-name`
  - B) PR from above to  `jupyterhub/mybinder.org-deploy:master`
PRO
  - The credentials needed for B) can be limited to only creating PRs
CONS
  -  The credentials needed for A) can push on master

Or:
  - A) push on `binderhub-ci-repos/mybinder.org-deploy:a-random-branch-name`
  - B) PR from above to  `jupyterhub/mybinder.org-deploy:master`
PRO
  - The credentials needed for A) can't touch main deployment
CONS
  -  The credentials needed for B) are not restricted to single repo and have write access.

Sorry if I"m unclear.

Carreau:
Also to be clearer:

> You can give a specific (new) ssh key push access to a
> specific branch...

I don't think you can do that; and AFAICT the protect branch does apply to Deploy Keys.

yuvipanda:
I meant specific repo, sorry.

I understand the problem better now! To me, the right solution now seems to be to create a new github user that's a bot. Then, the bot can:

1. Push to its own fork of mybinder.org-dpeloy on a branch
2. Make a PR to mybinder.org-deploy's staging branch
3. Human merges PR
4. Make a PR to mybinder.org-deploy's prod branch
5. Human merges this too!

This should work out rights-wise, right?

Apologies for the late response, @Carreau!

yuvipanda:
@Carreau do you think you can give me access to the bot account on github somewhere? I want to pick this up...

betatim:
I'll close this as we haven't worked on it for a long time. We now have a setup with https://github.com/henchbot/ that is working well

PR #230, opened 2017-10-27
Add fake provider and building classes to allow better work on the UI.

Carreau:
This is still imperfect as I'm not well aware of the various states the
Pods can be in. One thing we could do is actually capture a couple of
build cases and provide them for replay at different speeds.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,4 +71,6 @@ there is a simpler method!
 
 3. You can now access it locally at https://localhost:8585
 
-Note that building and launching will not work!
+Note that building and launching will not work, but the
+`testing/localonly/binderhub_config.py` setup a fake building process which
+allows you to work on the UI experience.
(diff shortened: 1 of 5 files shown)


choldgraf:
The main benefit of this over the current UI is that it lets you go through the full process yeah?

Carreau:
It does not change current UI, but fake the GitHub fetching and building
which should make working on the UI locally (when you do not have minikube)
easier. In the long term this should also allow you when working on the UI
to decide whether it should fail or succeed.

One thing I'd like to do for example is intercept the 'Step 13/39' do
display a more informative progress bar.

yuvipanda:
Thank you very much, @Carreau!

choldgraf:
will this "just work" if we're running the localonly `.py` script? Or should we write docs to explain how to use this?

choldgraf:
will this "just work" if we're running the localonly `.py` script? Or should we write docs to explain how to use this?

Carreau:
> Will this "just work" if we're running the localonly .py script? 

Yes.

PR #231, opened 2017-10-31
Prepare using git-crypt for key management.

Carreau:
Install git-crypt on travis, and setup the gitattributes to encrypt
everything that would end in a `.key`

Move to global variable instead of hardcoding the name of the travis key.

Should we also have all secrets in the subfolder ?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+secrets/** filter=git-crypt diff=git-crypt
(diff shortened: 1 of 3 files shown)


yuvipanda:
I'd also prefer a secrets/** line in the .giattributes and moving all secrets in there, instead of using *.key.

Carreau:
> I'd also prefer a secrets/** line in the .giattributes and moving all secrets in there, instead of using *.key.

Done.

.gitattributes
```diff
@@ -0,0 +1 @@
+secrets/* filter=git-crypt diff=git-crypt
```
yuvipanda: This needs to be secrets/**, otherwise it doesn't count subdirectories!

Carreau: Done.

yuvipanda:
ty, @Carreau! I'll check if this succeeds!

PR #232, opened 2017-11-01
support urlpath in addition to filepath

### LINKED ISSUE #128: Allow repos to specify initial URL
Right now a binder instances opens to the classic notebook which has a URL of `/tree`. It would be very helpful to allow a repo to specify the initial URL so that, for example, JupyterLab could open up (`/lab`). Combined with per repo `DOCKERFILE` support this would open the door to host just about anything this way.

minrk:
allows opening arbitrary URLs in addition to specific files

e.g. `/lab` for JupyterLab

closes #128

From URLs, etc. `urlpath` is treated the same as `filepath`, which should be fine.

I'm least confident about the UX part of choosing whether the path to open is a file or URL.

GIF:

binder

### DIFF
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -40,6 +40,7 @@ class ParameterizedMainHandler(BaseHandler):
             url=provider.get_repo_url(),
             ref=provider.unresolved_ref,
             filepath=self.get_argument('filepath', None),
+            urlpath=self.get_argument('urlpath', None),
             submit=True,
             google_analytics_code=self.settings['google_analytics_code'],
         )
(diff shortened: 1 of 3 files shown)


yuvipanda:
Awesome. Thanks, @minrk. Can you do a deploy when you get the chance?

choldgraf:
this is awesome - I'll do a deploy now. @yuvipanda are you around just in case we got the instructions wrong on the latest mybinder.org readme commit?

yuvipanda:
@choldgraf yup! go for it.

choldgraf:
looks to be working OK.

That said, I think the UI is getting to the point that we should have a high-level plan for it rather than glomming on elements here and there. I could see users not understanding the difference between `URL` and `filePath` etc...

yuvipanda:
Yeah, the JS is also getting to a point where we should start using something that provides a little more structure than just jquery.

PR #235, opened 2017-11-02
Make really sure that git URL comes at the end

yuvipanda:
Otherwise we do not actually ask repo2docker to push images...


yuvipanda:
Self-merging, since binder is broken right now and this should unbreak it. Apologies everyone, I think I caught everyone in a timezone dead zone.

choldgraf:
ah whoops - nice catch

PR #248, opened 2017-11-07
Rework dependencies of docs env.

Carreau:
I can't get it to work on my machine because of pygraphviz which refuses to install, 
and there is an actual dependency on MNE to create the flow diagram, which I suspect does not require Mayavi to actually do the diagram; but with conda that forces us to use Python 2.7 by indirect dependency. 

The diagram generated is this one : http://binderhub.readthedocs.io/en/latest/diagram.html

I think @choldgraf wrote the code to do that. Maybe instead of regenerating it everytime we should/could make the diagram once and store it in this repository.

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -6,8 +6,15 @@ dependencies:
 - traitlets>=4.1
 - sphinx>=1.4, !=1.5.4
 - sphinx_rtd_theme
-- mne
 - graphviz
+- recommonmark==0.4.0
+- pip
 - pip:
-  - recommonmark==0.4.0
   - pygraphviz==1.4rc1
+  - CommonMark<0.6
+  - mne # conda refuse to install as mayvi is 2.5 only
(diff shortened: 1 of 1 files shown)


choldgraf:
I'm +1 on removing the MNE dependency...that's an unnecessarily heavy dependency IMO.

willingc:
> Maybe instead of regenerating it everytime we should/could make the diagram once and store it in this repository.

Excellent idea.

Carreau:
> The -pip: line should remain (as it's how we do it everywhere else). If it's not working it's likely a bug with conda.

Also just to be clear I have not removed `- pip:`  I have added pip as a dependency, it [is a bug w/ conda](https://github.com/conda/conda/issues/5680).

choldgraf:
Didn't we remove the dependency on mne in another PR? This LGTM but can you resolve the merge conflict?

Carreau:
Nah, was superceeded by #250

PR #253, opened 2017-11-08
Add previous and next link to the bottom content in each page.

Carreau:
This is in addition to the content in the sidebar in order to be more
discoverable by users.

I'm also personally a fan of having it on top of each page as well; but
I don't think it fits well with _this_ theme, And the link in the
sidebar are sufficient.

--- 

Here is what it looks like:

---

screen shot 2017-11-08 at 13 16 37
---
If a title is too long it get elided:

--- 

screen shot 2017-11-08 at 13 16 45

---

This is not included, but we could also add link at the top of the page

screen shot 2017-11-08 at 13 15 50

---

cc @willingc

### DIFF
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -1,3 +1,28 @@
 img.logo {
   width:100%
 }
+
+.right-next {
+    float: right;
+    max-width: 45%;
+    overflow: auto;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+}
+
+.right-next::after{
+    content: ' »';
+}
+
+.left-prev {
+    float: left;
+    max-width: 45%;
+    overflow: auto;
+    text-overflow: ellipsis;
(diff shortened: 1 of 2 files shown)


willingc:
I have to say that I would like the links at the top too. Even if just "next" and "prev" in the upper righthand side.

choldgraf:
ack ok now I'm confused, I thought I just saw another PR where there were "previous" and "next" links...but it seems like this PR supersedes that?

FWIW I prefer the approach of this PR to have the full name of the previous/next section, I think it's a nice addition

Carreau:
There are in total 3 PRs, 

- One from @willingc to add the links in the navbar.
- One from me (this one) adding links at the bottom of the content (usefull on mobile pointed out somewhere else). This one does _not_ add on top. 
- And a third to argue as to whether add it at the top.

PR #254, opened 2017-11-09
Remove dependencies in Sphinx RTD Theme and work around conda bug.

Carreau:
The Sphinx RTD theme is AFAICT not useful.

Conda has a bug (https://github.com/conda/conda/issues/5680), with a
proposed fix (https://github.com/conda/conda/pull/6275). Waiting for it
we should should list pip as a conda dependency when using the `pip:`
section, as a temporary workaround.


willingc:
Yep. The Sphinx RTD theme is unused. Thanks for following up with the conda bug too.

Issue #249, opened 2017-11-07
Creating a static diagram


choldgraf:
Currently the diagram is being generated by code each time the docs build. As mentioned in a few issues, this is a sub-optimal solution, and it'd probably be easier to just use static diagrams that we can include in the documentation.

I took a stab at drafting what these might look like:

https://docs.google.com/presentation/d/1t5W4Rnez6xBRz4YxCxWYAx8t4KRfUosbCjS4Z1or7rM/edit?usp=sharing

There are two components in there:

* JupyterHub with Kubernetes
* BinderHub

Think we can iterate on this and get two "canonical" diagrams for our uses?

willingc:
>Think we can iterate on this and get two "canonical" diagrams for our uses? (one for jupyterhub one for binderhub)

Good idea @choldgraf. Let's iterate on these. Also, we've had a pretty standard visual for JupyterHub for a while. Let's scope this issue to:
- JupyterHub with Kubernetes
- BinderHub

choldgraf:
ah yes, good point, I meant JupyterHub k8s!

choldgraf:
I'll plan to freeze the diagrams in that google presentation and make a PR to add them to the docs soon, so I'd love to hear thoughts/feedback/suggestions etc!

choldgraf:
hey all - these diagrams are now semi-finished (in the always evolving but currently pretty good sense), so I'm closing this!

PR #255, opened 2017-11-09
Do not generate the flow diagram at docs rebuild.

Carreau:
- Remove graviz-like packages from dependencies
 - Do not regen the graph in conf.py
Plus
 - Pygraphviz is hard to install on some system (OS X).
 - The diagram may need some manual adjustments to avoid objects overlaying each other (mainly text)

That's the only way I manage to build the docs on my machine.

On top of #254 otherwise there is conflict.


choldgraf:
thanks for this @Carreau , I think it makes the most sense until we can figure out another diagram solution (see #249)

PR #256, opened 2017-11-09
Add top Previous/Next on each page.

Carreau:
As requested by willinc, to see what it looks like (and if we like it).
Text is literally `Previous` and `Next` not to visually change between
page and stay discreet.

I think it is useful, (I like that), but I tink it is only minimally
useful so would be ok if it's not merged.

screen shot 2017-11-08 at 18 14 14

### DIFF
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -26,3 +26,10 @@ img.logo {
     content: '« ';
 }
 
+.prev-next-bottom {
+  margin-top: 3em;
+}
+
+.prev-next-top {
+  margin-bottom: 1em;
+}
--- a/doc/_templates/page.html
+++ b/doc/_templates/page.html
@@ -7,17 +7,24 @@
    or symbols are handled by CSS classes in _static/custom.css
 #}
 
-{% block prev_next %}
+{% macro prev_next(prev, next, prev_title='', next_title='') %}
   {%- if prev %}
-     <a class='left-prev' href="{{ prev.link|e }}" title="{{ _('previous chapter')}}">{{ prev.title }}</a>
+     <a class='left-prev' href="{{ prev.link|e }}" title="{{ _('previous chapter')}}">{{ prev_title or prev.title }}</a>
   {%- endif %}
   {%- if next %}
-     <a class='right-next' href="{{ next.link|e }}" title="{{ _('next chapter')}}">{{ next.title }}</a>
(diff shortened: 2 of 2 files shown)


willingc:
I like it :smile: 

Thanks @Carreau. I would recommend merging since it gives the user an easy way to page through at the top as they are going through the tutorial (i.e. finding their place after a break) as well as having it on the bottom to navigate as well.

I approve, but will leave for others to merge.

Carreau:
Happy to have someone that share my thoughts about usability of documentation, and the project in general.

willingc:
What's cool is how it improves the mobile (and narrow screen width) experience. Handy for docs on one side of the screen and terminal on the other. 

screenshot 2017-11-08 at 6 31 55 pm

choldgraf:
I think this is cool as well! I have one question about how those links get assigned. How does the macro know which is the previous, and which is the next, page? I feel like there is some jinja magic going on here :-)

Carreau:
`prev` and `next` are automagically injected by sphinx/jinja into the template. THen you either call the macro with overwriting the text of next/prev page `{{ prev_next(prev, next, 'Previous', 'Next') }}`, or just let it figure it on its own. `{{ prev_next(prev, next) }}`, or you could for example not have a prev link `{{ prev_next(None, next) }}`

Carreau:
@choldgraf [said](https://github.com/jupyterhub/binderhub/pull/253#issuecomment-343041040) in #253

> FWIW I prefer the approach of this PR to have the full name of the previous/next section, I think it's a nice addition

At the bottom it is the full name; I think at the top it is too distracting to have the full names and why  I kept `Prev/Next`. You don't want to spoil it to the user. Imagine the following:

screen shot 2017-11-08 at 20 20 11

Oh now .....

choldgraf:
ah ok, I think I had only noticed the `prev` and `next` at the top and assumed it was the same in both. Are the variables `prev` and `next` set by the order in toctree then?

Carreau:
> Are the variables prev and next set by the order in toctree then?

g1f1ecn-1

choldgraf:
The reason I was asking is because I was trying to figure out if this would be useful for z2jh, which as a similar linear structure to one part of it. However it has multiple toctree sections on the index

willingc:
We probably should migrate the changes to z2jh too.

choldgraf:
I'm +1 on this

PR #259, opened 2017-11-10
Avoid spurious newlines in binderhub secret

yuvipanda:
I have absolutely no idea why this was perfectly fine until
today and suddenly decided to start giving me hell. helm
fails completely because there is a newline at the end of the
secret value, rejected by kubernetes.

I hate text templating to produce objects.


yuvipanda:
Apologies for self-merging, trying to deal with an outage.

Issue #262, opened 2017-11-10
Make binderhub chart work with latest master of JupyterHub chart


yuvipanda:
Currently it does not. Investigate why and fix it!

hydrosquall:
I originally got the following error in my console (and the site refuses to load) after following the directions in `CONTRIBUTING.md`, and attempting to visit the following link in either Chrome or Safari. (I am developing on Mac OSX). 

https://localhost:8585/

```bash
[I 171111 13:31:50 http1connection:253] Malformed HTTP message from ::1: Malformed HTTP headers: '\n\x01\x00\x00wZZ\x00\x00ÿ\x01\x00\x01\x00\x00\x00\x00\x0e\x00\x0c\x00\x00\tloca'
```

However, the error went away when I switched from `https` to `http` inside my web browser. Maybe this could be related?

hydrosquall:
#265 fixes the part where the I can now successfully pull down new repositories (which wasn't true before). However, after the container for the repo entered into the UI builds, the new server fails to launch.

choldgraf:
@yuvipanda does this work now? alternatively, do you think this is related to the RBAC challenges reported in #391

choldgraf:
It does now! thanks to #402

PR #264, opened 2017-11-11
Enable Branch Names to contain slashes (Fix #263)

### LINKED ISSUE #263: Setting build by branch won't work if name contains "/"
The git branch naming convention I've been using is something like `hotfix/name-of-issue` or `feature/name-of-issue`. However, I've been unable to build specific branches. I suspect that perhaps the branch reference is being split on `/`, and gets confused if the branch name contains this character.

For example, I've been unable to build my branch here `feature/add-requirements-file` on mybinder.org, so I've had to use the exact commit hash instead:

 https://github.com/hydrosquall/pandas-cookbook/tree/feature/add-requirements-file

I get the following error message:

```bash
Spec is not of the form "user/repo/ref", provided: "hydrosquall/pandas-cookbook/feature/add-requi
rements-file".
```

hydrosquall:
# Background

- When people follow the common `git-flow` [naming convention](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow), their branches will contain a `/` character. 
- In the present version of binderhub, such branches will not work (this should Fix #263)
- We'd like to be able to support building off of branches named with any naming convention, otherwise people will always have to build off of `develop` or `master`. 

# Changelog

- Adjust the `split` function so that binderhub supports branches named with `git-flow` conventions
- Minor change to a line that strips `.git` off the repo name with a more succinct + performant line.

# Notes

- I tried to test this locally by following directions in `CONTRIBUTING.md`, however I'm presently unable to get any repos to work. I think this may be related to #262 .

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,6 +52,12 @@ and run binderhub on the host system.
 
 All features should work, including building and launching.
 
+9. Running unit tests
+
+  ```bash
+  python setup.py test
+  ```
+
 ## Pure HTML / CSS / JS development
 
 If you do not want to set up minikube but just want to hack on the html / css / js,
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -21,6 +21,25 @@ from traitlets.config import LoggingConfigurable
 
 GITHUB_RATE_LIMIT = Gauge('binderhub_github_rate_limit_remaining', 'GitHub rate limit remaining')
 
+
+def tokenize_spec(spec):
+    """Tokenize a Git Spec into parts, error if spec invalid."""
+
+    spec_parts = spec.split('/', 2)  # allow ref to contain "/"
+    if len(spec_parts) != 3:
+        msg = 'Spec is not of the form "user/repo/ref", provided: "{spec}".'.format(spec=spec)
+        if len(spec_parts) == 2 and spec_parts[-1] != 'master':
+            msg += ' Did you mean "{spec}/master"?'.format(spec=spec)
+        raise ValueError(msg)
+
+    return spec_parts
+
+def strip_suffix(text, suffix):
+    if text.endswith(suffix):
+        text = text[:-(len(suffix))]
+    return text
+
+
 class RepoProvider(LoggingConfigurable):
     """Base class for a repo provider"""
     name = Unicode(
@@ -119,25 +138,6 @@ class GitHubRepoProvider(RepoProvider):
         super().__init__(*args, **kwargs)
         self.user, self.repo, self.unresolved_ref = self.process_spec(self.spec)
         self.repo = strip_suffix(self.repo, ".git")
-
-    @staticmethod
-    def process_spec(spec):
-        """Tokenizes a Github Spec into parts, error if spec invalid."""
-
-        spec_parts = spec.split('/', 2)  # allow ref to contain "/"
-        if len(spec_parts) != 3:
-            msg = 'Spec is not of the form "user/repo/ref", provided: "{spec}".'.format(spec=spec)
-            if len(spec_parts) == 2 and spec_parts[-1] != 'master':
-                msg += ' Did you mean "{spec}/master"?'.format(spec=spec)
-            raise ValueError(msg)
-
-        return spec_parts
-
-    @staticmethod
-    def strip_suffix(text, suffix):
-        if text.endswith(suffix):
-            text = text[:-(len(suffix))]
-        return text
 
     def get_repo_url(self):
         return "https://github.com/{user}/{repo}".format(user=self.user, repo=self.repo)
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -117,16 +117,27 @@ class GitHubRepoProvider(RepoProvider):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        spec_parts = self.spec.split('/', 2)  # allow ref to contain "/"
+        self.user, self.repo, self.unresolved_ref = self.process_spec(self.spec)
+        self.repo = strip_suffix(self.repo, ".git")
+
+    @staticmethod
+    def process_spec(spec):
+        """Tokenizes a Github Spec into parts, error if spec invalid."""
+
+        spec_parts = spec.split('/', 2)  # allow ref to contain "/"
         if len(spec_parts) != 3:
-            msg = 'Spec is not of the form "user/repo/ref", provided: "{spec}".'.format(spec=self.spec)
+            msg = 'Spec is not of the form "user/repo/ref", provided: "{spec}".'.format(spec=spec)
             if len(spec_parts) == 2 and spec_parts[-1] != 'master':
-                msg += ' Did you mean "{spec}/master"?'.format(spec=self.spec)
+                msg += ' Did you mean "{spec}/master"?'.format(spec=spec)
             raise ValueError(msg)
 
-        self.user, self.repo, self.unresolved_ref = spec_parts
-        if self.repo.endswith('.git'):
-            self.repo = self.repo[:-4]  # Strip .git suffix
+        return spec_parts
+
+    @staticmethod
+    def strip_suffix(text, suffix):
+        if text.endswith(suffix):
+            text = text[:-(len(suffix))]
+        return text
 
     def get_repo_url(self):
         return "https://github.com/{user}/{repo}".format(user=self.user, repo=self.repo)
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -117,7 +117,7 @@ class GitHubRepoProvider(RepoProvider):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        spec_parts = self.spec.split('/')
+        spec_parts = self.spec.split('/', 2)  # allow ref to contain "/"
         if len(spec_parts) != 3:
             msg = 'Spec is not of the form "user/repo/ref", provided: "{spec}".'.format(spec=self.spec)
             if len(spec_parts) == 2 and spec_parts[-1] != 'master':
@@ -126,7 +126,7 @@ class GitHubRepoProvider(RepoProvider):
 
         self.user, self.repo, self.unresolved_ref = spec_parts
         if self.repo.endswith('.git'):
-            self.repo = self.repo[:len(self.repo) - 4]
+            self.repo = self.repo[:-4]  # Strip .git suffix
 
     def get_repo_url(self):
         return "https://github.com/{user}/{repo}".format(user=self.user, repo=self.repo)
--- a/setup.py
+++ b/setup.py
@@ -1,20 +1,33 @@
 from setuptools import setup, find_packages
 
+install_requirements = [
+    'kubernetes==3.*',
+    'tornado',
+    'traitlets',
+    'docker',
+    'jinja2',
+    'prometheus_client'
+]
+
+setup_requirements = [
+    'pytest-runner'
+]
+
+test_requirements = [
+    'pytest'
+]
+
 setup(
     name='binderhub',
-    version='0.1',
-    install_requires=[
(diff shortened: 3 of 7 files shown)


choldgraf:
this seems reasonable to me and tests are passing...will merge later today if nobody raises a concern!

betatim:
Can you add a test so that this feature doesn't get lost again in the future? Otherwise LGTM.

hydrosquall:
@betatim sure, should I start a new folder for python unit tests, or should this be nestled somewhere in the `testing/` folder (which presently seems to mainly have config files)?

betatim:
Ouf, I hadn't realised there were no tests what so ever! 😁 https://imgur.com/gallery/8ECZb8L

I'd add tests in `binderhub/tests` and use pytest. https://github.com/jupyterhub/jupyterhub/tree/master/jupyterhub/tests has some examples for dealing with tornado and mocks, etc.

It might be worth to create a new PR to set all this up as it is probably 100% more work than this actual PR.

hydrosquall:
@betatim : I set up some basic tests, and hooked it up to `setup.py`. I also added basic documentation on how to run tests. 

Writing these tests inspired me to do some slight refactoring so these functions can be tested without involving the other parts of the `GithubRepoProvider` class :)

Let me know if you'd like to see further changes!

choldgraf:
wow, these are nice changes, thanks @hydrosquall ! I'll take a closer pass later today or tomorrow, though if @betatim has the time to take a look and thinks it's OK, I'm happy to merge

setup.py
```diff
@@ -1,20 +1,29 @@
...
-    version='0.1',
-    install_requires=[
-        'kubernetes==3.*',
-        'tornado',
-        'traitlets',
-        'docker',
-        'jinja2',
-        'prometheus_client'
-    ],
+    version='0.1.1',
```
betatim: is the change of version from 0.1 to 0.1.1 on purpose or typo? Just double checking

tests/repoproviders_test.py
```diff
@@ -0,0 +1,58 @@
...
+    ]
+)
+def test_spec_processing(spec, raw_user, raw_repo, raw_ref):
+    user, repo, unresolved_ref = GitHubRepoProvider.process_spec(spec)
+    assert raw_user == user
+    assert raw_repo == repo
+    assert raw_ref == unresolved_ref
+
+
+class TestSpecErrorHandling(TestCase):
```
betatim: What does using `TestCase` do for us? Asking because I've never had the need (or known) to use one before with pytest.

tests/repoproviders_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_too_short_spec(self):
+        spec = "nothing_to_split"
+        with self.assertRaisesRegexp(ValueError, "Spec is not of the form"):
```
betatim: Not sure if this is a technical or just stylistic choice: I try to stick as much as possible with the tools from pytest, so `with pytest.raises(...)` https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions instead of `unittest` things.

CONTRIBUTING.md
```diff
@@ -52,6 +52,12 @@ and run binderhub on the host system.
 
 All features should work, including building and launching.
 
+9. Running unit tests
+
+  ```bash
+  python setup.py test
```
betatim: Is there an advantage to running it like this over just `pytest`?

binderhub/repoproviders.py
```diff
@@ -117,16 +117,27 @@ def _default_auth(self):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        spec_parts = self.spec.split('/')
+        self.user, self.repo, self.unresolved_ref = self.process_spec(self.spec)
+        self.repo = strip_suffix(self.repo, ".git")
+
+    @staticmethod
+    def process_spec(spec):
```
betatim: Two thoughts: we could make these functions in the file instead of @staticmethod'ing them? And naming: `process_spec` makes me think "you mean a function that processes its argument? Crazy!" :) Maybe name it `_tokenize` or `_split`? Not sure `_tokenize` is much better :-/

binderhub/repoproviders.py
```diff
@@ -117,16 +117,27 @@ def _default_auth(self):
...
+        """Tokenizes a Github Spec into parts, error if spec invalid."""
```
betatim: -> "Tokenize a GitHub spec into user, repo, ref."? Mostly I am a sucker for "Tokenize" over "Tokenizes" (https://www.python.org/dev/peps/pep-0257/#specification)

binderhub/repoproviders.py
```diff
@@ -117,16 +117,27 @@ def _default_auth(self):
...
+                msg += ' Did you mean "{spec}/master"?'.format(spec=spec)
             raise ValueError(msg)
 
-        self.user, self.repo, self.unresolved_ref = spec_parts
-        if self.repo.endswith('.git'):
-            self.repo = self.repo[:len(self.repo) - 4]
+        return spec_parts
+
+    @staticmethod
+    def strip_suffix(text, suffix):
```
betatim: same comment about making it a private function in this file. I like the naming :)

setup.py
```diff
@@ -1,20 +1,29 @@
...
+setup_requirements = [
+    'pytest-runner'
+]
+
+test_requirements = [
+    'pytest'
+]
+
+install_requirements = [str(ir.req) for ir in parse_requirements(
+                        "requirements.txt", session=uuid.uuid1())]
```
betatim: Is there a good reason to keep `requirements.txt` which makes us have to do clever things to read it into `setup.py`? `pip install -e.` seems to work very well so even for local dev installs we don't need `requirements.txt` anymore. I'd vote for removing it.

binderhub/repoproviders.py
hydrosquall: @betatim I'm glad you raise this, as I had thought about this for a bit too. My thought is that `strip_suffix` definitely can be disconnected, but the `tokenize_spec` is tied to the way Github does things (I think, granted I haven't checked other VCS providers), so I would lean towards keeping that as a `staticmethod`.

I'll start by moving `strip_suffix` outside the class, and can move the other too if you feel strongly :)

hydrosquall: Also, appreciate the constructive critique about naming

CONTRIBUTING.md
hydrosquall: This way, it should install anything under `test_dependencies` before running the tests. However, it could easily be substituted for a combination of `pip install pytest; py.test -v tests`

betatim:
Thanks for all the work on setting up the testing!! My comments are a mix of nitpicks, opinions and selfish questions to make me smarter.

In the end the one thing I would really like is that typing `pytest` in the top level directory just works. I also have a slight preference for having `binderhub/binderhub/tests/` instead of `binderhub/tests/` (just realised my previous comment was ambiguous).

binderhub/repoproviders.py
betatim: We could reuse the spec-splitting for GitLab I reckon, if that makes a difference to you ;)

tests/repoproviders_test.py
hydrosquall: The main reason for my usage of `unittest.TestCase` was because I'm fond of the `assert.RaisesRegexp` method, from my quick reading of `pytest` it only allows for strict-equality-matching the error message.

I haven't used `pytest`'s assertions styles much, but I can convert if you'd like since I defer to you guys (the core project maintainers) for setting testing style standards :)

tests/repoproviders_test.py
hydrosquall: I use it mainly because I like `assertRaisesRegexp`, but it could go away if that gets replaced with pytest.raises.

In more complex situations, it also gives you a convenient place to define `setup` and `teardown` methods which will automatically run before and after a group of tests.

setup.py
hydrosquall: I bumped it intentionally, but I can change it back if necessary. I usually try to version bump when I file PRs that change some functionality in case people are depending on the existing behavior.

binderhub/repoproviders.py
hydrosquall: Ah, good catch

setup.py
hydrosquall: Yea, this is a good question. I guess I added one because most python modules I've seen have one, and because having one enables (free!) dependency version analysis tools like https://pyup.io to work with the project. (On a related note, I would vote for pinning the requirements to particular version numbers at some point, but thought that might merit a separate PR).

hydrosquall:
@betatim : ah, it turns out that just typing `pytest` in the top level directory works without needing the `/tests` argument, since pytest supports having the tests be in the `binderhub` directory, or a sibling of `binderhub` just as well. Would you still like to move the tests to be under binderhub? https://docs.pytest.org/en/latest/goodpractices.html

mariusvniekerk:
We may want to move branch into a query param?
Gitlab for example allows nested namespaces (eg `main/project/myrepo`) so relying on the number of slashes is not going to work.

CONTRIBUTING.md
merwok: Won’t using test_requires cause the dependencies to be installed as eggs dumped in the current dir, instead of regular package installation in your virtualenv?  IMO it’s best to have a test-requirements file and use tox to abstract virtualenv creation, deps installation and test runner options.

setup.py
merwok: This is an anti-pattern that will break in the next pip release.  pip doesn’t have a stable Python API, so using this kind of code is risky (and will be broken on purpose in pip 10).  It’s also confusing and a bit backward: setup.py was defined first (and can contain open requirements, i.e. without too strict version constraints, to maximize compatibility with other libraries), requirement files came up after (and these are useful with pinned versions to define known good sets for repeatable installs).

Links:
https://caremad.io/posts/2013/07/setup-vs-requirement/
http://nvie.com/posts/better-package-management/

A pretty great solution these days is to have unversioned dependencies in setup.py and use pip-tools to generate a requirements file with exact versions.

(Test dependencies don’t have to be in setup.py, they can be in their own requirements file.)

setup.py
minrk: since we haven't made a releas of binder yet, this should probably be `0.1.0dev`.

CONTRIBUTING.md
minrk: yes, test_requires and setup_requires can't be relied upon thanks to setuptools. A simple `test-requirements.txt` would be simpler and more reliable.

I strongly prefer to *avoid* the overhead and indirection of of tox, though.

minrk:
@mariusvniekerk for non-github providers, that may be necessary. One of the important things is that the *only* thing Binder itself encodes is that the first part (`gh`) registers the provider. The provider is responsible for interpreting the entire rest of the URL. If a URL path is ambiguous for a given provider (in that a corresponding git URL cannot be constructed via API requests), then additional info may be needed. I suspect the query params are not being passed at this time, though.

betatim:
Thanks for all the work and feeding my nitpick obsession :)

choldgraf:
@mariusvniekerk put another way, for something like GitLab we'd build a specific URL parser for it that'd be triggered by, e.g., `mybinder.org/v2/gl/myrepostructure/`.

@hydrosquall thanks for going above and beyond with this PR and satisfying @betatim's sick desire for testing suites! ;-)

CONTRIBUTING.md
merwok: Well the good thing with tox is that it builds on top of your existing settings and toolings.  In my projects I run commands such as flake8 or pytest directly most of the time (and they find their options in tox.ini), and only to make sure things are right in a clean isolated environment I call tox.

setup.py
hydrosquall: @merwok , I appreciate the background and links! I'll dive deeper into this later :)

merwok: This is a busy week but maybe the week after I could send you a PR!  Or feel free to tag me in the meantime.

PR #265, opened 2017-11-11
Use recommended template for creating docker config

thefotios:
This updates how the `config.json` in `secret.yaml` gets created to match `helms` "tips and tricks" example found [here](https://github.com/kubernetes/helm/blob/master/docs/charts_tips_and_tricks.md#creating-image-pull-secrets).

This may be related to #262 and the recent changes to the `secrets.yaml`. I have yet to test this, but posting it for @hydrosquall and others who may be running into the same issue.

### DIFF
--- a/helm-chart/binderhub/templates/secret.yaml
+++ b/helm-chart/binderhub/templates/secret.yaml
@@ -5,7 +5,7 @@ metadata:
 type: Opaque
 data:
   {{ if .Values.registry.enabled -}}
-  config.json: {{ b64enc (printf "{\"auths\": { \"https://%s\": { \"email\": \"not@val.id\", \"auth\": \"%s\" } } }" .Values.registry.host (b64enc (printf "%s:%s" .Values.registry.username .Values.registry.password) ) ) | quote }}
+  config.json: {{ template "imagePullSecret" . }}
   {{- end }}
   binder.hub-token: {{ .Values.jupyterhub.hub.services.binder.apiToken | b64enc | quote }}
   {{ if .Values.github.clientId -}}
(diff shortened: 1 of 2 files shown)


choldgraf:
@minrk and @yuvipanda I think you two have a better intuition on this one than I do!

minrk:
Seems AOK to me, but I'll give @yuvipanda last word to merge.

helm-chart/binderhub/templates/secret.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 type: Opaque
 data:
   {{ if .Values.registry.enabled -}}
-  config.json: {{ b64enc (printf "{\"auths\": { \"https://%s\": { \"email\": \"not@val.id\", \"auth\": \"%s\" } } }" .Values.registry.host (b64enc (printf "%s:%s" .Values.registry.username .Values.registry.password) ) ) | quote }}
+  config.json: {{ template "imagePullSecret" . }}
```
yuvipanda: Can you add a  `| quote` here?

yuvipanda:
Minor nit, can merge otherwise. Thank you for the patch!

yuvipanda:
if you want, I'm happy to amend + merge myself

thefotios:
I'll add a quote for safety, although it shouldn't be necessary since YAML *should* treat the `base64` data as a string.

yuvipanda:
Thank you very much for the patch! \o/

choldgraf:
LGTM! Thanks for the addition!!!

minrk:
Deployed to staging and this didn't work. I'll try to do some local testing.

PR #266, opened 2017-11-11
Add support for generic and gitlab git repositories.

mariusvniekerk:
Design
-------

This works by instead of passing a `{user}/{repo}/{branch}` spec we instead pass the 
full url to a git repo with a fragment as denoting the sha we wish to build

Todo
-----
- [x] Add tests
- [x] Discuss spec proposal
- [x] Enfore that the spec's fragment is actually a SHA1

Out of scope
-------------
Credential propagation.  All repos used in this way must be publicly accessible for now.
UI?

### DIFF
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -8,6 +8,7 @@ control services and providers.
 from datetime import timedelta
 import json
 import os
+import posixpath
 import time
 import urllib.parse
 
@@ -105,6 +106,12 @@ class GitRepoProvider(RepoProvider):
 
     def get_repo_url(self):
         return self.repo
+
+    def get_build_slug(self):
+        parsed = urllib.parse.urlparse(self.spec)
+        parts = [parsed.netloc]
+        parts.extend(parsed.path.split(posixpath.sep))
+        return '-'.join(parts)
 
 
 class GitHubRepoProvider(RepoProvider):
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -11,6 +11,7 @@ import os
 import posixpath
 import time
 import urllib.parse
+import re
 
 from prometheus_client import Gauge
 
@@ -22,7 +23,7 @@ from traitlets import Dict, Unicode, default
 from traitlets.config import LoggingConfigurable
 
 GITHUB_RATE_LIMIT = Gauge('binderhub_github_rate_limit_remaining', 'GitHub rate limit remaining')
-
+SHA1_PATTERN = re.compile(r'[0-9a-f]{40}')
 
 def tokenize_spec(spec):
     """Tokenize a GitHub-style spec into parts, error if spec invalid."""
@@ -35,6 +36,21 @@ def tokenize_spec(spec):
         raise ValueError(msg)
 
     return spec_parts
+
+
+def extract_string_from_arguments(arguments, key):
+    val = arguments.get(key)
+    if val:
+        if len(val) != 1:
+            raise ValueError("`{key}` can be specified only once!".format(key=key))
+        return val[0].decode('utf8').lower()
+    return None
+
+
+def sha1_validate(sha1):
+    if not SHA1_PATTERN.match(sha1):
+        raise ValueError("resolved_ref is not a valid sha1 hexadecimal hash")
+
 
 def strip_suffix(text, suffix):
     if text.endswith(suffix):
@@ -53,6 +69,12 @@ class RepoProvider(LoggingConfigurable):
     spec = Unicode(
         help="""
         The spec for this builder to parse
+        """
+    )
+
+    arguments = Dict(
+        help="""
+        Query parameters passed to the provider
         """
     )
 
@@ -84,21 +106,26 @@ class FakeProvider(RepoProvider):
         return '{user}-{repo}'.format(user='Rick', repo='Morty')
 
 
-class GitRepoProvider(RepoProvider):
-    """Bare bones git repo provider. Users must provide a resolve_ref in the following form
+class BasicGitRepoProvider(RepoProvider):
+    """Bare bones git repo provider.
 
-    http://hostname.com/path/repo_path#sha1
+    Users must provide a spec and has to specify a resolved_ref as a query parameter.
+
+    eg:
+    /build/git/https%3A//git.example.com/ns/repo.git?resolved_ref=a01ce557f868d7060a6099bd3c2d739734dcaf15
     """
 
-    name = Unicode("git")
+    name = Unicode("BasicGit")
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        parsed = urllib.parse.urlparse(self.spec)
-        if not parsed.fragment:
-            raise ValueError("spec used for raw git api must include the sha1 as a fragment")
-        self.resolved_ref = parsed.fragment
-        self.repo = urllib.parse.urlunparse(parsed._replace(fragment=''))
+        self.repo = self.spec
+
+        resolved_ref = extract_string_from_arguments(self.arguments, 'resolved_ref')
+        if not resolved_ref:
+            raise ValueError("`resolved_ref` must be specified as a query parameter for the basic git provider")
+        sha1_validate(resolved_ref)
+        self.resolved_ref = resolved_ref
 
     @gen.coroutine
     def get_resolved_ref(self):
@@ -110,8 +137,102 @@ class GitRepoProvider(RepoProvider):
     def get_build_slug(self):
         parsed = urllib.parse.urlparse(self.spec)
         parts = [parsed.netloc]
-        parts.extend(parsed.path.split(posixpath.sep))
-        return '-'.join(parts)
+        parts.extend(parsed.path.strip('/').split('/'))
+        # escape the name and replace dashes with something else.
+        return '-'.join(p.replace('-', '_-') for p in parts)
+
+
+class GitlabRepoProvider(RepoProvider):
+    """Simple gitlab provider.
+
+    Users must provide a spec and has to specify a resolved_ref as a query parameter.
+
+    eg:
+    /build/gl/user/namespace/repo?unresolved_ref=master
+    /build/gl/user/repo?unresolved_ref=a01ce557f868d7060a6099bd3c2d739734dcaf15
+    """
+
+    name = Unicode('GitLab')
+    hostname = Unicode('gitlab.com')
+
+    access_token = Unicode(config=True,
+        help="""Gitlab OAuth2 access token for authentication with the Gitlab API
+
+        For use with client_secret.
+        Loaded from GITHUB_ACCESS_TOKEN env by default.
+        """
+    )
+    @default('access_token')
+    def _oath_token_default(self):
+        return os.getenv('GITHUB_ACCESS_TOKEN', '')
+
+    private_token = Unicode(config=True,
+        help="""Gitlab private token for authentication with the Gitlab API
+
+        Loaded from GITLAB_PRIVATE_TOKEN env by default.
+        """
+    )
+    @default('private_token')
+    def _private_token_default(self):
+        return os.getenv('GITLAB_PRIVATE_TOKEN', '')
+
+    auth = Dict(
+        help="""Auth parameters for the GitHub API access
+
+        Populated from client_id, client_secret, access_token.
+    """
+    )
+    @default('auth')
+    def _default_auth(self):
+        auth = {}
+        for key in ('access_token', 'private_token'):
+            value = getattr(self, key)
+            if value:
+                auth[key] = value
+        return auth
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.namespace = self.spec
+        self.unresolved_ref = extract_string_from_arguments(self.arguments, 'unresolved_ref')
+        # TODO assertions
+        if self.unresolved_ref is None:
+            raise ValueError("A unresolved ref is required")
+
+    @gen.coroutine
+    def get_resolved_ref(self):
+        if hasattr(self, 'resolved_ref'):
+            return self.resolved_ref
+
+        namespace = urllib.parse.quote(self.namespace, safe='')
+        client = AsyncHTTPClient()
+        api_url = "https://{hostname}/api/v4/projects/{namespace}/repository/commits/{ref}".format(
+            namespace=namespace, ref=urllib.parse.quote(self.unresolved_ref, safe=''), hostname=self.hostname
+        )
+        self.log.debug("Fetching %s", api_url)
+
+        if self.auth:
+            # Add auth params. After logging!
+            api_url = url_concat(api_url, self.auth)
+
+        try:
+            resp = yield client.fetch(api_url, user_agent="BinderHub")
+        except HTTPError as e:
+            if e.code == 404:
+                return None
+            else:
+                raise
+
+        ref_info = json.loads(resp.body.decode('utf-8'))
+        self.resolved_ref = ref_info['id']
+        return self.resolved_ref
+
+    def get_build_slug(self):
+        # escape the name and replace dashes with something else.
+        return '-'.join(p.replace('-', '_-') for p in self.namespace.split('/'))
+
+    def get_repo_url(self):
+        return "https://{hostname}/{namespace}.git".format(hostname=self.hostname, namespace=self.namespace)
 
 
 class GitHubRepoProvider(RepoProvider):
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -8,7 +8,6 @@ control services and providers.
 from datetime import timedelta
 import json
 import os
-import posixpath
 import time
 import urllib.parse
 import re
@@ -25,6 +24,7 @@ from traitlets.config import LoggingConfigurable
 GITHUB_RATE_LIMIT = Gauge('binderhub_github_rate_limit_remaining', 'GitHub rate limit remaining')
 SHA1_PATTERN = re.compile(r'[0-9a-f]{40}')
 
+
 def tokenize_spec(spec):
     """Tokenize a GitHub-style spec into parts, error if spec invalid."""
 
@@ -36,20 +36,6 @@ def tokenize_spec(spec):
         raise ValueError(msg)
 
     return spec_parts
-
-
-def extract_string_from_arguments(arguments, key):
-    val = arguments.get(key)
-    if val:
-        if len(val) != 1:
-            raise ValueError("`{key}` can be specified only once!".format(key=key))
-        return val[0].decode('utf8').lower()
-    return None
-
-
-def sha1_validate(sha1):
-    if not SHA1_PATTERN.match(sha1):
-        raise ValueError("resolved_ref is not a valid sha1 hexadecimal hash")
 
 
 def strip_suffix(text, suffix):
@@ -72,12 +58,6 @@ class RepoProvider(LoggingConfigurable):
         """
     )
 
-    arguments = Dict(
-        help="""
-        Query parameters passed to the provider
-        """
-    )
-
     unresolved_ref = Unicode()
 
 
@@ -90,6 +70,11 @@ class RepoProvider(LoggingConfigurable):
 
     def get_build_slug(self):
         raise NotImplementedError("Must be overriden in the child class")
+
+    @staticmethod
+    def sha1_validate(sha1):
+        if not SHA1_PATTERN.match(sha1):
+            raise ValueError("resolved_ref is not a valid sha1 hexadecimal hash")
 
 
 class FakeProvider(RepoProvider):
@@ -106,7 +91,7 @@ class FakeProvider(RepoProvider):
         return '{user}-{repo}'.format(user='Rick', repo='Morty')
 
 
-class BasicGitRepoProvider(RepoProvider):
+class GitRepoProvider(RepoProvider):
     """Bare bones git repo provider.
 
     Users must provide a spec and has to specify a resolved_ref as a query parameter.
@@ -115,16 +100,15 @@ class BasicGitRepoProvider(RepoProvider):
     /build/git/https%3A//git.example.com/ns/repo.git?resolved_ref=a01ce557f868d7060a6099bd3c2d739734dcaf15
     """
 
-    name = Unicode("BasicGit")
+    name = Unicode("Git")
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.repo = self.spec
-
-        resolved_ref = extract_string_from_arguments(self.arguments, 'resolved_ref')
+        url, resolved_ref = self.spec.rsplit('/', 1)
+        self.repo = urllib.parse.unquote(url)
         if not resolved_ref:
             raise ValueError("`resolved_ref` must be specified as a query parameter for the basic git provider")
-        sha1_validate(resolved_ref)
+        self.sha1_validate(resolved_ref)
         self.resolved_ref = resolved_ref
 
     @gen.coroutine
@@ -135,21 +119,22 @@ class BasicGitRepoProvider(RepoProvider):
         return self.repo
 
     def get_build_slug(self):
-        parsed = urllib.parse.urlparse(self.spec)
+        parsed = urllib.parse.urlparse(self.repo)
         parts = [parsed.netloc]
         parts.extend(parsed.path.strip('/').split('/'))
         # escape the name and replace dashes with something else.
         return '-'.join(p.replace('-', '_-') for p in parts)
 
 
-class GitlabRepoProvider(RepoProvider):
-    """Simple gitlab provider.
+class GitLabRepoProvider(RepoProvider):
+    """GitLab provider.
 
     Users must provide a spec and has to specify a resolved_ref as a query parameter.
 
+    /build/gl/<url-escaped-namespace>/<unresolved_ref>
+
     eg:
-    /build/gl/user/namespace/repo?unresolved_ref=master
-    /build/gl/user/repo?unresolved_ref=a01ce557f868d7060a6099bd3c2d739734dcaf15
+    /build/gl/group%2Fproject%2Frepo/master
     """
 
     name = Unicode('GitLab')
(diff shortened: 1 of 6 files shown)


betatim:
So we would have URLs like: https://staging.mybinder.org/v2/git/github.com/betatim/appmode?filepath=%2Fapps%2Fexample_app.ipynb#someshahere ?? Maybe we could keep the format as before except that it does not support branch/tag names and instead only SHA1s?

mariusvniekerk:
Well almost like that, there is some more ugliness with adding in the scheme etc, which all needs to be escaped in any case.  We could also stick more things into query params instead of into URL components.
Given that under say, gitlab there can be namespaces that are nested as `a/b/repo` and users can have branches with a `/` in them we probably have to since otherwise it would be impossible to disambiguate

Currently, this is what I'm passing is something like this:
```python
repo_kind = 'git'
url = 'https://git.the.companyname.com/namespace/notebooks.git#f6f3fcab2b3b5358ebce0c944d1eb1aba400156a'
url = binder_url + ‘/build/{repo_kind}/{spec}’.format(repo_kind=repo_kind, spec=urllib.parse,quote(spec))
```

testing/minikube/binderhub_config.py
```diff
@@ -1,5 +1,7 @@
 # config file for testing with minikube-config.yaml
 
-c.BinderHub.hub_url = 'http://192.168.99.100:30123'
+import subprocess
+minikube_ip = subprocess.check_output(['minikube', 'ip']).decode('utf-8').strip()
+c.BinderHub.hub_url = 'http://{}:30123'.format(minikube_ip)
```
minrk: 👍

binderhub/repoproviders.py
```diff
@@ -64,6 +66,36 @@ def get_build_slug(self):
...
+    """
+
+    name = Unicode("git")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        parsed = urllib.parse.urlparse(self.spec)
+        if not parsed.fragment:
+            raise ValueError("spec used for raw git api must include the sha1 as a fragment")
+        self.resolved_ref = parsed.fragment
```
minrk: There should probably be some validation here, unless the errors out of a failed build to a bad ref are good enough. Plus, the fragment may not be reliable because the server doesn't tend to get that part of the URL, so it would have to be escaped somehow.

binderhub/repoproviders.py
```diff
@@ -64,6 +66,36 @@ def get_build_slug(self):
...
+    def get_resolved_ref(self):
+        return self.resolved_ref
+
+    def get_repo_url(self):
+        return self.repo
+
+    def get_build_slug(self):
+        parsed = urllib.parse.urlparse(self.spec)
+        parts = [parsed.netloc]
+        parts.extend(parsed.path.split(posixpath.sep))
```
minrk: Since it's a URL, we can use `/` explicitly, we don't need to rely on the fact that URL happen to have the same separator as posix.

binderhub/repoproviders.py
```diff
@@ -64,6 +66,36 @@ def get_build_slug(self):
...
+        return '-'.join(parts)
```
minrk: There's an ambiguity here, because replacing `/` with `-` gives `foo-bar/baz` the same slug as `foo/bar-baz`. Some form of escaping would need to be done to avoid collisions.

binderhub/repoproviders.py
```diff
@@ -64,6 +66,36 @@ def get_build_slug(self):
...
+    """Bare bones git repo provider. Users must provide a resolve_ref in the following form
+
+    http://hostname.com/path/repo_path#sha1
+    """
+
+    name = Unicode("git")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        parsed = urllib.parse.urlparse(self.spec)
```
minrk: I think there's going to need to be some help here, because a full URL in a URL path is going to be a little tricky, with escapes for `://`, etc. I think we will need to be a bit more explicit, such as specifying that the scheme is the first part, e.g.

    /v2/git/https/github.com/binder-examples/requirements.git@abcdef...

mariusvniekerk: yeah so currently i just escape everything and it works for me

binderhub/repoproviders.py
mariusvniekerk: yeah, i'm planning to add some basic checks to at least enforce that it is a sha1

mariusvniekerk:
@minrk @yuvipanda What are your thoughts on this?

binderhub/repoproviders.py
```diff
@@ -34,6 +37,21 @@ def tokenize_spec(spec):
 
     return spec_parts
 
+
+def extract_string_from_arguments(arguments, key):
```
yuvipanda: Add a docstring here? It's unclear what this does as is :)

binderhub/repoproviders.py
```diff
@@ -34,6 +37,21 @@ def tokenize_spec(spec):
...
+    val = arguments.get(key)
+    if val:
+        if len(val) != 1:
+            raise ValueError("`{key}` can be specified only once!".format(key=key))
+        return val[0].decode('utf8').lower()
+    return None
+
+
+def sha1_validate(sha1):
```
yuvipanda: This should probably be part of the BasicGitProvider class?

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
         return '{user}-{repo}'.format(user='Rick', repo='Morty')
 
 
+class BasicGitRepoProvider(RepoProvider):
```
yuvipanda: Why not just call this 'GitRepoProvider'?

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
...
+    """Bare bones git repo provider.
+
+    Users must provide a spec and has to specify a resolved_ref as a query parameter.
+
+    eg:
+    /build/git/https%3A//git.example.com/ns/repo.git?resolved_ref=a01ce557f868d7060a6099bd3c2d739734dcaf15
```
yuvipanda: Can we instead just use:

```
/build/git/<url-encoded-url>/<full-resolved-sha>
```

to keep consistent with the github provider? While not codified anywhere, I'd like us to reserve query parameters only for runtime parameters (that can change after build) rather than build parameters.

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
...
+
+    @gen.coroutine
+    def get_resolved_ref(self):
+        return self.resolved_ref
+
+    def get_repo_url(self):
+        return self.repo
+
+    def get_build_slug(self):
+        parsed = urllib.parse.urlparse(self.spec)
```
yuvipanda: I think there's a 63character limit here. Check what the github provider does?

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
...
+        return '-'.join(p.replace('-', '_-') for p in parts)
+
+
+class GitlabRepoProvider(RepoProvider):
+    """Simple gitlab provider.
+
+    Users must provide a spec and has to specify a resolved_ref as a query parameter.
+
+    eg:
+    /build/gl/user/namespace/repo?unresolved_ref=master
```
yuvipanda: Same comment about URL structure from earlier - keep build parameters part of the URL structure.

choldgraf:
(I also agree this is an awesome addition...will try to take a look at the code in the next few days though if @yuvipanda gives a +1 them I'm happy w/ it)

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
         return '{user}-{repo}'.format(user='Rick', repo='Morty')
 
 
+class BasicGitRepoProvider(RepoProvider):
+    """Bare bones git repo provider.
```
willingc: `"""Base git repo provider"""`

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
...
+
+    Users must provide a spec and has to specify a resolved_ref as a query parameter.
+
+    eg:
+    /build/git/https%3A//git.example.com/ns/repo.git?resolved_ref=a01ce557f868d7060a6099bd3c2d739734dcaf15
+    """
+
+    name = Unicode("BasicGit")
```
willingc: BaseGit

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
...
+
+    def get_build_slug(self):
+        parsed = urllib.parse.urlparse(self.spec)
+        parts = [parsed.netloc]
+        parts.extend(parsed.path.strip('/').split('/'))
+        # escape the name and replace dashes with something else.
+        return '-'.join(p.replace('-', '_-') for p in parts)
+
+
+class GitlabRepoProvider(RepoProvider):
```
willingc: Should this derive from `BaseGitRepoProvider`?  If not, I'm not sure that the generic GitRepoProvider class above is necessary.

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
...
+    """
+
+    name = Unicode('GitLab')
+    hostname = Unicode('gitlab.com')
+
+    access_token = Unicode(config=True,
+        help="""Gitlab OAuth2 access token for authentication with the Gitlab API
+
+        For use with client_secret.
+        Loaded from GITHUB_ACCESS_TOKEN env by default.
```
willingc: GITLAB ?

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
...
+        """
+    )
+    @default('access_token')
+    def _oath_token_default(self):
+        return os.getenv('GITHUB_ACCESS_TOKEN', '')
```
willingc: GITLAB ?

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
...
+
+        Loaded from GITLAB_PRIVATE_TOKEN env by default.
+        """
+    )
+    @default('private_token')
+    def _private_token_default(self):
+        return os.getenv('GITLAB_PRIVATE_TOKEN', '')
+
+    auth = Dict(
+        help="""Auth parameters for the GitHub API access
```
willingc: GitLab

binderhub/repoproviders.py
```diff
@@ -83,6 +106,135 @@ def get_build_slug(self):
...
+    def get_build_slug(self):
+        parsed = urllib.parse.urlparse(self.spec)
+        parts = [parsed.netloc]
+        parts.extend(parsed.path.strip('/').split('/'))
+        # escape the name and replace dashes with something else.
+        return '-'.join(p.replace('-', '_-') for p in parts)
+
+
+class GitlabRepoProvider(RepoProvider):
+    """Simple gitlab provider.
```
willingc: `"""GitLab Repo Provider`

mariusvniekerk:
@willingc I dont think that the deployed binderhub instance should support the `GitProvider`, however this is a useful thing for people deploying this behind a firewall and wanting to have it hit internal repositories if they aren't using gitlab/github.

mariusvniekerk:
@yuvipanda ready for round 3

binderhub/repoproviders.py
```diff
@@ -83,6 +91,138 @@ def get_build_slug(self):
...
+
+    def get_repo_url(self):
+        return self.repo
+
+    def get_build_slug(self):
+        parsed = urllib.parse.urlparse(self.repo)
+        parts = [parsed.netloc]
+        parts.extend(parsed.path.strip('/').split('/'))
+        # escape the name and replace dashes with something else.
+        return '-'.join(p.replace('-', '_-') for p in parts)
```
choldgraf: isn't this going to result in `--` in the build slug since we're also adding the `-` in the call to `replace`? Just making sure that's what we want

binderhub/repoproviders.py
```diff
@@ -83,6 +91,138 @@ def get_build_slug(self):
...
+    repo.  Users must provide a spec that matches the following form.
+
+    <url-escaped-namespace>/<unresolved_ref>
+
+    eg:
+    group%2Fproject%2Frepo/master
+    """
+
+    name = Unicode('GitLab')
+    hostname = Unicode('gitlab.com')
```
choldgraf: do gitlab repos ever live in other places than `gitlab.com`? Sorry if that's a dumb question just wanna make sure

binderhub/repoproviders.py
```diff
@@ -83,6 +91,138 @@ def get_build_slug(self):
         return '{user}-{repo}'.format(user='Rick', repo='Morty')
 
 
+class GitRepoProvider(RepoProvider):
+    """Bare bones git repo provider.
+
```
willingc: > this is a useful thing for people deploying this behind a firewall and wanting to have it hit internal repositories if they aren't using gitlab/github.

Thanks @mariusvniekerk. Let's add the use case in the docstring.

binderhub/repoproviders.py
choldgraf: (e.g. in the same way that `mybinder.org` is just a certain BinderHub deployment, is gitlab.com just a certain deployment of something?

binderhub/repoproviders.py
```diff
@@ -83,6 +91,138 @@ def get_build_slug(self):
...
+    access_token = Unicode(config=True,
+        help="""Gitlab OAuth2 access token for authentication with the Gitlab API
+
+        For use with client_secret.
+        Loaded from GITLAB_ACCESS_TOKEN env by default.
+        """
+    )
+    @default('access_token')
+    def _access_token_default(self):
+        return os.getenv('GITLABV_ACCESS_TOKEN', '')
```
willingc: Typo. Remove extra V

choldgraf: intentional `GITLABV_`? it seems to just be `GITLAB_` elsewhere

binderhub/repoproviders.py
```diff
@@ -83,6 +91,138 @@ def get_build_slug(self):
...
+
+    private_token = Unicode(config=True,
+        help="""Gitlab private token for authentication with the Gitlab API
+
+        Loaded from GITLAB_PRIVATE_TOKEN env by default.
+        """
+    )
+    @default('private_token')
+    def _private_token_default(self):
+        return os.getenv('GITLAB_PRIVATE_TOKEN', '')
```
choldgraf: @minrk could something similar to this be used for private github repos? On the assumption that the person deploying the binderhub would have those tokens for a specific private repo already on the machine from which they deploy?

binderhub/repoproviders.py
```diff
@@ -83,6 +91,138 @@ def get_build_slug(self):
...
+                auth[key] = value
+        return auth
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        quoted_namespace, unresolved_ref = self.spec.split('/', 1)
+        self.namespace = urllib.parse.unquote(quoted_namespace)
+        self.unresolved_ref = urllib.parse.unquote(unresolved_ref)
+        if self.unresolved_ref is None:
+            raise ValueError("A unresolved ref is required")
```
choldgraf: ~A~ An 

(sorry for the nitpick :-) )

binderhub/repoproviders.py
```diff
@@ -83,6 +91,138 @@ def get_build_slug(self):
...
+
+    eg:
+    group%2Fproject%2Frepo/master
+    """
+
+    name = Unicode('GitLab')
+    hostname = Unicode('gitlab.com')
+
+    access_token = Unicode(config=True,
+        help="""Gitlab OAuth2 access token for authentication with the Gitlab API
```
willingc: Let's use GitLab throughout this class to be consistent with the GitLab service's spelling.

examples/binder-api.py
```diff
@@ -14,6 +14,8 @@
 
 import requests
 
+#build_binder("https://github.com/binder-examples/demo", binder_url='http://192.168.99.100:31078/')
```
choldgraf: is this an accidental include?

binderhub/repoproviders.py
```diff
@@ -83,6 +91,138 @@ def get_build_slug(self):
         return '{user}-{repo}'.format(user='Rick', repo='Morty')
 
 
+class GitRepoProvider(RepoProvider):
+    """Bare bones git repo provider.
+
+    Users must provide a spec of the following form.
+
+    <url-escaped-namespace>/<resolved_ref>
```
choldgraf: The `GitHubRepoProvider` class doesn't require url escaping right? It just passes it to the tokenizer here:

https://github.com/mariusvniekerk/binderhub/blob/5675b40f9c8082b62a9cb5e3d749d23d28320706/binderhub/repoproviders.py#L28

would this mean that if people are using gitlab Binder, they'd need to escape characters etc, but if using github they don't?

examples/binder-api.py
willingc: Let's remove this or add a line above commenting it.

choldgraf:
As I mention before, this is a really useful addition so many thanks @mariusvniekerk ! A few comments in there to supplement whatever others suggest! Some questions are quite simple so if they're addressed feel free to delete!

binderhub/repoproviders.py
```diff
@@ -83,6 +91,138 @@ def get_build_slug(self):
...
+
+        namespace = urllib.parse.quote(self.namespace, safe='')
+        client = AsyncHTTPClient()
+        api_url = "https://{hostname}/api/v4/projects/{namespace}/repository/commits/{ref}".format(
+            namespace=namespace, ref=urllib.parse.quote(self.unresolved_ref, safe=''), hostname=self.hostname
+        )
+        self.log.debug("Fetching %s", api_url)
+
+        if self.auth:
+            # Add auth params. After logging!
```
willingc: 👍

binderhub/repoproviders.py
mariusvniekerk: I would probably say that we can be smart when generating these spec urls in the application javascript.  Right now we already convert full github urls to the gh/<spec> format.

binderhub/repoproviders.py
mariusvniekerk: it will only have a `--` if you have a part that ends with a `-` which is weirdish

binderhub/repoproviders.py
mariusvniekerk: for gitlab enterprise installs sure.

binderhub/repoproviders.py
yuvipanda: Yup, something similar to this needs to happen to support private repos on github :)

binderhub/repoproviders.py
yuvipanda: @choldgraf @mariusvniekerk this is for *git*, right? For git, we have to allow arbitrary URLs, so we need to do escaping. GitHub and GitLab both have more specific sturctures that ensure we don't need to do any escaping, since their username/reponame is already URL safe.

I think this should say `url-escaped-url` rather than `url-escaped-namespace`?

binderhub/repoproviders.py
mariusvniekerk: shamelessly copied from the github one

yuvipanda:
@willingc ty for the reviews!

> It feels a bit like we are scattering functionality among a number of classes (RepoProvider, GitHubRepoProvider, GitLabRepoProvider, BasicGitRepoProvider). Do we need a BasicGitRepoProvider since both GitLabRepoProvider and GitHubRepoProvider both inherit from RepoProvider?

I think it's ok right now, since I don't think there's any code duplication yet. Introducing another layer of inheritance is not worth it right now IMO. My usual cutoff for adding a layer to the inheritance chart is when you have *3* classes that share almost copy-pasted code. I don't think any of our providers have copy-pasted code yet, so we're fine I think.

yuvipanda:
@mariusvniekerk said:
> @willingc I dont think that the deployed binderhub instance should support the GitProvider, however this is a useful thing for people deploying this behind a firewall and wanting to have it hit internal repositories if they aren't using gitlab/github.

I actually think we *should* deploy this to the public mybinder.org instance too! It satisfies our technical criteria (fully cache-able by enforcing SHAs) and allows users from OSS projects that don't use gitlab or github or a supported provider yet to use it.

binderhub/repoproviders.py
```diff
@@ -100,6 +100,9 @@ class GitRepoProvider(RepoProvider):
 
     eg:
     https%3A%2F%2Fgithub.com%2Fjupyterhub%2Fzero-to-jupyterhub-k8s/f7f3ff6d1bf708bdc12e5f10e18b2a90a4795603
+
+    This provider is typically used if you are deploying binderhub yourself and you require access to repositories that
```
willingc: 🥇

mariusvniekerk:
@yuvipanda Good point.  There are likely several notebooks currently stuck on git repos that live on academic servers.  Binders for all!

binderhub/repoproviders.py
betatim: I think as a default using gitlab.com is good, then people who deploy binderhub for themselves can modify the hostname for their favourite gitlab instance

binderhub/repoproviders.py
mariusvniekerk: nothing :D

mariusvniekerk:
So the only remaining issue really is about long slug names.  This is a general problem and should probably be solved where we invoke repo2docker.  Not sure if we want to incorporate that in this PR

choldgraf:
IMO figuring out how to handle long build slugs doesn't need to be in this PR as it is something we'll need to think about for other providers as well, and I don't wanna block this useful contribution because of it. I opened an issue to talk about the slugs.

For this PR, how would long slugs be handled? Are they just cut off or are we just not handling them at all? I don't remember the details of the conversation but wasn't the problem that kubernetes would break if the slugs were too long?

examples/binder-api.py
```diff
@@ -63,7 +63,6 @@ def build_binder(repo,
         if evt.get('phase') == 'ready':
             url = "{url}?token={token}".format(**evt)
             print("Opening %s" % url)
-            webbrowser.open(url)
```
betatim: typo or on purpose? I quite liked this feature :-/

binderhub/repoproviders.py
betatim: This looks relevant https://github.com/jupyterhub/binderhub/blob/92972de4698cb26d2cfe8b28acddbfa524ea6896/binderhub/builder.py#L161 here we replace `_` with `-` again :-/ I need to think a bit/run some code to see what really ends up happening.

betatim:
Regarding long build slugs: where do you think they surface/cause a problem?

The only place I can think of is when they are used as pod names, which I think always goes through [this bit of code](https://github.com/jupyterhub/binderhub/blob/92972de4698cb26d2cfe8b28acddbfa524ea6896/binderhub/builder.py#L114-L120) where the slug is hashed and the short+hash version is used.

mariusvniekerk:
@betatim I believe docker images also have some limits on name length

examples/binder-api.py
mariusvniekerk: whoops

binderhub/repoproviders.py
```diff
@@ -83,6 +91,141 @@ def get_build_slug(self):
...
+    def get_resolved_ref(self):
+        return self.resolved_ref
+
+    def get_repo_url(self):
+        return self.repo
+
+    def get_build_slug(self):
+        parsed = urllib.parse.urlparse(self.repo)
+        parts = [parsed.netloc]
+        parts.extend(parsed.path.strip('/').split('/'))
```
yuvipanda: IMO, the right thing to do here is to simply return something like `url + hash`. The code in builder ensures that you are sized appropriately. Only counting netloc + path means that http / https will produce same image, which is not something we should do IMO.

binderhub/repoproviders.py
```diff
@@ -83,6 +91,141 @@ def get_build_slug(self):
...
+        try:
+            resp = yield client.fetch(api_url, user_agent="BinderHub")
+        except HTTPError as e:
+            if e.code == 404:
+                return None
+            else:
+                raise
+
+        ref_info = json.loads(resp.body.decode('utf-8'))
+        self.resolved_ref = ref_info['id']
```
yuvipanda: Will the ref_info always have 'id'? What if the given branch or commit / tag does not exist? Will gitlab return 404 for 'em?

mariusvniekerk: Yeah Gitlab 404s

binderhub/repoproviders.py
mariusvniekerk: Does the builder also escape things that are problematic?

mariusvniekerk:
I've been testing mostly the build endpoint directly

binderhub/repoproviders.py
yuvipanda: @mariusvniekerk hah, there's a TODO to make it so: https://github.com/jupyterhub/binderhub/blob/master/binderhub/builder.py#L92 but right now it doesn't.

I think we should fix that TODO by calling into escapism.escape from there so we can solve it centrally in one place though rather than per provider.

yuvipanda:
@mariusvniekerk can you tell me what you mean by that testing the build endpoint directly?

yuvipanda:
Ah, nvm me I understand what you meant now! I've been away from this too long lol

binderhub/repoproviders.py
yuvipanda: I made https://github.com/jupyterhub/binderhub/pull/298 for this!

yuvipanda:
@mariusvniekerk if we can iterate on https://github.com/jupyterhub/binderhub/pull/298 and simplify the build slugs here, I think this is ready to merge. We can work on the UI side of this in a different patch.

Excited to get this rolling!

choldgraf:
+1 to merge

willingc:
Congrats and thanks for the contribution @mariusvniekerk :cake:

choldgraf:
seriously! awesome contribution @mariusvniekerk !

betatim:
🦃 ! :)

PR #270, opened 2017-11-13
follow-up testing setup

### LINKED ISSUE #268: Setup Unit Testing for Binderhub
See this comment: https://github.com/jupyterhub/binderhub/pull/264#issuecomment-343750132

Reference for mocking + examples:

- https://github.com/jupyterhub/jupyterhub/tree/master/jupyterhub/tests

Additional tasks that could be separate PRs or part of this

- Including code coverage statistics in CI
- Add running unit tests to the `setup.py` + `travis.yml`

minrk:
follow-up to #264 

closes #268

- enable codecov
- add tests for basic entrypoint help
- restore install_requires in setup.py
- remove setup/test_requires in favor of `dev-requirements.txt` and `pytest`
- run tests prior to build, to avoid building when tests don't work
- install `git-crypt` from binary while at it
- make version 0.1.0dev, and accessible as `binderhub.__version__`


CONTRIBUTING.md
```diff
@@ -40,9 +40,9 @@ and run binderhub on the host system.
             --namespace=binder \
             -f testing/minikube/jupyterhub-helm-config.yaml
 
-6. Install binderhub:
+6. Install binderhub and its development requirements:
 
-        python3 -m pip install -e .
+        python3 -m pip install -e . -r dev-requirements.txt
```
betatim: 👍

.travis.yml
```diff
@@ -4,25 +4,28 @@ python:
 services:
 - docker
 install:
-- pip install --no-cache-dir ruamel.yaml
```
betatim: For my education: do you know what this used to do/was good for?

betatim:
LGTM.

.travis.yml
minrk: I know you deleted it, but for completeness:

This package is used in `build.py` for rewriting `values.yaml`. It has been moved to `dev-requirements`, along with the testing dependencies.

minrk:
Thanks!

choldgraf:
LGTM as well! Only question I have is related to PEP. Do we want to add a pep checker here? I'm like 40% in favor of the idea but wanted to float it in case somebody else feels strongly about it. I'm happy either way so feel free to merge

minrk:
Thanks! I'm not a big fan of style enforcement. But if we pick it up, I think the only solution is to pick an auto formatter. Style errors without a one-line auto-fix is pretty contributor-hostile in my experience.

choldgraf:
Sounds good! I agree.

--

minrk:
Thanks! I'll deploy this once #272 is merged, which fixes a regression that I missed in the original PR.

betatim:
-0 on pep8 violations making travis red. Recently added https://github.com/OrkoHunter/pep8speaks to scikit-optimize to see if that works out as a nice (not too noisy) way to encourage PEP8ing without a human reviewer. I do think some basic PEP8ing should be followed, it makes the code a lot easier to read, especially for people who are involved in several code bases (that otherwise would all have their own little idiosyncrasies).

choldgraf:
I'm +1 on finding technical ways to *encourage* PEP8 but not necessarily
*require* PEP8 :-)

Maybe we can write a Mr Meeseecks bot that scans for PEP8 and, if fails,
posts a picture of my mother making an "I'm disappointed in you" face.
--

PR #272, opened 2017-11-13
fix (and test!) GitHub provider resolution

minrk:
fixes failure introduced in #264


binderhub/repoproviders.py
```diff
@@ -136,7 +136,7 @@ def _default_auth(self):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.user, self.repo, self.unresolved_ref = self.process_spec(self.spec)
```
betatim: So the bug was calling a method that didn't exist anymore?

minrk: Yup. Last commit moved method to top-level function, but didn't keep up with the rename.

minrk:
Thanks! Will deploy to staging as soon as it's uploaded.

binderhub/repoproviders.py
hydrosquall: Oops, my bad! Sorry about that, thanks for catching this

minrk: @hydrosquall no problem at all! That's why we have tests (now). Thanks for your help setting that up.

PR #273, opened 2017-11-13
Adjust order in which we mention how to specify dependencies

betatim:
People might think Dockerfile is preferred because it is listed first
when the opposite is true.

### DIFF
--- a/binderhub/templates/index.html
+++ b/binderhub/templates/index.html
@@ -131,7 +131,7 @@
         <span class="point point-red">2</span>
       </div>
       <div class="col-md-7 front">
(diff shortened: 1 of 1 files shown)


hydrosquall:
Ah, I was under the impression from my previous reading of the docs (and from my experience this weekend) that if both a `requirements.txt` and `Dockerfile` are present in the root directory, that the Dockerfile would take precedence. Does "preferred" refer to "the practice that will minimize people's encounters with bugs", or to "which file will actually be read first"?

betatim:
`Dockerfile`s will always win. Preferred means we think people shouldn't have to use a `Dockerfile` to get what they want. A corollary to that is that I think people shouldn't use a `Dockerfile` hence mentioning env.yml and reqs.txt before the Dockerfile in this page.

PR #275, opened 2017-11-14
support urlpath in LegacyURLHandler and add testing bootstraps

minrk:
previous binder supported this and there are links in the wild relying on it

### DIFF
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -2,6 +2,7 @@
 Main handler classes for requests
 """
 from tornado import web
+from tornado.httputil import url_concat
 from tornado.log import app_log
 
 from .base import BaseHandler
@@ -49,6 +50,8 @@ class ParameterizedMainHandler(BaseHandler):
 class LegacyRedirectHandler(BaseHandler):
     """Redirect handler from legacy Binder"""
 
-    def get(self, user, repo):
+    def get(self, user, repo, urlpath=None):
         url = '/v2/gh/{user}/{repo}/master'.format(user=user, repo=repo)
+        if urlpath:
+            url = url_concat(url, dict(urlpath=urlpath))
         self.redirect(url)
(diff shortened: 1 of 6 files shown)


binderhub/main.py
```diff
@@ -49,6 +50,8 @@ def get(self, provider_prefix, spec):
 class LegacyRedirectHandler(BaseHandler):
     """Redirect handler from legacy Binder"""
 
-    def get(self, user, repo):
+    def get(self, user, repo, urlpath=None):
         url = '/v2/gh/{user}/{repo}/master'.format(user=user, repo=repo)
+        if url_path:
```
betatim: nit: `if url_path is not None:`

betatim:
LGTM. Test?

I can add a test as well if you point me to an example in jhub's tests that tests something like this.

minrk:
testing handlers is tricky without setting up a binderhub instance, which is tricky without a running kubernetes. I'm working on getting kubernetes deployed on Travis so we can start doing that.

betatim:
Ok. I don't think this feature should be held up by missing test, just thought it would be a good excuse to add some of the scaffolding so we can easily add tests for handlers in the future.

minrk:
I should have a running app for the tests by the time I'm done with #282

minrk:
While working on #282, I realized that I didn't need kubernetes and jupyterhub running to run tests with the app, so I moved that over here and added coverage for this handler.

I have to run, but hopefully these will pass now after chasing down some weird delays

minrk:
This ought to be good to go, then I can rebase and finish #282 to get tests running with a fully armed and operational binderhub setup.

choldgraf:
Looks like a great addition to me! I'll let @betatim give the final merge go-ahead on this.

willingc:
Thanks @minrk.

choldgraf:
Ah I was hoping after @minrk 's "fully armed and operational" comment, somebody would say "you may fire when ready" before merging :-)

willingc:
@choldgraf :smile:

@betatim I am leaving for you to merge into mybinder-deploy in case you see anything else you would like.

PR #277, opened 2017-11-14
remove quote in imagePullSecret template

minrk:
the additional quote produced incorrect structure for some reason I don't fully understand.

### DIFF
--- a/helm-chart/binderhub/templates/secret.yaml
+++ b/helm-chart/binderhub/templates/secret.yaml
@@ -5,7 +5,7 @@ metadata:
 type: Opaque
 data:
   {{ if .Values.registry.enabled -}}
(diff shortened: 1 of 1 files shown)


minrk:
this is needed to restore working deployments

yuvipanda:
That's strange! But if it works for you locally good enough for me. Can you paste what the structure is with and without quotes just for my curiousity?

minrk:
I'm not quite sure how to do that, but the error that resulted was:

```
Error: UPGRADE FAILED: render error in "mybinder/charts/binderhub/templates/secret.yaml": template: mybinder/charts/binderhub/templates/_helpers.tpl:19:62: executing "imagePullSecret" at <.Values.registry.hos...>: can't evaluate field Values in type string
```

PR #281, opened 2017-11-15
add missing https in auth helper

minrk:
typo introduced in #265 had registry as 'gcr.io' instead of 'https://gcr.io'

registry.host is just the hostname, not the full registry URL


minrk:
I was able to manually deploy this patch to staging and verified that it gets things back running.

Testing if a PR like this fixes an issue is a real bear without. Not super sure what to do about that, though. Even my plan for testing on Travis with minikube wouldn't have caught this because the registry wouldn't be enabled.

I guess we can try unittests for the helm chart where we run a verification over the config context, rather than launch binderhub.

minrk:
Thanks! With this, I think staging will be back to working order, and then we can deploy all the way to prod.

PR #282, opened 2017-11-15
run minikube on travis

minrk:
The goal of this PR is to have two things:

1. tests that verify that `helm install` works with the helm chart
2. a running binderhub as part of the Python test suite, so we can run proper tests

- [x] install minikube, helm
- [x] smoke-test `helm install binderhub`
- [x] verify that `helm install binderhub` produces a running binderhub
- [x] launch jupyterhub via helm
- [x] launch binderhub for tests, hooked up to jupyterhub

Something I don't plan to do, but might be nice in terms of integration coverage:

-  Run tests against binderhub launched via the helm chart,
   rather than hub via helm with local binder.
   The reason being error reporting and coverage reporting,
   which are greatly improved when the server is running in-process
   with the tests.

### DIFF
--- a/binderhub/tests/conftest.py
+++ b/binderhub/tests/conftest.py
@@ -18,8 +18,7 @@ here = os.path.abspath(os.path.dirname(__file__))
 root = os.path.join(here, os.pardir, os.pardir)
 minikube_testing_config = os.path.join(root, 'testing', 'minikube', 'binderhub_config.py')
 
-BUILD_NAMESPACE = os.environ.get('BINDER_TEST_BUILD_NAMESPACE') or 'binder-test'
-HUB_NAMESPACE = os.environ.get('BINDER_TEST_HUB_NAMESPACE')or 'binder-test-hub'
+TEST_NAMESPACE = os.environ.get('BINDER_TEST_NAMESPACE') or 'binder-test'
 KUBERNETES_AVAILABLE = False
 
 ON_TRAVIS = os.environ.get('TRAVIS')
@@ -32,7 +31,7 @@ REMOTE_BINDER = bool(BINDER_URL)
 def _binderhub_config():
     """separate from app fixture to load config and check for hub only once"""
     cfg = PyFileConfigLoader(minikube_testing_config).load_config()
-    cfg.BinderHub.build_namespace = BUILD_NAMESPACE
+    cfg.BinderHub.build_namespace = TEST_NAMESPACE
     if ON_TRAVIS:
         cfg.BinderHub.hub_url = cfg.BinderHub.hub_url.replace('192.168.99.100', '127.0.0.1')
     global KUBERNETES_AVAILABLE
@@ -155,7 +154,7 @@ def cleanup_binder_pods(request):
     if not KUBERNETES_AVAILABLE:
         # kubernetes not available, nothing to do
         return
-    cleanup = lambda : cleanup_pods(HUB_NAMESPACE,
+    cleanup = lambda : cleanup_pods(TEST_NAMESPACE,
                                     {'component': 'singleuser-server'})
     cleanup()
     request.addfinalizer(cleanup)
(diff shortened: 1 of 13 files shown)


minrk:
OK! This finally works. This got a bit complicated, so I split most of travis.yml into a few scripts

1. `ci/install.sh` does all of the kubectl/minikube/helm/etc. installation
2. `testing/minikube/install-hub` script will install the hub chart with minikube config and wait for it to come up
3. TEST=main sets up jupyterhub with helm and runs binderhub locally (better errors, debugging, mocking, coverage, etc.)
4. TEST=helm builds and installs the binderhub helm chart, and then runs a few tests (marked `remote`) against it.

If you have minikube and helm set up, you can run these tests locally with:

```bash
pip install -e . -r dev-requirements.txt
# only once, installs the hub for testing
./testing/minikube/install-hub
pytest
```

If you want to test a running binderhub, even a public one, you can use BINDER_TEST_URL, e.g. to run the tests against staging:

```bash
export ASYNC_TEST_TIMEOUT=15
BINDER_TEST_URL=https://staging.mybinder.org pytest -vsx -m remote
```

binderhub/tests/conftest.py
```diff
@@ -13,23 +18,56 @@
 root = os.path.join(here, os.pardir, os.pardir)
 minikube_testing_config = os.path.join(root, 'testing', 'minikube', 'binderhub_config.py')
 
+TEST_NAMESPACE = os.environ.get('BINDER_TEST_NAMESPACE') or 'binder-test'
+KUBERNETES_AVAILABLE = False
+
+ON_TRAVIS = os.environ.get('TRAVIS')
+
+BINDER_URL = os.environ.get('BINDER_TEST_URL')
+REMOTE_BINDER = bool(BINDER_URL)
```
willingc: Is this to check if Binder is on Travis or not?

binderhub/tests/conftest.py
```diff
@@ -13,23 +18,56 @@
...
+
+
 @pytest.fixture(scope='session')
 def _binderhub_config():
+    """separate from app fixture to load config and check for hub only once"""
```
willingc: Perhaps: 
```
Fixture, primarily for Travis testing, to load config and check for hub only once. 

This is separate from the 'app' test fixture.
```

binderhub/tests/conftest.py
```diff
@@ -55,16 +115,103 @@ def cleanup():
     bhub.url = 'http://127.0.0.1:%i' % bhub.port
     return bhub
 
+def cleanup_pods(namespace, labels):
+    """Cleanup pods in a namespace that match the given labels"""
+    kube = kubernetes.client.CoreV1Api()
+    def get_pods():
```
willingc: Perhaps add docstring or comment "return list of pods matching given labels"

binderhub/tests/conftest.py
```diff
@@ -55,16 +115,103 @@ def cleanup():
...
+            )
+        except kubernetes.client.rest.ApiException as e:
+            # ignore 404, 409: already gone
+            if e.status not in (404, 409):
+                raise
+    while pods:
+        print(f"Waiting for {len(pods)} binder pods to exit")
+        time.sleep(1)
+        pods = get_pods()
+
```
willingc: Perhaps add a print for all pods have been deleted.

minrk:
@willingc thanks for the review!

binderhub/tests/conftest.py
```diff
@@ -13,23 +18,62 @@
...
+    """Load the binderhub configuration
+
+    Currently separate from the app fixture
+    so that it can have a different scope (only once per session).
+    """
     cfg = PyFileConfigLoader(minikube_testing_config).load_config()
-    cfg.BinderHub.build_namespace = 'binder-test'
+    cfg.BinderHub.build_namespace = TEST_NAMESPACE
+    if ON_TRAVIS:
+        cfg.BinderHub.hub_url = cfg.BinderHub.hub_url.replace('192.168.99.100', '127.0.0.1')
```
betatim: https://github.com/jupyterhub/binderhub/blob/1951fe70957f7e1c4472e5cd64873d6ffb49dc15/testing/minikube/binderhub_config.py now asks minikube for the IP, on travis will we end up in the except clause of that or not?

betatim: I think this will work most of the time and fail rarely so if we are in a hurry merge first, fix later.

minrk: This line becomes obsolete thanks to [another PR](https://github.com/jupyterhub/binderhub/pull/266/files#diff-8f47f4b62ab8938d604ab0e0ffa506bd). I haven't rebased this PR since merging that one, so I will remove it in a subsequent one.

betatim:
This is a big PR and to properly review it would take me a lot of time because too many different things that I am not fluent enough in. However travis is currently happy, so I think we should merge this because having it is better than not having it and we can fix things up in the test suite while we are using it.

betatim:
Thanks for building all this infrastructure. Writing small simple tests is what keeps the project on track but someone needs to do all the heavy lifting first.

PR #288, opened 2017-11-19
cleanup SVG logo file

Carreau:
Remove illustrator cruft for a way smaller logo


minrk:
Hm. The new SVG is invalid and can't be rendered. #295 should fix it by removing the remaining references to the removed illustrator content.

PR #289, opened 2017-11-20
Included evaluating minikube docker-env

drorata:
Following a discussion with @minrk; there's one missing step which is added in this PR


minrk:
Thanks!

PR #293, opened 2017-11-20
Allow configuring launch timeout

ventilooo:
Hi, feel free to comment and give me your feedback in order to complete #293 
:fox_face:


choldgraf:
Good addition to make it configurable! However I don't think we want to remove the exponentially-increasing functionality. I'm not sure what's the best way to both allow users to configure timeout times, but also do the exponential increase. @yuvipanda what do you think?

minrk:
I added a commit here to load the exponential_backoff utility function from JupyterHub. If you import that function, you can pass your new configurable timeout to it.

yuvipanda:
Thanks for your first PR, @ventilooo! @minrk's comment is right - now you can import the exponential_backoff function and use it instead of gen.sleep!

ventilooo:
@choldgraf 
Is that PR still valuable ?

minrk:
@ventilooo sorry, I missed that comment. Yes, it would be great if you wanted to finish this PR!

minrk:
@ventilooo an example of the exponential_backoff is [here](https://github.com/ding-c3/jupyterhub/blob/932689f2f8a324088171331a385fd4230b0b6ece/jupyterhub/utils.py#L143).

In this example, it should look something like

```python
async def check_server():
    resp = await self.api_request(
        'users/%s' % username,
        method='GET',
    )

    body = json.loads(resp.body.decode('utf-8'))
    if body['server']:
        return True
    else:
        return False

await exponential_backoff(
    check_server,
    "Image %s for user %s took too long to launch" % (image, username),
    timeout=self.timeout,
)
```

Ping here if you'd like more guidance, or if you'd like someone else to finish the PR.

consideRatio:
Without any activity for more than two years, I'm closing this. It can certainly be reopened if this is still not implemented etc, which I don't know actually.

PR #294, opened 2017-11-21
bump jupyterhub chart

minrk:
get changes that should help reduce orphaned pods

### DIFF
--- a/helm-chart/binderhub/requirements.yaml
+++ b/helm-chart/binderhub/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: jupyterhub
-  version: "0.5.0-d1edb2e"
+  version: "0.5.0-7e2d5f3"
   repository: "https://jupyterhub.github.io/helm-chart"


yuvipanda:
This is unfortunately not enough. There has been a breaking change in the
Jupyterhub chart related to how ingresses are specified that needs changing
too. Can someone revert this for now? I'm on my phone...

choldgraf:
reverted just to be on the conservative side

willingc:
Out of curiosity...what was the breaking change on our side or k8s? If ours, how can I tell from the chart history.

yuvipanda:
@willingc it was broken in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/229 and I added it to the changelogs in - https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/249/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR42. I should've opened an issue here though, which I didn't :(

In the future, we should probably be writing the CHANGELOG as we go, which would help spot these things. Does that sound like an answer to the latter half of your question?

willingc:
Thanks @yuvipanda. It didn't occur to me to look at the still open changelog. I only cross scanned recently merged PRs. No worries on the issue opening.

Not sure re: CHANGELOG since it didn't occur to me to look there. Yet, it seems like a good idea.

PR #295, opened 2017-11-21
finish scrubbing illustrator from svg

minrk:
Previous #288 scrubbed illustrator content but not references to it, resulting in an invalid svg.

This should display the logo in browsers now.


Carreau:
Oops strange, worked here. Maybe cached versions  ?

PR #298, opened 2017-11-22
Escape characters in build slugs properly

### LINKED ISSUE #240: Prepend `build-` to build pod names
Makes it easier to find / debug.

yuvipanda:
We need to be DNS safe for use in kubernetes object names.

fixes #240


yuvipanda:
Done, @betatim!

PR #301, opened 2017-11-23
ensure build name is lowercase

minrk:
follow-up to #298

escapism can itself create uppercase escapes, so cast with `lower()` after escaping.

and ensure that errors in submit are at least logged

TODO: (for another PR because right now, this is blocking any builds from working at all)

- make sure the build fails if submit fails (it does not right now)


betatim:
LGTM. I won't have time till tonight to deploy this though. If you have time that would be good.

minrk:
Thanks! Deploy on its way.

PR #303, opened 2017-11-27
[MRG] truncate build slug after escaping

minrk:
long build names were getting too long because truncation happened before escaping, and the escapes increased the length of the string beyond the limit

- slugs are case-insensitive (uppercase are safe, cast to lowercase)
- only strings that come from providers are escaped (build_slug, ref)
- prefix, hash are assumed safe because they are internal

This should wait until #282 is merged so that builds can be tested.
I'll rebase with an update to the tests to ensure the build slug is long.

### DIFF
--- a/binderhub/tests/conftest.py
+++ b/binderhub/tests/conftest.py
@@ -221,7 +221,8 @@ def always_build(app, request):
     """
     if REMOTE_BINDER:
         return
-    session_id = b2a_hex(os.urandom(5)).decode('ascii')
+    # make it long to ensure we run into max build slug length
+    session_id = b2a_hex(os.urandom(16)).decode('ascii')
 
     def patch_provider(Provider):
         original_slug = Provider.get_build_slug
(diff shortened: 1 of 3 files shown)


betatim:
Should we add a test that has a build slug that is too long/add a comment to an existing test where the slug ends up being >63 to let future generations know that they shouldn't replace it with a repo that has a shorter name?

binderhub/tests/conftest.py
```diff
@@ -38,8 +38,6 @@ def _binderhub_config():
     """
     cfg = PyFileConfigLoader(minikube_testing_config).load_config()
     cfg.BinderHub.build_namespace = TEST_NAMESPACE
-    if ON_TRAVIS:
-        cfg.BinderHub.hub_url = cfg.BinderHub.hub_url.replace('192.168.99.100', '127.0.0.1')
```
betatim: :man_dancing:

binderhub/tests/conftest.py
```diff
@@ -221,7 +219,8 @@ def always_build(app, request):
     """
```
betatim: This is pretty clever, now that I better understand what it does/how to use it :)

betatim: Normally I dislike test related stuff that makes me go "aha, this is clever", but yeah. I might make a PR later that doesn't patch `get_build_slug` and somehow has a very long slug.

betatim:
LGTM

PR #305, opened 2017-11-28
update jupyterhub chart with ingress changes

minrk:
gets latest fixes, including init-container support and kubespawner fix that should cleanup pods after failed launch

There doesn't appear to be anything ingress-related to do here, but there will be on mybinder.org-deploy when we get there. <del>It proved cleaner to consolidate ingress at the binder chart level than at the deployment level.</del>

- [x] The nodePort config is going to need https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/296 for the helm tests to pass, I think.

### DIFF
--- a/ci/test-main
+++ b/ci/test-main
@@ -4,5 +4,7 @@
 
 set -ex
 ./testing/minikube/install-hub
+# DEBUG: give the hub a chance to wake up
+sleep 5
 pytest -vsx --cov binderhub
 helm delete --purge binder-test-hub
--- a/helm-chart/binderhub/requirements.yaml
+++ b/helm-chart/binderhub/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: jupyterhub
-  version: "0.5.0-d1edb2e"
+  version: "0.5.0-7e9089b"
   repository: "https://jupyterhub.github.io/helm-chart"
--- a/helm-chart/binderhub/templates/ingress.yaml
+++ b/helm-chart/binderhub/templates/ingress.yaml
@@ -14,17 +14,21 @@ metadata:
   {{- end }}
 spec:
   rules:
-    - http:
+    {{- range $host := .Values.ingress.hosts }}
+    - host: {{ $host }}
+      http:
         paths:
           - path: /
             backend:
               serviceName: binder
               servicePort: 8585
-      host: {{ .Values.ingress.host }}
+    {{- end }}
 {{- if and .Values.ingress.https.enabled (eq .Values.ingress.https.type "kube-lego") }}
   tls:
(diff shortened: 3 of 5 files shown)


minrk:
https://github.com/jupyterhub/mybinder.org-deploy/pull/169 demonstrates what this looks like in the end.  Binder's ingress routes for both binder and hub, so the Hub chart's ingresses are now ignored.

The problem I ran into with the new internal ingress: since it uses a different ingress controller (`ingress.class = jupyterhub-proxy-tls` as opposed to `ingress.class = nginx`), it seems to require that binder and hub.binder get separate load-balancer IPs. The only way I could figure to get a single public IP was to use a single ingress controller for hub and binder, which means either hooking binder's ingress into the jupyterhub chart's internal ingress. I'm not sure if what I did was the best solution, but it does work.

minrk:
@yuvipanda that's what I was trying to refer to above. The way the new internal ingress-controller works appears to require that the two separate ingress-controllers each get public IPs. Is there a way around that, so that one ingress-controller routes to another, only-internal ingress-controller that doesn't get a LoadBalancerIP?

The only way I was able to get it working using the internal ingress was to expose two LoadBalancerIPs, one for the hgub and another for binder, which seems less than ideal.

minrk:
Or is forcing separate IP addresses for the Hub and Binder by design?

yuvipanda:
@minrk which ingress controller are you talking about? The one that's bundled with the proxy? Or the one we install via [requirements.yaml](https://github.com/jupyterhub/mybinder.org-deploy/blob/staging/mybinder/requirements.yaml)? The former is an internal implementation detail and it won't pick up the ingress set up by https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/values.yaml#L143, since the proxy-ingress only looks for ingresses with the specific custom annotation. The ingress created by https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/values.yaml#L143 will only be picked up by the ingress controller we install via requirements.yaml, which is what we are currently using. That should allow you to replicate the same set up we have now.

It's confusing because we have an ingress controller embedded in the proxy now, but that is mostly immaterial in this case since we aren't actually using its features (primarily, the automatic HTTPS). The ingress controller embedded in the proxy is only a  convenience for people who don't want to set up an external requirements.yaml based ingress provider. For people who do (like the mybinder.org deploy) we can mostly ignore the embedded ingress controller.

Or I've totally misunderstood what you're saying :)

minrk:
I reverted the ingress changes. The only thing here is fixing `host` in the template to match the `hosts` in the values.yaml and in other places.

minrk:
Thanks @willing! And thanks @yuvipanda for helping me figure out the right thing to do with the ingresses.

PR #252, opened 2017-11-08
New Copy and Paste UI for Build Dropdown

hzarea:
Hi everyone, following up on Issue #132, here is an addition I designed for two new sections: URL copy and paste and Badge copy and paste. I've also created new icons for this. 

I haven't added the copy and paste features to these new sections, because I'm not sure how to work with that code, but here are the empty sections that we will be able to work with to replace the dropdown on the launch button above. 

screen shot 2017-11-08 at 12 22 47 pm

I kept all of the current code the same, I just added to it with this new section!

PR #306, opened 2017-11-28
Add support for github gists

mariusvniekerk:


binderhub/repoproviders.py
```diff
@@ -356,3 +364,54 @@ def get_resolved_ref(self):
...
+    <username>/<gist-id>[/<ref>]
+    """
+
+    def __init__(self, *args, **kwargs):
+        # We dont need to initialize entirely the same as github
+        super(RepoProvider, self).__init__(*args, **kwargs)
+        parts = self.spec.split('/')
+        self.user, self.gist_id, *_ = parts
+        if len(parts) > 2:
+            self.unresolved_ref = parts[1]
```
choldgraf: what's this usecase? A versioned gist? And just double-checking that here `unresolved_ref` would be set to `gist_id`...is that correct?

yuvipanda: gists are just git repositories, and so we should version them the same too!

binderhub/repoproviders.py
```diff
@@ -356,3 +364,54 @@ def get_resolved_ref(self):
...
     def get_build_slug(self):
         return '{user}-{repo}'.format(user=self.user, repo=self.repo)
+
+
+class GitHubGistRepoProvider(GitHubRepoProvider):
+    """GitHub gist provider.
+
+    Users must provide a spec that matches the following form (similar to github)
+
+    <username>/<gist-id>[/<ref>]
```
yuvipanda: Does this mean the ref is optional? I'd ideally like it to not be optional in the URL (we can ofc construct it in the JS). Git cloning says the default branch is master, so we should require a ref and go through the usual process of resolving it in the backend IMO.

binderhub/repoproviders.py
```diff
@@ -356,3 +364,54 @@ def get_resolved_ref(self):
...
+        else:
+            if self.unresolved_ref not in all_versions:
+                return None
+            else:
+                self.resolved_ref = self.unresolved_ref
+
+        return self.resolved_ref
+
+    def get_build_slug(self):
+        return self.gist_id
```
yuvipanda: This will coalesce multiple revisions of gists building at the same time, which isn't what we want. It should involve both the id and the resolved ref IMO.

binderhub/repoproviders.py
mariusvniekerk: that should be `parts[2]` doh

binderhub/repoproviders.py
mariusvniekerk: so there is no branch called `master` as far as the upstream gist repo is concerned.

binderhub/repoproviders.py
mariusvniekerk: we dont use the resolved ref for any of the other build slugs, that gets added later

binderhub/repoproviders.py
mariusvniekerk: we can use `master` as a sentinal value meaning latest in the history chain if we want

yuvipanda: Ah, hmm. I did a git clone and got to 'master'. Perhaps that's git clone's behavior? Am unsure. Am +1 using master as sentinel if that's not the case.

binderhub/repoproviders.py
yuvipanda: Ah, whoops. Sorry about that.

binderhub/app.py
```diff
@@ -19,7 +19,7 @@
 from .launcher import Launcher
 from .registry import DockerRegistry
 from .main import MainHandler, ParameterizedMainHandler, LegacyRedirectHandler
-from .repoproviders import GitHubRepoProvider, GitRepoProvider, GitLabRepoProvider
+from .repoproviders import GitHubRepoProvider, GitRepoProvider, GitLabRepoProvider, GitHubGistRepoProvider
```
betatim: nitpick: can this be `GitHubGistProvider` or even `GistProvider`? I don't know anywhere except GitHub that has Gists.

betatim:
We should add a flag that lets the owner of the binderhub enable/disable private gist support.

mariusvniekerk:
@betatim done

binderhub/repoproviders.py
```diff
@@ -356,3 +364,68 @@ def get_resolved_ref(self):
...
+        self.log.debug("Fetching %s", api_url)
+
+        resp = yield self.github_api_request(api_url)
+        if resp is None:
+            return None
+
+        ref_info = json.loads(resp.body.decode('utf-8'))
+
+        if (not self.allow_secret_gist) and (not ref_info['public']):
+            raise ValueError("gist is marked as secret, this is unsupported")
```
betatim: -> "Gist is marked as secret, to enable secret Gist support set GistRepoProvider.allow_secret_gist = True"?

willingc: Let's add a comment above line 382 or in the docstring that calls out the default behavior too.

For ValueError message:

You seem to want to use a secret Gist, but do not have permission to do so. To enable secret Gist support, set (or have an administrator set) 'GistRepoProvider.allow_secret_gist = True'

mariusvniekerk:
@willingc @betatim updated

yuvipanda:
Sorry about the delay in merging! TYVM for the PR!

willingc:
Congrats on the PR being merged @mariusvniekerk. We appreciate the contribution. 👍

betatim:
🎉 

Should we deploy (and enable) this to mybinder.org?

choldgraf:
I think we should deploy this, test it out a bit, and open an issue to write documentation for it (+ gitlab + raw git repos)

yuvipanda:
Note that users can't actually use any of this yet since we don't have any UI for it.

choldgraf:
yeah I know, but to me that's a feature rather than a bug :-) IMO we should
plan to document and/or add UI around it in a couple of weeks once we've
had time to see the feature in action, WDYT?

yuvipanda:
what do you mean by see the feature in action? :) The only way to use them right now is to write your own JavaScript...

choldgraf:
oops - maybe I was mistaken, I thought this PR adds the URL entrypoint for gists, does it not?

yuvipanda:
Nope, we don't have URL endpoints for gist nor gitlab :)

willingc:
@yuvipanda does #322 cover the remaining work needed for gist, gitlab, and other support?

binderhub/app.py
```diff
@@ -165,6 +165,7 @@ def _add_slash(self, proposal):
     repo_providers = Dict(
         {
             'gh': GitHubRepoProvider,
+            'gist': GistRepoProvider,
```
betatim: Doesn't this add the endpoint?

yuvipanda: only for the API. If you look at main.py's ParameterizedMainHandler and templates/index.html, you'll see that the code simply dumps it into the appropriate textboxes and triggers 'launch' by fake-clicking the launch button. And if you look at index.js it's hardcoded to use gh....

betatim: Ah! I hadn't read the code closely enough and assumed that the `ParametrizedMainHandler` kicks off the build :)

yuvipanda: It's not that smart yet :)

Since this work needs to touch the js, should we finish up and deploy #252 first?

PR #308, opened 2017-11-28
Add an issue template with guidance on what goes where

betatim:
via https://github.com/jupyterhub/binder/pull/35


choldgraf:
assuming we're OK with this issue template as well since we've iterated over at the binder repo...though we can iterate on improvements if folks think it should be different for this repo. I like the idea of having the same template for each though

willingc:
Nice! Thanks.

PR #309, opened 2017-11-28
Documenting Need for GitHub token in Contributing.md

### LINKED ISSUE #291: Document setting up your GitHub credentials with a local dev install
@jzf2101 was setting up a BinderHub with someone at a sprint and they ran into GitHub rate limit issues on their local test. We should document how to set up GitHub credentials so that BinderHub uses them in a local test.

@yuvipanda @minrk have either of you figured this out? Feel free to post barebones steps in this issue and I can make a PR to `CONTRIBUTING.md`.

jzf2101:
Documenting @minrk 's notes from #291. Closes #291

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,10 +2,10 @@
 
 Welcome! As a [Jupyter](https://jupyter.org) project, we follow the [Jupyter contributor guide](https://jupyter.readthedocs.io/en/latest/contributor/content-contributor.html).
 
-## Minikube
-
 To develop binderhub, you can use a local installation of JupyterHub on minikube,
-and run binderhub on the host system.
+and run binderhub on the host system.  You will also need a GitHub token for testing.
+
+## Installation
 
 1. [Install Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) and start it: `minikube start`.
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
Is this required? I thought this was optional, and the code seems to tell me it is. I haven't needed one for testing at all...

willingc:
How are you setting up your local test environment? @jzf2101 found that a number of folks were getting rate limited by GitHub when running tests locally during the PyCon Canada sprints. There is a Friday sprint at PyData.

willingc:
@yuvipanda ^

yuvipanda:
@willingc by default the limit is per IP, so depending on the network setup at PyCon Canada the limit is shared by all users. So users will hit it very quickly. I used to run into these at various university networks all the time. 

For sprints at specific location, it makes sense to ask users to get a token. We should make that clear, though. Perhaps augment #309 to mention sprints / IPs?

yuvipanda:
You can find out the rate limit you'll run into by going to 'https://api.github.com/rate_limit'. If `rate.remaining` is close to zero / not close to `rate.limit` then you need the token. Otherwise you don't.

willingc:
Thanks for the IP limit info. I'll bet that is what they ran into. Sounds good.

PR #310, opened 2017-11-29
Clarifying need for github token

choldgraf:
This is a little patch to clarify #309


choldgraf:
@yuvipanda does this work?

yuvipanda:
+1, I like this better! What do you think, @willingc / @jzf2101?

PR #319, opened 2017-12-01
Run a secondary docker daemon for building images

yuvipanda:
Add a dind (docker-in-docker) daemonset for building images

We currently mount the host's /var/run/docker.sock into
repo2docker for it to build. This gives us a few advantages:

1. Only one docker daemon per node, and we do not have to
    control it!
2. Better cache utilization, especially since many times the
    images are run on the same hosts they're built on. Much better
    than running a docker daemon per build

However, it also has massive disadvantages:

1. There's no security boundaries, so we're one exploit away
    from our entire infra being compromised
2. We are stuck to the version of docker that our underlying
    provider (GKE?) is running. This changes much slower than
    docker's release process. We can't have new Dockerfile
    features, such as --chown to the COPY command
3. We churn through a lot of containers when building images,
    leading to leakage issues like https://github.com/jupyterhub/mybinder.org-deploy/issues/189
4. We can't do things like apply networkpolicy or use iptables
    to block outgoing traffic in build containers, since those
    are not managed by k8s

This patch adds a compromise, where we:

1. Run a second docker daemon just for building as a daemonset
2. /var/lib/docker for new daemon is in an emptyDir volume, so
    is cleaned out every time there's a daemon restart (not often!)
3. The docker.sock file (default at /var/run/dind/docker.sock) is
    exposed in the host filesystem, so other containers on the same
    node can mount it. This also allows us to control access to this
    via PodSecurityPolicy.
4. Upgrade to latest version of docker for building, since the
    version of docker that's supported by docker for running inside
    docker (https://hub.docker.com/r/library/docker/) does not include
    the version that's running on GKE (1.13.x)

We still are mounting /var/lib/docker.sock for local dev & what not,
but the helm chart by default will deploy the second docker daemon.

Note that there's some build_limit work that got entangled in this PR, mostly
because it was hard to avoid conflicts otherwise (and github doesn't have a 
'dependent PR' mechanism yet :(. It isn't actively used in this patch, but when
it does, it requires a repo2docker with https://github.com/jupyter/repo2docker/pull/159


betatim:
Should we make binderhub depend on r2d in `setup.py`? We'd have to finish off releasing r2d, and then make the dependency >= that version.

yuvipanda:
@betatim I think that wouldn't matter, since binderhub only launches repo2docker inside a container. I'd need to bump the default container image tag...

binderhub/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        If it has one of the suffixes, it is converted into the appropriate
+        pure byte value.
+        """
+        if isinstance(value, (int, float)):
+            return int(value)
+
+        try:
+            num = float(value[:-1])
+        except ValueError:
+            raise TraitError('{val} is not a valid memory specification. Must be an int or a string with suffix K, M, G, T'.format(val=value))
```
minrk: TraitError not imported?

binderhub/app.py
```diff
@@ -125,6 +142,17 @@ class BinderHub(Application):
         config=True
     )
 
+    docker_api_url = Unicode(
```
minrk: docker_api_url looks like it would accept `tcp://1.2.3.4...`, is that true? I assume tcp sockets cannot be mounted in the way this is handled. If that's the case, `docker_socket` might be clearer, since it doesn't support any of the other forms. Plus, if it's docker_api_url, it should be `unix:///var/...`, and probably called `docker_host`, since that's what docker itself calls it (matching the DOCKER_HOST env).

Rather than docker_api_url, can we respect docker's own environment variables like `DOCKER_HOST` without our own logic? That should get all of docker's tcp/etc. support, right?

I'm not 100% sure how this would work here, since we want to talk to the same docker in both binderhub and the build, right? And docker environment variables can be paths that would presumably need to be mounted.

yuvipanda: Unfortunately we can't simply pass DOCKER_HOST around, since we have to do a hostPath volume mount if it's a unix socket. Since we currently only support doing this for unix sockets, I'm going to validate that it's a unix:/// url instead.

binderhub/utils.py
yuvipanda: Fixed

yuvipanda:
This works great now form my testing.

yuvipanda:
I added a significant security method to this PR and retitled it, since doing 'dependent PRs' in GitHub is hard :(

minrk:
This is great! I've looked it over and everything seems good for 'real' cases where a registry is in use. The only missing piece is that running *without* a registry (e.g. minikube for local or travis testing) no longer works because builds and runs are no longer talking to the same docker instance.

There are three places where we look for docker images when there isn't a registry:

1. with [docker_client](https://github.com/jupyterhub/binderhub/blob/2ea22d99436743bc6b005bd1fe57fbda3cbe3bd1/binderhub/builder.py#L186) in binderhub itself, where we use the docker env (this is typically the node docker via tcp using `$(minikube docker-env)`)
2. kube uses the node docker directory when launching images
3. build uses node docker (okay) or dind (now separate)

We need to make sure that when we aren't using a registry, all three of those are either talking to the same docker or at least using the same image list. I suspect the simplest way to do this is to eliminate support for running without a registery. When a *public* registry is not in use, we could require a local registry, which is easily added with minikube via `minikube addon enable registry`. 

This would simplify some things, because we could eliminate a branch in the runtime code and make it a branch in the setup code (i.e. setup public registry or launch internal one). The alternative is figuring out some way to move images from dind to the node docker, but I'm not sure there's a simpler way to do that than just using a registry.

yuvipanda:
@minrk the default should still be to mount /var/lib/docker.sock into it, no? shouldn't that continue to work for minikube? This should be entirely optional right now.

I do agree that only supporting the use case with the registry is sensible long term.

minrk:
@yuvipanda that makes sense! It's failing because dind is enabled by default in the helm chart. I switched the default to disabled based on your comment.

PR #320, opened 2017-12-01
bump jupyterhub chart

minrk:
gets https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/302 preventing restarting the Hub from shutting down user pods


minrk:
bumping minikube seemed to fix a weird error on helm install (that was reproducible trying to rebuild master that succeeded yesterday). No idea why.

PR #329, opened 2017-12-07
add base url for entire app

bitnik:


betatim:
Hey, thanks for the contribution. Could you add a bit of explanation what issue this fixes and why you did what you did? This way someone from the future will have an easier time figuring out what happened here :)

bitnik:
Hi, sorry. 

I wanted to run both JupyterHub and BinderHub apps under same domain. Both apps are running on kubernetes with service type as `NodePort`. JupyterHub was already running under `/jupyter/`.  And I wanted to run BinderHub app under `/binder/`.

For JupyterHub it was easy to do that by specifying `base_url` paremeter in [config file](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/values.yaml#L4). Once it is specified, urls of all necessary handlers are updated in [app.py](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/app.py#L328) during [initialization](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/app.py#L811).

So in this commit I wanted to do same for BinderHub.

choldgraf:
this seems like a reasonable contrib to me. @yuvipanda or @minrk does this seem like the right way to go about it per how jupyterhub does things?

Carreau:
That seem reasonable to me. I was thinking that something similar might be necessary at some point. 

So +1. This is the kind of thing were obviously we'll miss a few spots and will take a few iterations.

minrk:
Yup, this is the right thing to do, and looks like a solid implementation! Among other things, this is one of the things we will need to run binderhub behind jupyterhub as a service, since it will need to run at `/services/binder`. It does have merge conflicts, so if @bitnik could rebase, I think we can merge.

bitnik:
@minrk rebase is done.

minrk:
Thanks!

willingc:
Thanks @bitnik for the contribution. :cake:

choldgraf:
way to go @bitnik ! TYVM!!!

PR #334, opened 2017-12-08
Fix overlap of build logs.

Carreau:
Every col-md-x need to be wrapped in a `.raw` as hey are floats left.
Though `.row`s add a -15px margin on each side.

So add `.row` to all the siblings divs, and compensate by +15px margin on
container.

-- 
screen shot 2017-12-08 at 22 58 02


choldgraf:
looks good to me - hitting launch works fine?

choldgraf:
just tried it and LGTM!

choldgraf:
this is now deployed!

PR #336, opened 2017-12-09
Minimal CSS fixes.

Carreau:
Border radius of nested elements should differ by one pixel otherwise a
gap appear.

Align forms on narrow screen (less than 991 pixel). Bootstrap have a
longstaning bug that the narrow to wide transtition for some rules are
991, and otherrules 992px.

I also move the all page 32 pixel up because not everyone has lots of screen real estate. 

-- 

Edit, the 3px radius avoid this:

screen shot 2017-12-09 at 12 30 38

the media query give this on narrow screen:

screen shot 2017-12-09 at 12 45 55

### DIFF
--- a/binderhub/static/index.css
+++ b/binderhub/static/index.css
@@ -20,7 +20,6 @@
 }
 
 body {
-    padding-top: 32px;
     font-family: "ClearSans-Thin", sans-serif;
 }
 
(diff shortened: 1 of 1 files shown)


choldgraf:
seems reasonable to me - I'll merge at end of tomorrow unless somebody has thoughts! In the meantime @Carreau will have to keep looking at that extra pixel space :-)

PR #342, opened 2017-12-12
[WIP] require docker registry

minrk:
Follow-up to #319

This reduces the number of moving parts at runtime for docker-in-docker by requiring a docker registry to always be used. Instead of running without a registry, a local registry should be run when a public registry is not desired.

TODO:

- [x] remove use_registry flag
- [x] support insecure local registries (http, no auth)
- [x] run registry on travis
- [ ] add registry to helm chart so deploying with local registry is one-click
- [ ] test with d-in-d on travis now that we have a registry

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,7 @@ on GitHub if you don't have a token.
    ```bash
    helm init
    ```
+
 4. Add the JupyterHub helm charts repo:
 
    ```bash
@@ -42,15 +43,21 @@ on GitHub if you don't have a token.
 
         python3 -m pip install -e . -r dev-requirements.txt
 
-7. Before starting the local dev/test deployment run:
+7. Before starting the local dev/test deployment run,
+   start a local docker registry (in minikube):
 
         eval $(minikube docker-env)
+        # spawn registry in minikube docker
+        docker run -d -p 5000:5000 --restart=always --name registry registry:2
+        # forward localhost:5000 to minikube:5000 so both binderhub
+        # and kubernetes see the same registry at 127.0.0.1:5000
+        ssh -i $(minikube ssh-key) docker@$(minikube ip) -L5000:127.0.0.1:5000 -f -N
 
-7. Start binderhub with the testing config file:
+8. Start binderhub with the testing config file:
 
         python3 -m binderhub -f testing/minikube/binderhub_config.py
 
-8. Visit [http://localhost:8585](http://localhost:8585)
+9. Visit [http://localhost:8585](http://localhost:8585)
 
 All features should work, including building and launching.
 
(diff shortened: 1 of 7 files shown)


minrk:
In looking a bit more, maybe running a local registry isn't that simple. Running an insecure local registry is quite simple, but setting one up for 'real' use with authentication and everything is probably going to be *more* complicated, not less. I'm not 100% sure this is the right thing to do anymore.

choldgraf:
Is this going to change our start-up guide at all? Lemme know if so and I can write a follow-up docs PR

minrk:
> Is this going to change our start-up guide at all?

Yes, I updated CONTRIBUTING.md to include what should be necessary.

An *insecure* local registry is nice and easy, but after looking into what to do for production, running a properly secure local registry is a lot less so (it's not even supported in the official charts). For that reason, I'm reconsidering whether this is actually a net simplification for production, as opposed to testing.

My main goal here was to ensure that docker-in-docker / push / etc. got exercised in the tests, and I think I may revisit this to just *allow* for an insecure local registry for testing, and leave the logic for an altogether-disabled registry in place.

minrk:
Also, running docker-in-docker plus a local registry *triples* the amount of disk space each image takes:

1. image created by build in dind
2. push copy of image from dind to registry
3. pull copy from registry to host docker

This usually isn't a big deal (what's a few hundred GB between friendly SSDs?), but when you are testing on a minikube vm with limited disk space, this could get annoying (and super slow).

choldgraf:
FWIW, I definitley ran into space issues when testing binderhub on my laptop yesterday

pfisterer:
Is anybody still working on this issue? I want to deploy binderhub/jupyterhub in Kubernetes using a private registry with or without basic auth. 

From a discussion with @minrk on Gitter I understood that the changes proposed here can partially be used to update the current implementation of `DockerRegistry`.

I would like to contribute to the development if such an update could be integrated into binderhub.

minrk:
Closing this as stale, might pick it up some day

pfisterer:
Just in case someone is interested. I've manage to deploy a private registry and have configured it to be used by binderhub. 

However, this requires some effort to work seamlessly. I somehow didn't manage to get insecure registries working. So I used [external-dns](https://github.com/kubernetes-sigs/external-dns) + [cert-manager](https://cert-manager.io/docs/) to get "real" hostname + ACME certificates.

To run a "real" private registry, you need a `docker registry`(e.g., [stable/docker-registry](https://github.com/helm/charts/tree/master/stable/docker-registry)) + `docker degistry authentication server` (e.g., [cesanta/docker_auth](https://github.com/cesanta/docker_auth)).

I've compiled everything in an Ansible playbook here: [pfisterer/edsc-k8s-playbook](https://github.com/pfisterer/edsc-k8s-playbook).

PR #343, opened 2017-12-12
update build.py with upstream changes

minrk:
attempt to re-sync with https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/325

The following features are needed here that aren't there:

- build args need to derive from last_commit,
  so support templating values in image buildArgs
- last_commit needs to take a list of paths, not just one directory,
  so support path lists at both image and chart level.

These should hopefully be added in a way that still works for the jupyterhub chart.

While working on this I fixed the timestamp issues by using `helm repo index --merge` to only add new files, not modify already-indexed files. The result is that existing entries will not be modified.


helm-chart/build.py
```diff
@@ -2,149 +2,193 @@
...
+        paths = options.get('paths', []) + [image_path]
         if commit_range:
-            if not path_changed(IMAGE_FILES, commit_range):
-                print("Skipping {}, not touched in {}".format(image, commit_range))
+            if not path_touched(*paths, commit_range=commit_range):
+                print("Skipping {}, not touched in {}".format(name, commit_range))
                 continue
-        tag = last_git_modified(IMAGE_FILES)
-        image_spec = '{}{}:{}'.format(prefix, image, tag)
+        last_commit =last_modified_commit(*paths)
```
willingc: super nitpicky style: missing space after =

helm-chart/build.py
```diff
@@ -2,149 +2,193 @@
...
-            image_path
-        ])
         if push:
             subprocess.check_call([
                 'docker', 'push', image_spec
             ])
+    return value_modifications
+
+def build_values(name, values_mods):
+    rt_yaml = YAML()
```
willingc: `rt` Return? Something-type? It's not clear to me what the rt is shorthand for. Perhaps mention at first use in a comment.

helm-chart/build.py
```diff
@@ -2,149 +2,193 @@
...
+        rt_yaml.dump(values, f)
 
 
-def build_chart():
-    version = last_git_modified([BASEPATH] + IMAGE_FILES)
-    with open(os.path.join(CHARTPATH, 'Chart.yaml')) as f:
-        chart = yaml.load(f)
-    raw_version = chart['version']
-    chart['version'] = chart['version'].split('-')[0] + '-' + version
+def build_chart(name, version=None, paths=('.',)):
```
willingc: Out of curiosity why is `paths` set to a tuple as default? Is it to get the current path `.` for setting the commit later?

helm-chart/build.py
```diff
@@ -2,149 +2,193 @@
...
 
 def main():
-    argparser = argparse.ArgumentParser()
-    argparser.add_argument(
-        '--image-prefix',
-        default='jupyterhub/k8s-'
-    )
-    subparsers = argparser.add_subparsers(dest='action')
+    with open('chartpress.yaml') as f:
+        safe_yaml = YAML(typ='safe')
```
willingc: type?

helm-chart/build.py
minrk: yeah, I made it simpler by moving the default below.

helm-chart/build.py
minrk: I simplified the yaml initialization in the latest commit with a comment: `typ` (chosen presumably to avoid colliding with builtin `type`) is used to signal which type of yaml loader is used. We use `typ='rt'` for the roundtrip loader (which is also 'safe') since we are editing files.

helm-chart/build.py
```diff
@@ -55,27 +55,27 @@ def build_images(prefix, images, tag=None, commit_range=None, push=False):
...
         if tag is None:
             tag = last_commit
         image_name = prefix + name
         image_spec = '{}:{}'.format(image_name, tag)
+        value_modifications[options['valuesPath']] = {
+            'name': image_name,
+            'tag': tag
+        }
+
+        if commit_range and not path_touched(*paths, commit_range=commit_range):
```
willingc: This is so the build doesn't happen if it's a doc change. Is that correct?

minrk: yes, precisely

minrk:
Thanks, @willingc!

minrk:
Same file pushed to z2jh here: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/350

PR #346, opened 2017-12-13
dind: conditional daemonset, support initContainers

minrk:
Minor updates while deploying docker-in-docker:

- only run the daemonset when dind.enabled is True
- allow initContainers for dind itself, which should allow us to firewall builds like we do for runs

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -133,6 +133,7 @@ deployment:
 
 dind:
   enabled: false
+  initContainers:
   daemonset:
     image:
       name: docker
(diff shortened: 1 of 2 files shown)


helm-chart/binderhub/values.yaml
```diff
@@ -133,6 +133,7 @@ deployment:
 
 dind:
   enabled: false
+  initContainers:
```
yuvipanda: Can we default this to an array ([]) so it is clear it's an array type we expect here?

yuvipanda:
Thanks, @minrk!

PR #347, opened 2017-12-13
adding favicons for build + success

choldgraf:
This adds dynamic favicons for "building" and "success". I'            update_favicon("/favicon_success.ico");
m not 100% on the javascript since there was some weird cacheing going on with my binderhub server...not sure how to fix that so it'd be great if somebody else can give the code a shot...

cc @Carreau!


binderhub/static/index.js
```diff
@@ -258,7 +257,6 @@ $(function(){
         image.onStateChange('waiting', function(oldState, newState, data) {
             $('#phase-waiting').removeClass('hidden');
         });
-
         image.onStateChange('building', function(oldState, newState, data) {
             $('#phase-building').removeClass('hidden');
```
Carreau: If I add a update_fav here, seem to work for me. We can even just put it just after `$('#build-form').submit(function() {` (about line 230)

choldgraf: ok - updated moving it to the section you suggested

Carreau:
+1.

We'll have to take care of the base_url at some point now that there is one, but let's keep that for another issue.

choldgraf:
sounds good - let me know if there's anything else I should do on this PR...else we can iterate on subsequent PRs

Carreau:
I'll merge and deploy tomorrow. Getting late here.

PR #352, opened 2017-12-14
add deploy stage

minrk:
Fixes publishing of binderhub chart, which was accidentally skipped due to changes in #343

This adds a dedicated deploy stage, so it should be easier to find the deploy output without digging past the test output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,7 @@ after_success:
 jobs:
   include:
   - stage: deploy
-    branches:
-      only:
-      - master
+    if: type = push AND branch = master
     # set env to avoid deploy matrix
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -27,15 +18,24 @@ after_failure:
...
-    ./build.py --commit-range ${TRAVIS_COMMIT_RANGE} --push
-    cd ..
-  fi
+
+jobs:
+  include:
+  - stage: deploy
+    branches:
+      only:
+      - master
```
Carreau: That does not seem to work as right now travis tried to decrypt things and failed now waitign for DockerHub username.

PR #353, opened 2017-12-14
use ci/deploy.sh in sync with upstream

minrk:
should make it harder to get out of sync

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,12 +28,7 @@ jobs:
     if: type = push AND branch = master
     # set env to avoid deploy matrix
     env: TEST=helm
-    script:
-      - openssl aes-256-cbc -K $encrypted_d8355cc3d845_key -iv $encrypted_d8355cc3d845_iv -in travis.enc -out travis -d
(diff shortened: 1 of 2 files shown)


Carreau:
We are going to converge !

`./ci/deploy.sh: line 4: encrypted_c6b45058ffe8_key: unbound variable`

.travis.yml
```diff
@@ -28,12 +28,7 @@ jobs:
     if: type = push AND branch = master
     # set env to avoid deploy matrix
     env: TEST=helm
-    script:
-      - openssl aes-256-cbc -K $encrypted_d8355cc3d845_key -iv $encrypted_d8355cc3d845_iv -in travis.enc -out travis -d
```
minrk: I copied ci/deploy.sh from zero-to-jupyterhub, but forgot to replace the values with the ones here for this repo.

Carreau: Maybe a sign we should pass the variable as parameters if we want to keep things in sync.

willingc: +1 @Carreau

PR #354, opened 2017-12-14
Use correct envvar names.

Carreau:
I should have spoted that in review.


willingc:
I did a quick scan review of the other pr and I didn't catch this either.

PR #359, opened 2017-12-18
Fix the URL for readthedocs page

ibell:
RTD pages seem to have been shuffled, URL updated


choldgraf:
Great spot! You are correct that the docs shifted a bit...thanks for the patch!

PR #363, opened 2017-12-18
remove -deployment from binder pod name

minrk:
and apply common labels

matching upstream changes in jupyterhub chart, which is pulled in here as well


choldgraf:
this all seems pretty reasonable to me - tho travis isn't happy!

yuvipanda:
This needs the service object's label selector to be updated, since it is now only looking for a 'name: binder-pod' label which we have removed.

yuvipanda:
Nah I'll just deploy this now :)

PR #365, opened 2017-12-19
document build and launch code

willingc:


binderhub/builder.py
```diff
@@ -137,21 +137,33 @@ def escape(s):
         })
 
     async def get(self, provider_prefix, spec):
-        """Get a built image for a given GitHub user, repo, and ref."""
-        # We gonna send out event streams!
+        """Get a built image for a given spec (user, repo, and ref).
```
yuvipanda: This method actually launches a given repo of a given provider, building it if needed

The 'get' is tornado convention since it's a GET request.

binderhub/builder.py
```diff
@@ -169,6 +181,8 @@ def escape(s):
         if ref is None:
             await self.fail("Could not resolve ref for %s. Double check your URL." % key)
             return
+
+        # generate a complete build name `build-{user}-{repo}-{ref}`
```
yuvipanda: It's {user}-{repo}-{ref} only for github, it varies for other providers. There's a big docstring on _generate_build_api explaining what it tries to do...

binderhub/builder.py
```diff
@@ -226,6 +242,7 @@ def escape(s):
             docker_host=self.settings['build_docker_host']
         )
 
+        """----- In Progress Builds -----"""
```
yuvipanda: There's no code about finding in progress builds here, they're onlin in build.py.

binderhub/builder.py
```diff
@@ -280,19 +297,20 @@ def escape(s):
...
+        """"----- Launch after building an image -----"""
         if not failed:
             BUILD_TIME.labels(status='success').observe(time.perf_counter() - build_starttime)
             with LAUNCHES_INPROGRESS.track_inprogress():
                 await self.launch()
 
 
     async def launch(self):
-        """Ask the Hub to launch the image"""
+        """Ask JupyterHub's Hub to launch the image."""
```
yuvipanda: I think this should be just 'Ask JupyterHub to launch the image'

binderhub/builder.py
```diff
@@ -137,21 +137,33 @@ def escape(s):
...
-        """Get a built image for a given GitHub user, repo, and ref."""
-        # We gonna send out event streams!
+        """Get a built image for a given spec (user, repo, and ref).
+
+        Parameters
+        ----------
+            provider_prefix : str
+                the nickname for a repo provider (i.e. 'gh')
+            spec:
+                information that specifies the user, repo, and ref
```
yuvipanda: The spec varies per provider, it's user repo ref only for github.

willingc:
Ready to go. I think...

PR #369, opened 2017-12-20
mitigate eventstream reconnects

minrk:
By adding a sleep at the end of the eventstream handler, it should be less likely that an eventstream client will automatically reconnect, triggering a new build.


yuvipanda:
Thanks, @minrk

PR #370, opened 2017-12-20
implement per_repo_quota

### LINKED ISSUE #331: rate-limit individual repos
We had a significant spike in traffic earlier today, almost entirely from one repo.

screen shot 2017-12-08 at 14 53 47

We should consider limiting the impact one repo can have on the rest of the service. As it is now, one repo can easily consume the entirety of Binder. Setting an upper limit on the number of concurrent users running from a given repo would give us a knob for something like this.

On one hand, it's nice to be able to handle traffic peaks from things like the eLife announcement, or a norvig post. On another, it's an avenue for abuse of the service and can prevent others from being able to use it.

minrk:
sets the maximum number of pods that can be running from a given repo

default in the helm chart is 100, default for the app is 0 (no limit)

Currently, this fails the launch with an error that too many people are using the repo. Alternately, we could hold the launch in a queue and wait until the quota is no longer met.

closes #331

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -329,9 +329,15 @@ class BuildHandler(BaseHandler):
         # check quota first
         quota = self.settings.get('per_repo_quota')
 
+        # the image name (without tag) is unique per repo
+        # use this to count the number of pods running with a given repo
+        # if we added annotations/labels with the repo name via KubeSpawner
+        # we could do this better
         image_no_tag = self.image_name.rsplit(':', 1)[0]
         matching_pods = 0
         total_pods = 0
+
+        # TODO: run a watch to keep this up to date in the background
         pool = self.settings['build_pool']
         f = pool.submit(kube.list_namespaced_pod,
             self.settings["build_namespace"],
@@ -346,12 +352,16 @@ class BuildHandler(BaseHandler):
         for pod in pods.items:
             total_pods += 1
             for container in pod.spec.containers:
+                # is the container running the same image as us?
+                # if so, count one for the current repo.
                 image = container.image.rsplit(':', 1)[0]
                 if image == image_no_tag:
                     matching_pods += 1
                     break
 
         # TODO: allow whitelist of repos to exceed quota
+        # TODO: put busy users in a queue rather than fail?
+        # That would be hard to do without in-memory state.
         if quota and matching_pods >= quota:
             app_log.error("%s has exceeded quota: %s/%s (%s total)",
                 self.repo, matching_pods, quota, total_pods)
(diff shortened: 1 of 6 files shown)


binderhub/builder.py
```diff
@@ -312,11 +312,50 @@ def escape(s):
...
+                await self.launch(kube)
 
 
-    async def launch(self):
+    async def launch(self, kube):
         """Ask JupyterHub to launch the image."""
+        # check quota first
+        quota = self.settings.get('per_repo_quota')
+
+        image_no_tag = self.image_name.rsplit(':', 1)[0]
```
yuvipanda: Add a comment here on why this works?

binderhub/builder.py
```diff
@@ -312,11 +312,50 @@ def escape(s):
...
+        matching_pods = 0
+        total_pods = 0
+        pool = self.settings['build_pool']
+        f = pool.submit(kube.list_namespaced_pod,
```
yuvipanda: Longer term this should be a watch rather than making an individual API request for each launch, but is ok for now!

binderhub/builder.py
```diff
@@ -312,11 +312,50 @@ def escape(s):
...
+        # concurrent.futures.Future isn't awaitable
+        # wrap in tornado Future
+        # tornado 5 will have `.run_in_executor`
+        tf = Future()
+        chain_future(f, tf)
+        pods = await tf
+        for pod in pods.items:
+            total_pods += 1
+            for container in pod.spec.containers:
+                image = container.image.rsplit(':', 1)[0]
```
yuvipanda: Add a comment explaining why this works?

minrk:
Now that we've had a couple events of repos swamping mybinder.org, I think we should go with this. I've added TODO comments based on review for things that could be improved in the future and I think this is good to go now.

testing/minikube/binderhub_config.py
```diff
@@ -8,3 +8,4 @@
 c.BinderHub.hub_url = 'http://{}:30123'.format(minikube_ip)
 c.BinderHub.hub_api_token = 'aec7d32df938c0f55e54f09244a350cb29ea612907ed4f07be13d9553d18a8e4'
 c.BinderHub.use_registry = False
+c.BinderHub.build_namespace = 'binder-test'
```
betatim: Just double checking: On mybinder.org this gets set via an environment variable which is set via the helm chart or do we need to adjust something in mybinder.org-deploy before this works?

minrk: This was just missing from the testing environment. When you run minikube, it was doing builds in 'default' because this was unspecified. The chart is already doing everything right.

Issue #322, opened 2017-12-03
Add an API endpoint that resolves arbitrary URLs into specs


yuvipanda:
Riffing off the awesomebar concept that @choldgraf and @betatim (and maybe @mariusvniekerk?) have been talking about, I think it'll be useful to have an API endpoint that takes an arbitrary URL and returns a JSON response with possible spec info. 

So if you give it a GitHub url (long or short), it'll return the fact that it's a GitHub URL and a spec that can be fed into the rest of the API. Same for gist, or the specific git URLs, or gitlab (or future providers).

Responding with a JSON response allows clients to provide appropriate UI for this functionality before starting builds. In the mybinder.org UI, we can change an icon or something in the bar as soon as URL is pasted and before launch is clicked.

We can implement this with a 'detect' method on each of the providers that get passed this URL and can return True + a spec...

yuvipanda:
This would also keep provider code all python rather than split it into Python + JS, at the cost of an extra network request. However, if we're going to support other UI providers, making this be a API endpoint that can be hit rather than JS code to be included sounds great.

rgbkrk:
> I think it'll be useful to have an API endpoint that takes an arbitrary URL and returns a JSON response with possible spec info.

👍

mariusvniekerk:
if @rgbrk uses the build url directly, we could use it to fully validate the build parts of the new providers without having to have a UI yet.  :)

PR #372, opened 2017-12-20
Use webpack+npm for frontend js

yuvipanda:
Note that none of the functionality itself has changed. This just
starts using webpack+npm for dependency management + minification.

- Use npm for dependency management
- Use webpack to stitch the resources together into one file
- Upgrade xterm.js version when we are at it, to make life simpler
- Move xterm's fit.js addon to a different path. This is hacky,
  but no more hacky than what we had before
- Load one js file with all our JS and CSS, instead of multiple CSS
  and JS files.
- Fix build.py to allow easier building of docker image when developing
   locally
- Switch dockerfile to use a two staged build, so we can install npm in
  the build image but keep it out of the final image

Things to do:

- [x] Run this by someone who knows what they are doing
- [x] Fix setup.py to run npm & include the dist dir if needed
- [x] Add helper utils for automatic recompilation
- [x] Fix travis builds to pass
- [x] Fix docker build


yuvipanda:
This explicitly doesn't add babel or change the code much, since the intention is to just set up the build system and nothing else.

yuvipanda:
I suspect my setup.py foo is not good enough, since the JupyterHub one is far more complex.

rgbkrk:
Whooooa cool. I’ll help review this and chime in where I can.

binderhub/templates/index.html
```diff
@@ -2,14 +2,12 @@
 
 {% block head %}
 <meta id="base-url" data-url="{{base_url}}">
+<script src="{{static_url("dist/bundle.js")}}"></script>
+<link href="{{static_url("dist/styles.css")}}" rel="stylesheet"></link>
```
rgbkrk: Nice

binderhub/static/vendor/xterm.js-2.6.0/xterm.css
```diff
@@ -1,2247 +0,0 @@
...
-    background-color: #e4e4e4;
-}
-
-.terminal .xterm-color-255 {
-    color: #eeeeee;
-}
-
-.terminal .xterm-bg-color-255 {
-    background-color: #eeeeee;
-}
```
rgbkrk: So much was vendorized here, I’m glad this is switched to using packaging so it’ll be easier to review in the future.

webpack.config.js
```diff
@@ -0,0 +1,35 @@
...
+        ]
+    },
+    devtool: 'source-map',
+    plugins: [
+        new webpack.ProvidePlugin({
+            $: 'jquery',
+            jQuery: 'jquery',
+        }),
+        new ExtractTextPlugin("styles.css"),
+    ]
```
rgbkrk: Looking good to me

choldgraf:
this looks like a nice step towards getting iteratively closer to a more manageable JS codebase, 👍

choldgraf:
just another note that I think this will mean we can leverage https://github.com/IonicaBizau/git-url-parse to do some url parsing for git repos. I'll try to make some patches to that repo to support ref / file paths in the URL

yuvipanda:
@choldgraf I don't think we should use that library. Instead, I'm convinced that we should be using #322 for implementing awesome bar style UX.

yuvipanda:
I've ticked ' Run this by someone who knows what they are doing' based on the fact that @rgbkrk and @lgeiger have approved this :)

Am working on fixing the travis build now (which has caught some very legit errors!)

yuvipanda:
ok, I've fixed the travis build too, and this is ready for review + merge now.

binderhub/static/js/index.js
```diff
@@ -10,8 +10,17 @@
...
+import 'bootstrap';
 
-"use strict";
+import 'bootstrap/dist/css/bootstrap.min.css';
+import 'bootstrap/dist/css/bootstrap-theme.min.css';
+import '../index.css';
+
+
+// FIXME: Can not seem to import this addon from npm
+import {fit} from './vendor/xterm/addons/fit';
```
willingc: These docs have a different approach to loading add-ons:

https://xtermjs.org/docs/api/addons/fit/#undefined

https://github.com/xtermjs/xterm.js#addons

yuvipanda: Yep, except they don't seem to work with webpack. https://github.com/xtermjs/xterm.js/issues/1018 has more details.

willingc: Might as well toss the issue into the comment too. Yay, JS fun.

rgbkrk: Is the use of xterm here just for handling of stdout?

If you're less worried about full terminal support and just want to render basic color, deal with carriage returns, etc. there are packages available, namely:

* https://github.com/lgeiger/escape-carriage
* anser, ansi-to-react

yuvipanda: yeah, it's for handling the output of docker build. Unfortunately it often has ansi control characters in it, since it is just arbitrary programs outputting to stdout. Will take a look at the other packages!

rgbkrk: /cc @lgeiger

yuvipanda: @willingc I added the issue link as a comment!

helm-chart/images/binderhub/Dockerfile
```diff
@@ -1,12 +1,21 @@
...
+RUN apt-get update && \
+    apt-get install --yes nodejs python3 python3-pip python3-wheel python3-setuptools
 
-CMD python3 -m binderhub
+COPY . /tmp/binderhub
+WORKDIR /tmp/binderhub
 
+RUN python3 setup.py bdist_wheel
+
+FROM python:3.6-stretch
```
minrk: Adding a note that this multi-stage build (that I didn't know about!) sets minimum Docker to 17.05.

PR #378, opened 2017-12-22
Add user-name and user-id arguments to build command

betatim:
Follows up jupyter/repo2docker#172

We need to explicitly set the username and userid to use inside the
container now. This adds the corresponding commandline arguments to
the repo2docker call.

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -203,7 +203,7 @@ class BinderHub(Application):
     )
 
     builder_image_spec = Unicode(
-        'jupyter/repo2docker:235f0ad',
+        'jupyter/repo2docker:v0.4.1',
         help="""
         The builder image to be used for doing builds
         """,
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -203,7 +203,7 @@ class BinderHub(Application):
     )
 
     builder_image_spec = Unicode(
-        'jupyter/repo2docker:v0.4.1',
+        'jupyter/repo2docker:235f0ad',
         help="""
         The builder image to be used for doing builds
         """,
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -53,6 +53,8 @@ class Build:
             '--ref', self.ref,
             '--image', self.image_name,
             '--no-clean', '--no-run', '--json-logs',
+            '--user-name', 'jovyan',
+            '--user-id', '1000'
         ]
 
         if self.push_secret:
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -12,7 +12,7 @@ image:
   name: jupyterhub/k8s-binderhub
   tag: local
 
-repo2dockerImage: jupyter/repo2docker:v0.4.1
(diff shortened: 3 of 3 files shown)


yuvipanda:
@betatim I think you might need to update the repo2docker image tag in the helm chart too.

betatim:
Realized that after boarding ;)

Now I am undecided what we should have as default in binderhub, the binderhub helm chart and what we should test binderhub against on travis. I think we should test and use in the helm chart the latest tag. Unsure about the default in binderhub (could always point to a "released" version)?

yuvipanda:
@betatim I agree it should always point to a released version of code, but at a minimum it needs to point to a version that would work. So right now that would require we point to the latest commit has explicitly.

betatim:
Ready for review and merge.

choldgraf:
seems reasonable to me...just to be clear this doesn't change any functionality, just makes explicit the username and ID from the CLI, and makes sure this will keep working in the future, yeah?

I'll merge later today unless someone wants changes!

choldgraf:
I noticed that travis wasn't happy after merge, @betatim can you double check that it's happy whenever the builds start again? I gotta head off now!

yuvipanda:
This needs a slightly careful mybinder.org deploy, since I think we need to
co-ordinate a repo2docker bump too at the same time. If we aren't going to
deploy this right away can we revert this and leave it open until we can?

On Dec 27, 2017 2:23 PM, "Chris Holdgraf" <notifications@github.com> wrote:

I noticed that travis wasn't happy after merge, @betatim
<https://github.com/betatim> can you double check that it's happy whenever
the builds start again? I gotta head off now!

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/jupyterhub/binderhub/pull/378#issuecomment-354186436>,
or mute the thread

yuvipanda:
I've done so with #380. Feel free to revert the revert when anyone has time to do a deploy too! Thanks!

betatim:
What is special about this deploy? The repo2docker PR has already been merged. On my list I had:
* merge jupyter/repo2docker#172
* merge this
* deploy

Let me know what the missing step is and I can take care of deploying.

choldgraf:
I think it's just that they both need to be deployed to mybinder.org at the same time, probably w/ a special commit message saying that it is necessary to bump both the r2d and bh version at once. @yuvipanda just didn't have time to do a deploy. Wanna do this now?

yuvipanda:
@betatim @choldgraf I think the missing step is: Modify https://github.com/jupyterhub/mybinder.org-deploy/blob/staging/mybinder/values.yaml#L94 to point to new repo2docker image value as well. That is needed every time we make a change to repo2docker / binderhub that requires a specific version of the other. We should probably document this somewhere too!

Apologies for not being clear earlier! I feel spooked whenever master is in an undeployable and undeployed state, since it prevents quick deploys whenever there is a fire...

yuvipanda:
Deployed in https://github.com/jupyterhub/mybinder.org-deploy/pull/242

PR #384, opened 2017-12-28
Adding a loading page with simpler UI

choldgraf:
This is a prototype to build in loading page functionality. It's still WIP as I'm not super experienced in JS, so feel free to make suggestions. It essentially does the following two things:

* Adds a `loading.html` page that mimics `index.html` as much as possible, but makes the form invisible and adds a spinner (GIF of the spinner is below)
* Adds some CSS to define the loading spinner behavior
* Makes `ParameterizedMainHandler` serve `loading.html`

The JS under the hood is the exact same, since I figure we'll change that significantly after #372 + subsequent PRs are merged. IMO this could be a stop-gap solution to get the general structure there, and we could improve upon it in subsequent PRs. GIF of the loading page is below, lemme know what you think...

loading_demo

a couple questions:

* Should the page behave differently if a binder isn't already built, and has to go through the building process?
* Should it be possible to access logs from the loading page?
* What should happen if there is an error?

### DIFF
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -39,7 +39,7 @@ class ParameterizedMainHandler(BaseHandler):
             raise web.HTTPError(400, str(e))
 
         self.render_template(
-            "index.html",
+            "loading.html",
             base_url=self.settings['base_url'],
             url=provider.get_repo_url(),
             ref=provider.unresolved_ref,
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -256,6 +256,11 @@ $(function(){
         updateUrlDiv(url);
         update_favicon("/favicon_building.ico");
 
+        // Update the text of the loading page if it exists
+        if ($('div#loader-text').length > 0) {
+            $('div#loader-text p').text("Loading repository: " + repo)
+        }
+
         $('#build-progress .progress-bar').addClass('hidden');
         log.clear();
 
@@ -288,6 +293,12 @@ $(function(){
             // If we fail for any reason, we will show logs!
             if (!logsVisible) {
                 $('#toggle-logs').click();
+            }
+
+            // Show error on loading page
+            if ($('div#loader-text').length > 0) {
+                $('#loader').addClass("error");
+                $('div#loader-text p').html('Error loading ' + repo + '!<br /> See logs below for details.')
             }
             image.close();
         });
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -259,6 +259,11 @@ $(function(){
         // Update the text of the loading page if it exists
         if ($('div#loader-text').length > 0) {
             $('div#loader-text p').text("Loading repository: " + repo)
+            window.setTimeout( function() {
+                $('#loader').addClass("paused");
+                $('div#loader-text p').html("Repository " + repo + " is taking a long time to load!<br />See the logs for details.")
+                $('#loader').hover(function() {$('#loader').removeClass("paused")}, function() {$('#loader').addClass("paused")});
+            }, 30000)
         }
 
         $('#build-progress .progress-bar').addClass('hidden');
--- a/binderhub/static/loading.css
+++ b/binderhub/static/loading.css
@@ -74,6 +74,25 @@ https://ihatetomatoes.net for initial code templates.*/
         }
     }
 
+.error, .error:after, .error:before {
+    border-top-color: red !important;
+}
+
+.error {
+    -webkit-animation: spin 30s linear infinite !important; /* Chrome, Opera 15+, Safari 5+ */
+    animation: spin 30s linear infinite !important; /* Chrome, Firefox 16+, IE 10+, Opera */
+}
+
+.error:after {
+  -webkit-animation: spin 10s linear infinite !important; /* Chrome, Opera 15+, Safari 5+ */
+  animation: spin 10s linear infinite !important; /* Chrome, Firefox 16+, IE 10+, Opera */
+}
+
+.error:before {
+  -webkit-animation: spin 20s linear infinite !important; /* Chrome, Opera 15+, Safari 5+ */
+  animation: spin 20s linear infinite !important; /* Chrome, Firefox 16+, IE 10+, Opera */
+}
+
 #demo-content {
 	padding-top: 100px;
 }
@@ -84,7 +103,11 @@ https://ihatetomatoes.net for initial code templates.*/
     text-align: center;
 }
 
-/*To prevent the terminal from popping up on submit*/
-div.terminal {
+div#log-container {
+    width: 80%;
+    margin: 0% 10%;
+}
+
+.hidden {
     display: none;
 }
--- a/binderhub/static/loading.css
+++ b/binderhub/static/loading.css
@@ -12,10 +12,7 @@ https://ihatetomatoes.net for initial code templates.*/
     border-radius: 50%;
     border: 7px solid transparent;
     border-top-color: #f5a252;
-
-    -webkit-animation: spin 2s linear infinite; /* Chrome, Opera 15+, Safari 5+ */
-    animation: spin 2s linear infinite; /* Chrome, Firefox 16+, IE 10+, Opera */
-
+    animation: spin 2s linear infinite;
     z-index: 1001;
 }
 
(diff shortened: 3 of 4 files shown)


yuvipanda:
Great start! I love the visual and I think it's a much better UX!

Some comments:

1. I'd like to know what is actually being launched, so perhaps a indicator of the repo being launched?
2. What happens if loading errors?

yuvipanda:
We'll have to pretty much rewrite all of this to be a single page app after #372 lands however, jfyi :)

choldgraf:
answers below:

1. That's a great idea, will incorporate the repo being launched
2. Yep I agree, I'm hoping this gives us an early prototype that we can iterate on later (or learn from, totally demolish, and start from scratch :-) )

choldgraf:
OK, latest push makes some improvements per @yuvipanda 's suggestions...here's a GIF of a working page and an errored page:

loading_demo2

Carreau:
Can a click on the beautiful spinner toggle auto-launch to off with an indication ?

Also, Not that I am against CSS animations but it is know to have _some_ browser use hi-cpu/gpu while spinning, we should check for that, especially when building takes a long time.

If that's the case we can fallback on an actual GIF, heavier, but that does usually does not consume as much resources.

minrk:
> We'll have to pretty much rewrite all of this to be a single page app after #372 lands however, jfyi :)

We shouldn't have to do that. Each page can have its own entrypoint in the js bundle.

minrk:
This is great! After this lands, I think we can make the following improvements to the main form page, since it's no longer serving two purposes:

1. never redirect to the launched binder from the main form page
2. stop pushing the launch page to the history, instead create an explicit link after the binder is ready

minrk:
I merged #384 and resolved the resulting conflicts here. I think this is good to go (and a huge improvement!), but I won't merge myself now that I've contributed to the PR.

Carreau:
I fixed you comment from 384 (this PR) to #372 (I guess that's what you meant)

choldgraf:
Hey all - is there anything else for me to do on this one or are we ready to merge? I'm a bit uncertain as to what we should decide for "click the spinner to toggle off auto-launch", so could we punt that to another PR?

yuvipanda:
I'll take a look at this later today!

choldgraf:
cool! lemme know if there's anything else I should take care of here (or we can also open issues to iterate on this design)

Carreau:
So I'm am going to -1 on the PR, while I love the design and the implementation, when the spinner arrive on the page, Chrome jump from 3%-5% cpu usage to 13%-17% in a few second (when page is visible), be a tiny bit less when page is not the main visible tab. 

This is a known issue with CSS animation & browsers, and I really would not like binder to kill the battery of my laptop. 

Now I would be happy if we can find workaround, like make the  spinner a GIF (that should consume less), stop the spinning with a JS timeout if it takes more than say 15sec and/or have a windows focus-in/focus-out to stop/restart the spinning when the tab is invisible. 

So apologies but as is I'm going to not approve.

Carreau:
And OMG Firefox requires 50% CPU when the animation is visible !
This is the only tab I have open with FF. 
screen shot 2018-01-08 at 12 03 25

Carreau:
Possbile mitigations:

[transform: translateZ(0)](https://stackoverflow.com/questions/13176746/css-keyframe-animation-cpu-usage-is-high-should-it-be-this-way) forces composition on GPU which lower CPU usage a bit.

On chrome seem to lower CPU usage to 12-15%, on Firefox to 22-25%, Still really high IMHO.

minrk:
Sending it to the GPU may or may not make it more efficient. Using more GPU and less CPU may just be moving the work to an unmonitored energy consumer. Maybe we could make the css animations Safari-only, since there they use <0.2% CPU :)

Switching to a GIF to save energy makes sense, even if it makes me sad.

Carreau:
We can fallback to JS and/or canvas as well. I tried to replace `linear` by `steps(10)` (10 FPS) it still consumes a lot.

I would also be sad to use a GIF.

Carreau:
Here is it rendered in a canvas+p5js;

https://alpha.editor.p5js.org/mbussonn/sketches/SykwxzbVM
Not same proportions and colors but we can tweak that. 
Advantage is with p5 we can tweak FPS , so here I limit to 30 FPS. 
Not better than CSS (maybe even worse), except it is a Chrome worker that use CPU instead of Main process...

choldgraf:
ah - that is a good find @Carreau ... some questions below:

1. What's the benefit of using the chrome worker instead of the main CPU? Does that mean it'll only compete with other chrome processes and not with system processes?
2. Is it true that **all** CSS solutions will be this heavy? E.g. it seems like the spinners on this page [0] are a bit simpler...maybe they're also easier on resources?
3. What about using a pure javascript option like http://spin.js.org/ ? We could configure that to have three overlapping rings, or go with something different.

That said - users are ultimately expected to spend relatively little time on this page, as best-practices would be for people to build their Binder image before they distribute their Binder links, so I think it's safe to say that we don't want users sitting on this page for more than ~15 seconds.

I would also definitely be open to using this loading page as a stop-gap and then working on a more performance-friendly version in a subsequent PR...this is still beta after all and I imagine we may be making tweaks to the loading page with website rewrites in the future anyway


[0] https://codepen.io/alanshortis/pen/eJLVXr

Carreau:
> What's the benefit of using the chrome worker instead of the main CPU? Does that mean it'll only compete with other chrome processes and not with system processes?

None, it was just that I almost though "Yeah it does not use CPU" while it still does. 

> Is it true that all CSS solutions will be this heavy? E.g. it seems like the spinners on this page [0] are a bit simpler...maybe they're also easier on resources?

I'm pretty sure all css solution with `infinite` will be like that. A pure css spinner is ok if it's bounded in time, so we know we won't drain laptops. 

> What about using a pure javascript option like http://spin.js.org/ ? We could configure that to have three overlapping rings, or go with something different.

We'll have to try, I don't know how JS would do the animation, we could also investigate SVG+JS.

> That said - users are ultimately expected to spend relatively little time on this page, as best-practices would be for people to build their Binder image before they distribute their Binder links, so I think it's safe to say that we don't want users sitting on this page for more than ~15 seconds.

> I would also be open to using this loading page as a stop-gap and then working on a more performance-friendly version in a subsequent PR...this is still beta after all and I imagine we may be making tweaks to the loading page with website rewrites in the future anyway

Seem reasonable. I would say to add `.spinning` class to the element, and and have the annimation only when "Spinning" is applied, and remove ".spinning" after 10-15 sec. Re-add spinning on hover if you like in case users are wondering what's happening. 

I'm just afraid of really long job consuming resources in the BG.

choldgraf:
Yeah that could work, it also looks like there are some CSS tricks to minimize the resources needed:

https://stackoverflow.com/questions/13176746/css-keyframe-animation-cpu-usage-is-high-should-it-be-this-way

we could give that a shot. 

As an aside, what purpose does all the webkit etc code serve? For me the much simpler code here: https://codepen.io/choldgraf/pen/VyydbN seems to work fine and uses a pretty small amount of resources...could it be we're adding unnecessary amounts of baggage?

Carreau:
> Yeah that could work, it also looks like there are some CSS tricks to minimize the resources needed:
> 
> https://stackoverflow.com/questions/13176746/css-keyframe-animation-cpu-usage-is-high-should-it-be-this-way

Yes, that is similar  what I linked to before. 



> As an aside, what purpose does all the webkit etc code serve? For me the much simpler code here: https://codepen.io/choldgraf/pen/VyydbN seems to work fine and uses a pretty small amount of resources...could it be we're adding unnecessary amounts of baggage?

It's for old browsers, when animation was not standard and [vendor prefixed](https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix) I think for our target users it can likely be removed.

choldgraf:
> Yes, that is similar what I linked to before.

my bad - I missed that one as I saw all the posts after waking up :-)

> It's for old browsers, when animation was not standard and vendor prefixed I think for our target users it can likely be removed.

ok cool, that's what I suspected and am +1 on not writing a lot of extra code in order to support older browsers.

How about I push changes to:

1. Simplify the CSS as much as possible
2. Implement a 30 second animation timeout.
3. Implement as many CSS performance hacks as I can find on the internet :-)

And we can improve this incrementally in future PRs?

choldgraf:
ok, latest push addresses points 1,2, and 3 above...here's a GIF of what the timeout looks like (below it happens quickly but I've set the timeout to 30 seconds)

pause

choldgraf:
ah I should also note - I forgot that @minrk pushed the rebase changes and the latest commit actually contains a rebase + webpack fix in it. I think the push clobbered your addition, but I double checked between this commit and yours and the code for `loading.html` is the same, sorry!

Carreau:
Now deployed.

betatim:
🎉 !

betatim:
This is pretty cool.

PR #394, opened 2018-01-03
adding faq and section on github API

choldgraf:
Adds some language clarifying how to increase the github API on binderhub. Also adds a FAQ section for miscellaneous tips like this


CONTRIBUTING.md
```diff
@@ -54,7 +54,7 @@ on GitHub if you don't have a token.
 
 All features should work, including building and launching.
 
-## Testing
+## Increase your GitHub API
```
yuvipanda: Do you mean increase your github api *limit*?

doc/faq.rst
```diff
@@ -0,0 +1,26 @@
...
+Frequently Asked Questions
+==========================
+
+The following are some common questions and issues that arise when deploying
+your own BinderHub.
+
+How can I increase my GitHub API limit?
+---------------------------------------
+
+By default GitHub only lets you make a handful of requests each hour. To
```
yuvipanda: 60 requests an hour

doc/faq.rst
```diff
@@ -0,0 +1,26 @@
...
+increase this limit significantly, create an API access token that's attached
```
yuvipanda: 5000 requests an hour

CONTRIBUTING.md
willingc: Perhaps:

## Use a GitHub API token

choldgraf:
@yuvipanda comments addressed

CONTRIBUTING.md
```diff
@@ -54,22 +54,35 @@ on GitHub if you don't have a token.
 
 All features should work, including building and launching.
 
-## Testing
+## Increase your GitHub API limit
 
-It is recommended to create and enable your GitHub API token before running tests
+By default, GitHub has a limit of 60 API requests per hour. Thus,
```
willingc: Let's link to the GitHub page for clarity and more info: https://developer.github.com/v3/guides/getting-started/#authentication

CONTRIBUTING.md
```diff
@@ -54,22 +54,35 @@ on GitHub if you don't have a token.
...
+it is recommended to create and enable a GitHub API token before running tests
```
willingc: ~~it is recommended to create and enable~~ we recommend using

CONTRIBUTING.md
```diff
@@ -54,22 +54,35 @@ on GitHub if you don't have a token.
...
+
+To run unit tests, navigate to the root of the repository, then call:
 
   ```bash
-  export GITHUB_ACCESS_TOKEN=insert_token_value_here
   pytest
   ```
 
+We recommend increasing your GitHub API rate limit before running tests (see above).
+and increasing your
```
willingc: remove line 85

doc/faq.rst
```diff
@@ -0,0 +1,26 @@
+.. _faq:
+
+Frequently Asked Questions
+==========================
+
+The following are some common questions and issues that arise when deploying
+your own BinderHub.
+
+How can I increase my GitHub API limit?
```
willingc: Not sure that this is really needed in the FAQ. If included, it should be updated to match the info in CONTRIBUTING.md

choldgraf: Where would you recommend putting this info if not in the faq? The other thought I'd had was adding an `Extending your BinderHub` similar to what we have with z2jh, where we could put info like adding a github API token.

willingc: I would just leave it in the CONTRIBUTING.md unless folks have been asking about it in which case the FAQ is fine.

choldgraf: Don't you imagine they'd end up hitting the API limit pretty quickly on their own deployment, regardless of whether they're testing locally? It seems to me like the majority of BinderHub deployments would need to be linked against an API key, since 60 hits an hour isn't much

willingc: Perhaps in that case, the default should be to recommend the token as a default part of the BinderHub deployment when supporting GitHub (or GitLab...).

choldgraf: that's a good idea - lemme put something into a commit and see what you think about that

choldgraf: WDYT about latest commit?

doc/setup-binderhub.rst
```diff
@@ -168,5 +168,34 @@ Note the IP address in ``EXTERNAL-IP``. This is your BinderHub IP address.
...
-You now have a functioning BinderHub at the above IP address. For next
-steps, see :doc:`debug` and :doc:`turn-off`.
+You now have a functioning BinderHub at the above IP address.
+
+.. _api-limit:
+
+Increase your GitHub API limit
+------------------------------
+
+.. note::
```
willingc: Love the note.

PR #395, opened 2018-01-04
remove unused formats for ClearSans webfont

minrk:
we are only using .woff in CSS, so we don't need to bundle the rest since the are inaccessible.

These unused files were the vast majority of the size of binderhub repo and install.


choldgraf:
nice catch!

PR #397, opened 2018-01-04
Allow setting google analytics domain from helm chart

yuvipanda:
Makes it easier to deploy :)


choldgraf:
LGTM - I was gonna suggest that we document this somewhere, but I feel like a better approach would be to document it along w/ the other options someplace like http://binderhub.readthedocs.io/en/latest/reference/ref-index.html ...WDYT @yuvipanda ? Either way I don't think it's blocking on this PR so I'll merge it and we can iterate on docs later

PR #401, opened 2018-01-05
consolidate tips/notes

choldgraf:
I've found that over time as the install instructions have gotten more complex, some pages have a feeling of being cluttered because of the several boxes, indentation levels, etc. This is a quick PR to see if we could consolidate some of the tip/notes boxes in order to simplify the look. My hope is that this will make the text easier to follow.

Basically, any time that we have two tips/notes right after each other, this consolidates them into one box with a bulleted list. So

image

becomes

image

What do folks think about this? If people like it, I can make a similar PR to z2jh

### DIFF
--- a/doc/setup-binderhub.rst
+++ b/doc/setup-binderhub.rst
@@ -72,12 +72,9 @@ Create a file called ``secret.yaml`` and enter the following::
 
 .. tip::
 
-   The content you put just after ``password: |`` must all line up at the same
-   tab level.
-
-.. tip::
-
-   Don't forget the ``|`` after the ``password:`` label.
+   * The content you put just after ``password: |`` must all line up at the same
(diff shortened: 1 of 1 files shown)


willingc:
@choldgraf Totally cool with the change. Makes good sense. The PR looks good to merge.

Travis timed out :( FYI @yuvipanda @minrk

betatim:
[This is the log](https://gist.github.com/betatim/f98ece4512592d5c3460af46b1a3bcfd) from the travis job that timed out. Restarted it now to get a green light.

(Is there another way to keep the travis log when restarting a job besides copy&pasting to a gist?)

Carreau:
It's always the same error timeout of JHub, though info is mostly lost because of Asycn stuff

Issue #388, opened 2018-01-02
URL builder doesn't observe browser autocomplete


ryanlovett:
The URL builder catches everything I type, but it misses my autocomplete choices.

This is with Chrome:
![binder-complete](https://user-images.githubusercontent.com/10248125/34495406-db97c1f2-efa9-11e7-9829-55ff49c39158.gif)

choldgraf:
good catch! I think this would be fixed by adding "change" to the `on(` section of each field:

https://github.com/jupyterhub/binderhub/blob/master/binderhub/static/js/index.js#L200

choldgraf:
@ryanlovett can you check the mybinder.org behavior again now that #404 is merged? You will have to click outside of the text field after selecting your drop-down item, but maybe this is enough to consider this issue addressed?

ryanlovett:
I can confirm that the url updates after clicking outside of the text field, after tabbing, and even after clicking "launch" (!). The correct image was built and run. The url does not update after clicking the copy button however which I think is the one time where having the accurate url matters the most.

choldgraf:
ah so you mean that if you do:

start typing -> select your repo from drop-down -> immediately click copy, then the wrong URL is copied?

ryanlovett:
Yes, exactly. And then if I click anywhere outside the text field the url doesn't update.

choldgraf:
hmmmmm....ok let's leave this issue open then, and I can try giving it another shot later when I've got a moment. Thanks for the feedback @ryanlovett :-)

ryanlovett:
No problem. Once I get out from under the start of the semester I'll take a look too.

manics:
Is this still a problem?

choldgraf:
I'm not sure how to reproduce it - my browser doesn't give me drop-down options to choose from :-/

Maybe @ryanlovett can try it and see if it's still a problem, and if we don't hear back in a day or two we should close this

ryanlovett:
Looks like this has been fixed. Thanks!

PR #404, opened 2018-01-06
adding clipboard clicking

choldgraf:
Now that we're using npm, this uses a little npm package that lets you auto-copy to the clipboard. It does a couple of things:

1. Makes all the icons next to the text fields clickable, which copies the text to the clipboard
2. Updates the prompts on the build page to reflect this
3. Adds a clipboard badge to the main URL text box (and also fixes a bug where the box would run off of the screen if it was too narrow)
4. Mostly fixes the bug that @ryanlovett mentioned in #388

### DIFF
--- a/binderhub/static/index.css
+++ b/binderhub/static/index.css
@@ -276,3 +276,21 @@ h4.logo-subtext {
     }
 
 }
+
+/*Clipboard styling*/
+
+img.icon.clipboard{
+    order: 1;
+    border: thin solid silver;
+    border-radius: 5px;
+    border-bottom-left-radius: 0;
+    border-top-left-radius: 0;
+    border-left: none;
+}
+img.icon.clipboard:hover {
+    background: #f5f5f5;
+}
+
+img.icon.clipboard:active {
+    background: #ddd;
+}
--- a/binderhub/static/index.css
+++ b/binderhub/static/index.css
@@ -169,7 +169,7 @@ h4 {
 }
 
 
-.badge-snippet-row {
+.badge-snippet-row, .url-row {
     width: 100%;
     display: flex;
     flex-direction: reverse;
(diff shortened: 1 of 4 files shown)


Carreau:
Gnnn, how do you iterate on the html/css if you need to run npm now ? I get a page with 0-styling.

Carreau:
I think this needs to have visual feedback that the "clipboad" can be clicked with a `:hover` and `:active`  css state. The image  being on the left, outside of the "<pre>" (like the md and rst logo) convey more a "This is clipboard" than "You can click to copy", something more like that would IMHO be more obvious in the meaning:
screen shot 2018-01-06 at 14 05 03

Carreau:
This css, plus a couple of other tweaks:
```
img.icon.clipboard{
    order: 1;
    border: thin solid silver;
    border-radius: 5px;
    border-bottom-left-radius: 0;
    border-top-left-radius: 0;
    border-left: none;
}
img.icon.clipboard:hover {
    background: #f5f5f5;
}

img.icon.clipboard:active {
    background: #ddd;
}
```
clip

Carreau:
Side note, I think we should have that kind of clipboard icon _in addition_ to md and rst icon in front.

choldgraf:
ah I love the suggestions, will take a stab at adding them in (and agreed about the icons remaining...it took me like an hour to figure out how to enable this package in the first place so this was just a first stab)

choldgraf:
Changes made (I also fixed the caret while I was at it). Here's a GIF:

binder

The only problem I can see is that the text boxes get a little wonky when the text overflows the box, but I feel like that's for another PR

Carreau:
pre border-right-radius=0, and height of the second one is weird because of scrollbar. 
+1 once fixed i'll give it a shot.

Carreau:
I push a css fix (also your clipboard target were wrong).Mind having a look ? I also can't reproduce the issue with the overflow and the scrollbar on OSX, so you'r on your own if you want to fix it.

Carreau:
I just updated MyBinder Staging and Prod. You missed the window.

Carreau:
Ok, i'll merge this and deploy later on Staging.

betatim:
Close #388? The top comment says "mostly fixes" :)

PR #123, opened 2017-09-26
Fix and automate docbuild

willingc:
- fix failing doc build and automate with rtd
- uncovered a [bug within released pygraphviz](https://github.com/pygraphviz/pygraphviz/pull/104) that prevents python3 build; pin to rc version
- improved line continuation style

[rendered](http://test-binderhub.readthedocs.io/en/fix-docbuild/)

PR #405, opened 2018-01-07
Mention which PR triggered build in Helm Chart Commit Message

Carreau:
This should make slightly easier to figure out which PRs are in a helm
chart, and have descriptive PRs when updating mybinder.org-deploy

### DIFF
--- a/ci/test-helm
+++ b/ci/test-helm
@@ -13,7 +13,7 @@ echo -e "hub:\n  url: $HUB_URL" > helm-chart/travis-binder.yaml
 
 echo "building helm chart"
 cd helm-chart
-./build.py --commit-range ${TRAVIS_COMMIT_RANGE}
+./build.py --commit-range ${TRAVIS_COMMIT_RANGE} --extra "${TRAVIS_REPO_SLUG}$(git log -1 --pretty=%B | head -n1 | sed 's/^.*\(#[0-9]*\).*/\1/')"
 cd ..
 git diff
 docker images | sort
--- a/helm-chart/build.py
+++ b/helm-chart/build.py
@@ -155,7 +155,7 @@ def publish_pages(name, paths, git_repo, published_repo, extra_message=''):
             )
     subprocess.check_call(['git', 'add', '.'], cwd=checkout_dir)
     if extra_message:
-        exta = '\n\n%s' % extra_message
+        extra = '\n\n%s' % extra_message
     else:
         extra = ''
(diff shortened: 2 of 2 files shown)


helm-chart/build.py
```diff
@@ -154,10 +154,14 @@ def publish_pages(name, paths, git_repo, published_repo):
                 os.path.join(checkout_dir, f)
             )
     subprocess.check_call(['git', 'add', '.'], cwd=checkout_dir)
+    if extra_message:
+        exta = '\n\n%s' % extra_message
```
choldgraf: extra

ci/test-helm
```diff
@@ -13,7 +13,7 @@ echo -e "hub:\n  url: $HUB_URL" > helm-chart/travis-binder.yaml
 
 echo "building helm chart"
 cd helm-chart
-./build.py --commit-range ${TRAVIS_COMMIT_RANGE}
+./build.py --commit-range ${TRAVIS_COMMIT_RANGE} --extra "${TRAVIS_REPO_SLUG}$(git log -1 --pretty=%B | head -n1 | sed 's/^.*\(#[0-9]*\).*/\1/')"
```
choldgraf: I guess this magically only displays the commits that are being added?

choldgraf:
I think it's a good idea! Can you paste what a sample output looks like? I looked through the travis logs but it doesn't seem to be building there so maybe this only shows up in an actual deployment?

ci/test-helm
Carreau: Well, actually as binderhub is being build for each merged PR, then it only extract the PR number. 

- `${TRAVIS_REPO_SLUG}` is `jupyterhub/binderhub`
- `log -1` logs only the latest commit
- `--pretty=%B`  only the message

As the message is `Merge PR #123 into master`, `sed 's/^.*\(#[0-9]*\).*/\1/')"` extract the number. 

Thus `extra` is only `jupyterhub/binderhub#123`.

It will of course not work if we push commits directly.

yuvipanda:
Can you integrate this into the build.py of z2jh too, @Carreau?

yuvipanda:
(and thanks for the patch!)

Carreau:
> Can you integrate this into the build.py of z2jh too, @Carreau?

Good idea, I'll add on My todo list. 
I'm starting my journey back to the US starting today (with several stops and goodbyes), so it may takes some time for me to be back online.

PR #406, opened 2018-01-07
Include CSS on all pages.

Carreau:
Fixes the fact that error pages right now have no styles...

...  because right now it look like

screen shot 2018-01-07 at 18 40 34


Carreau:
Thanks !

PR #409, opened 2018-01-10
Add note about npm to CONTRIBUTING.md

### LINKED ISSUE #403: Need to run `npm run webpack` before dev now?
Now that #372 is merged, I found that Binder wouldn't build locally anymore because of a `<script>` loading error. I looked at the travis build and noticed that it runs `npm run webpack`. After I ran this, Binder displayed properly.

Is this something that will need to happen (`npm run webpack`) everytime the javascript changes? Either way, we should document this in `CONTRIBUTING.md` yeah?

yuvipanda:
Fixes #403

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,6 +71,18 @@ To run unit tests, navigate to the root of the repository, then call:
 
 We recommend increasing your GitHub API rate limit before running tests (see above).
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
I forgot to add this to the docs when we started depending on npm!

choldgraf:
thanks :-)

Issue #410, opened 2018-01-11
Ocassionally Flaky CI


yuvipanda:
Sometimes CI fails because of an error like:

```
binderhub/tests/test_build.py::test_build[gh/binderhub-ci-repos/requirements/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd] JupyterHub not available at http://127.0.0.1:30123: HTTPConnectionPool(host='127.0.0.1', port=30123): Read timed out. (read timeout=5)
ERROR [ 11%]
```

Restarting usually fixes it. This is quite annoying and needs to be looked into.

willingc:
[Pastebin of detailed Travis log](https://pastebin.com/GZyiG4sj)

yuvipanda:
@willingc nice! Although the errors I saw were for test-helm, rather than test-main I think! But definitely seen both flake.

PR #412, opened 2018-01-11
add async test timeout

willingc:
Addresses #410 

- Increase ASYNC_TEST_TIMEOUT from default to 15 which is what we have been using for Tornado on JupyterHub tests
- Increase sleep to 10s

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ install:
 - pip install . -r dev-requirements.txt
 script:
 - export BINDER_TEST_NAMESPACE=binder-test-$TEST
+- export ASYNC_TEST_TIMEOUT=15
 - ./ci/test-$TEST
 after_failure:
 - for pod in $(kubectl get pod --namespace=$BINDER_TEST_NAMESPACE | awk '{print $1}'); do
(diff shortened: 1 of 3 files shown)


yuvipanda:
Thanks @willingc. These should fix it, but I think a better longer-term approach would be similar to #413:

1. Use 'kubectl rollout' to check when hub is ready, rather than listing pods
2. Make a HTTP request to hub and make sure it responds positively before moving on.

I'm happy to merge this as is too, since it'll fix the issues right now. Mid-to longer term we should do both the above things tor test-main too though.

willingc:
@yuvipanda I think that we will likely need both this and #413. This one should help with the tests that use @pytest.gen while #413 should help with not starting until ready.

yuvipanda:
@willingc oh yeah, I was only talking about the sleep, rather than the ASYNC_TIMEOUT.

yuvipanda:
Going to merge this when tests pass unless you have objections!

willingc:
Go for it @yuvipanda. I'm cool with removing the sleep 10 as well if you want. It was just giving a little extra margin.

yuvipanda:
Thanks for the patch @willingc! I'm happy we're now treating all CI flakes seriously :)

PR #417, opened 2018-01-18
updating binderhub diagram

choldgraf:
This updates the BinderHub diagram that is currently being generated by a python script, the script itself is kinda buggy and results in less-than-ideal diagrams both aesthetically and structurally. 

This is the current diagram page:

https://binderhub.readthedocs.io/en/latest/diagram.html

and current diagram:

This changes it to:

Using an image generated from this slides presentation:

https://docs.google.com/presentation/d/1t5W4Rnez6xBRz4YxCxWYAx8t4KRfUosbCjS4Z1or7rM/edit#slide=id.g25dbc82125_0_53

feel free to give feedback / make edits to the google slide!


choldgraf:
Another option is that for things like images that may change semi-frequently as time goes on, we could put these online in a static location and link to them in the docs rather than embedding them in the git repository...that could save on space in the long-term

betatim:
I had to read this several times to grok it. So to check I got it my summary: remove the SVG which is being generated by a script. Use a PNG instead which is "generated" from the slides you linked.

If yes, then I agree with you. I'd keep the PNG in the repo and at the point where it is included in the docs make a comment (can you create comments in rST?) that links to the google slides that were used to make the image.

choldgraf:
@betatim yep, that's the plan...sounds good re: embedding it in the repo. I'll try to make the top-level comment more clear

choldgraf:
also, it looks like the new image didn't render, which probably contributed to confusion :-/

choldgraf:
@betatim the problem was that I had linked to my site w/ the updated diagram, but then I made another PR on binderhub that un-did all of those changes. Whoops! The image is now linked to my branch in github so it shouldn't change, LMKWYT!

betatim:
I like the new diagram. One thing that is a bit weird is that the entry point to the diagram is the bottom left. Together with an arrow pointing away from the GitHub logo that is where I started looking, to then quickly think: hold on a second, where is the user? My best idea is to make the users a brighter colour so they attract attention?

choldgraf:
@betatim I think that's a good idea! FWIW, the slides should be editable by anybody, so I totally welcome an update that hits what you're imagining

betatim:
What do you think of this splash of colour?

screen shot 2018-01-19 at 23 50 46

choldgraf:
I like it! What if we also added the Binder logo there:

image

betatim:
🎨! I removed the "(beta)" from the logo, it is super small and looks like "dirt" here IMHO. Also created a new slide as a place to collect the building blocks of the previous slides.

screen shot 2018-01-20 at 08 43 03

choldgraf:
Great idea, I agree

choldgraf:
ok, I updated the diagram in the latest commit, incorporating your changes + flipping the "users" pool and the legend, hopefully that'll make it clearer how to follow the flow. LMKWYT!

betatim:
Such art, much colour, wow.

choldgraf:
very flow chart
--

PR #418, opened 2018-01-18
Using z2jh cloud instructions for k8s and helm

choldgraf:
In #402 we realized that a lot of the k8s and helm instructions had become out of date. This PR adds a nifty rST feature to include parts of another text file. It grabs relevant parts of the latest version of the z2jh install instructions and injects them into our install instructions here (we should pin it to `v0.6-doc` when that is released). That way these docs will not get out-of-sync with z2jh like they are now.

I can confirm that following these docs resulted in a functioning BinderHub!

Here's an example of the docs:

http://predictablynoisy.com/binderhub/create-cloud-resources.html

Closes #402

PR #419, opened 2018-01-20
Update the version of the repo2docker image

betatim:
I think the failures of travis (like [this](https://github.com/jupyterhub/binderhub/pull/418)) at the moment is because binderhub is not yet using the [new version of repo2docker](https://github.com/jupyter/repo2docker/pull/197).

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -215,7 +215,7 @@ class BinderHub(Application):
     )
 
     builder_image_spec = Unicode(
-        'jupyter/repo2docker:235f0ad',
+        'jupyter/repo2docker:6e9088d',
         help="""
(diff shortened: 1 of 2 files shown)


betatim:
Could we configure a travis build that uses the latest repo2docker image and "master" of our other dependencies?

betatim:
Travis fails because `npm install` as part of `pip install binderhub` fails :-/

betatim:
Ready to merge

minrk:
> Could we configure a travis build that uses the latest repo2docker image and "master" of our other dependencies?

that should be doable. We ought to only need to:

1. set repo2docker tag to `:latest`
2. discover and set latest jupyterhub chart tag

PR #420, opened 2018-01-22
add details to CONTRIBUTING.md

### LINKED ISSUE #344: Improve CONTRIBUTING.md
I finally figured out how to install virtualbox on my Ubuntu OS (maybe we should document this, it took me like a day to figure out the right incantation).

Just went through the contributing.md code, here are some quick improvements:

## Install steps
1. Steps 5 and 6 should be swapped (we need to install the python requirements before `install-hub` since it needs `kubernetes`.
2. Step 5 and 8 could use more context, doesn't have to be a ton, but it'd be helpful to know what we're doing here.
3. In step 1, we should split up "install minikube" and "start minikube"

## Debugging tips
Not sure how common this is, but:

* I got a DiskAvailable error (something like that) that @yuvipanda said was because the minikube disk was full. In this case, I ran `minikube delete` followed by re-installing binderhub following the guide, and it worked. On this note, it may be worth mentioning the ``--disk-size`` parameter as it seems to be a common problem.

GladysNalvarte:
explain more about what each step does
and fix order of dev-requirements which must be installed before running install-hub

closes #344

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ on GitHub if you don't have a token.
 
       python3 -m pip install -e . -r dev-requirements.txt
           
-  This list of packages is necessary to create an environment that will generate the Docker image using the Git repository. Regardless of what is in the setup.py file, it will install what the user needs to build the Docker image.
+  This list of packages is necessary to create an environment that will generate the Docker image using the Git repository. Regardless of what is in the setup.py file, the requirements file will install what the user needs to build the Docker image.
   
 6. Install JupyterHub in minikube with helm
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,18 @@ To run unit tests, navigate to the root of the repository, then call:
 
 We recommend increasing your GitHub API rate limit before running tests (see above).
 
+## Building JS and CSS
+
+We use [npm](https://www.npmjs.com) for managing our JS / CSS dependencies and
+[webpack](https://webpack.js.org/) for bundling them together. You need to have
+a recent version of `npm` installed to run things locally.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -71,18 +86,6 @@ To run unit tests, navigate to the root of the repository, then call:
 
 We recommend increasing your GitHub API rate limit before running tests (see above).
 
-## Building JS and CSS
```
betatim: Do we not need this any more? Maybe we should leave a pointer here for how to rebuild just the JS and CSS for people exclusively working on the frontend.

CONTRIBUTING.md
```diff
@@ -34,26 +36,39 @@ on GitHub if you don't have a token.
...
-
-        ./testing/minikube/install-hub
-
-6. Install binderhub and its development requirements:
+5. Install binderhub and its development requirements:
 
-        python3 -m pip install -e . -r dev-requirements.txt
+      python3 -m pip install -e . -r dev-requirements.txt
+          
+  This list of packages is necessary to create an environment that will generate the Docker image using the Git repository. Regardless of what is in the setup.py file, it will install what the user needs to build the Docker image.
```
willingc: The list of packages in `dev-requirements.txt` will be used to create a suitable environment for generating the Docker image for the  Git Repository. Regardless ... ~it~  the requirements file will ...

choldgraf:
I went ahead and made @willingc 's patch to the branch. I think this is good to go! Thanks so much @GladysNalvarte - this kind of practical documentation is really helpful in building the community!

betatim:
🎉 !

(travis failed because you didn't wait for it or because it actually failed?)

choldgraf:
it was because of an eager merge on my part - travis passed and there was a one-word edit that I added to the doc before merging, my bad for not updating the final comment with that information

PR #431, opened 2018-01-31
Fix formatting issues in Contributing.md

jroggeman:
- Use consistent monospace styling for code blocks
 - Use consistent spacing for indentation of the code blocks (I stuck with three spaces as it was used consistently for the first four or so numbered items, but can change it if desired).


jroggeman:
Looks like build failed due to a transient networking issue in build script; can someone rerun for me? Rerun isn't available for non-contributors IIRC

PR #434, opened 2018-02-02
fix redirect url when given a url that already has a query

minrk:
join with '&' when '?' is already there starting an existing query.

cc @annefou @GladysNalvarte


choldgraf:
🎉

PR #436, opened 2018-02-05
add nodeSelector configuration for binder pod

bitnik:
Hi, currently I don't see a way to assign the binder pod to a specific node via helm config file. So this change makes it possible to set `nodeSelector` in helm configuration.

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -30,6 +30,7 @@ spec:
         checksum/config-map: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
         checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
     spec:
+      nodeSelector: {{ toJson .Values.nodeSelector }}
       {{- if .Values.rbac.enabled }}
       serviceAccountName: binderhub
       {{- end }}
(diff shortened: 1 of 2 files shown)


choldgraf:
Thanks for the addition - could you provide some simple explanation as docs / comments / etc to give context for the addition?

Also on a more general note, I'm a bit concerned that this will cause confusion for some people. For example, what if they choose a node that then gets deleted for some reason? We regularly have to cycle through nodes at `mybinder.org` as errors etc pop up.

bitnik:
Hi, in our case we have a kubernetes cluster created with kubeadm. In this cluster we have 3 nodes (1 master and 2 workers) and they mostly do not change. If we don't have `nodeSelector' option in helm configuration, binder pod is created on 1 of the nodes. But we want to have binder pod in the node which has more resources.

To overcome this problem currently we download the helm chart of BinderHub from [here](https://github.com/jupyterhub/helm-chart/tree/gh-pages), do the changes and install BinderHub.

Setting `nodeSelector` is already possible in installing JupyterHub with helm: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/values.yaml - so we only have problem with Binder pod.

Regarding your concern, if you don't want to specify the node, you can simply exclude 'nodeSelector' in config.yaml and it will behave as before.

yuvipanda:
Thanks for the patch! Look forward to more :)

PR #437, opened 2018-02-05
Better links on empty input.

Carreau:
If the repo links is emptied by the user, the various badges/links are
in inconsistent state.

Add logic to revert to original message "Fill in the fields to see a URL
for sharing your Binder." and similar.

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -161,9 +161,16 @@ function updateUrl() {
 
 function updateUrlDiv() {
   var url = updateUrl()
-  $('#basic-url-snippet').text(url);
-  $('#markdown-badge-snippet').text(markdownBadge(url));
-  $('#rst-badge-snippet').text(rstBadge(url));
+  if ((url||'').trim().length > 0){
+    $('#basic-url-snippet').text(url);
+    $('#markdown-badge-snippet').text(markdownBadge(url));
+    $('#rst-badge-snippet').text(rstBadge(url));
+  } else {
+    ['#basic-url-snippet', '#markdown-badge-snippet', '#rst-badge-snippet' ].map(function(item, ind){
+      var el = $(item)
+      el.text(el.attr('data-default'));
+    })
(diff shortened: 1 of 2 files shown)


betatim:
Can we make it so we don't have two copies of the same message (for example here https://github.com/Carreau/binderhub/blob/32b3a659e63c49db6013ce7e04cd01fc4d4a6659/binderhub/templates/index.html#L75) that we need to keep in sync?

Otherwise LGTM.

choldgraf:
I'm +1 on merge - won't do it since @betatim asked for changes but IMO that shouldn't be blocking on this PR since we'll be refactoring the javascript at some point anyway

Carreau:
What about now ?

choldgraf:
TBH this feels like over-engineering to me, but I am +1 on merge either way

Carreau:
> TBH this feels like over-engineering to me, but I am +1 on merge either way

I don't see really good way to get the value in only one place without something like that.

choldgraf:
> I don't see really good way to get the value in only one place without something like that.

I agree - that wasn't a comment on this particular implementation I just think it's fine to have two separate text box messages in this case :-)  but I don't feel strongly about it!

betatim:
Was just a suggestion/fishing for a cheap way to refactor this. If there isn't then let's merge.

choldgraf:
ah yeah - I wish GitHub had a way to convey "suggestion but feel free to merge without it" more effectively. Sorry if I came across as annoyed - I've just seen a lotta matplotlib PRs that were 99.9% done go stale for weeks because a well-meaning person suggested a minor non-blocking edit. I think I have a bit of PTSD from this :-)

restarted travis to see if it was a travis problem.

betatim:
Merged!

---

Generally all my comments are just suggestions. I am 99% happy if all I get is a reply "yeah, thought about that but can't find a good way"

choldgraf:
woot!  👍

PR #438, opened 2018-02-07
Make the timeout at which 'too long' message appears

yuvipanda:
Current timeout of 30s is too short, since a lot of repositories
definitely take longer than that

Ref #433


yuvipanda:
A bunch of other long term work needs to happen here, but this is a quick fix to make less people think binder has 'frozen' and does not work anymore.

/cc @jasongrout who also ran into this earlier today.

yuvipanda:
This is especially problematic for our new R support, since lot of R libraries compile from source and takes a long time.

choldgraf:
I think this is definitely a good idea, was meaning to do this sometime soon as well!

PR #439, opened 2018-02-07
Don't pause spinner when code is just loading slowly

yuvipanda:
Instead, pause it when launch actually fails

Ref #433

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -262,7 +262,7 @@ $(function(){
         if ($('div#loader-text').length > 0) {
             $('div#loader-text p').text("Loading repository: " + repo)
             window.setTimeout( function() {
-                $('div#loader-text p').html("Repository " + repo + " is taking a long time to load!<br />See the logs for details.")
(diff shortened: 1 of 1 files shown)


choldgraf:
I think @Carreau should comment on this one - the reason the change was made was in order to avoid using too much CPU, which he was concerned about

yuvipanda:
Updated wording, @betatim

yuvipanda:
If the spinner is taking up too much CPU, IMO we should remove the spinner or fix it. I think currently it is making mybinder.org look much more unstable than it is, which makes for a very sad panda.

Carreau:
Let's remove the timeout (for now), and try to add a metric for "after how long does a user leave the page if it's not launching".

That would give us some measure of what could be reasonable for an alternative UI or timeout.

We can also stop the spinner when the windows loose focus.

Infinite animation on the web sucks.

choldgraf:
stopping the animation when the window looses focus after the timeout seems like a good compromise to me, though I think it we should leave that for another PR...merging!

PR #444, opened 2018-02-08
[MRG] Add test for case insensitive repo ban matching

betatim:
Follow up #443 to add a test to stop the bug from coming back.


yuvipanda:
yay for moar tests

PR #445, opened 2018-02-08
[MRg] Update repo2docker version to 687788f

betatim:
This brings in the latest release of jupyterlab and the notebook.


yuvipanda:
Done! :)

Remember to also bump in mybinder.org-deploy!

PR #453, opened 2018-02-13
adding more buckets to prometheus launch times

choldgraf:
This gives us more fine-grained buckets for the launch / build times, as well as adds more repo information to the prometheus logging labels.

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -21,10 +21,9 @@ from prometheus_client import Histogram, Gauge
 
 from .base import BaseHandler
 from .build import Build, FakeBuild
-
-
-BUILD_TIME = Histogram('binderhub_build_time_seconds', 'Histogram of build times', ['status'], buckets=[1, 5, 10, 30, 60, 300, 600, float("inf")])
-LAUNCH_TIME = Histogram('binderhub_launch_time_seconds', 'Histogram of launch times', ['status'], buckets=[1, 5, 10, 30, 60, 300, 600, float("inf")])
+BUCKETS = [2, 5, 10, 15, 20, 25, 30, 60, 120, 240, 480, 960, 1920, float("inf")]
+BUILD_TIME = Histogram('binderhub_build_time_seconds', 'Histogram of build times', ['status'], buckets=BUCKETS)
+LAUNCH_TIME = Histogram('binderhub_launch_time_seconds', 'Histogram of launch times', ['status'], buckets=BUCKETS)
 BUILDS_INPROGRESS = Gauge('binderhub_inprogress_builds', 'Builds currently in progress')
 LAUNCHES_INPROGRESS = Gauge('binderhub_inprogress_launches', 'Launches currently in progress')
 


choldgraf:
hmm...I'm not sure why travis isn't happy, it's returning

`ValueError: Incorrect label names`

I assume on the call to this:

https://github.com/jupyterhub/binderhub/pull/453/files#diff-fc58e34e6c3a4b92c753d3608b98abf3R391

betatim:
Travis is now happy?! Merge?

choldgraf:
_shrug_ I'm not really sure what the team thinks about this change haha, I've just been on a prometheus bender lately...

binderhub/builder.py
```diff
@@ -22,9 +22,9 @@
 from .base import BaseHandler
 from .build import Build, FakeBuild
 
-
-BUILD_TIME = Histogram('binderhub_build_time_seconds', 'Histogram of build times', ['status'], buckets=[1, 5, 10, 30, 60, 300, 600, float("inf")])
-LAUNCH_TIME = Histogram('binderhub_launch_time_seconds', 'Histogram of launch times', ['status'], buckets=[1, 5, 10, 30, 60, 300, 600, float("inf")])
+BUCKETS = [int(round(1.6**ii)) for ii in range(15)] + [float("inf")]
```
minrk: In prometheus/grafana, is there value in these being round numbers? i.e. is it useful to be able to say "build time < 60"?

This would change the buckets to:

    [1, 2, 3, 4, 7, 10, 17, 27, 43, 69, 110, 176, 281, 450, 721, inf]

For instance, [this chart](https://grafana.mybinder.org/dashboard/db/service-level-objectives?refresh=1m&panelId=2&fullscreen&orgId=1) references each bucket by name, so changing the buckets here would require whitelisting each new bucket in that chart data. Is there a way in the charts to do thresholding on the `le` value as a number? So that the 'builds completed in 60 seconds' chart wouldn't need to change whenever we update the buckets?

choldgraf: I'm a bit unclear on what you're suggesting, but the clunky thing is that grafana / prometheus can't think of the label names as "numbers" only as strings. If you have an intuition for a way to improve on that I'm all for it. I agree w/ you that it's annoying these labels are treated as unique names instead of as numbers :-/

minrk: Sorry. I've run into the issue you are describing, where histogram bucket names are only useful as strings, so I couldn't figure out how to do a good '% > 60' graph like you made on the SLO page with these updated buckets. Adding lots of buckets of non-round numbers makes producing that graph more challenging (especially since 60 isn't a bucket boundary anymore), unless someone has a good mechanism to do numerical aggregation on prometheus bucket values that I haven't figured out.

I was mainly wondering which is more valuable: more resolution in the histogram, or nice round numbers on the y-axis of hists + relative ease of maintaining the over-under graph.

choldgraf: TBH I think it'd be better for us to just manually add a few more buckets in this case - I agree w/ you this is suboptimal but is maybe a better compromise. FWIW I looked into this on the grafana repo and it go anywhere :-/ https://github.com/grafana/grafana/issues/10954

choldgraf:
the latest push reverts to manually specifying buckets, but adds more. It's roughly linear in increments of 5 up to 30 seconds, then doubles each time after that. @minrk what do you think?

choldgraf:
I think we are too often approving PRs without merging :-)

PR #454, opened 2018-02-14
Ensure build and image names are valid

GladysNalvarte:


choldgraf:
Hey @GladysNalvarte - could you:

1. Add a description to this PR so that we get an idea for what it's accomplishing and why
2. Add a short docstring to your function to convey mostly the same information as 1 ?

minrk:
Relevant notes:

1. this catches a bug where build slug may not produce a valid image name (e.g. repo name ending with `_`, which results in image name ending with '-' which is invalid)
2. this ensures that the image name is always safe, regardless of build slugs returned by providers (rather than requiring all providers to return image-safe build slugs below a certain length)
3. since this changes the image name generation scheme, all existing cached images will be invalidated.

GladysNalvarte:
Thanks Min that was very precise!

binderhub/builder.py
```diff
@@ -5,7 +5,7 @@
 import hashlib
 from http.client import responses
 import json
-import string
+import  string
```
betatim: spurious space between import and string

binderhub/builder.py
```diff
@@ -111,25 +111,42 @@ def _generate_build_name(self, build_slug, ref, prefix='', limit=63, hash_length
...
         # escape parts that came from providers (build slug, ref)
         # only build_slug *really* needs this (refs should be sha1 hashes)
         # build names are case-insensitive because ascii_letters are allowed,
         # and `.lower()` is called at the end
         safe_chars = set(string.ascii_letters + string.digits)
         def escape(s):
             return escapism.escape(s, safe=safe_chars, escape_char='-')
-        build_slug = escape(build_slug)
+        
+        build_slug = self._safe_build_slug(build_slug, limit= limit - len(prefix) - ref_length - 1)
```
betatim: `limit= limit` -> `limit=limit`

betatim:
Looks good to me. If you have a case with which you can reproduce the bug you mentioned I'd add a test so that we will notice in the future if it ever comes back.

GladysNalvarte:
@betatim, when you write the name of your repo in GitHub if you use any of these characters  ( / , : *) at the end of the name, all those will be automatically substituted by - (dash) but the underscore was not substituted by GitHub so you can have repo names like (GladysNalvarte/mean_) and in builder.py  the repo name was not expected to end in any other character than the normal ones or dash.  Therefore, the underscore at the end of the repo name caused binder to stop building the image. Now Min fixed it and I hope that all the options are covered now.

minrk:
Thanks @GladysNalvarte!

PR #455, opened 2018-02-16
add nodeSelector configuration for build pod

bitnik:
Hi, with this change it is possible to assign build pods to a specific node by setting `buildNodeSelector`. The reason of the change is same as https://github.com/jupyterhub/binderhub/pull/436. If `buildNodeSelector` is not specified in config.yaml, it will behave as before.

With this change it is also possible to run binderhub in multi-node mode without using a container registry by assigning all binderhub pods to the same node.

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -234,6 +234,15 @@ class BinderHub(Application):
         config=True
     )
 
+    build_node_selector = Dict(
+        default_value=None,
+        allow_none=True,
+        config=True,
+        help="""
+        Select the node where build pod runs on.
+        """
+    )
+
     repo_providers = Dict(
         {
             'gh': GitHubRepoProvider,
@@ -326,6 +335,7 @@ class BinderHub(Application):
             'launcher': self.launcher,
             "build_namespace": self.build_namespace,
             "builder_image_spec": self.builder_image_spec,
+            'build_node_selector': self.build_node_selector,
             'build_pool': self.build_pool,
             'per_repo_quota': self.per_repo_quota,
             'repo_providers': self.repo_providers,
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -235,8 +235,7 @@ class BinderHub(Application):
     )
 
     build_node_selector = Dict(
-        default_value=None,
-        allow_none=True,
+        {},
         config=True,
         help="""
         Select the node where build pod runs on.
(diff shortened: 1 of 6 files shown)


bitnik:
[Key-value pairs are the labels of the nodes](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector). And yes if the node is deleted, helm config must be updated or the new joined node must have the same label. Btw if you do not specify the `buildNodeSelector`, it will work as before (build pod will run on any node). Actually this is same as  and as dangerous as specifying `nodeSelector` for [jupyterhub](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/values.yaml#L9), [singleuser](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/values.yaml#L129) and [binderhub](https://github.com/jupyterhub/binderhub/blob/master/helm-chart/binderhub/values.yaml#L10).

I think danger of specifying this field depends on kubernetes setup. For us, we have a kubernetes cluster created with kubeadm and our nodes rarely changes and assigning pods to a node is sometimes very useful. But if in a setup nodes change frequently, probably it is dangerous as you said.

binderhub/app.py
```diff
@@ -234,6 +234,15 @@ def _add_slash(self, proposal):
         config=True
     )
 
+    build_node_selector = Dict(
+        default_value=None,
+        allow_none=True,
```
minrk: rather than default=None, allow_none, we can leave the default as an empty dict, right?

bitnik: right. i am going to change it.

helm-chart/images/binderhub/binderhub_config.py
```diff
@@ -28,6 +28,7 @@ def get_config(key, default=None):
 c.BinderHub.per_repo_quota = get_config('binder.per-repo-quota', 0)
 
 c.BinderHub.builder_image_spec = get_config('binder.repo2docker-image')
+c.BinderHub.build_node_selector = get_config('binder.build-node-selector', None)
```
minrk: Now it should be `{}` here, I think.

bitnik: sorry, I missed it

minrk:
Thanks!

PR #459, opened 2018-02-19
support repo2docker appendix

### LINKED ISSUE #458: When I can't access a pod, tell me how to create a new pod from the same public repo
(originally suggested [on Twitter](https://twitter.com/RaphaelDasG/status/964641858054844419))

When someone [has published a personal pod URL<sup>1</sup>](https://twitter.com/pythonsummit/status/964442816637296640) (or has shared it with me, see #160), I won't (and probably shouldn't) be able to access that pod.

Chances are, that they didn't want to give me access to their personal pod, but instead to a new pod based on the same publicly hosted notebook file or repo, which they could have achieved by sharing the ["binder link"<sup>2</sup>](https://mybinder.readthedocs.io/en/latest/introduction.html#generating-a-sharable-binder-link).

Thus, the error page I land on when following the pod URL should explain the difference between these two (if possible, without requiring prior knowledge of kubernetes lingo or other implementation details) **and tell me where to get to a new pod for that notebook file or repo** (i.e. the corresponding sharable binder link).

(Could be combined with #457, but either suggestion could be implemented without the other, so I'm filing them as separate issues.)

---
<sup>1</sup>On the mybinder service, an URL beginning with <code>https://<strong>hub.</strong>mybinder.org</code> &hellip;
<sup>2</sup>On the mybinder service, the URL of form `https://mybinder.org/v2/<provider-name>/<org-name>/<repo-name>/<branch|commit|tag>?filepath=<path/to/notebook.ipynb>`

minrk:
Raw repo2docker appendix can be set via `c.BinderHub.appendix`.

A default appendix ships in the binderhub repo in the `appendix` directory. `Setting BinderHub.binderhub_appendix_url` to a git archive URL,
e.g. `https://github.com/minrk/binderhub/archive/appendix.tar.gz`, will run the default appendix from the binderhub repo, which does:

- download binderhub_appendix_url archive
- untar `/appendix` subdirectory
- execute `/appendix/run-appendix`

In the appendix added here, that results in some extra config that sets a custom `login` page, which is the page you are shown when you don't have access to a Binder (e.g. someone shared a link to a running pod, as in #458). The custom login page eliminates any token/password form, replacing it with a message and link to a new Binder for the current repo. It could also redirect to that link automatically, or do so after some delay, giving people a chance to read the message.

Happy to have some help defining what should be communicated on that page. It's pretty sparse right now.

closes #458 for the case where the pod is still running, but not the case where the pod has stopped, in which case JupyterHub serves a 403 page and doesn't have enough info to reconstruct the original URL.

TODO:

- [x] decide if the appendix belongs in this repo, standalone, or in a deployment-specific repo, like mybinder.org-deploy
- [x] add appendix support and defaults to helm charts


minrk:
This also sets the stage for installing a Jupyter extension that has a copyable share link that will point to the right Binder URL.

minrk:
I think this should be ready. While there is an appendix in this repo, it is not enabled by default (in large part because there's no easy way to build the URL for the right version).

appendix/templates/login.html
```diff
@@ -0,0 +1,10 @@
+{% extends "templates/login.html" %}
+{% block site %}
+
+<div id="ipython-main-app" class="container">
+    <h1>You don't have access to this Binder!</h1>
```
betatim: ! -> . ? (change ! to . maybe?)

Other ideas that might work: "This isn't the binder you are looking for." or "You can't access this Binder." or "This Binder belongs to someone else."

appendix/templates/login.html
```diff
@@ -0,0 +1,10 @@
...
+    <h2>
+        You can get a new Binder for this repo by clicking <a href="{{binder_url}}">here</a>
```
betatim: Prefix with: "If you followed a link here that someone shared with you, you can get a new ..."? I'm wondering if someone who just followed a link here (and doesn't know that is the wrong link -> a newcomer) will understand that what they want is a new binder. How do we phrase it so that someone who is new, now confused because they see an "error page" follows the link instead of shrugging and closing the tab.

binderhub/app.py
```diff
@@ -98,6 +110,41 @@ def _valid_base_url(self, proposal):
...
+        Appendix to pass to repo2docker
+
+        A multi-line string of Docker directives to run.
+        Since the build context cannot be affected,
+        ADD will typically not be useful.
+
+        This should be a Python string template.
+        It will be formatted with at least the following names available:
+
+        - binder_url: the binder URL for (e.g. for linking)
```
betatim: too many fors? Or "the binder URL for this repository"?

binderhub/app.py
```diff
@@ -98,6 +110,41 @@ def _valid_base_url(self, proposal):
...
+        - repo_url: the repository URL used to build the image
+        """,
+        config=True,
+    )
+
+    binderhub_appendix_url = Unicode(
+        help="""url of binderhub git-archive tarball for using the default appendix
```
betatim: "URL of the git-archive tarball to use as the default appendix"?

Why default? The user/repo creator can't change this so this is _the_ appendix for this binderhub instance no?

I'd change L133 to "the contents of `appendix/` from the Binderhub repository will be used if this is not set"

binderhub/app.py
```diff
@@ -98,6 +110,41 @@ def _valid_base_url(self, proposal):
...
+
+        appendix/appendix.sh in this repo will be used if set.
```
betatim: should be `run-appendix` no? (If we don't replace it due to above comment)

binderhub/builder.py
```diff
@@ -215,6 +215,7 @@ def escape(s):
                 image_found = False
             else:
                 image_found = True
+            image_found = False
```
betatim: left over from debugging?

helm-chart/images/binderhub/binderhub_config.py
```diff
@@ -29,6 +29,12 @@ def get_config(key, default=None):
 
 c.BinderHub.builder_image_spec = get_config('binder.repo2docker-image')
 c.BinderHub.build_node_selector = get_config('binder.build-node-selector', {})
+appendix = get_config('binder.appendix')
+if appendix:
+    c.BinderHub.appendix = appendix
+else:
+    c.BinderHub.binderhub_appendix_url = get_config('binder.appendix-url', '')
```
betatim: Why do we only set this `if not appendix`? My first guess would have been that you can set appendix and `appendix_url` independently of each other. A custom appendix might not use `appendix_url` but ok. I think I have a slight preference for letting people set a custom appendix and use the appendix_url config option. But I couldn't justify it with a use-case right now :-/

If we don't change it add a comment saying "We assume that a custom appendix already contains all the right URLs and information".

betatim:
Should a failure to download/run the appendix lead to a failed build? Maybe we should be forgiving especially in the case where we append it to a custom `Dockerfile`. Just thinking that some files might not be where we think they should be or that `wget` isn't installed or something crazy like that. Right now it will lead to very weird errors and "what the whats??" from the user who tested their stuff wit h`repo2docker` and now sees errors from things that "don't exist in my Dockerfile".

I am confused about the default behaviour of binderhub wrt appendix. The helm chart sets it to empty values, but the default appendix in the config/traitlet isn't empty, however the default URL is, I think. So default bhub without helm chart will crash? Will read this again tomorrow.

minrk:
> Should a failure to download/run the appendix lead to a failed build?

Yes, I believe so. Failure to run the appendix ought to mean there's a problem. We could disable the appendix for Dockerfiles.
Yes, I believe so. `wget` being unavailable isn't that uncommon, but *both* wget and curl being unavailable is exceedingly rare. Making the append

> I am confused about the default behaviour of binderhub wrt appendix. 

The default is no appendix. The default appendix and appendix URL are both empty. There are two ways to enable an appendix:

1. set a raw appendix URL directly
2. specify a URL to a tarball, which will generate an appendix

I'm going to remove the shortcut from binderhub and require specifying the raw appendix. I think that will be clearer.

minrk:
OK, should be ready again. Removed support for the appendix-from-url since I think that was confusing. Instead, I replaced that with an example. I've been able to test and confirm that this works.

The default appendix behavior is no appendix at all. I don't see a good way for the appendix to come from this repo by default because there's no way to add to the docker build context in the build pod. We can only add docker directives, which means any files must come from URLs.

betatim:
LGTM modulo https://github.com/jupyterhub/binderhub/pull/459/files#r172680992, silly comment or overlooked comment?

binderhub/builder.py
minrk: oh geez, yes. This would be a problem! Fixed.

minrk:
Missed that one, thanks!

examples/appendix/extra_notebook_config.py
```diff
@@ -0,0 +1,3 @@
+import os
+c.NotebookApp.extra_template_paths.append('/etc/jupyter/binder_templates')
+c.NotebookApp.jinja_template_vars.update({'binder_url': os.environ.get('BINDER_URL', 'https://mybinder.org')})
```
willingc: We may wish to add comment to the example here clarifying if the 'BINDER_URL' is not set within the environment then mybinder.org will be used as a default. We can address this in a follow on PR with the UI text updates.

examples/appendix/templates/login.html
```diff
@@ -0,0 +1,25 @@
+{% extends "templates/login.html" %}
+{% block site %}
+
+<div id="ipython-main-app" class="container">
+  <h1>You can't access this Binder</h1>
```
willingc: I would like to iterate on the text a bit in a follow up PR for end user clarity/usability.

PR #460, opened 2018-02-20
ensure binderhub has pycurl

minrk:
using pycurl solved timeout issues with jupyterhub in the past when there are many concurrent requests, so hopefully it will help us out with current timeout issues


minrk:
This may help with timeout errors we are seeing now that binder is under load

willingc:
Looks good. Thanks @minrk

PR #461, opened 2018-02-20
use a semaphore to limit concurrent requests to the Hub API

minrk:
Too many concurrent AsyncHTTPClient requests can result in TimeoutErrors

limit enforced with a Semaphore on Launcher

switching to pycurl may have alleviated this, but let's double-up the protection.

use `c.Launcher.concurrency` config to change the limit. Default is 10.

TODO:

- [x] expose limit config via helm chart


minrk:
I don't think we need this

PR #463, opened 2018-02-22
Switch JupyterHub chart version to 0.6

yuvipanda:
The version we are using suffered from a severe performance
bug, causing very high CPU usage!

See https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/461
and https://github.com/jupyterhub/kubespawner/pull/125

### DIFF
--- a/helm-chart/binderhub/requirements.yaml
+++ b/helm-chart/binderhub/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: jupyterhub
-  version: "0.6.0-dcb92ba"
+  version: "0.6.0"
   repository: "https://jupyterhub.github.io/helm-chart"


helm-chart/binderhub/requirements.yaml
```diff
@@ -1,4 +1,4 @@
 dependencies:
 - name: jupyterhub
-  version: "0.6.0-dcb92ba"
+  version: "0.6.0"
```
betatim: So this pins us to the 0.6.0 chart from Jan 31st right? Why not jupyterhub-v0.6-935f86d which is the latest dev chart and what I would have gone to per the deployment instructions after updating something in the jhub chart.

yuvipanda: Because we know 0.6.0 works and is the 'released' one. Any further charts should actually be v0.7- (we should do a bump). We can bump those later, but I generally prefer minimal changes

betatim: Actually, ignore this. I was thinking about the mybinder.org deployment procedure which is all about updating the version of the binderhub chart mybinder.org depends on. Not what version of the jhub chart the binderhub chart depends on.

PR #468, opened 2018-02-23
Update JupyterHub helm chart version

betatim:
The new version of the chart brings in a new culler that is better at
handling users that are slow to stop. See
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/531

### DIFF
--- a/helm-chart/binderhub/requirements.yaml
+++ b/helm-chart/binderhub/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: jupyterhub
-  version: "0.6.0"
+  version: "v0.7.0-2cdcb44"
   repository: "https://jupyterhub.github.io/helm-chart"


helm-chart/binderhub/requirements.yaml
```diff
@@ -1,4 +1,4 @@
 dependencies:
 - name: jupyterhub
-  version: "0.6.0"
+  version: "v0.7-091be8d"
```
betatim: "v0.7..." or "0.7..."? Checking.

minrk:
There's a bug in our build.py when building multiple images. It's reusing the last commit from the first image as the tag for subsequent images. This is fixed in the new standalone chartpress, which is picked up in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/533 . If we merge that PR, then the next build of the jupyterhub chart should have the right images.

PR #476, opened 2018-03-03
support handcrafted links to non-github binders

### LINKED ISSUE #475: GitRepoProvider seems to redirect to github?
Hi,

I have a binder-ready repository which you can clone `https://gitlab.lrde.epita.fr/adl/binder-test.git`, and I would like to try commit `304d0e858f7ad031fc83cf0eb288b954c2d7123e`.

For the [documentation of `class GitRepoProvider'](https://github.com/jupyterhub/binderhub/blob/master/binderhub/repoproviders.py#L114), I understand I should use the following link to launch the binder:

https://mybinder.org/v2/git/https%3A%2F%2Fgitlab.lrde.epita.fr%2Fadl%2Fbinder-test.git/223304d0e858f7ad031fc83cf0eb288b954c2d7123e

However, clicking on this link unexpectedly redirects me to `https://mybinder.org/v2/gh/https://gitlab.lrde.epita.fr/adl/binder-test.git/master` which (unsurprisingly) fails to load.

adl:
Here is my attempt to fix #475.

The `index.html` form still support only GitHub repositories, but I believe we should now be able to handcraft links to binders with the gitlab or git providers, as I was trying to do.

I guess that someone else can later upgrade the form to support other git providers by making it possible to change the `provider_prefix` input.

Now, I've only tested this by following the [pure HTML/CSS/JS dev settup](https://github.com/jupyterhub/binderhub/blob/master/CONTRIBUTING.md#pure-html--css--js-development) and adding and completing the `repo_providers` in `testing/localonly/binderhub_config.py`.  Then I tried the link given in #475 and made sure it would remain unchanged by the javascript code.  After that the loading fails with `Internal Server Error`, which I guess is expected.  (Setting up the full testing environment seems way too complicated for some evening hack.)

### DIFF
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -41,6 +41,7 @@ class ParameterizedMainHandler(BaseHandler):
         self.render_template(
             "loading.html",
             base_url=self.settings['base_url'],
+            provider_prefix=provider_prefix,
             url=provider.get_repo_url(),
             ref=provider.unresolved_ref,
             filepath=self.get_argument('filepath', None),
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -135,6 +135,7 @@ class GitRepoProvider(RepoProvider):
             raise ValueError("`resolved_ref` must be specified as a query parameter for the basic git provider")
         self.sha1_validate(resolved_ref)
         self.resolved_ref = resolved_ref
+        self.unresolved_ref = resolved_ref
 
     @gen.coroutine
     def get_resolved_ref(self):
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -114,13 +114,13 @@ Image.prototype.launch = function(url, token, filepath, pathType) {
     window.location.href = url;
 };
 
-function v2url(repository, ref, path, pathType) {
-  // return a v2 url from a repository, ref, and (file|url)path
+function v2url(provider_prefix, repository, ref, path, pathType) {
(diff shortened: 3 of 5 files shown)


choldgraf:
nice - I will give this a shot on a binderhub local deployment once we're done with JupyterDays stuff!

minrk:
@adl thanks so much! I did some testing and this doesn't quite get GitLab building working because we have some bugs in how we are escaping URLs. I've updated the PR with the fixes, which turned out a lot more involved than I had intended.

minrk:
This PR now has two significant changes in addition to @adl's original fixes for generalizing the provider prefix:

1. don't let tornado decode URL parts that are to be provider specs. This is relevant, because the spec for GitLab needs to be "user%2Frepo/ref" and we get "user/repo/ref" from tornado. This was causing a problem because ParameterizedMainHandler serves a page at `/v2/spec` that will request `/build/spec` and `spec` needs to be *exactly the same* in both. Tornado's unquoting caused the values to be different and builds to fail.
2. split the single entrypoint function in index.js so that we aren't submitting a hidden form on the 'loading' page. Instead, we're triggering the build directly with a 'build' function. This is the biggest change. It's mostly moving sections of the big main function into two or three separate functions. This was required, because in order to ensure that the spec passed from `/v2/:spec` is identical to the original spec, it needed an entrypoint that took the spec itself, not parsing it out into the build form and then reconstituting it, which we only do correctly for github URLs.

minrk:
Thanks, @adl!

PR #480, opened 2018-03-07
tornado 5 compatibility in io_loop fixture

minrk:
cherry-picked out of #459


willingc:
Good catch. I'm reviewing the appendix pr now.

PR #485, opened 2018-03-08
pin tornado to 4.5.3

minrk:
in case tornado 5 is what's causing the recent loss of build output on mybinder.org


minrk:
This wasn't it. It was an unhleathy node. tornado 5 is fine.

betatim:
Tornado v5 was also a thing that came up in https://github.com/jupyter/repo2docker/pull/252. Should we make a coordinated move (update r2d and bhub) to v5?

minrk:
Reopening this one since I think there's a *chance* that our recent 'event stream lost' uptick was related to the tornado 5 update. It could also be related to issues in our nginx ingress, which is hard to debug since we exclude all of the nginx ingress logs from stackdriver for cost reasons.

minrk:
I don't think we need to coordinate moves to tornado 5, since these are separate environments and applications where tornado 5 issues may come up. I'm surprised that it's an issue here, and it may still not be, but I want to pin it to at least control for that being a factor.

PR #488, opened 2018-03-15
helm chart: load appendix as regular file

minrk:
rather than trying to parse as yaml, which eliminates newlines

This is needed to use the appendix in the helm chart (https://github.com/jupyterhub/mybinder.org-deploy/pull/502)


choldgraf:
looks good! Should we do some degree of validation on the appendix when reading in? e.g. just make sure it adheres to some particular form?

also, what user-facing documentation should we create to describe how to utilize this?

PR #489, opened 2018-03-15
add GITHUB_ACCESS_TOKEN to travis

### LINKED ISSUE #484: GitHub rate limit on Travis
Tests failed several times on Travis today due to hitting the GitHub rate limit and needed to be retried. We should investigate loading a GitHub token into the travis env so we have a higher limit.

minrk:
closes #484 for tests run from master/tags (not pull requests, which don't have access to encrypted variables)


minrk:
I would expect this PR to fail due to hitting the rate limit, but after merge master should stop failing. Unfortunately, this PR cannot verify that it fixes itself before merging.

choldgraf:
so.....should I just merge it then and we'll find out? :-)

minrk:
Let's let it finish, at least. If it fails for some reason other than the rate limit, I might have broken something. Never underestimate the likelihood of bash typos.

minrk:
Yeah, both failed with rate limit. That's expected. If this works after merge, then we know it's fixed.

minrk:
master passed non-helm, but failed helm, so I got it half-right, at least.

choldgraf:
Hehe

PR #493, opened 2018-03-16
get kubespawner fixes via hub chart

minrk:
for deleting stopped pods: https://github.com/jupyterhub/kubespawner/pull/147


minrk:
Failure was due to github rate limit, should be fine once merged.

Issue #495, opened 2018-03-18
Trailing slash breaks binder link


djsegal:
What I mean is:

+ `https://mybinder.org/v2/gh/binder-examples/jupyterlab/master` - works
+ `https://mybinder.org/v2/gh/binder-examples/jupyterlab/master/` - fails

Is there a reason for this?

choldgraf:
are you deploying your own binderhub server? or is this on mybinder.org?

choldgraf:
(reason I ask is that I seem to remember we've fixed this but it may be in a newer version of the helm chart than the one we list in the docs)

djsegal:
Very interesting! Do you know which version.

I'm self-hosting, so I could update it

djsegal:
@choldgraf, what version should I use then? 

I realize they're all [development versions](https://jupyterhub.github.io/helm-chart#development-releases-binderhub), but is binderhub-0.1.0-e54c217 (the most recent one) really working?

I don't want to trade the version in the docs (that basically works) for one that obviously doesn't work (i just don't know it yet).

// this kind of requires your in-knowledge of the state of deployment versions

djsegal:
Just tried the most recent version and it still breaks on slashes:

<img width="1061" alt="screen shot 2018-03-20 at 1 14 47 am" src="https://user-images.githubusercontent.com/3156114/37637086-26252e12-2bdc-11e8-83ac-5adc833e36dd.png">

choldgraf:
thanks @djsegal for the useful patch - it is much appreciated!

PR #498, opened 2018-03-20
Remove trailing slash in urls automatically

djsegal:
Attempts to resolve:

+ https://github.com/jupyterhub/binderhub/issues/495


choldgraf:
can't believe we hadn't gotten to this one yet - thanks so much for the patch :-)

Issue #486, opened 2018-03-09
Using a non-gcr.io registry


choldgraf:
We should document how to use a registry other than `gcr.io`. I think giving DockerHub as an example should suffice here. From the suggestion that @willingc gave, I believe we need to tell people how to fill in this part of the chart:

https://github.com/jupyterhub/binderhub/blob/master/helm-chart/binderhub/values.yaml#L26

With the right values for DockerHub. In particular, we also need to give context for what this "JSON authentication" is which I believe is explained here:

https://docs.docker.com/engine/reference/commandline/login/#privileged-user-requirement

does this look correct?

cc @minrk

willingc:
A few more links from the gitter discussion:

DockerHub for the registry, but these docs look pretty good: https://docs.docker.com/registry/deploying/

Probably user set in values.yaml and credentials in secrets.yaml under registry. https://binderhub.readthedocs.io/en/latest/setup-binderhub.html#create-secret-yaml-file

minrk:
From what I recall from working on #342 we make assumptions about how auth works with gcr.io that's not generalized to other registries. I know that local un-authenticated registries won't work without changes there. I haven't tried with the DockerHub registry, so it might be that it does work and we need to be clear what config to use, or it's possible that we need to make some changes. It should be a matter of testing with the hub.docker.com and documenting and/or fixing what's needed. We definitely need changes to allow the local wide-open registry case, though.

PR #499, opened 2018-03-20
update registry, so that docker hub registry works

bitnik:
Hi,

Firstly I am not sure if this is a good implementation but I saw https://github.com/jupyterhub/binderhub/issues/486 and wanted share how we managed to use docker hub public registry. gcr should still work without changing any configuration.

Configuration to use docker hub registry:

- `secret.yaml`:
```
registry:
  username: <docker_hub_username>
  password: <docker_hub_password>
```
- `config.yaml`:
```
registry:
  enabled: true
  prefix: <user/organization_name>/<image_prefix>
  authHost: index.docker.io/v1
  authTokenUrl: https://auth.docker.io/token?service=registry.docker.io&scope=repository:{image}:pull
  imageManifestUrl: https://registry.hub.docker.com/v2/{image}/manifests/{tag}
```

### DIFF
--- a/helm-chart/images/binderhub/binderhub_config.py
+++ b/helm-chart/images/binderhub/binderhub_config.py
@@ -20,6 +20,11 @@ def get_config(key, default=None):
 c.BinderHub.debug = True
 
 c.BinderHub.docker_image_prefix = get_config('binder.registry.prefix')
+if get_config('binder.use-registry'):
+    c.BinderHub.docker_registry_host = get_config('binder.registry.host')
+    if get_config('binder.registry.auth-host'):
+        c.BinderHub.docker_auth_host = get_config('binder.registry.auth-host')
+    c.BinderHub.docker_token_url = get_config('binder.registry.auth-token-url')
 
 c.BinderHub.docker_push_secret = get_config('binder.push-secret')
 c.BinderHub.build_namespace = os.environ['BUILD_NAMESPACE']
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -13,7 +13,7 @@ import tornado.ioloop
 import tornado.options
 import tornado.log
 import tornado.web
-from traitlets import Unicode, Integer, Bool, Dict, observe, validate, TraitError
+from traitlets import Unicode, Integer, Bool, Dict, observe, validate, TraitError, default
 from traitlets.config import Application
 
 from .base import Custom404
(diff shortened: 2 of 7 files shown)


minrk:
Nice! It seems like we can simplify this a bit.

What if we had:

```yaml
registry:
   host: https://gcr.io
   authTokenUrl: https://gcr.io/v2/token?service=gcr.io
```

and we add the `scope=image:{image}:pull` as part of our implementation? Is there any case where this can vary if it's a docker registry?

The imageManifestUrl should always be `{host}/v2/{image}/manifests/{tag}` for any docker registry, right? So that doesn't need to be configurable beyond `host` as well. And there are only two hosts: the one in authTokenUrl and the actual registry  host, right?

bitnik:
yes, we can add the `scope=image:{image}:pull` as part of the implementation.

And also yes, the imageManifestUrl should always be `{host}/v2/{image}/manifests/{tag}` for any docker registry as far as I know. But there are actually 3 hosts for docker hub registry: the one in `authTokenUrl` (`auth.docker.io`), the one in `config.json` (`index.docker.io/v1`) for authentication and the actual docker registry host (`registry.hub.docker.com`).

So actually we need:
```yaml
registry:
  host: https://registry.hub.docker.com
  authHost: https://index.docker.io/v1
  authTokenUrl: https://auth.docker.io/token?service=registry.docker.io
```

and `host` and `authHost` are same for gcr (`https://gcr.io`).

minrk:
@bitnik thanks! Let's do that, and make the default `BinderHub.auth_host` be the same as `BinderHub.host`, so that only .host needs to be specified if they are the same:

```python
auth_host = Unicode(...)

@default('auth_host')
def _default_auth_host(self):
    return self.host
```

bitnik:
@minrk I updated the commit. Can you check it?

bitnik:
@minrk?

betatim:
(restarted travis, broke because GH limit exceeded)

minrk:
@bitnik sorry for the delay! I think this looks good, just need to check if the Travis failure is because of a typo somewhere, or a problem in the tests themselves. Then I think it should be set.

helm-chart/images/binderhub/binderhub_config.py
```diff
@@ -20,6 +20,10 @@ def get_config(key, default=None):
 c.BinderHub.debug = True
 
 c.BinderHub.docker_image_prefix = get_config('binder.registry.prefix')
+c.BinderHub.docker_registry_host = get_config('binder.registry.host')
```
minrk: This is the source of the helm test failure. When registry is not enabled, these will be undefined and thus None, which produces an error. If you add `, ''` here and below for docker_token_url it will default to an empty string and be a safe value. You can also do the same with `if get_config` that you have for auth-host to only load the config if it's defined.

bitnik:
@minrk I added `if get_config('binder.use-registry'):` in `helm-chart/images/binderhub/binderhub_config.py`. Sorry, i didn't think that having a new separate commit would be much better for reviewing.

minrk:
Thanks!

PR #502, opened 2018-03-22
get max-age culler from jupyterhub chart

minrk:


minrk:
Failure was github rate limit

PR #507, opened 2018-03-29
Fix launching fav icons when base_url is used

bitnik:
Hi, 

Currently fav icons (building, fail and succes) during launching don't work, if `base_url` is specified.

And there were some minor problems (redundant or missing closing tags) with html templates.


choldgraf:
Thanks for the patches / cleanup @bitnik !!

the travis fail was a github rate limit error, restarting to see if that fixes it

can you confirm that this correctly makes the favicon load/update when you test it locally?

PR #508, opened 2018-04-01
disable git shallow-clone

minrk:
Same as https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/621

shallow-clone can result in incorrect image tagging if image directories go without updates for a bit.

Additionally:

- show `git diff` after chartpress for easier debugging
- add nice heading comments for multi-line travis commands (style copied from mybinder.org-deploy)
- add a few travis files to .gitignore
- pin chartpress version so that updates there will not be used here except via explicit update

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,13 +21,16 @@ script:
 - export BINDER_TEST_NAMESPACE=binder-test-$TEST
 - ./ci/test-$TEST
 after_failure:
-- for pod in $(kubectl get pod --namespace=$BINDER_TEST_NAMESPACE | awk '{print $1}'); do
+- |
+  # get pod logs
+  for pod in $(kubectl get pod --namespace=$BINDER_TEST_NAMESPACE | awk '{print $1}'); do
     echo $pod
     kubectl logs --namespace=$BINDER_TEST_NAMESPACE $pod
   done
 after_success:
 - codecov
 - |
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -21,20 +23,24 @@ script:
...
 - |
+  # publish helm chart
   if [[ "$TRAVIS_PULL_REQUEST" == "false" && "$TEST" == "helm" ]]; then
     openssl aes-256-cbc -K $encrypted_d8355cc3d845_key -iv $encrypted_d8355cc3d845_iv -in travis.enc -out travis -d
     chmod 0400 travis
     export GIT_SSH_COMMAND="ssh -i ${PWD}/travis"
     docker login -u ${DOCKER_USERNAME} -p "${DOCKER_PASSWORD}"
     cd helm-chart
     chartpress --commit-range "${TRAVIS_COMMIT_RANGE}" --push --publish-chart
+    git diff
```
betatim: should we add a comment saying what the expected output of `git diff` is when everything is fine?

minrk: Good call. Done.

minrk:
There is no solution to the GitHub API limit issue other than not testing GitHub reps, since we can't make credentials accessible to pull requests. It should pass once merged, since master has a GitHub token.

PR #509, opened 2018-04-03
typo in bad spec log message

minrk:
forgot to include log format args


choldgraf:
Hmm, Travis errored  twice now (I restarted it a bit ago)

PR #513, opened 2018-04-06
enable ip-anonymization in google analytics

minrk:


minrk:
also disabled the google-analytics cookie. The result of disabling the ga cookie is that there is no user-id and the session tracking is lost. I don't think we care more about tracking session activity than we do about avoiding setting any cookies for users. See more info [here](https://support.google.com/analytics/answer/3123663?hl=en).

betatim:
Does this mean we won't be able to tell if a user is a new vs a returning user? That might be an interesting metric for grants?

minrk:
@betatim I believe so. It seems we have to do this regardless. Identifying returning users for analytics, even by UUIDs in cookies, seems to require affirmative consent. So our choices appear to be:

1. disable session tracking always
2. Add opt-in to tracking dialog. Either disable session-tracking or disable analytics altogether until users opt-in
3. disable analytics and run our own basic, anonymous pageview counter

betatim:
For tracking less is more :) So I think we should go with this and if we find out/really miss the session/returning user tracking we can invest the effort to make a opt-in UI etc

choldgraf:
I am +1 on being conservative about what we collect until we get some clarity on what is/isn't going to be an issue.

choldgraf:
I think we have enough data (unique repositories, languages, several binders that have made it to hacker news, etc) to make a good case for Binder :-)

betatim:
Merged now. We can always re-enable it if we discover we need it or get some clarity on the regulation/compliance bit.

PR #515, opened 2018-04-11
add provider, repo labels to build/launch metrics

minrk:
gives us better access to popular repos in prometheus

and lets us track repos/providers that fail a lot

this adds an 'unbounded' label (repo), which is discouraged in prometheus, but kubernetes metrics already have pod names which have strictly more unique values than the number of repos since they get one for each launch, so this at least shouldn't make anything worse and this is valuable/useful info. Plus, the number of unique repos in a recent window is on the order of tens to hundreds, not millions, which is the range I've seen when searching for "how many unique labels is too many" for prometheus.

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -208,6 +208,12 @@ class BuildHandler(BaseHandler):
 
         repo = self.repo = provider.get_repo_url()
 
+        # labels to apply to build/launch metrics
+        self.metric_labels = {
+            'provider': provider.name,
+            'repo': repo,
+        }
+
         try:
             ref = await provider.get_resolved_ref()
         except Exception as e:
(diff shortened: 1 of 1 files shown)


choldgraf:
I think from a user's perspective, this sounds great. I'm a bit concerned since @yuvipanda has urged caution when it comes to adding new labels etc, but it looks like @minrk has looked into this a bit and this should be OK for us. We can always remove it if prometheus slows down a bunch.

That said, travis seems unhappy!

minrk:
The kube stats metrics already create a label for each *pod*, and repos are strictly less varied than pods, since each launch of a repo gets a new pod name. So this at least can't make our worst case worse.

Travis failures were due to hitting the unuathenticated GitHub rate limit. This ought to be fixed by #519. Rebased and hopefully tests should pass now.

choldgraf:
🎉

PR #516, opened 2018-04-11
stop waiting for launches if they are not pending anymore

minrk:
should avoid failed launches appearing to always take more than ten minutes to fail

If a launch is not finished, the user model will have `server=None` and `pending='some event'`. This is what we should wait for. If the launch fails, it will have `server=None` and `pending=None`. We haven't been checking `pending`, so if the launch actually fails we will always wait until the maximum timeout before giving up, even though we know that no event is pending.

I think this is the reason all failed launches appear to take the same long time.


choldgraf:
nice catch!!! restarting travis to see what's up

choldgraf:
hmmm, travis seems to be genuinely unhappy - here's the error as far as I could tell:

`FAILED [ 10%]base_events.py            1260 ERROR    Task was destroyed but it is pending!`

https://travis-ci.org/jupyterhub/binderhub/jobs/365058120#L968

seems related to this PR as well, no?

minrk:
I don't believe the failure was related to this PR. All PRs have been failing for a while due to hitting the GitHub API rate limit. #519 should fix this. I've rebased the PR, so hopefully Travis should become trustworthy again.

PR #517, opened 2018-04-11
retry launch failures

minrk:
Retries occur at two levels:

1. individual API requests, which will be retried in case of connection problems,
   which is ~all current failures, I think.
2. retry the full launch process (new user, etc.) in case of actual failure

TODO:

- [x] expose config in helm chart
- [x] decide metric updates. Should the launch timing be the user-visible time (including all retries in one time measurement), or time each launch attempt, e.g. adding 'retry' status for failed launches that will be attempted again.

### DIFF
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -40,12 +40,12 @@ class Launcher(LoggingConfigurable):
         """
     )
     retry_delay = Integer(
-        10,
+        4,
         config=True,
         help="""
         Time (seconds) to wait between retries for Hub API requests.
 
-        Time is scaled by the retry attempt (i.e. first retry after 1 * delay, second after 2 * delay)
+        Time is scaled exponentially by the retry attempt (i.e. 2, 4, 8, 16 seconds)
         """
     )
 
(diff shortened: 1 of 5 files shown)


minrk:
Right now, the launch time metric has:

- status = success, retry (failed but will retry), failure (all attempts failed)
- retries = number of retries (0 if success/failure on first attempt)
- time = duration of latest attempt. Alternately, this could be the total time (i.e. time from first request to resolution), which may be desirable, as it's the time users actually experience.

minrk:
Thanks. Giving rebase a try. I was able to run this locally.

I think this will help our success rate across hub restarts and occasional hiccups.

binderhub/launcher.py
```diff
@@ -30,15 +30,47 @@ class Launcher(LoggingConfigurable):
...
+        4,
+        config=True,
+        help="""Number of attempts to make on Hub API requests.
+
+        Adds resiliency to intermittent Hub failures,
+        most commonly due to Hub, proxy, or ingress interruptions.
+        """
+    )
+    retry_delay = Integer(
+        10,
```
choldgraf: 10 seems like quite a lot, no? That means that by the 4th delay it will be waiting for 40 seconds before proceeding right?

choldgraf:
This is a clever idea, it looks good to me. One nitpick question about the default time but that shouldn't block on the PR. @minrk you think it is ready to do? I'll hold off to see if @willingc has any other thoughts!

binderhub/launcher.py
minrk: switched to exponential backoff with an initial time of 4 seconds, so: 4, 8, 16, 32.

PR #519, opened 2018-04-11
Mock GitHub API requests in tests

minrk:
- [x] Adds a MockAsyncHTTPClient, which may use recorded responses, if provided
- [x] use recorded responses for `api.github.com` if GITHUB_ACCESS_TOKEN is not set (e.g. pull requests on Travis)
- [x] add records for `api.github.com` from a local run to the repo so they can be used in PRs
- [x] gitignore http-response files created by tests
- [x] skip GitHub API tests when running in helm without GitHub API token

Once finished, this should result in PRs passing on Travis again, as the GitHub rate limit will no longer be hit. The cached responses will not be used from master, which has GitHub credentials.


binderhub/tests/utils.py
```diff
@@ -1,11 +1,69 @@
...
+
+from tornado.httputil import HTTPHeaders
+from tornado.httpclient import AsyncHTTPClient, HTTPError, HTTPRequest, HTTPResponse
+
+
+class MockAsyncHTTPClient(AsyncHTTPClient.configurable_default()):
+    mocks = {}
+    records = {}
+
+    def url_key(self, url):
```
willingc: Nice!

minrk:
Tests passed! That doesn't necessarily confirm that the github rate limit issue is entirely fixed, since we could have been lucky, but I think it is.

choldgraf:
woot! here's to fewer "known failing" tests!

willingc:
Fantastic work @minrk.

PR #530, opened 2018-04-19
add missing unix:// to DOCKER_HOST

minrk:
DOCKER_HOST always has the `proto://host` form, even for unix sockets

You can try it on your own machine:

```bash
DOCKER_HOST=/var/run/docker.sock docker ps # fails
DOCKER_HOST=unix:///var/run/docker.sock docker ps # ok
```

Also instantiate the docker client immediately on starting the image cleaner, which would have caught this bug sooner.


minrk:
I think this will get mybinder working with the latest versions of everything

choldgraf:
@minrk we're ready to deploy this to binder, right?

minrk:
I think this is deployed, but there were further issues preventing the image-cleaner from working. #534 will hopefully get it going.

PR #533, opened 2018-04-22
Fix broken reference to architecture diagram

ablekh:


choldgraf:
Much appreciated! Thanks!

PR #534, opened 2018-04-23
Fixes for image-cleaner

minrk:
- get image list at startup, to ensure that it works
- fix package name for docker (it's `docker` not `docker-py`)
- fix percentage display (multiply by 100 to turn fraction into percent)


choldgraf:
@minrk is this ready to go?

minrk:
Yes, I believe so

minrk:
You'll want to set the inode threshold to something when you deploy. Maybe 0.2? I think we'll want to keep an eye on the logs of the image cleaner to figure out what a good value is.

PR #535, opened 2018-04-23
include base_url in binder_url

bitnik:
Currently the `binder url` doesn't contain `base url`. So if `base_url` is set as different than default value ("/"), `binder_url` becomes wrong.


minrk:
Thanks!

PR #536, opened 2018-04-25
document how to use docker hub registry

### LINKED ISSUE #486: Using a non-gcr.io registry

bitnik:
Hi, after https://github.com/jupyterhub/binderhub/pull/499, I also want to document how to use docker hub registry.

I was not sure where to document it exactly, so I just made an extra file for it. I can update it according to your feedback.

This is related to https://github.com/jupyterhub/binderhub/issues/486.

* closes #486


choldgraf:
This looks great - I think it'd actually be better to have a separate page for creating a registry, and then have sub-sections for `gcr.io` vs. `dockerhub`. Could I make a push to your branch that does this?

bitnik:
Nice, I just sent you an invitation.

choldgraf:
ok, made a few rearrangements and edits so that both registries are in-line with one another. Can somebody else take a look at this to make sure it makes sense? @willingc ?

choldgraf:
thanks for the docs update @bitnik !!! Super helpful!

PR #537, opened 2018-04-25
adding known deployments page

choldgraf:
Adds a page where we can include known deployments that we hear of


betatim:
Opened #540 for licensing topic. I think the text from that should be in our READ/deployment guide so merging this now.

PR #541, opened 2018-04-27
rework image cleaner a bit

minrk:
- clean images on both dind and host docker,
  since host docker contains the vast majority of used images.
- run image cleaner as its own daemonset, rather than inside dind, since it should run even when dind doesn't. This also prevents the need to relaunch dind when the image cleaner changes, which takes a long time.
- prioritize deleting untagged images so we remove images such as failed build products first
- match image-gc config to kubernetes' own as much as possible:
  - units are percent instead of fractions (80 instead of 0.8)
  - units refer to 'used' rather than 'available' (80 means )
  - add high/low thresholds so when high threshold is exceeded, images are reclaimed until the low threshold is passed
  - config is now `imageGCThreshold[High|Low]`, matching the names in kubernetes

The default thresholds are 80/60: when 80% of inodes are used, cull images until the number drops below 60%

includes #539

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -6,6 +6,7 @@ import logging
 import os
 from urllib.parse import urlparse
 
+import kubernetes.client
 import kubernetes.config
 from jinja2 import Environment, FileSystemLoader
 from tornado.httpclient import AsyncHTTPClient
@@ -360,6 +361,8 @@ class BinderHub(Application):
                 kubernetes.config.load_incluster_config()
             except kubernetes.config.ConfigException:
                 kubernetes.config.load_kube_config()
+            self.tornado_settings["kubernetes_client"] = kubernetes.client.CoreV1Api()
+
 
         # times 2 for log + build threads
         self.build_pool = ThreadPoolExecutor(self.concurrent_build_limit * 2)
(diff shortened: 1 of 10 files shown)


minrk:
I've tested this manually with:

    gcloud compute --project=binder-prod ssh gke-prod-a-ssd-pool-32-134a959a-lbw0
    lbw0 $ docker run -v /var/run/docker.sock:/var/run/docker.sock -v /var/lib/docker:/var/lib/docker --rm -it minrk/k8s-image-cleaner:76b09ad

and it successfully cleaned out a number of large images from the host docker on that node. YOu can see the results in the inodes-available chart (that uptick in the green line):

screen shot 2018-04-27 at 16 08 32

So at least the image seems to be correct.

I've also tested deploying this locally with minikube, so the new daemonset seems to work as well.

Deleting images takes a pretty long time (on the order of 30 minutes to cull from 80 to 60% of used inodes).

choldgraf:
travis restart made it happy...it sounds like you did all the testing we could try without actually deploying to prod...so I'm gonna merge this :-)

re: the amount of time it takes to clean images, do you still think this is a preferable solution to just recycling nodes occasionally? That happens relatively instantaneously (though it also creates some disruptions and isn't automatable right now). WDYT?

choldgraf:
actually, looking through the changes, I'm gonna hold off on merging cuz this touches on a bunch more k8s components than I had thought, and I wanna make sure somebody with more k8s knowledge than me makes sure it's 👍

PR #544, opened 2018-05-03
Cache GitHub ref resolutions

### LINKED ISSUE #543: Cache GitHub refs
mybinder.org was recently down because we exhausted the GitHub API rate limit. ~all of these requests were for the same repo with the same version. If we used a cache to remember the answer to previous requests, we could make these API requests with an `If-None-Match` header, which doesn't consume the GitHub API if nothing has changed.

We do this on nbviewer, and it was the most important thing we did for avoiding the github rate limit. Since we only use the GitHub API to resolve refs, we don't have to do the full response caching that nbviewer does, we only need to cache ETag + resolved ref. Because of this, I think we can use a simple in-memory cache rather than what nbviewer does, which is run memcache in a separate container.

minrk:
Since the majority of launches are of stable/popular repos (e.g. ipython/ipython-in-depth/master)
Using ETag / If-None-Modified allows us to cache responses and reduce consumption of the API limit

No reduction in HTTP requests is made, but GitHub does not count cache hits against the rate limit. Cache is least-recently-used (LRU) with a max size of 1024 entries. Since each entry is a two-key dict, this should still be quite small.

Cache is only in memory, so relaunching BinderHub resets the cache.

closes #543

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -348,7 +348,7 @@ class BinderHub(Application):
         # hook up tornado logging
         if self.debug:
             self.log_level = logging.DEBUG
-        tornado.options.logging = logging.getLevelName(self.log_level)
+        tornado.options.options.logging = logging.getLevelName(self.log_level)
         tornado.log.enable_pretty_logging()
         self.log = tornado.log.app_log
 
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -15,11 +15,13 @@ import re
 from prometheus_client import Gauge
 
 from tornado import gen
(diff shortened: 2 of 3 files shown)


binderhub/app.py
```diff
@@ -348,7 +348,7 @@ def initialize(self, *args, **kwargs):
         # hook up tornado logging
         if self.debug:
             self.log_level = logging.DEBUG
-        tornado.options.logging = logging.getLevelName(self.log_level)
+        tornado.options.options.logging = logging.getLevelName(self.log_level)
```
betatim: `options.options` isn't a typo?

Issue #260, opened 2017-11-10
Some way to allow for Binder branding w/in notebooks etc


choldgraf:
This will be tricky as we support more and more UIs, but I wonder if it'd be good to have some kind of visual cue within Binder sessions to tell people that they're working on Binder. Maybe a favicon? Or alternatively we had discussed making a configurable logo that's displayed in, e.g., the notebook. This might also be something that organizations want if they start deploying their own BinderHubs.

choldgraf:
another ping on this one since it came up a few times over the last few weeks, in particular @enolfc and @gouarin and maybe @nthiery who mentioned this at the orsay meeting

betatim:
We now have support for appending to a `Dockerfile` (see #459). This means that we have the infrastructure for making this happen inside the live binder. We can inject our own themes for the single user notebook that include some binder branding/advice/link to original repo/GA tracking

Has someone tried modifying the look of the singleuser notebook server already?

bitnik:
For numbers 1 and 2, do you think is it a good idea to enable having custom/extended templates similar to https://github.com/jupyterhub/jupyterhub/pull/1601? If you think so, I can try to bring the same feature to BinderHub.

choldgraf:
Good question - @minrk do you have an intuition for whether we'd be well-off to make templates for BinderHub pages?

minrk:
Yes, I think we can and should. We're already using jinja, so it's mostly a matter of exposing the right configuration, and good use of blocks in our templates, to make it easy to customize the right stuff.

bitnik:
@minrk I made a PR for this.

@betatim I didn't really change the look of single user server but while trying out build appendix feature, I added a button by using `custom.js`. This button (next to logout button) copies sharable user server link (with token) into clipboard. An example repo built after this feature is https://notebooks.gesis.org/binder/v2/gh/binder-examples/requirements/master.

bitnik:
@choldgraf I had myself some difficulties about sharing binder links, so that was the first thing I tried when I saw `build.appendix` is implemented :)

The code is [here](https://github.com/gesiscss/orc/tree/master/binderhub/appendix).

[`run-appendix`](https://github.com/gesiscss/orc/blob/master/binderhub/appendix/run-appendix#L6-L11) script adds the content of [`static/custom.js`](https://github.com/gesiscss/orc/blob/master/binderhub/appendix/static/custom.js) into `~/.jupyter/custom/custom.js` inside user notebook server:
```bash
[...]

PR #545, opened 2018-05-03
Enable template extension

bitnik:
https://github.com/jupyterhub/binderhub/commit/6e194dde00f196edf31a095ef65572ac8c4db5df enables extension of base binder templates by supplying new templates under `template_path` and also adding new templates variables with `template_variables`.  `template_path` is where new templates are located. Base templates are loaded both with and without `templates/` prefix, so new templates can extend base templates by including `{% extends "templates/<template_name>.html" %}`, otherwise it overwrites the base template. An example configuration:
```bash
template:
  variables:
    var1: value1
    var2: value2
  path: /etc/binderhub/templates
```

I am not sure about https://github.com/jupyterhub/binderhub/commit/281b1cb2b172aed7f30847b2b1b9c187dff27de8. The idea was that users can supply extra static files to be used in new templates, so they are also served by binder app as other static files. To do that `extra_static_path` (where extra static files are located) must be defined and in templates source of extra static files must be defined as `"/<base_url>/<extra_static_url_prefix>/<static_file_path>"`. Do you think that this is a good idea?

related to https://github.com/jupyterhub/binderhub/issues/260
This PR is based on https://github.com/jupyterhub/jupyterhub/pull/1601

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -8,13 +8,13 @@ from urllib.parse import urlparse
 
 import kubernetes.client
 import kubernetes.config
-from jinja2 import Environment, FileSystemLoader
+from jinja2 import Environment, FileSystemLoader, PrefixLoader, ChoiceLoader
 from tornado.httpclient import AsyncHTTPClient
 import tornado.ioloop
 import tornado.options
 import tornado.log
 import tornado.web
-from traitlets import Unicode, Integer, Bool, Dict, observe, validate, TraitError, default
+from traitlets import Unicode, Integer, Bool, Dict, validate, TraitError, default
 from traitlets.config import Application
 
 from .base import Custom404
@@ -25,8 +25,6 @@ from .main import MainHandler, ParameterizedMainHandler, LegacyRedirectHandler
 from .repoproviders import GitHubRepoProvider, GitRepoProvider, GitLabRepoProvider, GistRepoProvider
 from .metrics import MetricsHandler
 from .utils import ByteSpecification, url_path_join
-
-TEMPLATE_PATH = [os.path.join(os.path.dirname(__file__), 'templates')]
 
 
 class BinderHub(Application):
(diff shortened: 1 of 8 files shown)


bitnik:
To use this feature in a deployment, one has to first extend binderhub docker image by copying templates and static files as in this example:
```bash
FROM jupyterhub/k8s-binderhub:<tag>

COPY templates /etc/binderhub/templates
COPY static /etc/binderhub/static
```
And use this image in helm config.

I was thinking how to get rid of this step. Maybe we could find a solution by using gitRepo volume or [initContainers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) to clone/download these files and mount to a specified path where binder app looks for them?

minrk:
This is really great! If you can come up with an example that uses a volume without rebuilding the image, that would be pretty cool

bitnik:
@minrk https://github.com/jupyterhub/binderhub/pull/545/commits/51cddfd9fd894c22bb43e5300641e9dabf617885 enables `initContainers`, `extraVolumes` and `extraVolumeMounts` configuration in binder deployment. By using them, templates and static files can be added into a volume and don't need to rebuild the image. Here is an example configuration:

```yaml
template:
  path: "/etc/binderhub/templates/<path_to_templates_in_repo>"
  static:
    path: "/etc/binderhub/templates/<path_to_static_files_in_repo>"
    urlPrefix: extra_static/

initContainers:
  - name: git-clone-templates
    image: alpine/git
    args:
      - clone
      - --single-branch
      - --branch=<branch_name>
      - --
      - <repo_url>
      - /etc/binderhub/templates
    securityContext:
      runAsUser: 0
    volumeMounts:
      - name: custom-templates
        mountPath: /etc/binderhub/templates
extraVolumes:
  - name: custom-templates
    emptyDir: {}
extraVolumeMounts:
  - name: custom-templates
    mountPath: /etc/binderhub/templates
```

minrk:
Very nice!

PR #547, opened 2018-05-04
Stop the 'show logs' button from starting a new job.

neprune:
The type for the show logs button wasn't set and since buttons have [submit type by default](https://www.w3.org/TR/2011/WD-html5-20110525/the-button-element.html#the-button-element), clicking it would submit the form and cause a new job to start.

### DIFF
--- a/binderhub/templates/index.html
+++ b/binderhub/templates/index.html
@@ -123,7 +123,7 @@
         <div id="log-container" class="panel panel-default on-build hidden row">
           <div id="toggle-logs" class="panel-heading">
             Build logs
-            <button class="btn btn-link btn-xs pull-right">show</button>
+            <button type="button" class="btn btn-link btn-xs pull-right">show</button>
           </div>
           <div class="panel-body hidden">
             <div id="log"></div>


betatim:
This looks like a good change to make!

I tried to find when we show the show logs button without yet having launched a build and can't find that. Can you help me out and describe what I need to do to make this "bug" happen?

neprune:
Sorry, I should clarify that the problem occurs when you want to view the logs while a build is still in progress. You can reproduce it by choosing a repo you haven't already built (so the build is slow enough) and clicking show logs straight after you click launch. You will see the details of two builds displayed in the logs.

neprune:
Weird, I'd expect it to be visible on any browser that implements that part of the HTML5 standard correctly; I can see it on Chrome and Firefox for `mybinder.org`. Have you tried using a 'fresh' repo? It can be difficult to spot when the image has already been built because the launch is so quick. Another thing to look at is the network activity in developer tools. In Chrome, I can see two open `EventSource`s after the show logs button is clicked.

choldgraf:
ahhh I see the problem - for me I was able to reproduce, but only if I actually click the "show/hide" _text_ on the right side of the button. If you click the button itself, the behavior happens but I don't think the submit happens. Gonna marge this, thanks for spot-checking our crappy UX @neprune :-)

betatim:
I finally managed to see it :)

Thanks for spotting this!

PR #549, opened 2018-05-08
set PYTHONUNBUFFERED=1 in image-cleaner

minrk:
Python seems to produce no output in some container scenarios without this env

I'm not sure why logging doesn't have this issue while printing does.


willingc:
@minrk Perhaps the print statements do not end in a new-line. If they don't, flush=True would be needed on the print statement to clear the buffer.

betatim:
Do you want to experiment with this or happy enough that setting the env var makes it work?

minrk:
I'm happy as-is.

print statements include newlines by default, so I don't think that would be it. Googling suggested taht this is a common issue with print statements on kubernetes (and docker daemon images in general).

PR #550, opened 2018-05-08
avoid debug-level logging from curl_httpclient

minrk:
curl_httpclient logs way too much (all requests and responses) at debug-level.

- Avoids setting log-level for curl lower than INFO, even if the rest of the app has debug logging enabled
- Stops setting log-level to debug in the helm chart


willingc:
Go for it @betatim

PR #563, opened 2018-05-15
try to fix cordoning

minrk:
- restores cordoning
- while in rbac, add standard chart labels
- switches rolebindings to cluster roles. Since nodes aren't namespaced resources, I think we have to use ClusterRole, not Role. It seems bizarre that the non-functional role was allowed to be created instead of producing an error immediately.


minrk:
This took a lot more fiddling to get fixed than I wanted. In the end, I couldn't get minikube 0.25 or 0.26 to start on Travis.

I did improve the debugging a bit along the way for when deploys fail.

PR #564, opened 2018-05-16
Use logging module to get timestamps for image cleaner

betatim:
Get timestamps for the output of the image cleaner. Useful for trying to correlate with other events.


choldgraf:
Seems like a good idea! Can you confirm that this works when you run it locally?

PR #568, opened 2018-05-18
forcing node to be cordoned in image cleaner

choldgraf:
Adds a quick `cordon(..` to the top of the image cleaner "while" loop so that we ensure the node is cordoned while cleaning is happening


choldgraf:
@minrk @willingc

PR #570, opened 2018-05-21
reduce docker load in image cleaner

minrk:
- add IMAGE_GC_DELAY delay between deletions
- wait to move on to the next image after a deletion times out

This should hopefully prevent the image cleaner rendering docker itself being unresponsive for extended periods, which seems to happen when we greedily schedule image deletions one after another even when the requests timeout (itself an indication that docker isn't keeping up).

### DIFF
--- a/helm-chart/binderhub/templates/image-cleaner.yaml
+++ b/helm-chart/binderhub/templates/image-cleaner.yaml
@@ -42,6 +42,8 @@ spec:
               fieldPath: spec.nodeName
         - name: PATH_TO_CHECK
           value: /var/lib/docker
+        - name: IMAGE_GC_DELAY
+          value: {{ $Values.imageCleaner.delay | quote }}
         - name: IMAGE_GC_THRESHOLD_HIGH
           value: {{ $Values.imageCleaner.imageGCThresholdHigh | quote }}
         - name: IMAGE_GC_THRESHOLD_LOW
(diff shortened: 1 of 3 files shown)


betatim:
Do we have any stats on how many images the cleaner deletes per cleaning session?

We briefly discussed in gitter now if we could end up in a situation where all nodes are being cordoned because they are being cleaned. We concluded "unlikely because all nodes would have to be cleaned at the same time". Now we wait 5s between each clean increasing the total time a node is cordoned. If this reduces the load on docker/time it is unresponsive enough do we still need to cordon the nodes?

choldgraf:
that's a good question - I'm gonna merge this, but would love to hear @minrk 's thoughts on whether this is a potential problem we'll run into...

minrk:
We definitely can run into all nodes being cordoned, but that can only last a short while because it will immediately trigger an autoscale event when the first pod is requested while all nodes are cordoned. Autoscale triggers immediately when a requested pod is "unschedulable". Marking all nodes with NodeNotSchedulable means that this will happen immediately on the next pod request.

 With four nodes, this is still *unlikely* because cleaning takes ~30-60 minutes (it will take longer after this PR, depending on how many images). Assuming each node culls for two hours every twenty four, there is a ~ [1/20k chance](https://nbviewer.jupyter.org/gist/minrk/45b52e83fb6a5bbd4bf25545be713e79) that all four nodes will be cordoned at once.

choldgraf:
lol this nbviewer link makes me want `gist` support in the Binder UI even more :-D

PR #576, opened 2018-05-24
dont create crole, crolebinding and service account if image-cleaner is not enabled

bitnik:
when image cleaner is disabled, helm upgrade returns `no ServiceAccount with the name "<release-name>-image-cleaner" found`.

### DIFF
--- a/helm-chart/binderhub/templates/rbac.yaml
+++ b/helm-chart/binderhub/templates/rbac.yaml
@@ -42,6 +42,7 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
   name: binderhub
+{{ if .Values.imageCleaner.enabled -}}
 ---
 # image-cleaner role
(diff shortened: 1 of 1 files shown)


choldgraf:
Ah, so if the image cleaner is not enabled, it tries to create a role for an account that doesn't exist?

bitnik:
Sorry, I just tested it again and I didn't get this error. I think that that was my fault.

But still I have cluster role and service account created for image-cleaner when it is disabled. Shouldn't they be removed when image-cleaner is disabled?

choldgraf:
It seems reasonable to me just for a cleaner implementation...I'm trying to think if this would create confusion if a deployment turned the image cleaner on/off/on etc. @minrk does this seem fine to you?

minrk:
This makes sense, thanks!

choldgraf:
woot, thanks @bitnik !!

PR #590, opened 2018-06-15
known deployments, GESIS correct name

bitnik:


choldgraf:
@bitnik thanks for the clarification :-)

Issue #501, opened 2018-03-21
LOADING.md - Implementation specifications needed


consideRatio:
I was thinking that temporarily showing a server-detached notebook rendering using nbviewer might be appreciated when a specific notebook file to be loaded is known (when a index.ipynb exist or the filepath query param is set for example).

I really appreciate the interactive fun graphics as it is now though, but at the same time a web developer friend recently told me about how Amazon had concluded that every 100ms of loading made them loose a lot of users. This idea might be a way to combat the comparably huge loading time.

Another idea would be to support the choice of some loading screen content based on a file in the repo named "loading.ipynb" or similar, or perhaps by some setting file.

minrk:
That's a very interesting idea! Running nbconvert on binder might be a level of complexity we aren't prepared for *yet*, but it could be a nice feature.

betatim:
Could we use an iframe with nbviewer inside? Downside is that we now depend on an external service, upside is that the additional complexity for mybinder.org is low.

choldgraf:
I'm +0 on this, leaning towards -.1 :-) 

The mean reason other than the increase in complexity is that I think a big feature of Binder is precisely that it *doesn't* make a ton of special assumptions about the workflows we're enabling. When I tell people "you can totally run RStudio, JupyterLab, OpenRefine, etc" in Binder they think it's really cool, so I'm a bit hesitant to start treating notebooks as "special" relative to all the other workflows that are out there...

consideRatio:
That is a point that makes a lot of sense to me @choldgraf !

What about being able to have a rendered "LOADING.md" provided by the repo or similar shown during loading? I'm just tossing the ideas out here for more than me to process.

choldgraf:
`LOADING.md` sounds cool! It'd also be cool if you could make it a gif or something like this. I'm not sure from a technical perspective how difficult it'd be though...

minrk:
LOADING.md is cool, but we should be careful if it can do things like reference other files in the repo (e.g. a relative URL for an inlined screenshot). That can get tricky real quick because to fully render the markdown, we need to have URL handlers that serve the whole repo as static files (this is a significant fraction of what nbviewer does).

choldgraf:
So, are we +1 on `LOADING.md`? If so, @consideRatio could you edit the top-level comment / issue title (or open a new issue) with this more actionable path forward? :-)

consideRatio:
#### TODO
- [ ] Specify the wanted behavior of LOADING.md

[...]

PR #593, opened 2018-06-20
nbviewer preview for github notebooks

minrk:
Load nbviewer.jupyter.org in an iframe for quicker previews of notebooks while waiting for the Binder to launch:

screen shot 2018-06-20 at 23 18 17

only enabled for GitHub right now, which is the only provider currently supported by both services.

Still a WIP for various UI questions.

This implements a proposal in #501 where there is discussion about whether notebooks should get special treatment. I think it *is* appropriate to treat notebooks specially, as long as others don't suffer as a result. For instance, we could restrict this to only `if filepath.endswith('.ipynb')` so that we don't load nbviewer for repos that run something like rstudio. That said, the nbviewer file navigator is still useful for browsing any repo of files, not just notebooks, so maybe it's still useful.

cc @nthiery

### DIFF
--- a/binderhub/static/loading.css
+++ b/binderhub/static/loading.css
@@ -92,6 +92,7 @@ div#log-container {
 
 .preview {
   margin-top: 40px;
+  width: 70%;
 }
 
 #nbviewer-preview > iframe {
--- a/binderhub/static/loading.css
+++ b/binderhub/static/loading.css
@@ -89,3 +89,13 @@ div#log-container {
 .hidden {
     display: none;
 }
+
+.preview {
+  margin-top: 40px;
+}
+
+#nbviewer-preview > iframe {
+  width: 100%;
+  height: 80vh;
+  border: 1px solid #aaa;
+}
--- a/binderhub/templates/loading.html
+++ b/binderhub/templates/loading.html
@@ -24,6 +24,25 @@
   </div>
 </div>
 
+{% block preview %}
+<p class="preview-text text-center">
+Here's a preview on
+<a href="https://nbviewer.jupyter.org">nbviewer</a>
+while we start a server for you.
+</p>
+{# we can only produce an nbviewer URL for github right now #}
+{% set prefix, repo_ref = provider_spec.split('/', 1) %}
+{% if prefix == "gh" %}
+{% set repo, ref = repo_ref.rsplit('/', 1) %}
+<div id="nbviewer-preview">
+    <iframe
+        src="https://nbviewer.jupyter.org/github/{{repo}}/tree/{{ref}}/{% if filepath %}{{ filepath }}{% endif %}"
+        style="width: 80%; margin-left: 10%; height: 80vh; border: 1px solid #aaa;"
+    >
+</div>
+{% endif %}
+{% endblock preview %}
+
 {% endblock main %}
 
 {% block footer %}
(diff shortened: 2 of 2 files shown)


choldgraf:
I think this is super cool! That said, I'm hesitantly +1 on this, with a couple of precautions:

1. We don't want to confuse people into thinking they can "interact" with the nbviewer image (many users, esp new ones, may not understand this without some kind of visual cue)
2. What do we do if there are multiple notebooks in the repo?
3. Are we using up screen space that might one day be needed for something else? (e.g. for logos, sponsorships, etc)?
4. Agreed about giving notebooks special treatment so long as it doesn't harm others. That said, maybe for non-notebook repos we just display the files as you describe? Or let people specify an image if there's a file called "binder/PREVIEW.png" or something?

betatim:
Nice work. I'm wondering how useful people would find it vs how annoyed they'd be that they should not click any links in the notebook they are seeing/will navigate away mid sentence? Maybe the latter we can mitigate by not auto navigating to the launched server but putting a "your binder is now ready for you" notice?

choldgraf:
I like your UI mockup - I think the language makes it relatively clear that people can't click.

Re: images and such, what if instead we just defaulted to showing `README.md` if it exists, and no filepath is given? Would that be less clunky?

betatim:
> Re: images and such, what if instead we just defaulted to showing README.md if it exists, and no filepath is given?

I like the idea of showing the README. Especially when you just followed a binder link from somewhere-on-the-internet it would give you an idea what it is you are about to see.

My vote would be to show filePath if given, otherwise README.md. With a small note saying that "please beware that we will auto navigate you to your binder once it is ready". And then we deploy it and see what people say. If it leads to outrage we can just remove it again :)

minrk:
> I like your UI mockup

Just to be clear - these aren't mockups, they are screenshots of this PR's current behavior. Here's the latest screenshot with an updated message:

screen shot 2018-06-22 at 14 29 27




> what if instead we just defaulted to showing README.md if it exists, and no filepath is given?

Any "if exists" logic means adding significant logic to binderhub, because it means querying the repo to find files (for GitHub), or cloning the repo (if we want to support other providers in general). It also means that we would need to implement the rendering logic in binderhub. Right now, binderhub never has access to the contents of the repo, only repo2docker. There is no significant difference in complexity between README.md vs binder/PREVIEW.md. Both require cloning the repo, and serving all files from the repo via binderhub to work correctly for things like relatively-linked images.

choldgraf:
Gotcha - in that case I think I'm +1 on merging this and seeing how the community takes to it. @minrk when you're happy w/ it remove the WIP and we can give another review

willingc:
@minrk A design thought. I wonder if adjusting the size of the frame around the nbviewer contents to be smaller in width than the Build Logs bar would be an improvement. I think it's nice to offer additional information but the nbviewer ui seems to be competing a bit for visual attention of the primary UI goal of the page, the build state of a current binder. By making the nbviewer frame smaller, I believe it sends the user the message "hey here's some additional information while you wait" without detracting from the page's primary goal.

minrk:
Makes sense. It's a bit narrower now:

screen shot 2018-06-28 at 11 11 13

choldgraf:
I'm +1 on a merge if @willingc thinks the new push improved things!

willingc:
Looks great to me. Thanks @minrk.

betatim:
Let's deploy this during the week

PR #595, opened 2018-06-26
reorganized index to take out warning and notes

jzf2101:
The warning seemed too harsh and we have a lot of tips and notes on the page, so I've reorganized to make it easier to read.


willingc:
Thanks @jzf2101

Issue #323, opened 2017-12-03
Support Authentication for BinderHub


yuvipanda:
Since binderhub is already a JupyterHub service, we should be able to support authentication there without too much trouble (I think?). This should be great for private installs, and allow us to leverage all the JupyterHub authenticators. It'd also remove our strict requirement on NullAuthenticator in the hub. 

Authentication should be optional though, so we can still run fully open binders, and API users of these don't have to go through a fake auth scheme.

What are the missing pieces, @minrk?

betatim:
The auth would have to happen before the user can click build so we need a handler in binderhub right? This handler would then talk to the jupyterhub auth system. Does the jhub API let me get things that are in the `auth_state`? I'm thinking GitHub tokens that then allow repo2docker to clone a private repo. Or using e.g. the openhumans OAuth to auth users and then inject tokens into the running pod so they can access private APIs for data.

yuvipanda:
http://jupyterhub.readthedocs.io/en/latest/api/services.auth.html is the docs on this I believe.

The OpenHumans use case should already be doable once we have auth, since it's just injecting stuff into the running container (which we can already do in JupyterHub). Am unsure about auth_state availability in the API though.

minrk:
I'll have to do a little checking for how to run a Hub-authenticated service outside the Hub. I think right now, they are confined to services run at `hub.url/services/:name`, but this is mostly artificial, so we can think carefully about what we want to allow.

One question: is this an authenticated Binder, totally dedicated to binder applications, or is it Binder as an additional functionality on top of a regular Hub deployment? I think coexisting would be challenging.

The main things that will need work are:

1. make sure hub-authenticated services can run on other domains/urls (this requires changes to JupyterHub), or run binderhub at `hub.url/services/binder`
2. the launch sequence will need to change from creating temporary users to creating named servers for real, existing users (retrieved from the authenticated launch request)
3. switch culling to servers-only, not users
4. switch launch, ideally, to `jupyterhub-singleuser` instead of `jupyter-notebook`, but this isn't *strictly* necessary.

betatim:
(Can we use bhub and jhub instead of hub and hub? Not always but often I get lost in which hub we are talking about :) )

yuvipanda:
My hope is that we can merge binderhub helm chart into z2jh at some point, completely eliminating the 'do I need a binderhub or jupyterhub?' confusion.

betatim:
[...]

Issue #377, opened 2017-12-21
Persistent storage


choldgraf:
I've had a number of people (especially at universities) ask me if it'd possible to enable persistent storage with BinderHub. The most common use case seems to be to use a BinderHub server to let teachers create repositories/environments for their classes / bootcamps / etc, but they'd like students to have their own "space" where things will persist over time.

I anticipate these requests to just increase with time, but it's also a bit unclear to me exactly how this functionality would be combined w/ BinderHub.  Maybe this could behave in the same way that `nbgitpuller` does.

Just opening this issue since I noticed we don't have another place in this repo where we discuss the topic. I'm curious if people have thoughts on a path forward for something like this.

betatim:
You could use this to provide access to data or longer term storage. Providing a mechanism/example config for people to setup read-only storage that gets mounted to say `~/data` and a read-write storage mounted to `~/personal` would be useful I think. One thing that is a bit tricky is that a repo could contain a directory with the same name as `personal/` or `data/`. Maybe mount the repo at `~/repo` instead of directly at `~` (https://github.com/jupyter/repo2docker/pull/134)

yuvipanda:
The primary thing preventing us from using persistent storage with binder is Authentication. We have no idea which disk should be mounted for which user. #323 should help fix that.

@betatim Mounting the repo in a path under $HOME won't help, since mounting persistent volumes in $HOME will just overwrite everything under. What you need is ability to mount the persistent volume somewhere else, which is already supported. We could also consider a postStart hook that copies files over. We aren't dependent on that repo2docker PR for this particular feature IMO

betatim:
You could mount the repository at `~/repo` and a read only volume at `~/data` and they'd happily coexist no?

You could also mount the shared/persistent volumes somewhere like `/data` but then you can't navigate there with the jupyter tree view because that is rooted in `~/`.

yuvipanda:
I agree! It's not necessary for supporting persistent volumes, but very nice to have! Authentication is a blocker though.

choldgraf:
ok cool - I've updated the top-level comment w/ the current state of this issue so we can keep track of what needs to be done

ctb:
Curious - this is handled by JupyterHub, right? The difference is JupHub doesn't built new images/use multiple images...?

choldgraf:
[...]

PR #596, opened 2018-06-29
[WIP] support authentication

bitnik:
Hi, last days I was working on adding an optional jhub authentication for binder. Firstly, comments on https://github.com/jupyterhub/binderhub/issues/323 helped a lot, thanks!

It is not completed yet and I won't be available next 2 weeks. So I thought I can make this PR and maybe get some feedback from you if I am going on the right way. It would be great for me, if you can review my changes.

Current situation: 
- if `auth.custom.className`set as "nullauthenticator.NullAuthenticator" (default configuration), authentication is not enabled and binderhub works with fake users as before.
- if `auth` set to anything different from `NullAuthenticator`, authentication is enabled and when user comes to binder page, it is redirected to jhub login page. But after user logs in there is `Too many redirections` error. I think this is because `HubAuthenticated.get_current_user` can't gets the user and redirects back to login page and log in page redirects back to binder page (because user is logged in).

Sorry if I miss something to mention in this PR. I had today less time than planned to work on this.

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -266,6 +266,10 @@ class BinderHub(Application):
         help="""API token for talking to the JupyterHub API""",
         config=True,
     )
+
+    hub_api_url = Unicode(os.getenv('JUPYTERHUB_API_URL'),
+                          help="The URL of the Hub API")
+
     hub_url = Unicode(
         help="""
         The base URL of the JupyterHub instance where users will run.
@@ -452,6 +456,7 @@ class BinderHub(Application):
         self.launcher = Launcher(
             parent=self,
             hub_url=self.hub_url,
+            hub_api_url=self.hub_api_url,
             hub_api_token=self.hub_api_token,
             create_user=not self.auth_enabled,
         )
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -267,9 +267,6 @@ class BinderHub(Application):
         config=True,
     )
 
-    hub_api_url = Unicode(os.getenv('JUPYTERHUB_API_URL'),
-                          help="The URL of the Hub API")
-
     hub_url = Unicode(
         help="""
         The base URL of the JupyterHub instance where users will run.
@@ -456,7 +453,6 @@ class BinderHub(Application):
         self.launcher = Launcher(
             parent=self,
             hub_url=self.hub_url,
-            hub_api_url=self.hub_api_url,
             hub_api_token=self.hub_api_token,
             create_user=not self.auth_enabled,
         )
@@ -466,8 +462,6 @@ class BinderHub(Application):
             "docker_image_prefix": self.docker_image_prefix,
             "github_auth_token": self.github_auth_token,
             "debug": self.debug,
-            'hub_url': self.hub_url,
-            'hub_api_token': self.hub_api_token,
             'launcher': self.launcher,
             'appendix': self.appendix,
             "build_namespace": self.build_namespace,
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -18,6 +18,7 @@ import tornado.log
 import tornado.web
 from traitlets import Unicode, Integer, Bool, Dict, validate, TraitError, default
 from traitlets.config import Application
+from jupyterhub.services.auth import HubOAuthCallbackHandler
 
 from .base import Custom404
 from .builder import BuildHandler
@@ -101,6 +102,11 @@ class BinderHub(Application):
         help="""If JupyterHub authentication enabled, 
         require user to login (don't create temporary users during launch) and 
         start the new server for the logged in user.""",
+        config=True)
+
+    use_oauth = Bool(
+        False,
+        help="""If oauth is used, use `HubOAuth` as hub authentication class.""",
         config=True)
 
     port = Integer(
@@ -487,7 +493,10 @@ class BinderHub(Application):
             'template_variables': self.template_variables,
             'executor': self.executor,
             'auth_enabled': self.auth_enabled,
+            'use_oauth': self.use_oauth,
         })
+        if self.auth_enabled and self.use_oauth:
+            self.tornado_settings['cookie_secret'] = os.urandom(32)
 
         handlers = [
             (r'/metrics', MetricsHandler),
@@ -521,6 +530,11 @@ class BinderHub(Application):
                                  tornado.web.StaticFileHandler,
                                  {'path': self.extra_static_path}))
         handlers = self.add_url_prefix(self.base_url, handlers)
+        if self.auth_enabled and self.use_oauth:
+            oauth_redirect_uri = os.getenv('JUPYTERHUB_OAUTH_CALLBACK_URL') or \
+                                 url_path_join(self.base_url, 'oauth_callback')
+            oauth_redirect_uri = urlparse(oauth_redirect_uri).path
+            handlers.insert(-1, (oauth_redirect_uri, HubOAuthCallbackHandler))
         self.tornado_app = tornado.web.Application(handlers, **self.tornado_settings)
 
     def stop(self):
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -96,6 +96,11 @@ class BinderHub(Application):
             proposal.value = proposal.value + '/'
         return proposal.value
 
+    auth_enabled = Bool(
+        False,
+        help="If JupyterHub authentication enabled, don't create users during launch.",
+        config=True)
+
     port = Integer(
         8585,
         help="""
@@ -448,6 +453,7 @@ class BinderHub(Application):
             parent=self,
             hub_url=self.hub_url,
             hub_api_token=self.hub_api_token,
+            create_user=not self.auth_enabled,
         )
 
         self.tornado_settings.update({
@@ -479,6 +485,7 @@ class BinderHub(Application):
             'static_url_prefix': url_path_join(self.base_url, 'static/'),
             'template_variables': self.template_variables,
             'executor': self.executor,
+            'auth_enabled': self.auth_enabled,
         })
 
         handlers = [
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -233,18 +233,6 @@ class BinderHub(Application):
         config=True
     )
 
-    # TODO: Factor this out!
-    github_auth_token = Unicode(
-        None,
-        allow_none=True,
-        help="""
-        GitHub OAuth token to use for talking to the GitHub API.
-
-        Might get throttled otherwise!
-        """,
-        config=True
-    )
-
     debug = Bool(
         False,
         help="""
@@ -468,7 +456,6 @@ class BinderHub(Application):
         self.tornado_settings.update({
             "docker_push_secret": self.docker_push_secret,
             "docker_image_prefix": self.docker_image_prefix,
-            "github_auth_token": self.github_auth_token,
             "debug": self.debug,
             'launcher': self.launcher,
             'appendix': self.appendix,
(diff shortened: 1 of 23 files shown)


minrk:
This is a really great start, thank you!

I think the main thing missing to get the basics started for the auth to work is setting up the oauth client configuration for the Binder service, as illustrated [in this example](https://github.com/jupyterhub/jupyterhub/tree/e81eb9a5f88c3f25e7b83373ba4d8fed05c51fdc/examples/external-oauth). We'll also want to make sure the environment variables defined [here](https://jupyterhub.readthedocs.io/en/stable/reference/services.html) are available to the binder process

binderhub/launcher.py
```diff
@@ -25,11 +25,14 @@
 # Set length of suffix. Number of combinations = SUFFIX_CHARS**SUFFIX_LENGTH = 36**8 ~= 2**41
 SUFFIX_LENGTH = 8
 
+
 class Launcher(LoggingConfigurable):
     """Object for encapsulating launching an image for a user"""
 
     hub_api_token = Unicode(help="The API token for the Hub")
     hub_url = Unicode(help="The URL of the Hub")
+    hub_api_url = Unicode(help="The URL of the Hub API")
```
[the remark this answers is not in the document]
bitnik: yes, that was unnecessary. i removed it in the next commit. thanks :)

bitnik:
@minrk firstly thank you for the feedback and links.

I tested the current state with `NullAuthenticator` (no authentication), `DummyAuthenticator`, `ShibbolethAuthenticator` and `GitHubOAuthenticator` and authentication for binder worked. But there are still some problems and restrictions:

- I had to use user id instead of name in pod and pvc names. `Launcher.username_from_repo` is used to create server names and it restricts the name to be max 30 chars length. But user name limit is 255. That means username + servername usually exceeds 63 char limit of pod and pvc names.
- Another problem is that `KubeSpawner._expand_user_properties` makes username safe but not server name. And some repos had invalid pod and pvc names. PR to fix this: https://github.com/jupyterhub/kubespawner/pull/226
- When auth is enabled `jupyterhub-singleuser` command is used to launch user servers and this requires JupyterHub to be installed in user images.

The extra configuration I used to enable auth with github:
```yaml
jupyterhub:
  cull:
    users: false
  hub:
    baseUrl: /jupyter/
    extraConfig:
      binder: |
        from kubespawner import KubeSpawner
        class BinderSpawner(KubeSpawner):
          def start(self):
              if 'image' in self.user_options:
                # binder service sets the image spec via user options
                self.image_spec = self.user_options['image']
                # to disabled pvc creation
                # self.storage_pvc_ensure = False
                # self.volumes = []
                # self.volume_mounts = []
              return super().start()
        c.JupyterHub.spawner_class = BinderSpawner
        c.JupyterHub.allow_named_servers = True
    services:
      binder:
        url: "/base_url/services/service_name"

  auth:
    type: github
    github:
      clientId: "client id from GitHub"
      clientSecret: "client secret from GitHub"
      callbackUrl: "host/jupyter/hub/oauth_callback"

  singleuser:
    cmd: jupyterhub-singleuser
    storage:
      type: dynamic
      dynamic:
        pvcNameTemplate: claim-{userid}{servername}
        volumeNameTemplate: volume-{userid}{servername}
      homeMountPath: /home/pv

baseUrl: /jupyter/services/binder/
```

As you see in the conf persistent storage is activated (https://github.com/jupyterhub/binderhub/issues/377) and mounted to `/home/pv`.

With this configuration user can launch servers under `/jupyter/services/binder/` (with unique server names) and launch a server with name '' under `/jupyter/hub/home`. In the future I would like to use `KubeSpawner.profile_list` and list user's launched servers (`Spawner` table) under `/hub/home` and manage them. But this will be a new issue.

bitnik:
Before last commit https://github.com/jupyterhub/binderhub/pull/596/commits/630e55d3d202eee185c85c07bed2e5fb0af9d854 named servers was required when authentication is enabled. With https://github.com/jupyterhub/binderhub/pull/596/commits/630e55d3d202eee185c85c07bed2e5fb0af9d854, it is possible to have authentication enabled without named servers (this means each logged in user can only launch 1 server at a time and they can stop/start this server in JupyterHub home).

bitnik:
I am done for now and it would be nice to have a review :)

I think my previous comments were not really explanatory. Here I try to explain the status again.

Currently it is possible to have BinderHub running in 3 ways:

#### 1. Without authentication: 

This is the default configuration. It works in this way if `.Values.jupyterhub.auth.custom.className` is set to `NullAuthenticator` (which is the default). For each launch, BinderHub creates a temporary user and starts a server for that user. 

#### 2. With authentication and without named servers:

BinderHub is configured in this way when `.Values.jupyterhub.auth.custom.className` is set to sth different than `NullAuthenticator`. In this way Binderhub limits each authenticated user to start one server at a time. When user already has a running server, BinderHub returns an error.

This way requires:

- to run BinderHub under `<jupyterhub_base_url>/services/binder/`, because `jupyterhub-services` cookie, which is used by the service for authentication, is available under `<jupyterhub_base_url>/services/`
- to set `jupyterhub.cull.users` to `False`

When authentication is enabled, it is not required to use `jupyterhub-singleuser` command to start user notebook servers but it is better. Because then notebook server is aware of JupyterHub, so JupyterHub can start/stop it with default JupyterHub home actions and users don't need a token to reach a running server (if logged out). But using this command requires 2 things:

- to have `JupyterHub` installed in user images which is not installed by `repo2docker` by default. But `.Values.build.appendix` can be used to install it in each build.

- to use another `BinderSpawner`, e.g.:

```python
from kubespawner import KubeSpawner
class BinderSpawner(KubeSpawner):
  def start(self):
      if 'image' in self.user_options:
        self.image_spec = self.user_options['image']
      return super().start()
```

#### 3. With authentication and with named servers:

This is one step further of way 2. Authenticated users can launch multiple servers by using binder service. BinderHub generates a unique server name by using `Launcher.username_from_repo` method. This way requires enabling named servers (`.Values.jupyterhub.hub.allowNamedServers`). But JupyterHub home doesn't have a UI to manage these named servers yet.

Some further comments:

- When 2 or 3 is used, persistent storage can be activated. But then the question is where to mount it.

- When 2 or 3 is used, `state` field of `Spawner` table and `KubeSpawner.profile_list` feature can be used to list already launched images of user in JupyterHub home. And user can spawn one of the options.

- Soon I also want to try to move binder form into JupyterHub home and test it.

I really like to hear your comments!

binderhub/base.py
```diff
@@ -2,9 +2,45 @@
...
+from urllib.parse import urlencode
 
 
-class BaseHandler(web.RequestHandler):
+def authenticated(method):
+    """Copied from tornado.web.authenticated and `auth_enabled` condition is added.
+    If authentication in not enabled, this decorator doesn't do anything.
+    """
+    @functools.wraps(method)
+    def wrapper(self, *args, **kwargs):
```
minrk: Rather than reimplementing `@web.authenticated`, we can define a `.get_current_user` that returns `'anonymous'` if auth is not enabled.

binderhub/base.py
```diff
@@ -2,9 +2,45 @@
...
+    return wrapper
+
+
+class BaseHandler(HubAuthenticated, web.RequestHandler):
+    """HubAuthenticated by default allows all successfully identified users (see allow_all property)."""
+
+    def initialize(self):
+        super().initialize()
+        if self.settings['auth_enabled'] and self.settings['use_oauth']:
+            self.hub_auth_class = HubOAuth
```
minrk: Shouldn't this be `self.hub_auth = HubOAuth()`, instantiating the object instead of setting the class?

dev-requirements.txt
```diff
@@ -7,3 +7,4 @@ pytest-tornado
 requests
 ruamel.yaml>=0.15
 https://github.com/jupyterhub/chartpress/archive/271c75e.tar.gz
+jupyterhub==0.9.2
```
minrk: I don't think we want to pin jupyterhub in dev-requirements.

binderhub/app.py
```diff
@@ -96,6 +97,23 @@ def _valid_base_url(self, proposal):
...
         return proposal.value
 
+    auth_enabled = Bool(
+        False,
+        help="""If JupyterHub authentication enabled, 
+        require user to login (don't create temporary users during launch) and 
+        start the new server for the logged in user.""",
+        config=True)
+
+    use_oauth = Bool(
```
minrk: It seems we don't nee to have separate `auth_enabled` and `use_oauth` flags. We can have a single `auth_enabled`  flag that sets up `HubOAuth`, perhaps? Or even run with a `jupyterhub_service` flag that will load the full service config (URL, etc.)?

binderhub/launcher.py
```diff
@@ -101,60 +117,67 @@ def username_from_repo(self, repo):
...
+                app_log.error("Error creating user %s: %s\n%s",
+                    username, e, body,
+                )
+                raise web.HTTPError(500, "Failed to create temporary user for %s" % image)
+        elif server_name == '':
+            # authentication is enabled but not named servers
+            # check if user have a running server ('')
+            user_data = await self.get_user_data(username)
+            if server_name in user_data['servers']:
+                raise web.HTTPError(500, "User %s already has a running server." % username)
```
minrk: The error code here ought to be `409` if a server is requested but can't be started because it's running.

binderhub/app.py
bitnik: but don't we need 2 flags to distinguish if an OAuthenticator or another authenticator (e.g. `ldapauthenticator`, `jhub_shibboleth_auth`) is used?

bitnik: I mean that when I use `jhub_shibboleth_auth`, `auth_enabled` is True but `user_oauth` is False. And then `HubAuth` is used (not `HubOAuth`).

bitnik:
thanks a lot for the review. I made some changes according to your comments.

And yes I really like to add tests for authentication. But first I have to learn how to do that. Could you help me about this? I will start with checking the existing tests once again and doing changes if necessary.

bitnik:
sorry, the last commit e1272af04d3f40a6efe92f8b187b1c218ca234be actually doesn't do anything for authentication. i was going through existing tests and end up with doing all these changes. if you wish, i can move this commit to another PR.

in order to test some cases for nbviwer i used `berndweiss/gesis-meta-analysis-2018` repo. if you want, i can change it with https://github.com/binderhub-ci-repos/requirements but then some changes are needed in that repo.

jhamman:
@bitnik - thanks for working on this. I'm really excited to try it out so let us know when you think its ready for some beta testing.

bitnik:
@minrk i started writing tests for auth. could you check https://github.com/jupyterhub/binderhub/pull/596/commits/ab190eec3c985780b40c3260ca22de21339169df if I started correctly?

If looks correct, first problem is that i don't know how to pass binder host ip programatically for `hub.services.binder.url`: https://github.com/jupyterhub/binderhub/pull/596/files#diff-bc01dfc82637c05d27bf80a26ed7d576R5
When I do it manully, there is another problem that `HubAuth.api_token` is not filled with `JUPYTERHUB_API_TOKEN` env variable by default. To fix it, I had to add these lines
```python
    @default('api_token')
    def _default_api_token(self):
        return os.getenv('JUPYTERHUB_API_TOKEN', '')
```

bitnik:
@jhamman I think you can already test current status. I am testing it on our staging server. [Here](https://github.com/gesiscss/orc/blob/jbhub/binderhub/config_test.yaml) is the configuration I used last time.

bitnik:
@minrk when can you review here? I really want to finish this asap :)

bitnik:
this pr contains many changes unrelated to authentication. i am going to restructure it into smaller ones. sorry that i didn't have time to do it before.

bitnik:
moved to https://github.com/jupyterhub/binderhub/pull/666.

PR #601, opened 2018-07-04
use blob urls if filepath is defined

minrk:
nbviewer redirects /blob/ requests for directories to /tree/ but not reliably vice versa

/tree/ is required when the root of a repo is requested, though (i.e. no filepath given)


betatim:
Supersedes #600

PR #602, opened 2018-07-04
Update GitHub unknown ref resolve test case

### LINKED ISSUE #597: Build of master broken
The build fails after merging the nbpreviewer. Apparently because resolving https://travis-ci.org/jupyterhub/binderhub/jobs/398523763 gives a HTTP error. When I visit that URL In my browser I get the expected reply, so not quite sure what is broken :-/

betatim:
Fixes #597 

This updates the recorded output and makes it so that a status of 422 doesn't raise an exception.

One thing I don't understand is why for one test setup this worked (we used the recorded response so tests passed) and one set of tests didn't use the recorded response (those are the ones that failed).


minrk:
Nice!

> One thing I don't understand is why for one test setup this worked

We only use the mock responses when testing pull requests, because they don't have github credentials for rate limit reasons. We use the real GitHub API from master because we have credentials available there.

PR #603, opened 2018-07-06
copybutton extension

choldgraf:
Adds the copybutton extension that I put together to add copy buttons to code blocks. Here's what the built site looks like:

https://predictablynoisy.com/binderhub/setup-binderhub.html#install-binderhub

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -47,6 +47,7 @@ extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.napoleon',
     'autodoc_traits',
+    'sphinx_copybutton'
 ]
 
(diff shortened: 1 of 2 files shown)


betatim:
LGTM.

willingc:
@choldgraf Very nice 👍 

Let's make sure to add this to `doc/environment.yml` so that everyone can benefit (the docs use the environment.yml to build on RTD).

Is it possible to always display the copy icon? I'm not sure people will know it is there if it's initially hidden.

choldgraf:
@willingc re: display, yep that's doable with a CSS over-ride, to the `.copybtn` class, I believe!

PR #604, opened 2018-07-09
adding copybutton dependency

choldgraf:
I believe that this will allow the buttons to show up in readthedocs. does this look correct @willingc ?


willingc:
Thanks @choldgraf

PR #607, opened 2018-07-13
[MRG] Make different repo providers accessible via the UI

betatim:
Getting started on making other repo providers accessible from the UI.

This is what it currently looks like:
repo-providers

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -118,7 +118,7 @@ class BuildHandler(BaseHandler):
         We also ensure that the returned value is DNS safe, by only using
         ascii lowercase + digits. everything else is escaped
         """
-        
+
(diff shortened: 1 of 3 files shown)


betatim:
I think this is good for some reviewing now. The JS is getting pretty ugly/duplicated :-/

repo-providers2

minrk:
Tested locally and it works great!

choldgraf:
how'd I miss this?! cool!

to be clear, this is *just* the UI right? We still don't support a `/v2/gl` request?

betatim:
This is super cool, not just cool because we have supported `/v2/gl/...` for a while now so this adds the missing piece to make it user friendly :)

choldgraf:
really?! man I am behind the times

Issue #611, opened 2018-07-18
Following BinderHub install instructions results in 400 error


choldgraf:
Hey all - I'm running through the BinderHub instructions on our documentation to make sure they work before we present it at PEARC. I got almost everything working, but when I try to build a repo on my BinderHub, I get this error in the Binder pod:

```
      File "<string>", line 4, in raise_exc_info
      File "/usr/local/lib/python3.6/site-packages/tornado/gen.py", line 1063, in run
        yielded = self.gen.throw(*exc_info)
      File "/usr/local/lib/python3.6/site-packages/binderhub/registry.py", line 34, in get_image_manifest
        auth_resp = yield client.fetch(auth_req)
      File "/usr/local/lib/python3.6/site-packages/tornado/gen.py", line 1055, in run
        value = future.result()
      File "/usr/local/lib/python3.6/site-packages/tornado/concurrent.py", line 238, in result
        raise_exc_info(self._exc_info)
      File "<string>", line 4, in raise_exc_info
    tornado.httpclient.HTTPError: HTTP 400: Bad Request
```

Any ideas what's up? If you wanna see the deployment, it's in `binder-sandbox`. 

ps, I'm using the latest tagged binderhub release, not the one in the docs

cc @minrk @yuvipanda @betatim @willingc @jzf2101 in case any of you have deployed binderhub recently for workshops etc

jzf2101:
We didn't get that error right @matthew-brett ?

betatim:
@matthew-brett can you extract the version of the JupyterHub and BinderHub charts you used from your shell history? The `helm install ...` commands should contain them.

Then we could determine if this is a recent bug or not that @choldgraf is seeing.

What registry are you using Chris? There were some changes to add support for dockerhub as registry so maybe they broke something?

choldgraf:
ok I got that figured out (the solution was to stop being an idiot and delete the lines starting in `<`)

I've got a new error:

```
Token exchange failed for project 'binder-sandbox-194621'. Caller does not have permission 'storage.buckets.create'. To configure permissions, follow instructions at: https://cloud.google.com/container-registry/docs/access-control
```

@betatim @jzf2101 @matthew-brett did you run into this in the workshop?

betatim:
Forgetting to remove the lines like `<this says something and should be replaced>` was also a mistake we made. So might be worth thinking about how we can make it more obvious that those lines need removing. Maybe turn them into comments?

matthew-brett:
[...]

PR #614, opened 2018-07-18
updating gcr registry info section

choldgraf:
I realized that the problem in #611 was that I'm an idiot and didn't delete the `<XXX>` lines from the code snippet above. Since we ask users to copy/paste that into their config.yml, I tried modifying the text a bit to reduce the chance that other users would encounter this error. In my case it was because I wasn't paying enough attention, but I think that people who don't understand yaml or JSON particularly well could be confused by this.


betatim:
LGTM.

PR #615, opened 2018-07-19
Fix some syntax errors, remove unused include

betatim:
I removed the `k8s.txt` include because it was broken (and not being included at the moment). From the contents of `k8s.txt` and where it was being included I couldn't quite work out what should be included/in `k8s.txt`.

This also sets up travis so it runs the docs build as part of a PR to make sure we don't break the docs and don't notice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,15 +15,15 @@ before_install:
 install:
 - export PATH=$PWD/bin:$PATH
 - ./ci/install.sh
-- pip install --no-cache-dir . -r dev-requirements.txt
+- pip install -r dev-requirements.txt
 - npm install
 - npm run webpack
 # sphinx requirements
-- pip install --no-cache-dir . -r docs/doc-requirements.txt
+- pip install -r doc/doc-requirements.txt
 script:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -18,9 +18,12 @@ install:
 - pip install --no-cache-dir . -r dev-requirements.txt
 - npm install
 - npm run webpack
+# sphinx requirements
+- pip install --no-cache-dir . -r docs/doc-requirements.txt
```
minrk: We should remove the `--no-cache` here. That's only for docker. There's also no `docs/doc-requirements.txt`, but there is a `docs/environment.yml`.

betatim: but there is `doc/doc-requirements.txt`, fixed that typo.

betatim: I find it annoying/confusing that we have both in the docs. It always leaves me guessing which should I use, who uses which file etc :_/

minrk:
> sphinx-build: error: unrecognized arguments: -M

betatim:
Huzzah!

PR #625, opened 2018-07-31
truncate initial build logs

minrk:
to the last N lines prior to connecting if there are more than N lines of logs already.

Applies when connecting to running builds with a lot of logs already. Shouldn't affect builds that the current request is starting, but should mitigate the cost of many watchers connecting to watch a long-running build that produces a lot of output, which caused the latest [mybinder.org outage](https://github.com/jupyterhub/mybinder.org-deploy/pull/680).

adds `BinderHub.log_tail_lines` config, which is passed as `tail_lines` to the kubernetes log command. In the helm chart as `build.logTailLines`.

Default is 100 lines.

This is the equivalent of the command-line:

    kubectl logs --tail 100 -f build-pod-name

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -10,7 +10,6 @@ import time
 import escapism
 
 import docker
-from kubernetes import client
 from tornado.concurrent import chain_future, Future
 from tornado import gen, web
 from tornado.queues import Queue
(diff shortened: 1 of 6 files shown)


binderhub/builder.py
```diff
@@ -308,6 +310,7 @@ def escape(s):
             docker_host=self.settings['build_docker_host'],
             node_selector=self.settings['build_node_selector'],
             appendix=appendix,
+            log_tail_lines=self.settings['log_tail_lines'],
```
minrk: @choldgraf here is where we pass the arg to kubernetes. It’s the `--tail` arg to kubectl logs.

choldgraf:
woooo iterative bug improvements!

PR #635, opened 2018-08-14
Add directory change to docs build command

betatim:
We need to change back to the top level directory after building the
documentation.

The result of not changing back to the top level directory was that our automated chart releasing feature was broken and no coverage was calculated. This should not happen again as I added a command to the deploy part of travis that changes to "the correct" directory by force.

What we should think about is how we could have caught this error earlier.


betatim:
Will merge this myself to get our chart publishing machinery started again.

PR #637, opened 2018-08-14
Keep watching build pod status

betatim:
Do not stop watching a build pod's state changes

If a client disconnects from the build log we have to continue to watch
for status changes of the build pod in order to clean it up when it
completes or fails.

I think this is the source of the behaviour we are seeing since the last deploy:

pod-activity

where over time build pods in "Completed" state pile up. I think this happens when someone starts a build and then disconnects/closes their browser tab. We stop the watch and so never notice that the pod is now done and clean it up. We introduced the behaviour of stopping watches recently to cope with very long running/never ending builds that are also very popular. They would use up all the threads in the thread pool. Having many threads also used a lot of resources.

Not sure this solution is ideal as it can now happen that there is one space left in the thread pool to start the build but not another one to stream the logs because a watch is still on for a previously created build pod's status. This will lead to confusing behaviour for a user.

A way out would be to make two thread pools: one for build pod state changes and one for log streaming, with the build pod state pool being smaller than the log streaming pool.

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -155,11 +155,9 @@ class Build:
                     if f['type'] == 'DELETED':
                         self.progress('pod.phasechange', 'Deleted')
                         return
-                    if self.stop_event.is_set():
-                        app_log.info("Stopping watch of %s", self.name)
-                        return
                     self.pod = f['object']
-                    self.progress('pod.phasechange', self.pod.status.phase)
+                    if not self.stop_event.is_set():
+                        self.progress('pod.phasechange', self.pod.status.phase)
                     if self.pod.status.phase == 'Succeeded':
                         self.cleanup()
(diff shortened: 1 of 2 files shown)


minrk:
Reverting this does mean that every viewer of a build still takes up a slot in the build thread pool until the build finishes, defeating part of the purpose of the stop event (stopping logs is more important, but this still allows a popular slow build to block all other builds, just half as fast). Maybe we can avoid increasing the build watches by keeping a dict of the currently active Build objects and only starting the logs-watch per request, avoiding duplicate watches of the pod status?

minrk:
Merging since this is an improvement. Going to take a stab at tracking currently watched builds

betatim:
Thanks!

PR #638, opened 2018-08-15
Bump chartpress dependency to work around ruamel.yaml changes

betatim:
Bring in a newer version of chartpress to fix build errors like https://travis-ci.org/jupyterhub/binderhub/jobs/416092817 due to ruamel.yaml changing the type of objects.

The changes this brings in https://github.com/jupyterhub/chartpress/compare/271c75e...2f26733f6ebf9e2b76bd9cfa3b8a8439daa0b16c

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,4 +6,4 @@ pytest-cov
 pytest-tornado
 requests
 ruamel.yaml>=0.15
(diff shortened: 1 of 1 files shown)


betatim:
We broke the build https://travis-ci.org/jupyterhub/binderhub/jobs/416427639 after merging. Going to restart travis.

choldgraf:
Ah this is the same bug that bit us in the littlest jupyterhub I think

PR #642, opened 2018-08-20
Mentioned installation dependencies

Miniland1333:


Miniland1333:
Hmm I don't know why this build failed. This change does not affect the application logic since it only modifies CONTRIBUTING.md

betatim:
LGTM, waiting on travis to rebuild.

CONTRIBUTING.md
```diff
@@ -8,6 +8,12 @@ on GitHub if you don't have a token.
 
 ## Installation
 
+Before you begin, the following utilities are necessary for installation:
+```bash
+python3, python3-pip, npm, curl, socat, docker
```
minrk: socat?

Miniland1333:
When I was first installing binderhub, it failed since I did not have socat installed. While I don't see it as a direct dependency it might be tornado that relies on it.

CONTRIBUTING.md
Miniland1333: When I was first installing binderhub, it failed since I did not have socat installed. While I don't see it as a direct dependency it might be tornado that relies on it.

betatim: Could you try and reconstruct in which situation it is needed? I've not had to install it and can't think where it would be used.

Miniland1333:
```bash
root@reinstall:~/binderhub-master# ./testing/minikube/install-hub
"jupyterhub" has been added to your repositories
Hang tight while we grab the latest from your chart repositories...
...Skip local chart repository
...Successfully got an update from the "jupyterhub" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈ Happy Helming!⎈
E0826 23:00:56.887935   10861 portforward.go:331] an error occurred forwarding 33177 -> 44134: error forwarding port 44134 to pod 0b55e4f53cb7b9838d85915d2169577537b22d3df4d069cc5ac8ad024faf3d87, uid : unable to do port forwarding: socat not found.
Error: transport is closing
  File "./testing/minikube/install-hub", line 51, in <module>
    is_running = name in check_output(['helm', 'list', '-q']).decode('utf8', 'replace').split()
  File "/usr/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['helm', 'list', '-q']' returned non-zero exit status 1.
```

Miniland1333: The revelant line being:
```bash
unable to do port forwarding: socat not found
```

minrk: Maybe add it as a caveat, then, rather than listed as a basic dependency. e.g. `On Linux, if you see an error like:... you may need to install socat`

Miniland1333: It looks like socat is a required dependency for kubernetes/kubelet. https://github.com/kubernetes/kubernetes/issues/26093
https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/pkg/kubelet/dockershim/docker_streaming.go


Since Jupyterhub still relies on kubernetes and socat is not installed by default, I would prefer that socat is listed as a dependency. It looks like on the kubernetes issue board that [not having socat is a common problem](https://github.com/kubernetes/kubernetes/issues/19765) when setting up kubernetes projects.

Miniland1333:
Would you guys think that this statement is too basic to include?
```bash
Make sure that your current working directory is a copy of this repository.
```
While this step is probably automatic for Github veterans, I was confused during my installation since this step is never mentioned in the instructions. Based on this, I had assumed that the instructions would have included something like a curl command or git clone to download binderhub like you did with jupyterhub and helm.

CONTRIBUTING.md
minrk: We should at least clarify that this is a linux-only dependency. I don't have socat on my mac, for instance, and there aren't errors (maybe because it's present in the minikube vm).

Maybe we should switch this to a copy/pasteable `apt-get install` command that we know works on Ubuntu. As it is now, I think the details are too sparse and it looks like I need a `socat` package that I don't actually have or need.

minrk:
@Miniland1333 adding a `git clone` as the first step is great idea. Let's start with cloning the repository:

    git clone https://github.com/jupyterhub/binderhub

Miniland1333:
Okay I made both the socat and git clone changes. In the "before you begin" section, I am not sure if docker and socat would be better off in their own discreet step rather than with the "before you begin".
@minrk I gave you push access if you'd like to make additional formatting changes.

Carreau:
LGTM, we can refine later on if needed.

PR #643, opened 2018-08-22
avoid stopping build-pod watch when nobody’s looking

minrk:
Finishes #637 by commenting-out remaining check of stop_event

We should switch this over to a dedicated thread watching *all* build pods and deleting the stopped ones, rather than a watch thread for each build pod, but this will fix the accumulating build pods for now.


betatim:
LGTM.

PR #653, opened 2018-09-11
bump jupyterhub chart 7310d99...7174419

minrk:
adds nodeselector for user scheduler, configurable minAvailable to pdbs


betatim:
Looks good to me.

Issue #622, opened 2018-07-27
Getting filepath right for notebook preview


betatim:
via #616 

Works:
https://mybinder.org/v2/gh/engineersCode/EngComp1_offtheground/master?filepath=notebooks_en%2F2_Jupyter_strings_and_lists.ipynb

Doesn't work:

https://mybinder.org/v2/gh/engineersCode/EngComp1_offtheground/master?filepath=%2Fnotebooks_en%2F2_Jupyter_strings_and_lists.ipynb 

This is quite hard to spot for users. Maybe we can add a tooltip/hint to the urlpath form field that asks users if they want to add a leading slash?

choldgraf:
I made a very tiny rearranging edit to make it easier to parse :-)

also, is there *ever* a time that we want to have a leading slash there? If not, could we just sanitize it?

betatim:
I don't *think* we need a leading slash ever. But I am not sure if it is true. I'd merge a PR that sanitises or warns.

minrk:
clsoed by #660

Issue #456, opened 2018-02-16
Filepath not allowing path to subdirectory


SylvainCorlay:
- not specifying a filepath starts with the root directory, as expected.
 - specifying a full path to a notebook behaves as expected.

However, we cannot specify a path to a subdirectory (where the notebooks would typically be).

choldgraf:
Hmmm - I just tried this URL:

https://mybinder.org/v2/gh/binder-examples/zero-to-binder/master?filepath=binder

and it **did** put me in the `binder/` folder.

@SylvainCorlay can you share the link you're using that isn't working?

betatim:
I think the problem is
```
https://mybinder.org/v2/gh/binder-examples/zero-to-binder/master?filepath=binder
```
vs
```
https://mybinder.org/v2/gh/binder-examples/zero-to-binder/master?filepath=binder/
```
(trailing slash). Should we sanitise that or just declare it illegal?

choldgraf:
IMO we should sanitize something like that as it seems like a pretty reasonable assumption that users will make

choldgraf:
@SylvainCorlay can you confirm that it works w/o the trailing slash?

choldgraf:
I think this has now been fixed, #495 is closed as well. Please re-open if there is more to be done!

PR #660, opened 2018-09-13
fix file/url path handling for nbviewer and server launch url

bitnik:
with https://github.com/jupyterhub/binderhub/commit/5c993cf4f21d1c2d0274aa18d23259baa213cbae I tried to make js code easier to read and follow. 

https://github.com/jupyterhub/binderhub/commit/c03211f13a4cce7308bd06b1e8c1024fbf7e5199 fixes https://github.com/jupyterhub/binderhub/issues/456 and https://github.com/jupyterhub/binderhub/issues/622

https://github.com/jupyterhub/binderhub/commit/23e94ede3d9230426236c9c4bb9dfc656ef13236 updates loading page test and adds more cases. There I used `berndweiss/gesis-meta-analysis-2018` repo for some cases. If you want, I can change it (and also `minrk/ligo-binder` repo) with https://github.com/binderhub-ci-repos/requirements but then some changes are needed there.

### DIFF
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -27,6 +27,7 @@ class ParameterizedMainHandler(BaseHandler):
     def get(self, provider_prefix, _unescaped_spec):
         prefix = '/v2/' + provider_prefix
         spec = self.get_spec_from_request(prefix)
+        spec = spec.rstrip("/")
         try:
             self.get_provider(provider_prefix, spec=spec)
         except web.HTTPError:
@@ -48,7 +49,7 @@ class ParameterizedMainHandler(BaseHandler):
             nbviewer_url = 'https://nbviewer.jupyter.org/github'
             org, repo, ref = org_repo_ref.split('/', 2)
             # NOTE: tornado escapes query arguments too -> notebooks%2Findex.ipynb becomes notebooks/index.ipynb
-            filepath = self.get_argument('filepath', '')
+            filepath = self.get_argument('filepath', '').lstrip('/')
             blob_or_tree = 'blob' if filepath else 'tree'
             nbviewer_url = f'{nbviewer_url}/{org}/{repo}/{blob_or_tree}/{ref}/{filepath}'
         self.render_template(
@@ -56,7 +57,7 @@ class ParameterizedMainHandler(BaseHandler):
             base_url=self.settings['base_url'],
             provider_spec=provider_spec,
             nbviewer_url=nbviewer_url,
-            urlpath=self.get_argument('urlpath', None),
+            # urlpath=self.get_argument('urlpath', None),
             submit=True,
             google_analytics_code=self.settings['google_analytics_code'],
             google_analytics_domain=self.settings['google_analytics_domain'],
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -25,12 +25,8 @@ class ParameterizedMainHandler(BaseHandler):
     """Main handler that allows different parameter settings"""
 
     def get(self, provider_prefix, _unescaped_spec):
-        # re-extract spec from request.path
-        # get the original, raw spec, without tornado's unquoting
-        # this is needed because tornado converts 'foo%2Fbar/ref' to 'foo/bar/ref'
         prefix = '/v2/' + provider_prefix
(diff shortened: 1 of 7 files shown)


binderhub/main.py
```diff
@@ -25,12 +25,9 @@ class ParameterizedMainHandler(BaseHandler):
...
 
     def get(self, provider_prefix, _unescaped_spec):
-        # re-extract spec from request.path
-        # get the original, raw spec, without tornado's unquoting
-        # this is needed because tornado converts 'foo%2Fbar/ref' to 'foo/bar/ref'
         prefix = '/v2/' + provider_prefix
-        idx = self.request.path.index(prefix)
-        spec = self.request.path[idx + len(prefix) + 1:]
+        spec = self.get_spec_from_request(prefix)
+        spec = spec.rstrip("/")
```
bitnik: `spec = spec.rstrip("/")` also fix for https://github.com/jupyterhub/binderhub/issues/622.

binderhub/main.py
```diff
@@ -44,12 +41,22 @@ def get(self, provider_prefix, _unescaped_spec):
...
             raise web.HTTPError(400, str(e))
 
+        provider_spec = f'{provider_prefix}/{spec}'
+        nbviewer_url = None
+        if provider_prefix == "gh":
+            # we can only produce an nbviewer URL for github right now
+            nbviewer_url = 'https://nbviewer.jupyter.org/github'
+            org, repo, ref = spec.split('/', 2)
+            # NOTE: tornado escapes query arguments too -> notebooks%2Findex.ipynb becomes notebooks/index.ipynb
+            filepath = self.get_argument('filepath', '').lstrip('/')
```
bitnik: `filepath = self.get_argument('filepath', '').lstrip('/')` fixes #622

binderhub/main.py
bitnik: that line also does the same for nbviewer as done for build in https://github.com/jupyterhub/binderhub/blob/master/binderhub/builder.py#L205

binderhub/static/js/index.js
```diff
@@ -90,36 +90,36 @@ Image.prototype.close = function() {
...
 };
-
-Image.prototype.launch = function(url, token, filepath, pathType) {
+Image.prototype.launch = function(url, token, path, pathType) {
     // redirect a user to a running server with a token
-    if (filepath) {
+    if (path) {
       // strip trailing /
       url = url.replace(/\/$/, '');
+      path = decodeURIComponent(path);
```
minrk: Need to be careful with decoding/encoding path. There are two patch kinds and two path sources, which will have different encoding:

- from url param (encoded)
- filepath from form (not encoded)
- urlpath from form (encoded)

Need to make sure that the input path is always treated correctly.

Calling decode here means that encode also needs to be called below in order to preserve the right value.

bitnik:
@minrk I deleted the line with decode.

But if I understood your comment correctly, there is probably still sth wrong. [This part of the js](https://github.com/jupyterhub/binderhub/blob/master/binderhub/static/js/index.js#L408-L418) returns the decoded path value. Maybe we should get the raw query arguments in python code from `request.query`. What do you think?

minrk:
It's tricky! Let's try to define APIs carefully:

How about this:

- filepaths are not encoded because they should reflect the logical path in the repo (e.g. `/path to/file.ipynb`)
- urlpaths are encoded because they are URLs (e.g. `/notebooks/path%20/file.ipynb`)

When coming in from query params, both are encoded by necessity (`?filepath=/path%20file.ipynb&urlpath=/notebooks/path%20file.ipynb`).

I think that means that in general, our code should act as if filepaths are not encoded and urlpaths are, and when filepath is coming from the url param, we should decode it immediately first, so that it can be handled appropriately throughout, the same as if it had been typed in the form.

That means in `build`, taking a path, *filepath* must call `encodeURIComponent` prior to building the URL and *urlpath* must NOT encode the URI.

bitnik:
is it still not correct? @minrk

minrk:
I think it's all set now, thanks!

bitnik:
thank you @minrk!

PR #665, opened 2018-09-18
use binderhub-ci-repos/requirements in tests

minrk:
instead of user-owned repos

### DIFF
--- a/binderhub/tests/test_main.py
+++ b/binderhub/tests/test_main.py
@@ -10,11 +10,11 @@ from .utils import async_requests
 
 @pytest.mark.parametrize(
     "old_url, new_url", [
-        ("/repo/minrk/ligo-binder", "/v2/gh/minrk/ligo-binder/master"),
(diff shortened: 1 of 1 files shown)


betatim:
There was a bit of diversity in what the external repos used to setup stuff. Now we have just a requirements file. Pro: much faster Con: we might miss a few edge cases.

Is it as easy as creating more/new repos in the binderhub-ci org? Then we could make an issue to add a few targeted repos to exercise R and apt-get?

choldgraf:
I agree w/ @betatim that we should make sure to replicate at least the same amount of complexity/diversity in the test files. That said, I think that can be done in a separate PR. What do you think?

betatim:
L:et's merge this once travis is happy and then make a new PR for more (complex) repos.

minrk:
Travis is happy.

The tests changed here aren't really meant for testing builds, they are testing URL/ref resolution, so should be insensitive to what happens once the build starts.

Other tests (test_build) are where we should make sure to have a variety of build types to test, but that is not affected by this PR.

PR #666, opened 2018-09-18
support authentication

bitnik:
Clean version of https://github.com/jupyterhub/binderhub/pull/596/.

From https://github.com/jupyterhub/binderhub/pull/596/:
> Mainly I think the separate `use_oauth` config is unnecessary, since `auth_enabled` seems to never be true without it.

> It seems we don't nee to have separate `auth_enabled` and `use_oauth` flags. We can have a single `auth_enabled` flag that sets up `HubOAuth`, perhaps? Or even run with a `jupyterhub_service` flag that will load the full service config (URL, etc.)?

Maybe I don't understand what @minrk meant here. I still think that we need 2 flags: `auth_enabled` is to hold if authentication is enabled and `use_oauth` is to hold if an authenticator from https://github.com/jupyterhub/oauthenticator is used.

> I would love to get some tests for the authenticated cases. Do you think you are up for that?

I started writing tests but I am not sure what I did is right.

There is already a problem while running auth tests locally. The problem is that `HubAuth.api_token` is not filled with `JUPYTERHUB_API_TOKEN` env variable by default (https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/services/auth.py#L172). As a workaround I added the dynamic default value generator for `api_token` in `jupyterhub/services/auth.py`:

```python
    @default('api_token')
    def _default_api_token(self):
        return os.getenv('JUPYTERHUB_API_TOKEN', '')
```
When I do this, tests work locally. Is this a bug in JupyterHub or is the way I did auth tests wrong?


binderhub/tests/conftest.py
```diff
@@ -118,6 +118,23 @@ def _binderhub_config():
...
 
+@pytest.fixture(scope='session')
+def _binderhub_auth_config():
+    cfg = _binderhub_config()
+    cfg.BinderHub.base_url = '/services/binder/'
+    cfg.BinderHub.auth_enabled = True
+    cfg.BinderHub.use_oauth = False
+    cfg.BinderHub.use_named_servers = False
+
+    os.environ['JUPYTERHUB_API_TOKEN'] = cfg.BinderHub.hub_api_token
```
minrk: Since the class has been imported already, setting the environment variable here is too late. Instead, we can set `cfg.HubAuthenticated.api_token = cfg.BinderHub.hub_api_token` directly. Same, I imagine, for the other environment variables. Set config values directly instead of relying on environment variables.

minrk:
> I still think that we need 2 flags: auth_enabled is to hold if authentication is enabled and use_oauth is to hold if an authenticator from https://github.com/jupyterhub/oauthenticator is used.

I don't think so. The switch here doesn't seem to be related to jupyterhub's own authenticator choice. `use_oauth` selects between the old `HubAuthenticated` (relies on cookies, fragile and inflexible) and the newer `HubOAuthenticated` (uses oauth to authorize with the Hub). I was recommending that we eliminate the `use_oauth` flag and always authenticate with the Hub via `HubOAuthenticated`.

binderhub/launcher.py
```diff
@@ -80,8 +83,16 @@ class Launcher(LoggingConfigurable):
...
 
+    async def get_user_data(self, username):
+        resp = await self.api_request(
+            'users/%s' % username,
+            method='GET',
+        )
+        body = json.loads(resp.body.decode('utf-8'))
+        return body
+
     def username_from_repo(self, repo_url):
```
betatim: Maybe rename it to `unique_name_from_repo` now that it generates more than just a username.

binderhub/launcher.py
```diff
@@ -106,60 +117,67 @@ def username_from_repo(self, repo_url):
...
+        data = {'image': image,
+                'repo_url': repo_url,
+                'token': base64.urlsafe_b64encode(uuid.uuid4().bytes).decode('ascii').rstrip('=\n')}
+
+        # server name to be used in logs
+        _server_name = " {}".format(server_name) if server_name else ''
 
         # start server
-        app_log.info("Starting server for user %s with image %s", username, image)
+        app_log.info("Starting server%s for user %s with image %s", _server_name, username, image)
```
betatim: "Starting server%s for user" -> "Starting server %s for user" (extra space)

binderhub/launcher.py
```diff
@@ -106,60 +117,67 @@ def username_from_repo(self, repo_url):
...
+            except HTTPError as e:
+                if e.response:
+                    body = e.response.body
+                else:
+                    body = ''
+                app_log.error("Error creating user %s: %s\n%s",
+                    username, e, body,
+                )
+                raise web.HTTPError(500, "Failed to create temporary user for %s" % image)
+        elif server_name == '':
```
betatim: `elif server_name == '':` -> `elif server_name:` should be enough to check if a string is not empty.

binderhub/launcher.py
```diff
@@ -174,14 +192,9 @@ def username_from_repo(self, repo_url):
             else:
                 body = ''
 
-            app_log.error("Error starting server for %s: %s\n%s",
-                username, e, body,
-            )
+            app_log.error("Error starting server{} for user {}: {}\n{}".
```
betatim: missing space between `server{}`

bitnik:
@minrk @betatim thanks a lot for reviews!

@minrk sorry, you were right, I was thinking wrong about `HubOAuthenticated` and `HubAuthenticated`. I changed it to use only `auth_enabled` flag and authenticate with JupyterHub through `HubOAuthenticated`.

binderhub/launcher.py
bitnik: `_server_name` already contains the space

binderhub/launcher.py
bitnik: `_server_name` already contains the space

binderhub/tests/conftest.py
bitnik: I think reading https://traitlets.readthedocs.io/en/latest/config.html will help me to figure out how to do this

testing/minikube/jupyterhub-helm-auth-config.yaml
```diff
@@ -0,0 +1,12 @@
+cull:
+  users: false
+hub:
+  services:
+    binder:
+      url: http://10.0.2.2:8585
```
minrk: This hardcoded ip may not be stable for any given deployment

bitnik:
@minrk tests passed after I replaced `10.0.2.2` with host IP.

choldgraf:
boom! this one has been a long time coming! really nice @bitnik :-)

bitnik:
finally :) thanks to all of you!

meeseeksmachine:

https://discourse.jupyter.org/t/a-persistent-binderhub-deployment/2865/1

PR #668, opened 2018-09-19
stop trimming trailing / on files

### LINKED ISSUE #456: Filepath not allowing path to subdirectory

minrk:
regression in #660 to fix #456 applied to urlpath in addition to filepath, when it should only have applied to filepath

Trailing / on a urlpath can be important to routing, so we shouldn't strip it. If there is an error with a urlpath that doesn't work, it is not the responsibility of Binder to fix it.

closes https://github.com/jupyterhub/mybinder.org-deploy/issues/731

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -95,10 +95,11 @@ Image.prototype.launch = function(url, token, path, pathType) {
     if (path) {
       // strip trailing /
       url = url.replace(/\/$/, '');
-      // trim trailing and leading '/'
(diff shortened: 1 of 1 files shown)


betatim:
Nice!

Should we add a test to stop this from coming back?

minrk:
It's pure js and we don't have js-in-the-browser test scaffolding yet. So yes, we definitely should, but it's a bigger job than "just add a test".

PR #669, opened 2018-09-19
Allow admins to add extra scripts to footer

yuvipanda:
This lets us add tracking & other arbitrary JS without having
to add support for each of them specifically.

The Google Analytics code should be moved to use this instead
too.

Ref https://github.com/jupyterhub/mybinder.org-deploy/issues/725

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -86,6 +86,16 @@ class BinderHub(Application):
         config=True
     )
 
+    extra_footer_scripts = Unicode(
+        '',
+        help="""
+        Extra bits of JavaScript that should be loaded in footer of each page.
+
+        Omit the <script> tag. This should be primarily used for
+        analytics code.
+        """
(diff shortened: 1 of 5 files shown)


betatim:
Can you move the GA code too?

And does this let people have yaml like:
```yaml
jscode:
  - GA code goes here
  - some other snippet
```
so multiple snippets are multiple items in the yaml or do we have to mush them all together?

yuvipanda:
@betatim I don't wanna move the GA code just now, will do as a separate PR.

Good point re multiple snippets. Currently this requires you to smush them all together, but we should allow multiple ones. I'll amend.

yuvipanda:
@betatim it takes a dict now. I used a dict rather than list so multiple config files in helm can add scripts.

betatim:
Nice! Dict is indeed much better.

PR #672, opened 2018-09-21
Mark footer scripts as safe

yuvipanda:
This is admin provided JS, so we gotta count it as safe.
Currently, this gets double escaped.

Followup #670


choldgraf:
restarted travis and it became happy, so merging :-)

PR #673, opened 2018-09-24
Mark extra_footer_scripts as safe

yuvipanda:
dictsort returns tuples (like .items()), not values.

Followup #670


yuvipanda:
My local testing setup was borked, but it has been unborked now. I actually verified this one this time. Apologies for the previous two PRs!

choldgraf:
Checks are happy now....so I'm gonna take your word that this is right incantation :-)

Issue #676, opened 2018-10-02
Why does BinderHub Helm chart force the use of kube-lego?


sc250024:
Greetings,

I'm curious why the `ingress.yaml` template was only written for the usage of `kube-lego` ? The below `ingress.yaml` template is taken from version `binderhub-0.1.0-230bbde` of the BinderHub Helm chart:

```yaml
{{- if .Values.ingress.enabled -}}
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: binderhub
  {{- if or (and .Values.ingress.https.enabled (eq .Values.ingress.https.type "kube-lego")) .Values.ingress.annotations }}
  annotations:
    {{- if eq .Values.ingress.https.type  "kube-lego" }}
    kubernetes.io/tls-acme: "true"
    {{ end -}}
    {{- range $key, $value := .Values.ingress.annotations }}
    {{ $key }}: {{ $value | quote }}
    {{- end }}
  {{- end }}
spec:
  rules:
    {{- range $host := .Values.ingress.hosts }}
    - host: {{ $host }}
      http:
        paths:
          - path: /
            backend:
              serviceName: binder
              servicePort: 8585
    {{- end }}
{{- if and .Values.ingress.https.enabled (eq .Values.ingress.https.type "kube-lego") }}
  tls:
    - secretName: kubelego-tls-binder-{{ .Release.Name }}
      hosts:
        {{- range $host := .Values.ingress.hosts }}
        - {{ $host }}
        {{- end }}
{{- end }}
{{- end -}}
```

I understand the desire to have things "just work" out of the box, but this seems inconvenient for anyone who either uses their own TLS certificate, or is using [cert-manager](https://github.com/jetstack/cert-manager/) (like we are) with a cluster-wide wildcard certificate. The kube-lego project will not be maintained in the near future, and mentions cert-manager as a replacement; the [README.md](https://github.com/jetstack/kube-lego/blob/master/README.md) for the kube-lego project even mentions this:

> kube-lego is in maintenance mode only. There is no plan to support any new features. The latest Kubernetes release that kube-lego officially supports is 1.8. The officially endorsed successor is cert-manager.

Any reasons for a very opinionated ingress template?

sc250024:
Opened a PR above to solve this issue.

minrk:
BinderHub doesn't force the use of kube-lego, but it provides shortcuts to more easily enable https with letsencrypt by setting a single flag, rather than requiring lots of custom configuration.

[...]

PR #677, opened 2018-10-02
Less opinionated ingress template

sc250024:
## Description
Changes the `ingress.yaml` template file to be less opinionated, particularly about the use of the `kube-lego` project for TLS . The template is based off of the ingress template generated from `helm create <chart-name>` from Helm client v2.11.0. 

## Motivation and Context
Resolves issue https://github.com/jupyterhub/binderhub/issues/676. Basically, I thought it was strange that the ingress template required usage of `kube-lego`.

## How Has This Been Tested?

Locally tested with the following values:

```yaml
ingress:
  annotations:
    external-dns.alpha.kubernetes.io/hostname: binderhub.company.com
    kubernetes.io/ingress.class: nginx-external
    nginx.ingress.kubernetes.io/configuration-snippet: |
      more_set_headers "Referrer-Policy: strict-origin-when-cross-origin";
      more_set_headers "X-Content-Type-Options: nosniff";
      more_set_headers "X-Frame-Options: SAMEORIGIN";
      more_set_headers "X-XSS-Protection: 1; mode=block";
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
    nginx.ingress.kubernetes.io/ingress.class: nginx-external
  enabled: true
  path: /
  hosts:
  - binderhub.company.com
  tls:
  - hosts:
    - binderhub.company.com
    secretName: star-company-com-tls
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

### DIFF
--- a/helm-chart/binderhub/templates/ingress.yaml
+++ b/helm-chart/binderhub/templates/ingress.yaml
@@ -1,34 +1,32 @@
 {{- if .Values.ingress.enabled -}}
+{{- $ingressPath := .Values.ingress.path -}}
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: binderhub
-  {{- if or (and .Values.ingress.https.enabled (eq .Values.ingress.https.type "kube-lego")) .Values.ingress.annotations }}
+{{- with .Values.ingress.annotations }}
   annotations:
-    {{- if eq .Values.ingress.https.type  "kube-lego" }}
-    kubernetes.io/tls-acme: "true"
-    {{ end -}}
-    {{- range $key, $value := .Values.ingress.annotations }}
-    {{ $key }}: {{ $value | quote }}
-    {{- end }}
-  {{- end }}
+{{ toYaml . | indent 4 }}
+{{- end }}
 spec:
+{{- if .Values.ingress.tls }}
+  tls:
+  {{- range .Values.ingress.tls }}
+    - hosts:
+      {{- range .hosts }}
+        - {{ . | quote }}
+      {{- end }}
+      secretName: {{ .secretName }}
+  {{- end }}
+{{- end }}
   rules:
-    {{- range $host := .Values.ingress.hosts }}
-    - host: {{ $host }}
+  {{- range .Values.ingress.hosts }}
+    - host: {{ . | quote }}
       http:
         paths:
-          - path: /
+          - path: {{ $ingressPath }}
             backend:
               serviceName: binder
               servicePort: 8585
-    {{- end }}
-{{- if and .Values.ingress.https.enabled (eq .Values.ingress.https.type "kube-lego") }}
-  tls:
-    - secretName: kubelego-tls-binder-{{ .Release.Name }}
-      hosts:
-        {{- range $host := .Values.ingress.hosts }}
-        - {{ $host }}
-        {{- end }}
+  {{- end }}
 {{- end }}
-{{- end -}}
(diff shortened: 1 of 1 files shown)


minrk:
The primary goal here is that setting up https should be as easy as possible, i.e. a matter of setting "enable https and it just works". This means that for the default case, you should only need to set:

1. enable https
2. https host
3. letsencrypt email

No knowledge of implementation details like annotations, etc. should be required. That's not to say that these more generic hooks that you've added shouldn't exist, but the letsencrypt shortcut and default behavior should not be removed.

sc250024:
@minrk Noted, I'll make another commit. I agree with your point, I only filed the issue because the ingress template, as it was before, **only** allowed `kube-lego` for TLS.

minrk:
Perfect, thanks! Yes, we don't want to exclude other use cases, especially as kube-lego is on its way out.

sc250024:
If you all are trying to preserve existing `kube-lego` functionality, then I think the new commit I have made is enough. Really what I was after was being able to use my own TLS certificate, so I changed the logic slightly to allow that.

helm-chart/binderhub/templates/ingress.yaml
```diff
@@ -23,12 +23,22 @@ spec:
...
   tls:
+  {{- if eq .Values.ingress.https.type "kube-lego" }}
     - secretName: kubelego-tls-binder-{{ .Release.Name }}
       hosts:
         {{- range $host := .Values.ingress.hosts }}
         - {{ $host }}
         {{- end }}
+  {{- else }}
+    {{- range .Values.ingress.tls }}
+    - hosts:
```
minrk: Not `{{ .Values.ingress.tls | toYaml | nindent 6 }}` ? That should avoid duplicating the structure here in the template.

sc250024: Not sure what you mean here; they do the same thing in the end. As I mentioned, I took the `ingress.yaml` from what's created using `helm create`, which is the closest to a standard you could get since it's in the official binary. The template in full is:

```yaml
{{- if .Values.ingress.enabled -}}
{{- $fullName := include "test-chart.fullname" . -}}
{{- $ingressPath := .Values.ingress.path -}}
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: {{ $fullName }}
  labels:
    app.kubernetes.io/name: {{ include "test-chart.name" . }}
    helm.sh/chart: {{ include "test-chart.chart" . }}
    app.kubernetes.io/instance: {{ .Release.Name }}
    app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- with .Values.ingress.annotations }}
  annotations:
{{ toYaml . | indent 4 }}
{{- end }}
spec:
{{- if .Values.ingress.tls }}
  tls:
  {{- range .Values.ingress.tls }}
    - hosts:
      {{- range .hosts }}
        - {{ . | quote }}
      {{- end }}
      secretName: {{ .secretName }}
  {{- end }}
{{- end }}
  rules:
  {{- range .Values.ingress.hosts }}
    - host: {{ . | quote }}
      http:
        paths:
          - path: {{ $ingressPath }}
            backend:
              serviceName: {{ $fullName }}
              servicePort: http
  {{- end }}
{{- end }}
```

And it's generated from Helm version:
```
Client: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}
```

minrk: No, it's fine if there's a reason. Just an improvement I thought I saw.

PR #678, opened 2018-10-03
Bump default r2d version

yuvipanda:
BinderHub by default seemed to be using a much older
version of repo2docker by default than what we are
using in mybinder.org. This commit brings them to
sync.


betatim:
Restarted the builds.

PR #680, opened 2018-10-03
passthrough BinderHub config

minrk:
This is analogous to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/941 but goes one step further, which I think is appropriate for the different audiences of BinderHub and z2jh. This completely eliminates the helm-specific structure in values.yaml for setting BinderHub configuration.

The result is that previous config structure that exclusively applied to the helm chart:

```yaml
build:
  repo2dockerImage: '...'
```

which the chart and image then mapped to `c.BinderHub.repo2docker_image`. Instead, users set the class config directly in yaml:

```yaml
config:
  BinderHub:
    repo2docker_image: '...'
```

This has the following benefits:

- documentation of BinderHub class configuration *is* documentation of the helm chart configuration. There is only one name/structure to learn. The only thing to know for applying BinderHub configuration in helm is that the traits go under `config.BinderHub`.
- defaults live in exactly one place (the class defaults), no duplication or divergence of defaults.
- adding or modifying a configurable requires no modifications to the chart whatsoever unless it should be handled as a secret.

TODO:

- [x] finish implementation
- [x] make sure registry secret still works
- [x] update config docs
- [x] ensure configurables show up in autodoc output so that we ensure all configuration options are always documented

### DIFF
--- a/binderhub/tests/test_registry.py
+++ b/binderhub/tests/test_registry.py
@@ -12,8 +12,8 @@ from binderhub.registry import DockerRegistry
 
 def test_registry_defaults(tmpdir):
     registry = DockerRegistry(docker_config_path=str(tmpdir.join("doesntexist.json")))
-    assert registry.registry_host == "https://registry.hub.docker.com"
-    assert registry.auth_host == "https://index.docker.io/v1"
+    assert registry.url == "https://registry.hub.docker.com"
+    assert registry.auth_config_url == "https://index.docker.io/v1"
     assert (
         registry.token_url == "https://auth.docker.io/token?service=registry.docker.io"
     )
@@ -37,6 +37,7 @@ def test_registry_username_password(tmpdir):
     registry = DockerRegistry(docker_config_path=str(config_json))
     assert registry.username == "user"
     assert registry.password == "pass"
+    assert registry.url == "https://registry.hub.docker.com"
 
 
 def test_registry_gcr_defaults(tmpdir):
@@ -53,8 +54,8 @@ def test_registry_gcr_defaults(tmpdir):
             f,
         )
     registry = DockerRegistry(docker_config_path=str(config_json))
-    assert registry.registry_host == "https://gcr.io"
-    assert registry.auth_host == "https://gcr.io"
+    assert registry.url == "https://gcr.io"
+    assert registry.auth_config_url == "https://gcr.io"
     assert registry.token_url == "https://gcr.io/v2/token?service=gcr.io"
     assert registry.username == "_json_key"
     assert registry.password == "{...}"
@@ -138,10 +139,10 @@ async def test_get_image_manifest(tmpdir, request):
             f,
         )
     registry = DockerRegistry(
-        docker_config_path=str(config_json), token_url=url + "/token", auth_host=url
+        docker_config_path=str(config_json), token_url=url + "/token", url=url
     )
-    assert registry.registry_host == url
-    assert registry.auth_host == url
+    assert registry.url == url
+    assert registry.auth_config_url == url
     assert registry.token_url == url + "/token"
     assert registry.username == username
     assert registry.password == password
--- a/binderhub/tests/test_registry.py
+++ b/binderhub/tests/test_registry.py
@@ -1,0 +1,149 @@
+"""Tests for the registry"""
+import base64
+import json
+import os
+
+import pytest
+
+from tornado.web import Application, RequestHandler, HTTPError
+
+from binderhub.registry import DockerRegistry
+
+
+def test_registry_defaults(tmpdir):
+    registry = DockerRegistry(docker_config_path=str(tmpdir.join("doesntexist.json")))
+    assert registry.registry_host == "https://registry.hub.docker.com"
+    assert registry.auth_host == "https://index.docker.io/v1"
+    assert (
+        registry.token_url == "https://auth.docker.io/token?service=registry.docker.io"
+    )
+    assert registry.username == ""
+    assert registry.password == ""
+
+
+def test_registry_username_password(tmpdir):
+    config_json = tmpdir.join("dockerconfig.json")
+    with config_json.open("w") as f:
+        json.dump(
+            {
+                "auths": {
+                    "https://index.docker.io/v1": {
+                        "auth": base64.encodebytes(b"user:pass").decode("ascii")
+                    }
+                }
+            },
+            f,
+        )
+    registry = DockerRegistry(docker_config_path=str(config_json))
+    assert registry.username == "user"
+    assert registry.password == "pass"
+
+
+def test_registry_gcr_defaults(tmpdir):
+    config_json = tmpdir.join("dockerconfig.json")
+    with config_json.open("w") as f:
+        json.dump(
+            {
+                "auths": {
+                    "https://gcr.io": {
+                        "auth": base64.encodebytes(b"_json_key:{...}").decode("ascii")
+                    }
+                }
+            },
+            f,
+        )
+    registry = DockerRegistry(docker_config_path=str(config_json))
+    assert registry.registry_host == "https://gcr.io"
+    assert registry.auth_host == "https://gcr.io"
+    assert registry.token_url == "https://gcr.io/v2/token?service=gcr.io"
+    assert registry.username == "_json_key"
+    assert registry.password == "{...}"
+
+# Mock the registry API calls made by get_image_manifest
+
+class MockTokenHandler(RequestHandler):
+    """Mock handler for the registry token handler"""
+
+    def initialize(self, test_handle):
+        self.test_handle = test_handle
+
+    def get(self):
+        scope = self.get_argument("scope")
+        auth_header = self.request.headers.get("Authorization", "")
+        if not auth_header.startswith("Basic "):
+            raise HTTPError(401, "No basic auth")
+        b64_auth = auth_header[6:].encode("ascii")
+        decoded = base64.decodebytes(b64_auth).decode("utf8")
+        username, password = decoded.split(":", 2)
+        if username != self.test_handle["username"]:
+            raise HTTPError(403, "Bad username %r" % username)
+        if password != self.test_handle["password"]:
+            raise HTTPError(403, "Bad password %r" % password)
+        self.test_handle["token"] = token = base64.encodebytes(os.urandom(5)).decode(
+            "ascii"
+        ).rstrip()
+        self.set_header("Content-Type", "application/json")
+        self.write(json.dumps({"token": token}))
+
+
+class MockManifestHandler(RequestHandler):
+    """Mock handler for the registry token handler"""
+
+    def initialize(self, test_handle):
+        self.test_handle = test_handle
+
+    def get(self, image, tag):
+        auth_header = self.request.headers.get("Authorization", "")
+        if not auth_header.startswith("Bearer "):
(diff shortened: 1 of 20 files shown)


minrk:
This is all set and I verified that it works with both dockerhub and gcr.io registries.

I had to consolidate the registry configuration in order to make things sensible, now that we are passing through the true configuration rather than the. This meant moving the registry config to the DockerRegsistry class, off of the base BinderHub class. I also gave it more intelligent defaults, reading values from docker config and inferring e.g. the urls from docker/gcr.io.

minrk:
Additionally, I added to NOTES.txt a big informative failure method if it sees the removed config, so it should be easier to upgrade without silently losing your configuration.

binderhub/tests/test_registry.py
```diff
@@ -0,0 +1,149 @@
...
+                }
+            },
+            f,
+        )
+    registry = DockerRegistry(docker_config_path=str(config_json))
+    assert registry.registry_host == "https://gcr.io"
+    assert registry.auth_host == "https://gcr.io"
+    assert registry.token_url == "https://gcr.io/v2/token?service=gcr.io"
+    assert registry.username == "_json_key"
+    assert registry.password == "{...}"
```
willingc: Do we need an additional test case where the user sets auth_host to docker and then sets registry host to gcr?

doc/setup-binderhub.rst
```diff
@@ -112,40 +113,74 @@ If you are using ``gcr.io``
 To configure BinderHub to use ``gcr.io``, simply add the following to
 your ``config.yaml`` file::
 
-  registry:
-    prefix:  gcr.io/<google-project-id>/<prefix>
-    enabled: true
+  config:
+    BinderHub:
+      use_registry: true
+      docker_image_prefix: gcr.io/<google-project-id>/<prefix>-
```
willingc: Perhaps give a heads up in the doc text about the trailing dash after <prefix>

willingc: Nevermind, it's a few lines below :wink:

doc/setup-binderhub.rst
```diff
@@ -112,40 +113,74 @@ If you are using ``gcr.io``
...
+      use_registry: true
+      docker_image_prefix: <docker-id|organization-name>/<prefix>-
 
 .. note::
 
-   * **``<docker-id/organization-name>``** is where you want to store Docker images.
+   * **``<docker-id|organization-name>``** is where you want to store Docker images.
      This can be your Docker ID account or an organization that your account belongs to.
    * **``<prefix>``** can be any string, and will be prepended to image names. We
      recommend something descriptive such as ``binder-dev`` or ``binder-prod``.
```
willingc: Update the comment about a trailing dash here too.

doc/setup-binderhub.rst
```diff
@@ -112,40 +113,74 @@ If you are using ``gcr.io``
...
 
+If you are using a custom registry
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Authenticating with a Docker registry is slightly more complicated.
+BinderHub knows how to talk to gcr.io and DockerHub,
+but if you are using another registry, you will have to provide more information, in the form of three different urls:
+
+- auth host (added to ``docker/config.json`` from ``registry.host``)
```
willingc: Perhaps (find URL in registry.host and add to docker/config.json)

doc/setup-binderhub.rst
```diff
@@ -112,40 +113,74 @@ If you are using ``gcr.io``
...
+        docker_image_prefix: "your-registry.io/<prefix>-"
+      DockerRegistry:
+        token_url: "https://myregistry.io/v2/token?service="
+        registry_host: "https://registry.myregistry.io"
+
+The two URLs will come from your registry.
+``registry_host`` only needs to be specified if it is different
+from the ``registry.host`` that you specified above.
+
+The ``registry.host`` is the host used in docker ``config.json``,
```
willingc: Make this paragraph a note. `...can differ when the auth host and registry host differ.`

doc/setup-binderhub.rst
```diff
@@ -187,12 +222,13 @@ of the JupyterHub we just deployed.::
...
-  hub:
-    url: http://<IP in EXTERNAL-IP>
+  config:
+    BinderHub:
+      hub_url: http://<IP in EXTERNAL-IP>
 
 Next, upgrade the helm chart to deploy this change::
 
-  helm upgrade <name-from-above> jupyterhub/binderhub --version=v0.1.0-85ac189  -f secret.yaml -f config.yaml
+  helm upgrade <name-from-above> jupyterhub/binderhub --version=v0.1.0-...  -f secret.yaml -f config.yaml
```
willingc: where `...` is the hash of the version's commit.

helm-chart/binderhub/templates/NOTES.txt
```diff
@@ -1,3 +1,148 @@
...
+  BinderHub:
+    use_registry: false
+
+or on GitHubRepoProvider:
+
+config:
+  GitHubRepoProvider:
+    access_token: "..."
+
+
```
willingc: In these examples, where `use_registry` and `access_token` are traits to be configured.

helm-chart/binderhub/templates/NOTES.txt
```diff
@@ -1,3 +1,148 @@
...
+    build_max_age: {{ .Values.build.maxAge }}
+    build_namespace: {{ .Values.build.namespace }}
+    build_node_selector: {{ .Values.build.nodeSelector }}
+    log_tail_lines: {{ .Values.build.logTailLines }}
+    per_repo_quota: {{ .Values.perRepoQuota }}
+{{- end }}
+
+{{- /* end removedConfig */ -}}
+{{- end }}
+
```
willingc: Would the user have ever configured `builder_image` that changed?

helm-chart/binderhub/templates/configmap.yaml
```diff
@@ -3,80 +3,18 @@ apiVersion: v1
...
-  {{ if .Values.build.nodeSelector -}}
-  binder.build-node-selector: {{ toJson .Values.build.nodeSelector | quote }}
-  {{- end }}
-  {{ if .Values.build.appendix -}}
-  binder.appendix: {{ .Values.build.appendix | quote }}
-  {{- end }}
-  binder.log-tail-lines: {{ .Values.build.logTailLines | quote }}
-  binder.build-max-age: {{ .Values.build.maxAge | quote }}
-  binder.build-cleanup-interval: {{ .Values.build.cleanupInterval | quote }}
+{{- /* only pick relevant subset of values. Including too much here can trigger unnecessary binderhub pod restarts */}}
```
willingc: important: only pick...

helm-chart/binderhub/values.yaml
```diff
@@ -15,38 +12,17 @@ resources:
...
-googleAnalyticsDomain:
-
 registry:
-  enabled: false
-  prefix: binderhub-local/
-  host: https://gcr.io
-  authHost:
-  authTokenUrl: https://gcr.io/v2/token?service=gcr.io
-  username: _json_key
+  # docker config.json hot for auth
```
willingc: docker config.json for auth host

binderhub/tests/test_registry.py
minrk: I don't think so. That doesn't really make sense and won't work, but I'd rather let the underlying errors raise than try to preemptively detect such things, since these can be totally arbitrary URLs (private registries, amazon, etc.)

helm-chart/binderhub/templates/_helpers.tpl
```diff
@@ -15,7 +15,32 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
-{{- define "imagePullSecret" }}
-{{- $authHost := default .Values.registry.host .Values.registry.authHost -}}
-{{- printf "{\"auths\": {\"%s\": {\"auth\": \"%s\"}}}" $authHost (printf "%s:%s" .Values.registry.username .Values.registry.password | b64enc) | b64enc }}
+{{/*
+Render docker config.json for the registry-publishing secret.
+*/}}
+{{- define "registryDockerConfig" -}}
+
+{{- /* default auth host */ -}}
+{{- $authHost := (default "https://index.docker.io/v1" .Values.registry.host) }}
```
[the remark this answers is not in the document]
minrk: Since it is changing how these are set in a breaking way, now might be a good time to do it:

How about:

- auth_key: the key in docker config.json to find authentication data for the registry (typically `https://registry.host`)
- registry_url: the URL for the registry (typically `https://registry.host`), set from auth_key by default.

?

minrk:
Thanks for the reviews, folks! All great feedback. Working on implementing it all now.

As for the diff on `c`, the only reason I think we can't really do that on Travis, at least, is that dumping `c` will include some credentials (e.g. github). I can, however, do it by hand using some `kubectl exec` before and after. I'll investigate that before deploying tomorrow.

helm-chart/binderhub/templates/_helpers.tpl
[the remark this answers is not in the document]
minrk: On reading more about the docker login API, this should be a URL, too.

I also think that docker hub may be the *only* registry where the auth_url (the key in docker config) and registry url can differ.

So it might be something that could be simplified to not have two configurables here, and only special-case transforming `index.docker.io` -> `registry.hub.docker.com`.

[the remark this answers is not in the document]
minrk: Reworked to be simplified to emphasize that these values ~never differ for user-specified registries.

The keys are now:

- Registry.url (the actual url of the registry)
- Registry.auth_config_url (the url used to find auth in docker config.json, only in the extremely unlikely situation that it differs from Registry.url)

doc/setup-binderhub.rst
minrk: Ended up reworking this entirely, since the extra URL should be a super special case, I think.

helm-chart/binderhub/templates/NOTES.txt
minrk: This would have been set by `build.repo2dockerImage`, which is covered in the deprecation check for top-level `build`.

binderhub/tests/test_registry.py
willingc: I agree. I was looking at it from a slightly different perspective. The user gets confused about the registry host vs auth host and transposes them. Do we give reasonable messages for the user to not do it this way?

helm-chart/binderhub/templates/_helpers.tpl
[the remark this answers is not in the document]
willingc: +1 to moving to URL instead of host since it's easier for users to comprehend.

minrk:
@willingc Thanks! It took me a while to understand it, since docker hub's urls are all weird (I think for legacy v1 registry api reasons). I think it's actually very hard (or perhaps impossible) for the key in the auth to differ from the registry URL when using anything but the default docker registry. That should mean that users will ~never have to set this, and perhaps that it shouldn't even be configurable.

yuvipanda:
This is also a good time to bump versions on both the package and helm chart. Makes the life of other deployments much easier in knowing when the breaking changes are...

minrk:
Going to deploy this to staging.

PR #682, opened 2018-10-04
Expend the appendix example

bitnik:
This PR:

- updates the login template from https://github.com/jupyterhub/mybinder.org-deploy/blob/master/appendix/templates/login.html
- adds page.html which extends the base page template and removes logout button. I think logout button is confusing and there is already Quit button to stop binder session
- adds new buttons next to Quit button:
  - `Go to repo`: opens the source repo url in new tab
  - `Copy binder link`: copies the binder launch link into clipboard
  - `Copy session link`: copies the binder session link into clipboard. when this link is shared with another user, that user will reach to the same binder session. it will not start a new launch.


bitnik:
I also tried to document the exmaple.

bitnik:
@pdehaye thanks! actually I thought I already did this but I was appending buttons to wrong html element. now they are also in  notebook view page.

pdehaye:
great!

minrk:
Nice, thanks!

PR #685, opened 2018-10-10
track launches in progress if image already exists

bitnik:
Now only launches after successful builds are tracked with `LAUNCHES_INPROGRESS`. This PR makes launches without build (if image already exists) to be tracked too and so they are also written into prometheus database.


minrk:
Thanks!

PR #687, opened 2018-10-11
bump jupyterhub chart

minrk:
https://github.com/jupyterhub/zero-to-jupyterhub/compare/7174419...5ce02a2

This gets the values-passthrough changes in jupyterhub/zero-to-jupyterhub#941, which removes the workaround for using extraConfigMap, instead relying on the passthrough of .Values.custom.


minrk:
Failure is due to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/970

PR #688, opened 2018-10-11
publish chart prior to tests

minrk:
test timing or failure shouldn't prevent or delay chart publication from master. This should shorten the time for deploying changes to mybinder.org.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,20 +38,7 @@ after_success:
 # what previous commands might have done or not
 - cd $TRAVIS_BUILD_DIR
 - codecov
-- |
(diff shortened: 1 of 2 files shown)


minrk:
> You don't waiting for the tests will help catch some typos and such in the charts?

It's not skipping the tests, it's just changing the order, so it shouldn't affect whether errors are caught. Additionally, since this is only for deploying from master, it has already passed tests and review in a pull request, so the odds are *much* higher that a test failure in master after passing in a PR is spurious rather than serious. If there's an issue, we should still fix it, but that's not related to whether publish should be cancelled.

PR #692, opened 2018-10-16
authentication fix

bitnik:
This PR fixes the check if authenticated user has already a running server (''). And also updates required JupyterHub version.

### DIFF
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -145,9 +145,9 @@ class Launcher(LoggingConfigurable):
                     username, e, body,
                 )
                 raise web.HTTPError(500, "Failed to create temporary user for %s" % image)
-        elif server_name:
+        elif server_name == '':
             # authentication is enabled but not named servers
-            # check if user have a running server ('')
+            # check if user has a running server ('')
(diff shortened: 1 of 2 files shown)


choldgraf:
Thanks for the fix!

Is this something testable? Was it a bug or is this just a more explicit way of testing for a running server?

bitnik:
this is better way to inform users if they already have a running server and they try to launch a new binder (this is for the case if named servers are not enabled and each user can only launch one binder at a time). otherwise they get a "Failed to launch image ..." message.

Actually I already implemented it in this way, but after [this review](https://github.com/jupyterhub/binderhub/pull/666/files/a6235c0ec44fa7ab6ae70d84b8c9430890e19bb4#diff-28deb80e5e4fbd96b8b1bdcd3ce1cc3eR148) (the comment is wrong, it actually checks if server name is empty), I changed it wrongly [in this commit](https://github.com/jupyterhub/binderhub/pull/666/commits/3d228a8e5aa52130920f8e66b6c2fb8024b0ba2d#diff-28deb80e5e4fbd96b8b1bdcd3ce1cc3eR148).

PR #693, opened 2018-10-16
reorganizing front page of docs

choldgraf:
This reorganizes the front page of the documentation similar to what we've done with the other JupyterHub projects. It also loads the CSS from teh `binder` repository as a CDN to show how this could work for other projects as well.

Finally, it sets up circleci so that we can demo the documentation before merging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,40 @@
+version: 2
+jobs:
+  build_docs:
+    docker:
+      - image: circleci/python:3.6-stretch
+    steps:
+      # Get our data and merge with upstream
+      - run: sudo apt-get update
+      - checkout
+      # Python env
+      - run: echo "export PATH=~/.local/bin:$PATH" >> $BASH_ENV
+
+      - restore_cache:
+          keys:
+            - cache-pip
+      # Install packages needed for sphinx build
+      - run: pip install --user alabaster traitlets tornado sphinx recommonmark sphinx_copybutton escapism kubernetes prometheus_client jupyterhub docker python-json-logger
+      - save_cache:
+          key: cache-pip
+          paths:
+            - ~/.cache/pip
+
+      # Build the docs
+      - run:
+          name: Build docs to store
+          command: |
+            cd doc
+            make html
+
+      - store_artifacts:
+          path: doc/_build/html/
+          destination: html
+
+
+workflows:
+  version: 2
+  default:
+    jobs:
+      - build_docs
+
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
           keys:
             - cache-pip
       # Install packages needed for sphinx build
-      - run: pip install --user alabaster traitlets tornado sphinx recommonmark sphinx_copybutton escapism kubernetes prometheus_client jupyterhub docker python-json-logger
+      - run: pip install --user -r doc/doc-requirements.txt
       - save_cache:
           key: cache-pip
           paths:
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1,3 +1,4 @@
+=========
 BinderHub
 =========
 
@@ -6,7 +7,7 @@ BinderHub
    BinderHub is under active development and subject to breaking changes.
 
 Getting started
----------------
+===============
 
 The primary goal of BinderHub is creating custom computing environments that
 can be used by many remote users. BinderHub enables an end user to easily
@@ -19,14 +20,15 @@ your BinderHub deployment.
 
 To get started creating your own BinderHub, start with :doc:`create-cloud-resources`.
 
-Extending JupyterHub
---------------------
+.. note::
+   
+   BinderHub uses a JupyterHub running on Kubernetes for much of its functionality.
+   For information on setting up and customizing your JupyterHub, we recommend reading
+   the `Zero to JupyterHub Guide <https://zero-to-jupyterhub.readthedocs.io/en/latest/index.html#customization-guide>`_.
 
-If you’d like to extend your JupyterHub setup, see
-`Zero to JupyterHub <https://zero-to-jupyterhub.readthedocs.io/en/latest/index.html#customization-guide>`_.
 
 BinderHub Deployments
----------------------
+=====================
 
 Our directory of BinderHubs is published at :doc:`known-deployments`.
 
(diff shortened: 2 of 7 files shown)


choldgraf:
Example of the new front page: https://8-89419368-gh.circle-artifacts.com/0/html/index.html

doc/index.rst
```diff
@@ -20,13 +21,13 @@ your BinderHub deployment.
 To get started creating your own BinderHub, start with :doc:`create-cloud-resources`.
 
 Extending JupyterHub
---------------------
+====================
```
willingc: Not directly related to this PR but the "Extending JupyterHub" section feels a bit out of place here.

choldgraf: Ah that's a good point - do you have a suggestion for a different wording / title here? I'm happy to add it into this PR

willingc: Honestly, I think that I would remove this section completely and link to the z2jh in the create-cloud-resources doc. :shrug: It feels out of place here and dilutes the emphasis on BinderHub.

choldgraf: ok, see the latest push. I removed the header and replaced it with a `.. note` directive on the index page and the cloud resources page. Does that work?

choldgraf:
So I put together a really lightweight jupyterhub-alabaster theme here: https://github.com/jupyterhub/alabaster-jupyterhub

All it currently does is add an extra CSS file (and sources from Alabaster).

It worked for me locally, though for some reason circle doesn't seem happy with it because the theme file isn't copied over. I'm not sure why this is...@willingc any ideas for what's up? (btw, sphinx theming was surpringly easy! thanks for the tip)

.circleci/config.yml
```diff
@@ -0,0 +1,41 @@
...
+      - checkout
+      # Python env
+      - run: echo "export PATH=~/.local/bin:$PATH" >> $BASH_ENV
+
+      - restore_cache:
+          keys:
+            - cache-pip
+      # Install packages needed for sphinx build
+      - run: pip install --user alabaster traitlets tornado sphinx recommonmark sphinx_copybutton escapism kubernetes prometheus_client jupyterhub docker python-json-logger
+      - run: pip install -e git+git://github.com/jupyterhub/alabaster_jupyterhub.git#egg=master
```
betatim: Not sure why this fails, I'd try `pip install https://github.com/jupyterhub/alabaster-jupyterhub/archive/master.zip` and see what happens

willingc: Added some modifications to the theme https://github.com/jupyterhub/alabaster-jupyterhub/pull/3

.circleci/config.yml
```diff
@@ -0,0 +1,41 @@
...
+      - run: sudo apt-get update
+      - checkout
+      # Python env
+      - run: echo "export PATH=~/.local/bin:$PATH" >> $BASH_ENV
+
+      - restore_cache:
+          keys:
+            - cache-pip
+      # Install packages needed for sphinx build
+      - run: pip install --user alabaster traitlets tornado sphinx recommonmark sphinx_copybutton escapism kubernetes prometheus_client jupyterhub docker python-json-logger
```
minrk: This should be in a docs/requirements.txt, no?

choldgraf:
woot - docs are building with the new theme:

https://23-89419368-gh.circle-artifacts.com/0/html/index.html

betatim:
Restarted travis job. It feels like in recent days the number of "randomly failed" binderhub CI jobs has gone up. Does anyone know off the top of their head where to look for a change we made that could cause this?

choldgraf:
Hmmm, I'm getting tornado timeout errors on travis...I don't see how they'd be related to this PR though :-/

betatim:
I think it is because we [install the latest version](https://travis-ci.org/jupyterhub/binderhub/jobs/442963667#L832) of the Python kubernetes client library which a few days ago release a new version](https://github.com/kubernetes-client/python/releases) which as a new API https://travis-ci.org/jupyterhub/binderhub/jobs/442963667#L1188

betatim:
However in https://github.com/kubernetes-client/python/blob/3fb2be14e18d84edef094bbd908b6bb3e39aafe6/kubernetes/client/models/v1_host_path_volume_source.py#L43 the `type` keyword argument still shows up, so 🤷‍♂️ ?

choldgraf:
so.......should we wait until master is fixed and I rebase on it?

betatim:
I think getting master fixed is the most important task to work on right now. In my experience the chances of someone taking the time to dive into why the tests fail will only decrease it if we keep making exceptions and merge things despite travis being red. The problem is no one is making the time to work on fixing master which shows that our team is stretched too thinly.

In my mind there is no advantage to a PR being merged into a code base where the tests don't pass as we won't release or deploy till that is fixed.

minrk:
Master appears to be fixed by @willingc and @yuvipanda's recent PRs. This one should be good to go now, no?

betatim:
Probably?

What confuses me is that restarting travis this morning did not make this pass even though I thought it would now do a merge of this branch and the new master. Might just be my misunderstanding of how travis works though :-/

minrk:
Restarting Travis does not update the merge commit, only pushing a new commit does that. But in any case, this PR should not be able to affect the Travis status. That does make me think of something relevant for this PR, though! We should remove the calls to build the docs from .travis.yml, since doc build & test is now done on Circle.

choldgraf:
The latest commit rebases onto master and also removes docs building from Travis (we're only building on Python 3.6 on travis, so I'm doing the same for the docs in Circle). Let's see if things become happy and if not I'll debug

choldgraf:
Travis is still unhappy, here's where things seem to go wrong:

https://travis-ci.org/jupyterhub/binderhub/jobs/445169922#L1623

error is:

```bash
task: <Task pending coro=<get() running at /home/travis/build/jupyterhub/binderhub/binderhub/builder.py:381> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f78383cef78>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/tornado/ioloop.py:719]>
base_events.py            1260 ERROR    Task was destroyed but it is pending!
task: <Task pending coro=<BinderHub.watch_build_pods() running at /home/travis/build/jupyterhub/binderhub/binderhub/app.py:587> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f78383971f8>()]>>
base_events.py            1260 ERROR    Task was destroyed but it is pending!
task: <Task pending coro=<BuildHandler.keep_alive() running at /home/travis/build/jupyterhub/binderhub/binderhub/builder.py:75> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f78383ceeb8>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at /home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/tornado/ioloop.py:719]>
```

It seems something gets deleted that the tests expect to still exist?

choldgraf:
Travis is passing on Master (and this is rebased on master)...but I can't see what in this PR would be breaking things :-/

willingc:
@choldgraf It looks like travis is passing now. I'll merge and we can see how it goes from here.

choldgraf:
huh strange - the travis builds on 3.X took 11 and 12 minutes, respectively. Previously, they were taking like 18-19 minutes before the error (one of them passed at ~18 minutes). Maybe this is an error because of resource availability somewhere and it was just taking too long?

PR #696, opened 2018-10-19
increase async test timeout

willingc:


willingc:
pytest-tornado plugin may be related to the recent Travis flakiness.

PR #702, opened 2018-10-23
update requirements for doc build

willingc:


willingc:
Build and deployment works on my test account: https://readthedocs.org/projects/test-binderhub/builds/7987182/

Let's go with building on pip for RTD for the moment.

PR #704, opened 2018-10-24
require jupyterhub-0.8-d462e1c

bitnik:
There are some JupyterHub configuration fixes: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/compare/5e087da...d462e1c


minrk:
Thanks!

PR #708, opened 2018-11-02
split per-repo and timing metrics

minrk:
each unique value for each label has a cost in prometheus

The timing metrics right now have lots of axes (status, retries, bucket, repo),
resulting in a bit of an explosion.

It is especially an impediment when attempting to look at summary statistics where the repo is not relevant because prometheus must sum over many per-repo vectors.

This removes the repo label from timing metrics and adds counter metrics for the repos, allowing us to hugely reduce the cost of overview metrics, such as build time and success. Additionally, it reduces the resolution of our timing buckets in the histograms.

In particular, the `_count` metrics of the histograms that we use for our success rate will go from ~3k vectors right now to just a few, which should be the biggest improvement we see on the Overview page. This could solve https://github.com/jupyterhub/mybinder.org-deploy/issues/761

Changes:

- reduce buckets for histograms, with separate buckets for builds and launches
- don't resolve builds that take less than a minute or launches that take more than 10
- remove provider, repo metrics from build/launch time metrics
- add (launch|build)_count Counter metrics, with status, provider, repo

The only downside I can think of:

- It was technically possible (though we never did this) to identify which repos were the slowest. This will no longer be possible.

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -240,7 +240,7 @@ class BuildHandler(BaseHandler):
         repo_url = self.repo_url = provider.get_repo_url()
 
         # labels to apply to build/launch metrics
-        self.repo_labels = {
+        self.repo_metric_labels = {
             'provider': provider.name,
             'repo': repo_url,
         }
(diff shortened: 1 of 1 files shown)


binderhub/builder.py
```diff
@@ -17,22 +17,35 @@
...
 from tornado.ioloop import IOLoop
 from tornado.log import app_log
-from prometheus_client import Histogram, Gauge
+from prometheus_client import Counter, Histogram, Gauge
 
 from .base import BaseHandler
 from .build import Build, FakeBuild
 
-BUCKETS = [2, 5, 10, 15, 20, 25, 30, 60, 120, 240, 480, 960, 1920, float("inf")]
+BUILD_BUCKETS = [60, 120, 300, 600, 1800, 3600, 7200, float("inf")]
```
yuvipanda: Add a comment here on why we have different buckets?

binderhub/builder.py
```diff
@@ -224,7 +237,7 @@ def escape(s):
         repo_url = self.repo_url = provider.get_repo_url()
 
         # labels to apply to build/launch metrics
-        self.metric_labels = {
+        self.repo_labels = {
```
yuvipanda: repo_metric_labels maybe?

PR #710, opened 2018-11-06
Validate & flatten analytics events

yuvipanda:
- Use JSONSchema to validate events before emitting them.
  The schema also acts as documentation for the events.
- Flatten the capsule. This makes analytics much easier,
  since you can more easily use dataframe tools like pandas.
  We prevent capsule added fields from being redefined in
  schemas

Ref https://github.com/jupyterhub/mybinder.org-deploy/issues/789

### DIFF
--- a/binderhub/event-schemas/launch.json
+++ b/binderhub/event-schemas/launch.json
@@ -1,0 +1,25 @@
+{
+    "$id": "binderhub.jupyter.org/launch",
+    "version": 1,
+    "title": "BinderHub Launch Events",
+    "description": "BinderHub emits this event whenever a new repo is launched",
+    "type": "object",
+    "properties": {
+        "provider": {
+            "enum": [
+                "GitHub",
+                "GitLab",
+                "Git"
+            ],
+            "description": "Provider for the repository being launched"
+        },
+        "spec": {
+            "type": "string",
+            "description": "Provider specification for the repo being launched. Usually, <reponame>/<commit-specification>"
+        },
+        "status": {
+            "enum": ["success", "failure"],
+            "description": "Success/Failure of the launch"
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


binderhub/event-schemas/launch.json
```diff
@@ -0,0 +1,25 @@
...
+    "version": 1,
+    "title": "BinderHub Launch Events",
+    "description": "BinderHub emits this event whenever a new repo is launched",
+    "type": "object",
+    "properties": {
+        "provider": {
+            "enum": [
+                "GitHub",
+                "GitLab",
+                "Git"
```
betatim:
```suggestion
                "Git",
                "Gist"
```
Do we also need the gist provider?

binderhub/repoproviders.py
```diff
@@ -4,6 +4,8 @@
 Subclass the base class, ``RepoProvider``, to support different version
 control services and providers.
 
+Note: When adding a new repo provider, add it to the allowed values for
+      repo providers in event-schemas/launch.json.
```
betatim: Can we replace this with a test that looks at all classes in the repo provider module and checks that their name is in the list?

binderhub/event-schemas/launch.json
yuvipanda: We don't, it seems to just override github provider without changing name.

helm-chart/images/binderhub/requirements.txt
```diff
@@ -2,5 +2,6 @@ pycurl==7.43.0.1
 tornado==5.0.*
 kubernetes==7.0.*
 jupyterhub==0.9.4
+jsonschema==2.6.0
 # Logging sinks to send eventlogging events to
-git+https://github.com/googleapis/google-cloud-python@d72f443#subdirectory=logging
\ No newline at end of file
+git+https://github.com/googleapis/google-cloud-python@d72f443#subdirectory=logging
```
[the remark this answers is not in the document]
yuvipanda: @minrk I've just reverted this instead: https://github.com/jupyterhub/binderhub/pull/716

binderhub/repoproviders.py
yuvipanda: @betatim done!

PR #711, opened 2018-11-06
Streamline CircleCI documentation building

yuvipanda:
- Reduce duplication in specifying dependencies. 
- Add jupyterhub as a dependency (was missing)
- Streamline .circleci/config, stealing from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1000

Fixes doc error in https://github.com/jupyterhub/binderhub/pull/710

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,12 @@ here = os.path.dirname(__file__)
 with open(os.path.join(here, 'binderhub', '_version.py')) as f:
     exec(f.read(), {}, version_ns)
 
+with open(os.path.join(here, 'requirements.txt')) as f:
+    requirements = [
+        l.strip() in f.readlines()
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -9,6 +9,12 @@
 with open(os.path.join(here, 'binderhub', '_version.py')) as f:
     exec(f.read(), {}, version_ns)
 
+with open(os.path.join(here, 'requirements.txt')) as f:
+    requirements = [
+        l.strip() in f.readlines()
```
betatim:
```suggestion
        l.strip() for l in f.readlines()
```

betatim: Apparently I can't apply my own suggestions :(

PR #715, opened 2018-11-07
Prepare release v0.1

betatim:
Final touches before tagging v0.1

I will not try to setup automatic deploying to PyPI as that has failed a few times for me in the past. Instead will follow

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,0 +1,23 @@
+Version x.y.z (unreleased)
+==========================
+
+Release date: the future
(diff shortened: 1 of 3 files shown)


betatim:
Procedure after merging this, following https://packaging.python.org/guides/distributing-packages-using-setuptools/#uploading-your-project-to-pypi

* python setup.py sdist
* python setup.py bdist_wheel
* edit `$HOME/.pypirc` to use the binder team account

PR #717, opened 2018-11-07
Back to development mode

betatim:
Switches the version tag back to development mode and adds a short checklist to follow for making releases.


choldgraf:
I think those instructions should be much improved, but in a subsequent PR so it doesn't block on the tag being changed. I'll open an issue about that then I'm +1 on merge if the tests pass

PR #725, opened 2018-11-12
fix environment variables for authentication

bitnik:
fix after new helm chart configuration (https://github.com/jupyterhub/binderhub/pull/680)

re-added `JUPYTERHUB_API_URL`, it is needed to connect to hub API during authentication (https://github.com/jupyterhub/jupyterhub/blob/0.9.4/jupyterhub/services/auth.py#L158)

Btw I really like the new BinderHub configuration in helm, thanks a lot!

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -91,8 +91,6 @@ spec:
         {{- if .Values.config.BinderHub.auth_enabled }}
         - name: JUPYTERHUB_API_URL
           value: {{ (print (.Values.config.BinderHub.hub_url | trimSuffix "/") "/hub/api/") }}
-        - name: JUPYTERHUB_SERVICE_PREFIX
-          value: {{ .Values.config.BinderHub.base_url | quote }}
         - name: JUPYTERHUB_BASE_URL
           value: {{ .Values.jupyterhub.hub.baseUrl | quote }}
         - name: JUPYTERHUB_CLIENT_ID
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -88,10 +88,11 @@ spec:
             secretKeyRef:
               name: binder-secret
               key: "binder.hub-token"
-        {{- if or (ne .Values.jupyterhub.auth.type "custom") (ne .Values.jupyterhub.auth.custom.className "nullauthenticator.NullAuthenticator") }}
-        # auth is enabled
+        {{- if .Values.config.BinderHub.auth_enabled }}
+        - name: JUPYTERHUB_API_URL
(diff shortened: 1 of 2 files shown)


helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -88,10 +88,11 @@ spec:
...
               name: binder-secret
               key: "binder.hub-token"
-        {{- if or (ne .Values.jupyterhub.auth.type "custom") (ne .Values.jupyterhub.auth.custom.className "nullauthenticator.NullAuthenticator") }}
-        # auth is enabled
+        {{- if .Values.config.BinderHub.auth_enabled }}
+        - name: JUPYTERHUB_API_URL
+          value: {{ (print (.Values.config.BinderHub.hub_url | trimSuffix "/") "/hub/api/") }}
         - name: JUPYTERHUB_SERVICE_PREFIX
-          value: {{ .Values.baseUrl | quote }}
+          value: {{ .Values.config.BinderHub.base_url | quote }}
```
minrk: Note: this will usually be null. Is that okay?

minrk: I think perhaps instead, this should be moved to binderhub_config.py in the image which has more information available, rather than environment variables.

bitnik: do you mean to have `c.HubOAuth.base_url = get_value('config.BinderHub.base_url ', '/')` in binderhub_config.py?

minrk: `.Values.config.BinderHub.base_url` can definitely be null. It is undefined by default.

This is the difference between setting these things in configuration vs setting them in the chart.

`BinderHub.base_url` on the *instantiated object* cannot be null. But that doesn't mean that it must be specified either in the chart or in 

So, instead, we can gate this with:

```yaml
{{- $default "/" .Values.config.BinderHub.base_url }}
...
```

But I think it would still be better to do that in the image than in the helm env.

bitnik: @minrk I pushed a new commit. But I am not sure if this is what you wanted.

helm-chart/images/binderhub/binderhub_config.py
```diff
@@ -83,6 +83,9 @@ def get_value(key, default=None):
 if os.getenv('BUILD_NAMESPACE'):
     c.BinderHub.build_namespace = os.environ['BUILD_NAMESPACE']
 
+if c.BinderHub.auth_enabled:
```
minrk: Can you add `and 'base_url' in c.BinderHub` here? `base_url` is not typically going to be defined.

minrk:
One more little change, then I think it's set. Thanks!

minrk:
Thanks!

Issue #724, opened 2018-11-12
low contrast of new badge on the form


minrk:
The new badge is nice! But currently our front page shows it on a background of ~the same color as the new blue:

<img width="239" alt="screen shot 2018-11-12 at 12 36 44" src="https://user-images.githubusercontent.com/151929/48345235-c37c4c80-e677-11e8-9f15-2dd9afc63d27.png">

We should either consider a different color for the badge, a different color for the heading backgrounds, or putting the badge somewhere else, so it's not on the blue heading background.

betatim:
<img width="971" alt="screen shot 2018-11-12 at 13 28 01" src="https://user-images.githubusercontent.com/1448859/48347536-d1819b80-e67e-11e8-9678-d87c5ac509d4.png">


A quick experiment with using the gold as background colour for the headings. I think I'd be Ok with that for a while. Maybe change the colour of the text in the heading to be darker?

I think there was a feeling that the old red badge looked too much like "this thing failed". But we could try out a golden badge (your ticket to the chocolate factory ;) ).

betatim:
<img width="980" alt="screen shot 2018-11-12 at 13 31 54" src="https://user-images.githubusercontent.com/1448859/48347718-5bc9ff80-e67f-11e8-9437-29ceda475767.png">

headings are the same colour as those of the other form fields. Feels like that puts a lot of emphasis on those two headings :-/

choldgraf:
Argg good catch @minrk...I agree that the contrast thing needs to be fixed, though I also agree that the "bootstrap blue" we're using is better than bright orange. 

Here's one with the same grey the button uses:

![image](https://user-images.githubusercontent.com/1839645/48361170-ca926300-e655-11e8-9570-6a5a07a48124.png)

thoughts?

betatim:
I like that more than making it yellow/orange/golden. So I'd vote to go with Chris' suggestion for now.

Makes me wonder if we need any background colour at all or if we could style it like the input fields. Something for a new PR though.

choldgraf:
maybe @agahkarakuzu or @tgeorgeux have thoughts as well?

I'd be +1 on making them grey as a stopgap, then iterating more slowly on the right color after that

betatim:
Let's do that. It can't be worse than the low contract situation we have right now :D

choldgraf:
I'll make a PR for this now, but we should double-check that we're all OK with it before merging!

agahkarakuzu:
[...]

PR #726, opened 2018-11-12
updating css for logo contrast

choldgraf:
The current front-page has a fairly poor logo contrast...this updates the background for these sections to the same grey that's used in another button, so that it stands out better.

It's a stop-gap measure so that people can see the new "blue" for the logo, so we should keep the conversation going about what the "right" color is long-term!


betatim:
For some discussion see #724

PR #729, opened 2018-11-13
Add about page to BinderHub

betatim:
This adds a simple about page to BinderHub and the possibility to extend it with a short message.

The idea is to use the space for the message to put some details about who is operating the page and how to contact them. In some jurisdictions you have to put that kind of information on a web property. For mybinder.org we can use it to insert a message about how to get hold of the team, abuse, seucrity reports, etc.

This is my idea for implementing https://github.com/jupyterhub/binder/pull/139#issuecomment-438377209

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
             make html
 
       - store_artifacts:
-          path: doc/build/html/
+          path: doc/_build/html/
           destination: html
 
 workflows:
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -25,7 +25,7 @@ from traitlets import Unicode, Integer, Bool, Dict, validate, TraitError, defaul
 from traitlets.config import Application
 from jupyterhub.services.auth import HubOAuthCallbackHandler
 
-from .base import Custom404
+from .base import AboutHandler, Custom404
 from .build import Build
 from .builder import BuildHandler
 from .launcher import Launcher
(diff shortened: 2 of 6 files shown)


betatim:
This is what the default page looks like with an extra "Hello world" message:
screen shot 2018-11-13 at 22 00 34

If repo2docker was a direct dependency that is installed in the same container we could even dispaly its version:
screen shot 2018-11-13 at 21 58 43
It would be quite useful for debugging to have that kind of information and maybe even at a later point BinderHub might want to expose that in a machine-readable format for the purposes of federation/compatibility. Not quite sure how to get it though without asking the builder to spawn a pod and start repo2docker in it :-/ Maybe sth for a new PR.

choldgraf:
very cool! would it make sense to include a "...running JupyterHub version XXX" in there too?

betatim:
Added a short bit of docs. Is this the place you had in mind?

Showing the JupyterHub version as well would be super. Again, not quite sure how we could obtain it in a computationally cheap fashion :-/

Also made a small change to the circle config as apparently the build directory for the docs has changed.

betatim:
https://115-89419368-gh.circle-artifacts.com/0/html/customizing.html

choldgraf:
yep, that's a good start, we can iterate in subsequent PRs as it becomes clearer how to expose this to users. I'm +1 once travis is happy!

choldgraf:
ok, mergy merge merge! now the fun process of figuring out what this language should be on mybinder.org :-)

@betatim, interested in opening a PR to get the ball rolling there? I imagine we could merge that PR and the binder docs PR at the same time once we like the language

PR #730, opened 2018-11-14
[MRG] Fix RTD builds

### LINKED ISSUE #728: Update the docs build which is missing traitlets
The docs builds are failing for the past 5 days due to missing traitlets. Likely due to refactoring of docs build in #711.

I don't know why the badge is not updating on the README here, but the [builds are failing on RTD](https://readthedocs.org/projects/binderhub/builds/).

betatim:
Closes #728 

Install things the way RTD does in CircleCI to make sure this doesn't come back.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,11 +6,11 @@ jobs:
     steps:
       - checkout
 
-      # Install documentation requirements
+      # Install documentation requirements in the same way that ReadTheDocs
+      # does to check our dependencies are all specified
       - run:
           name: Install doc dependencies
           command: |
-            sudo pip install -r requirements.txt
             sudo pip install -r doc/doc-requirements.txt
 
       # Build the docs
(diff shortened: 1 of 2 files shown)


betatim:
In #711 the CircleCI build was fixed but I think it meant that we had a different environment in CircleCI than in RTD. As a result the CI to test if our documentation was building or not made us think it was working but RTD had stopped working.

choldgraf:
This LGTM - we don't need requirements.txt in the circle build because we're only building the docs, yeah?

choldgraf:
tests are passing and the docs are currently broken, so I'm gonna merge!

betatim:
The last line of the doc's requirements file installs BinderHub which uses the `setup.py` which pulls in all the requirements the BinderHub thinks it needs. Seemed easier than using the `requirements.txt` from BinderHub (and I thought requirements files were on their way out anyway :-) )

PR #735, opened 2018-11-23
Add pod priorities to image builder and cleaner

betatim:
The DIND and image cleaner pods have a priority of zero which I think means they can get evicted when there isn't enough room for pods with higher priority. This is my guess for the build up of pending build pods over night.

Is there a way to add a pod priority to all pods except certain ones? My guess is that we'd want a pod priority of 10 for everything except the placeholder pods. Right now we have to go and annotate matomo, dind, etc by hand to make sure they have a priority above zero.

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -33,6 +33,9 @@ spec:
       initContainers:
 {{ toYaml .Values.initContainers | indent 6 }}
       {{- end }}
+      {{- if .Values.jupyterhub.scheduling.podPriority.enabled }}
+      priorityClassName: {{ .Release.Name }}-default-priority
+      {{- end }}
       nodeSelector: {{ toJson .Values.nodeSelector }}
       {{- if .Values.rbac.enabled }}
       serviceAccountName: binderhub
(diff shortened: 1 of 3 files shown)


betatim:
@consideRatio do you remember why in the default-priority class `globalDefault: false` is set? From a quick look at https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass I don't think there is a downside to doing this and it would save a lot of manual work of adding pod priority annotations to pods.

betatim:
Instead of adding the priority class explicitly I will change the global flag to true.

minrk:
Working this out on gitter: I believe this is a bug in the jupyterhub chart, not setting the default priority as the *global* default priority. Applying that fix and deploying it here should have the desired effect. We may need to restart pods to get it, though.

consideRatio:
Pods does indeed need to be restarted for their priority to be updated.

The new proposed default priority:
```
podPriority:
  enabled: false
  globalDefault: true
  defaultPriority: 10
  userPlaceholderPriority: 0
```

And my recommended config for mybinder.org:
```
podPriority:
  enabled: true
  globalDefault: false
  userPlaceholderPriority: -10
```

minrk:
@consideRatio sounds right to me!

betatim:
Taken care of elsewhere.

PR #738, opened 2018-11-26
fix url to fail and success launch fav icons when base url is different than /

bitnik:
Hi, this was already fixed in https://github.com/jupyterhub/binderhub/pull/507 but I just realised it came back with https://github.com/jupyterhub/binderhub/pull/476 by mistake.

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -275,7 +275,7 @@ function build(providerSpec, log, path, pathType) {
 
     $("#loader").addClass("paused");
     $('div#loader-text p').html("Repository " + spec + " has failed to load!<br />See the logs for details.");
-    update_favicon("/favicon_fail.ico");
+    update_favicon(BASE_URL + "favicon_fail.ico");
     // If we fail for any reason, we will show logs!
     log.show();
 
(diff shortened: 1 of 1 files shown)


betatim:
Do you want to add a test to prevent this from coming back yet again in the future :)?

bitnik:
yes, that would be really good! but I really don't know how to test favicon changes during launch :/

betatim:
Good question. When I wrote my comment this morning I was thinking "ah, just check if you can request the favicons at the correct URL" but the bug was actually in the "client side" code calling the wrong URL, which I am not sure how to test either. We'd need to have something that runs client side JS, so https://github.com/GoogleChrome/puppeteer or similar. That sounds like a lot of work to setup. Would be super cool though.

choldgraf:
Hmmm travis errors are timeouts, so I'm restarting them to see if that fixes things.

re: adding a test, could we make the BASE_URL thing a parameter for the `update_favicon` function, and then test for it with pure JS instead of hard-coding it into the input? We'd have to document this option, in this case...

@betatim it's a bit unclear to me from your last comment whether you'd prefer not merging this PR until a test is merged, or if this is an "open an issue about this and merge this PR" kinda thing...

betatim:
Merge away. It would be good to add a test but unless someone knows how to do it without it requiring some research and experimentation let's get this merged first.

PR #739, opened 2018-11-26
bump hub chart f2d579e...151be76

minrk:
for priority class fixes

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/compare/f2d579e…151be76

### DIFF
--- a/helm-chart/binderhub/requirements.yaml
+++ b/helm-chart/binderhub/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: jupyterhub
-  version: "0.8.0-f2d579e"
+  version: "0.8.0-151be76"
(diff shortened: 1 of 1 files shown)


betatim:
I lost track of what our conclusion was on making a priority class the default or lowering the cut-off. Before we hit merge on this one: GKE with its default configuration (which we are using) will continue to assign 0 as priority for all pods that have no explicit priority class, our user place holder pods will get -10. This means that the placeholders will get evicted before pods which have no explicit priority?

If yes: merge away :)

minrk:
@betatim exactly right. There is a global default (0 by default). The placeholders have a priority class (-10 by default). For deployments with an auto scale  cutoff of 0, you can enable a global default from our chart with a higher value and bump the lace holder priority to 0.

PR #745, opened 2018-12-02
[WIP] Switch to using kubernetes_asyncio rather than kubernetes

### LINKED ISSUE #744: Switch to using kubernetes_asyncio client
Currently we use the official kubernetes python API Client. This is a purely synchronous client, and we must use threadpools / threads to use it from our tornado app.

We should experiment with using https://github.com/tomplus/kubernetes_asyncio instead, which is the asynchronous version of the library. This should (hopefully) simplify our code and let us stop thinking about threads.

yuvipanda:
This works when manually tested, but automated tests need fixing.

Fixes #744

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -11,8 +11,8 @@ import json
 from glob import glob
 from urllib.parse import urlparse
 
-import kubernetes.client
-import kubernetes.config
+import kubernetes_asyncio.client
+import kubernetes_asyncio.config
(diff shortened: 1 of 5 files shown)


yuvipanda:
This will be useful in the long run, but  I can't put too much effort into this right now. Can come back later - I think a lot of the `Build` class queue stuff can instead be made into a async generator.

PR #771, opened 2019-01-16
set HubOAuth.hub_host correctly

bitnik:
@minrk I just realized that I made a mistake in https://github.com/jupyterhub/binderhub/pull/770: I set `JUPYTERHUB_HOST` env variable with hub url, which also contains paths (such as base url). Here I removed that env var and set `HubOAuth.hub_host` in binderhub_config.py.


bitnik:
@minrk could you restart the tests here?

PR #773, opened 2019-01-20
Add liveness probe to binderhub helm chart

### LINKED ISSUE #772: Add a health check for binderhub
On mybinder.org we've had a few instances of binderhub pods not responding even though the hub is. This has required us to manually restart the pods.

Instead, we should have a [readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) set up for binderhub pods that can do this restarting automatically.

betatim:
Defines a liveness probe for BinderHub.

This will fail when the hub stops responding to requests for `/about`. Maybe in a follow up PR we can introduce a "alive and healthy" endpoint that can trigger a restart earlier. Might also implement the "consecutive failed launches" logic in that endpoint.

closes #772

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -103,3 +103,9 @@ spec:
         ports:
           - containerPort: 8585
             name: binder
+        livenessProbe:
+          httpGet:
+            path: /about
+            port: binder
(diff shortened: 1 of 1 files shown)


minrk:
Let's deploy this tomorrow!

PR #775, opened 2019-01-24
add base_url into livenessProbe.httpGet.path

bitnik:
Liveness probe fails with 404 when base_url is set for BinderHub. So I added base_url into `livenessProbe.httpGet.path` if base_url is set, default is "/".


betatim:
Merged!

PR #776, opened 2019-01-25
Move Image class to separate file

captainsafia:
This PR does the following.

- Moves Image class to separate file
- Renames Image to BinderImage to avoid conflicting with Image global
- Adds state transition validation to onStateChange


yuvipanda:
@captainsafia awesome! no rush :)

captainsafia:
This change is deployed to staging and prod now.

PR #781, opened 2019-01-30
Docs link in building page

choldgraf:
This PR adds a short sentence w/ a link to the Binder documentation from the loading page. Many BinderHub users (particularly new ones) will first experience Binder by clicking a link, not from going to `mybinder.org`. This makes sure that we show them a link to the docs on the loading page itself.

### DIFF
--- a/binderhub/templates/loading.html
+++ b/binderhub/templates/loading.html
@@ -14,7 +14,7 @@
 	<p>Loading your Binder...</p>
 </div>
 <div id="loader-links">
-  <p>New to Binder? Check out the <a target="_blank" href="https://mybinder.readthedocs.io/en/latest/">Binder Documentation</a> for more information</p>
+  <p class="text-center">New to Binder? Check out the <a target="_blank" href="https://mybinder.readthedocs.io/en/latest/">Binder Documentation</a> for more information</p>
 </div>
 
 <div id="log-container" class="panel panel-default on-build hidden row">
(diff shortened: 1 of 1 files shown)


binderhub/templates/loading.html
```diff
@@ -13,6 +13,9 @@
 <div id="loader-text">
 	<p>Loading your Binder...</p>
 </div>
+<div id="loader-links">
+  <p class="text-center">New to Binder? Check out the <a target="_blank" href="https://mybinder.readthedocs.io/en/latest/">Binder Documentation</a> for more information</p>
```
betatim:
```suggestion
  <p class="text-center">New to Binder? Check out the <a target="_blank" href="https://mybinder.readthedocs.io/en/latest/">Binder Documentation (opens in new window)</a> for more information</p>
```
What is the current UX recommendation about `target=_blank`? Personally I find it annoying when sites force me to open something in a new tab, as I am perfectly able to click "open in new tab" when I mean to do that. I can see that for new users it would be good to open it in a new window by default though maybe "new" in this context isn't new to binder but new to using browsers?

betatim:
(restarted the build, it was a timeout error)

binderhub/templates/loading.html
choldgraf: The main reason I did it here was because this is a temporary "loading" screen. So in a sense we know what users are expecting: they're expecting to be taken to a running Binder. But if they get redirected to a new page first, then that will never happen.

choldgraf:
looks like travis is happy now, LMK what you think about the target=blank thing...I am not super opinionated on that one

PR #783, opened 2019-01-31
docs for private repo access

minrk:
Also adds missing `r` prefix on git credential env when using access tokens instead of client id/secret. Turns out the fix in https://github.com/jupyter/repo2docker/pull/572 was not quite sufficient (though still correct).


choldgraf:
This is great Min, thanks for the PR! This should close #783 as well!

PR #790, opened 2019-02-14
relax default liveness probe to 10s

minrk:
in case something is being bogged down, 1s may not be enough

mybinder.org failed liveness probes this morning, despite being fully responsive. Possible due to this being a too-short timeout.


betatim:
Merged this despite travis failures as I don't think any of our tests would notice if this was wrong. Deploying this once the image is ready to improve mybinder.org.

PR #791, opened 2019-02-14
updating instructions and adding badges

choldgraf:
A couple small updates:

* removes the ``** sections because they weren't being formatted properly. It replaces this with just \`\`
* Adds badges for the discourse page to the index
* Makes a clarification about < and > symbols


betatim:
LGTM. Weird was that the badges weren't rendering in the preview :-/

PR #792, opened 2019-02-14
badge typo

choldgraf:
@betatim this is why the badge wasn't rendering :-)


betatim:
😂

I looked at the HTML and saw a `img` tag and was puzzled :)

PR #795, opened 2019-02-20
bump dind image to 18.09.2

minrk:
for CVE-2019-5736

users don't appear to be affected since we run them as uid 1000 and the vulnerability [appears to require pods to run as uid 0](https://kubernetes.io/blog/2019/02/11/runc-and-cve-2019-5736/)

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -130,7 +130,7 @@ dind:
   daemonset:
     image:
       name: docker
-      tag: 17.11.0-ce-dind
+      tag: 18.09.2-dind
   storageDriver: overlay2
   resources: {}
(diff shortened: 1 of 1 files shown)


yuvipanda:
@minrk users can still run as uid0 during the docker build itself, so we should still deploy this.

yuvipanda:
Ah, I think I'm saying the same thing you're saying!

minrk:
Yup, I could have been clearer. For "users" I meant jupyterhub user pods ("launches" I guess in Binder terms), so only builds were affected.

This was already deployed to mybinder.org: https://github.com/jupyterhub/mybinder.org-deploy/pull/899 this PR only changes the default for binderhub.

yuvipanda:
Awesome, thanks @minrk

PR #801, opened 2019-03-03
Use different command name in Windows setup

captainsafia:
I ran into this issue while setting up BinderHub for development on Windows.

On Windows, the `npm` command is a wrapper, not a real executable. As a result, calling `subprocess.check_call` with `"npm"` as a parameter fails as an executable with that name does not exist.

The solution is to invoke the actual executable, called `npm.cmd`, when running on Windows systems.


consideRatio:
LGTM! =)

PR #802, opened 2019-03-05
update gist repoprovider

bitnik:
This PR:

- adds user name into gist repo url. repo url is sent to metrics, ~~events~~ and spawner. gist url without user name works too but i think full url is more useful
- changes gist provider name from "GitHub" to "Gist"
- in UI allows giving full gist url

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -62,7 +62,8 @@ function updateRepoText() {
     text = "GitLab.com repository or URL";
   }
   else if (provider === "gist") {
-    text = "Gist ID (username/gistId)";
+    text = "Gist ID (username/gistId) or URL";
+    tag_text = "Git commit SHA";
   }
   else if (provider === "git") {
     text = "Arbitrary git repository URL (http://git.example.com/repo)";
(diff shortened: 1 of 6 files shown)


betatim:
LGTM. The travis failure is an actual failure because we have a test checking the names of the repo providers :)

binderhub/static/js/index.js
```diff
@@ -62,7 +62,8 @@ function updateRepoText() {
     text = "GitLab.com repository or URL";
   }
   else if (provider === "gist") {
-    text = "Gist ID (username/gistId)";
+    text = "Gist ID (username/gistId) or URL";
+    tag_text = "Git commit SHA";
```
betatim: Where does `tag_text` end up getting used?

betatim: Ah, nevermind.

betatim:
Maybe while we wait for the test to get fixed, does anyone remember why we didn't include the username from the beginning? It seems like a good idea so I wonder if historical us figured out a reason not to do it (and we have forgotten it in the mean time).

bitnik:
The test is fixed.

I also checked the related issues and PR for supporting gist but couldn't find out why username was excluded.

bitnik:
I have 1 question: Does extending `provider` list in `event-schemas/launch.json` require version upgrade?

betatim:
What do you think on bumping the schema version @yuvipanda ?

I think it makes sense to bump it.

yuvipanda:
I like it!

I think bumping the version in this case makes sense.

We should make sure our documentation is top notch about the various versions, since - version 1 will probably hang around forever

betatim:
The only docs I could find relating to event logging are https://github.com/jupyterhub/binderhub/blob/master/doc/eventlogging.rst

We could add a link to each schema from those docs as a starting point for documenting them. I'd link directly the the schemas in GitHub (with a SHA1 in the URL).

bitnik:
Launch schema version is bumped to 2 and documented. I also updated builder to emit the new version.

betatim:
Merged! Thanks for this and I'll try and deploy this next Tuesday.

Issue #712, opened 2018-11-07
Query parameter naming scheme?


betatim:
Should we agree a "standard" way of extending the BinderHub URL schema?

The particular use case for me is that I'd like to be able to pass extra information to the spawner via the launch URL. For example which of a set of resource constraints to use or what set of extra volumes to mount on launch.

Right now thinking something like:
* example.com/v2/gh/binder-examples/requirements/master?resources=lots-more&volumes=all-of-them
* example.com/v2/gh/binder-examples/requirements/master?resources=small-cpu&volumes=none

with the spawner knowing how to convert "lots-more", "small-cpu", etc into a set of actual constraints /volume specs.

Right now I can pick essentially any name but I was wondering if we should agree that (say) the `x-` prefix will be reserved for people to do "what they want" and BinderHub itself will only use none prefixed query parameters.

minrk:
I think at least recommending that extensions use `x-` is a good start.

jhamman:
Reviving this conversation again as it seems to address most of #731 and #759. I'm personally not very experienced in the area of building REST-like api. Maybe someone who has some experience in this area could suggest a path toward evaluating schemas?

yuvipanda:
There's also the option of passing JSON as a 'resourceRequests' parameter.

yuvipanda:
Kubernetes resources are pretty much key value pairs of string -> integer, with some special casing. If you read https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/, you'll see that:

1. CPU, RAM and 'ephemeral storage' are special cased
2. Other plugins can define and use any arbitrary values here - see https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#consuming-extended-resources

I think we should do something similar, and define in the 'binderhub runtime parameters spec' a key-value pair mapping for resources, with some well known keys as special case. I like that the values are integers with a small base unit (byte for memory or microcpu for CPU) - this keeps us out of floating point precision issues, and translates well to how most resource managers see these resources. 

These can then be encoded in the URL in many ways. My favorite way is to just repeat the parameter, and use [get_query_arguments](https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.get_query_arguments) to get it as a list. So you'd end up with:

`?resourceRequest=cpu:5000&resourceRequest=memory:1024` in the URL, and can read that properly. This also makes it pretty extensible.

[...]

PR #805, opened 2019-03-13
 pass allowed query parameters to spawner via user options

bitnik:
Firstly, this pull request can wait until a naming scheme for query parameters is decided (https://github.com/jupyterhub/binderhub/issues/712). I just want to propose a way to pass query parameters from launch url through `user_options` to spawner.

This PR introduces a new config `query_parameter_names` which is to be filled with list of allowed query parameter names. With this PR launch url query is appended to build url (`/build/<prefix>/<spec>`). And BinderHub uses `query_parameter_names` before launch and checks if they exist in the build url query and passes found ones with their value to spawner's `user_options` (through JupyterHub's api). 

This PR doesn't do anything with spawner, so default `BinderSpawner` doesn't know how to interpret these new user options. But one can define `c.KubeSpawner.pre_spawn_hook` similar to  https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1178#issuecomment-469813124:

```yaml
hub:
  extraConfig:
    myExtraConfig: |
      async def my_pre_spawn_hook(spawner):
          value = spawner.user_options.get(<parameter_name>)
          # set a spawner setting according to value
          c.spawner.<any KubeSpawner (or Spawner) setting> = <any value>

      c.KubeSpawner.pre_spawn_hook = my_pre_spawn_hook
```

I also added an empty hidden input `user_options` into binder form. When user launches by using form, its value is appended to build url in javascript code. This might be useful when someone has customised binder form template.

When a naming scheme is decided, I can make an update to validate `query_parameter_names` config.

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -21,7 +21,7 @@ import tornado.options
 import tornado.log
 from tornado.log import app_log
 import tornado.web
-from traitlets import Unicode, Integer, Bool, Dict, validate, TraitError, default
+from traitlets import Unicode, Integer, Bool, Dict, validate, TraitError, default, Set
 from traitlets.config import Application
 from jupyterhub.services.auth import HubOAuthCallbackHandler
 
@@ -402,6 +402,16 @@ class BinderHub(Application):
         config=True,
     )
 
+    query_parameter_names = Set(
+        trait=Unicode(),
+        default_value=set(),
+        help="""
+        List of allowed names. Before launch, BinderHub checks if they exist in the query and 
+        pass found ones with their value to spawner's user_options.
+        """,
+        config=True
+    )
+
     @staticmethod
     def add_url_prefix(prefix, handlers):
         """add a url prefix to handlers"""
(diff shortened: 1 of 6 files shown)


bitnik:
Btw I deployed this version of BinderHub on our staging server (https://notebooks-test.gesis.org) with `query_parameter_names: ['run-nbgitpuller']` and a new `BinderSpawner` which knows how to interpret that option (https://github.com/gesiscss/example-binderhub-deployments/blob/master/persistent_storage/config.yaml). So when user launches a repo from "Your projects" table `nbgitpuller` is not executed after server start unless user checks "Update" checkbox. And when launching a repo with binder form, `nbgitpuller` is always executed.

rochaporto:
I've done something similar to allow custom images directly on the jupyterhub profile selection in a deployment relying on KubeSpawner. Looking at allowing the profile selection from the binderhub it seems the same should work, as it's done in this PR.

Is there an update on how Binder is planning to provide this kind of thing? I saw the tickets regarding the naming scheme for query parameters (#712) but relying on user_options seems to do the job as well.

PR #807, opened 2019-03-18
Make clear that it work with GitHub and GitLab

rgaiacs:
Improve on text and form.

Screenshot from 2019-03-18 14-55-10

The drop-down menu on the left should help users to know that GitLab and self-host are allow.

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -69,9 +69,7 @@ function updateRepoText() {
     tag_text = "Git commit SHA";
   }
   $("#repository").attr('placeholder', text);
-  $("label[for=repository]").text(text);
   $("#ref").attr('placeholder', tag_text);
-  $("label[for=ref]").text(tag_text);
 }
 
 function getBuildFormValues() {
--- a/binderhub/templates/index.html
+++ b/binderhub/templates/index.html
@@ -25,9 +25,8 @@
         <h4 id="form-header" class='row'>Build and launch a repository</h4>
         <input type="hidden" id="provider_prefix" value="gh"/>
         <div class="form-group row">
-          <label for="repository">GitHub repo or Gist name or GitLab.com URL</label>
+          <label for="repository">Git repository URL (github.com, gitlab.com or self-host)</label>
           <div class="input-group">
-            <input class="form-control" type="text" id="repository" data-lpignore="true" placeholder="GitHub repository name or link"/>
             <div class="input-group-btn" id="url-type-btn">
               <button type="button" class="btn btn-secondary dropdown-toggle"
                 data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"
@@ -45,6 +44,7 @@
                 <li class="dropdown-item" value="git"><a href="#">Git repository</a></li>
               </ul>
             </div>
+            <input class="form-control" type="text" id="repository" data-lpignore="true" placeholder="GitHub repository name or link"/>
           </div>
         </div>
         <div class="form-row row">


betatim:
I am not quite sure what you mean/want to change compared to what is already there.

Kapture 2019-03-18 at 16 20 23

The textbox label already changes depending on what you select from the dropdown menu.

We are thinking about a new tagline for BinderHub (https://discourse.jupyter.org/t/brainstorm-a-tag-line-for-binderhub/491/4) maybe we can also rework the text at the top of the page:

Kapture 2019-03-18 at 16 20 23

and mention gitlab and github explicitly? Right now it just says "Git repository" with github as the default in the dropdown menu. How to make it more obvious to the impatient reader is going to be tricky because we can't mention everything in the space above :-/ -> needs some wordsmithing I think.

rgaiacs:
> I am not quite sure what you mean/want to change compared to what is already there.

When I read "GitHub repository" I will assume that it only works with GitHub (no GitLab, no Bitbucket, no self host, ...) This might be just me. But this is the reason that I sent the pull request to use the language "Git repository URL" that doesn't change when you select another provider.

The other change is to move the selector from right to left. I think this make clear that other host are available. But might be just me. I appreciate the words of some UX expert here.

I will comment about on the tagline on Discourse to keep the conversation in one place.

choldgraf:
My intuition is also that putting the button on the left will make it more discoverable (and thus easier to infer that multiple repository providers are supported). But I am also not a UI/UX expert :-)

That said, I think I'm +1 on moving it to the left haha

drj11:
I agree with @rgaiacs now that he has explained himself a bit :). There is no reason to "cleverly" change the text, make it static.

In the binder workshop in Manchester I do remember feeling vaguely annoyed that it was for github, and only my personal sense of outrage that someone should tie a service to github that led me to discover the switching dropdown.

psychemedia:
> When I read "GitHub repository" I will assume that it only works with GitHub (no GitLab, no Bitbucket, no self host, ...) 

So does the `Git repository` allow you to run a repo from Bitbucket? (I couldn't figure out what Bitbucket patterned URL to use if so? eg on the repo: https://bitbucket.org/ueacomputervision/image-labelling-tool/src/master/ )

betatim:
The launch link is: https://mybinder.org/v2/git/https%3A%2F%2Fbitbucket.org%2Fueacomputervision%2Fimage-labelling-tool.git/f3ddb33e4839f8a0fe73c168993b405adc13daf0

and when you fill in the form it looks like this:

Screen Shot 2019-05-06 at 19 46 10

You need to explicitly specify the revision you want, you can't write `master` for the "arbitrary git URL" mode as we have no way to resolve that "word" into a revision, which is what we need to check the build cache. For GitHub and GitLab there is a little bit of magic to talk to the respective APIs to resolve it first. Could add this for bitbucket as well if someone is interested enough to open a PR.

meeseeksmachine:

https://discourse.jupyter.org/t/how-to-use-bitbucket-repositories-on-mybinder-org/1000/1

psychemedia:
@betatim ah, yep, thanks... doh! Should have added the revision, I think there were prompts to do so; not concentrating..!

My crude bookmarklet hack would be to scrape the URL from most recent commit on `commits/all` page....

meeseeksmachine:

https://discourse.jupyter.org/t/how-to-use-bitbucket-repositories-on-mybinder-org/1000/1

betatim:
Closed in favour of #1002. Moving the dropdown menu still makes sense but we should update the static texts to something that includes not only Git repositories (hosted anywhere) but also Zenodo and Figshare like things.

Issue #799, opened 2019-03-01
`urlpath` launch parameters available on web interface [feature request]


rgaiacs:
Would be great to have options to launch JupyterLab or RStudion from the web interface.

# Current Web Interface

![screenshot from 2019-03-01 13-55-59](https://user-images.githubusercontent.com/1506457/53643087-154ae180-3c2b-11e9-8ae1-82674206cc74.png)

# Proposal Web Interface

![screenshot from 2019-03-01 13-55-59-alt](https://user-images.githubusercontent.com/1506457/53643096-1aa82c00-3c2b-11e9-9407-050be2315957.png)

sgibson91:
:+1: for this! A check box or drop-down as suggested here would be amazing!

betatim:
Thanks for this! I like the idea of having a button/drop down that acts as a short cut for launching one of the different "standard" UIs.

You can already specify it, but the fact that you opened this issue means it is too well hidden:

![kapture 2019-03-01 at 17 19 14](https://user-images.githubusercontent.com/1448859/53651127-4dabe900-3c46-11e9-8684-2cfde840d4ce.gif)

fm75:
I was planning to add an issue asking that the binder badge section be expanded to give the markdown for classical notebook **and** jupyterlab. Having seen this issue how about just adjusting the UI to show
- In the "Copy the URL... section 
  - classical jupyter notebook
  - JupyterLab 
  - rstudio

In the badge generating section, do the same. Show the markdown and restructured text for all options.

Sooner or later, I expect that the classical notebook URL will be dropped in favor of the JupyterLab option. I have started to put both in, but it would be nice not to have to generate the JupyterLab markdown by hand.

PR #808, opened 2019-03-18
WIP: Improve User Interface to allow selection of interface

rgaiacs:
Related to jupyterhub/binderhub#799.

Screenshot from 2019-03-18 15-20-11

Future work:

- [ ] Integrate Javascript

### DIFF
--- a/binderhub/templates/index.html
+++ b/binderhub/templates/index.html
@@ -47,39 +47,31 @@
             </div>
           </div>
         </div>
+
         <div class="form-row row">
           <div class="form-group col-md-4">
             <label for="ref">Git branch, tag, or commit</label>
             <input class="form-control" type="text" id="ref" placeholder="master"/>
           </div>
-          <div class="form-group col-md-6">
-            <label for="filepath"></label>
-            <div class="input-group">
-              <input class="form-control" type="text" id="filepath"
-                placeholder=""
-              />
-              <div class="input-group-btn" id="url-or-file-btn">
-                <button type="button" class="btn btn-secondary dropdown-toggle"
-                  data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"
-                  title="Specify whether the path to open is a URL or a file"
-                >
-                <span id="url-or-file-selected">
-                {{ 'URL' if urlpath else 'File' }}
-                </span>
-                <span class="caret"></span>
-                </button>
-                <ul class="dropdown-menu">
-                  <li class="dropdown-item"><a href="#">File</a></li>
-                  <li class="dropdown-item"><a href="#">URL</a></li>
-                </ul>
-              </div>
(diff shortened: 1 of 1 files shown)


consideRatio:
Woooo oh I think this is very relevant UX improvement! :heart: @rgaiacs :heart:

betatim:
I think we need to keep the UI element for setting the `urlPath` parameter "by hand" as it is used to do all sorts of things in addition to selecting between the possible UIs we can list.

Maybe the way forward is to create two views of this launch form. The default one would show a simplified view that focusses on "I am a user and these are the goals I want to achieve" (for example selecting a UI from a premade dropdown selection). A "advanced" or "expert" one which shows things from the view we currently have "urlPath" and "filePath" etc that let you set them to what ever value you want but also require you to know what they do and how they interact etc.

rgaiacs:
@betatim I like the idea of simplified vs "advanced". I can devote a few hours on 3rd April to work on this.

betatim:
Sounds great. If someone else has time before then are you Ok with them creating a PR for that?

rgaiacs:
> If someone else has time before then are you Ok with them creating a PR for that?

Absolutely! Anyone is welcome to take over this PR. :heart:

drj11:
why bother with the "advanced" version? Document the URL structure and advanced users can splice the URL together themselves.

betatim:
> why bother with the "advanced" version?

because even expert users enjoy a bit of UI candy to help them to what they want to do.

drj11:
yes, with infinite budget, of course.

consideRatio:
Hey @rgaiacs! I'm going through various PRs hoping to help them resolve one way or another or perhaps get an updated description about the current status of things. What do you think about this PR?

The current ui looks like this.

Screenshot from 2021-09-23 01-38-40

Is it really worth having a choice of UIs? I'm thinking it could indicate to the user that the listed UIs will be available, which the binderhub UI can't really say. Currently we count on JupyterLab to be available and have a fallback to Jupyter Notebook after #1383 is merged.

rgaiacs:
Hey @consideRatio! Thanks reaching out.

This pull request was open during one event organised by [The Turing Way](https://the-turing-way.netlify.app/welcome). We were using R Studio from Binder and I wanted to open an R project with Jupyter Notebook.

I'm happy for you to close this pull request based in that Binder can't upfront which UI will be available and users requesting an unavailable UI will lead to poor user experience.

consideRatio:
Thanks for the quick response @rgaiacs! I'll go ahead with a close then.

One thing I take with me going onwards is that it would be valuable if we could do some inspection work of built repo's or similar, as such data could be used to enable features like this in a reliable way.

Thank you for being an active member of the community @rgaiacs! :heart:

PR #818, opened 2019-04-02
Support basic (htpasswd) authentication for registry

dylex:
For small, local deployments, registries can use [basic authentication](https://docs.docker.com/registry/deploying/#native-basic-auth) which doesn't support tokens. This allows using such registries by just including the user/pass directly in the registry request, rather than doing an extra token step. Tested with a local kube cluster and registry.

It's enabled by just setting token_uri to None, but could also add an explicit/separate setting for this.

### DIFF
--- a/doc/setup-binderhub.rst
+++ b/doc/setup-binderhub.rst
@@ -177,6 +177,10 @@ Second, you will need to instruct BinderHub about the token URL::
       DockerRegistry:
         token_url: "https://myregistry.io/v2/token?service="
 
+If you setup your own local registry using
+`native basic HTTP authentication <https://docs.docker.com/registry/deploying/#native-basic-auth>`__
+(htpasswd), you can set ``token_url`` to ``None``.
+
 .. note::
 
     There is one additional URL to set in the unlikely event that docker config.json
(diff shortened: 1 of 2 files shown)


doc/setup-binderhub.rst
```diff
@@ -177,6 +177,8 @@ Second, you will need to instruct BinderHub about the token URL::
       DockerRegistry:
         token_url: "https://myregistry.io/v2/token?service="
 
+If you are using basic HTTP authentication (htpasswd), you can set token_url to None.
```
betatim:
```suggestion
If you are using basic HTTP authentication (htpasswd), you can set ``token_url`` to ``None``.
```

Will people know why/if they are using basic HTTP auth? Maybe we can expand this a bit to say what you said in the PR description "small local deployments with ..." so that people have an easier time realising when this is appropriate to use.

betatim:
Looks good to me. One small comment on the docs. Do you know how easy it would be to setup a test for this so that the functionality doesn't break in the future?

dylex:
Updaed the documentation with a bit more and a link to the registry docs (I think I got that syntax right).

It's not too hard to run a registry in a pod in minikube, but it is a bit tricky getting the service and TLS certs setup correctly so clients are actually willing to talk to it (I haven't gotten plain http to work), so I'm not exactly sure the best way to test it. It's possible some existing cloud registry works with basic auth as well as token.

betatim:
Thanks for the PR!

Issue #794, opened 2019-02-19
a binder deployment with authentication and persistent storage


bitnik:
Currently we are working on a binder deployment with authentication and persistent storage enabled and with a user interface in JupyterHub home page, where users can manage their repositories/projects.

For this purpose we have now a deployment running on https://notebooks-test.gesis.org/jupyter/. When you first login, you will see the JupyterHub home page (https://notebooks-test.gesis.org/jupyter/hub/home) with 2 parts: "Your projects" table and the classical binder form with some parts hidden:

![ss1](https://user-images.githubusercontent.com/6059799/52999000-f0818d80-3424-11e9-9367-ac68de7e3ac4.png)

Binder is running under https://notebooks-test.gesis.org/jupyter/services/binder/ and you can also use it but in this deployment the idea is that you don't need to use it directly.

## How it works
Firstly some preliminary information:
- Each user can start 1 server at a time (named servers are not activated)
- Each user gets 1 persistent volume and it is mounted on `/home/jovyan`

### Binder form
It is the classical form with 'share url' and 'badge url' parts are hidden. And it has 1 limitation: branch/tag/commit field is readonly and always "master". When user launches a repo via form:
- always the latest version of the repo is built (last commit in master branch) and server is started with this image
- `nbgitpuller` is used to pull the code under a sub directory `/home/jovyan/{repo_dir}`. `repo_dir` is  generated by using provider name, user/org name and repo name. And server is started on that sub directory (you can start a new terminal and there you can list all directories of projects). `nbgitpuller` is not executed for the default repo (`gesiscss/data_science_image`).
- each new launched repo is added into "Your Projects" table. This list is saved in `state` field of `Spawners` table and only last 10 launched repos are saved.

In short, binder form is used to create a new project and update it from remote. 

### Your Projects
When first login, user has there only the default repo (`gesiscss/data_science_image`). Each repo which is built and launched via binder form is added in this table and user can re-start that repository by using the start buttons on each row. When user clicks on a start button in the table:
- A server started by using the image (commit) that user last time worked with
- Right now it is not working but we want to skip `nbgitpuller` command execution  on server start when server is started from projects table, so that user can continue working on where they left. We can do this by passing an option to spawner (I think this is very related to https://github.com/jupyterhub/binderhub/issues/712)
- We are also thinking about having a `delete` button in the actions of table which removes the repository from the table and deletes the folder of the repo in user's persistent volume. Right now we have the button in the actions column but it doesn't do anything.

[...]

PR #819, opened 2019-04-03
[WIP] possible solution to propagate informative spawn failure messages from spawner to bhub ui

bitnik:
First of all I need help here :) I don't fully understand what is happening (specially on JupyterHub side) but here is what I understand so far:

1. `slow_spawn_timeout` setting is by default 0:
 - Code: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/8ed2f8111b5575dc5df29afb114a8ee5906f9a96/images/hub/jupyterhub_config.py#L30-L32
- PR: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/884
2. This means any kind of failure during spawn is caught by jhub as `TimeoutError` and jhub always raises 500 error (even maybe spawner sends 409 error): https://github.com/jupyterhub/jupyterhub/blob/e89836c035f79a44cb5ebc1126e53c6f605464c1/jupyterhub/handlers/base.py#L887-L924
3. BinderHub catches this 500 error of API request and  retries same API request 4 more times: https://github.com/jupyterhub/binderhub/blob/1835d07222388da2a23765899cd006e6f6462827/binderhub/launcher.py#L75-L84
4. After retrying, launch fails with https://github.com/jupyterhub/binderhub/blob/1835d07222388da2a23765899cd006e6f6462827/binderhub/launcher.py#L197
5. Then BinderHub retries launch process (in total 3 times by default)
6. In total BinderHub makes 12 API requests and then user gets the standard error ("Failed to launch image ...") on the UI. During these 12 requests, hub restarts 2 times because `consecutive_failure_limit` is by default 5 (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/8ed2f8111b5575dc5df29afb114a8ee5906f9a96/jupyterhub/values.yaml#L17)

The same process happens regardless of type of error from spawner. In this PR to solve this issue,  `slow_spawn_timeout` is set to 10 seconds (default value), so BinderHub gets actual error from spawner. I also set `consecutiveFailureLimit` to 0, so hub doesn't restart after informative failures of spawner. But this is actually not good, because hub also ignores real errors that it has to restart to get rid of them.

I updated build and launch code too. So now it propagates 409 error messages from JupyterHub API to UI and doesn't retry to launch. And as I wrote before I need help here, this part can be wrong or missing. So I don't mind if we totally changed it.

related to https://github.com/jupyterhub/binderhub/issues/712 and  https://github.com/jupyterhub/binderhub/pull/805

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -12,7 +12,7 @@ import escapism
 import docker
 from tornado.concurrent import chain_future, Future
 from tornado import gen
-from tornado.web import Finish, authenticated
+from tornado.web import Finish, authenticated, HTTPError
 from tornado.queues import Queue
 from tornado.iostream import StreamClosedError
 from tornado.ioloop import IOLoop
@@ -518,8 +518,8 @@ class BuildHandler(BaseHandler):
                         status=status, **self.repo_metric_labels,
                     ).inc()
 
-                if i + 1 == launcher.retries:
-                    # last attempt failed, let it raise
+                if i + 1 == launcher.retries or (isinstance(e, HTTPError) and e.status_code == 409):
+                    # last attempt failed or 409 client error, let it raise
                     raise
 
                 # not the last attempt, try again
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -64,10 +64,6 @@ class Launcher(LoggingConfigurable):
             try:
                 return await AsyncHTTPClient().fetch(req)
             except HTTPError as e:
-                # swallow 409 errors on retry only (not first attempt)
-                if i > 1 and e.code == 409 and e.response:
-                    self.log.warning("Treating 409 conflict on retry as success")
-                    return e.response
                 # retry requests that fail with error codes greater than 500
                 # because they are likely intermittent issues in the cluster
                 # e.g. 502,504 due to ingress issues or Hub relocating,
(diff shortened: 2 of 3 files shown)


betatim:
For a historical perspective it would be good to hear from @yuvipanda

betatim:
I think the default path is (this is pretty tricky code and I've gotten it wrong before :-/) to go here https://github.com/jupyterhub/jupyterhub/blob/e89836c035f79a44cb5ebc1126e53c6f605464c1/jupyterhub/handlers/base.py#L894-L902 when spawning. The `slow_spawn_timeout` is (if I remember correctly) more about giving the spawner "a second or two" to become fully ready so users can go straight to their notebook server instead of first being redirected to a "waiting for your spawner" page to then be redirected to their notebook server. So maybe a better name would be something like `extra_patience_for_fast_spawners`.

A default for the `consecutive_failure_limit` tuned more to lower traffic hubs than mybinder.org is a good idea. If we change it we need to remember to update the mybinder.org config so that when this is deployed the value doesn't change.

While I read and think: when do spawns fail for reasons that are "interesting" to the user? Also what does a 409 represent?

bitnik:
> A default for the consecutive_failure_limit tuned more to lower traffic hubs than mybinder.org is a good idea. If we change it we need to remember to update the mybinder.org config so that when this is deployed the value doesn't change.

Now I think setting `consecutive_failure_limit` 0 was bad idea. I should revert it. And it should be documented, so each deployment can configure it for their need. What do you think? 

> While I read and think: when do spawns fail for reasons that are "interesting" to the user? Also what does a 409 represent?

For example (a future case), when user has custom resource request more than limit, spawner can fail and send informative error message as "Requested amount of resource is not allowed, limit is ...". I think when https://github.com/jupyterhub/binderhub/issues/712 (https://github.com/jupyterhub/binderhub/issues/712#issuecomment-455536850) is implemented, different BinderHub deployments would need this for different reasons, e.g., I need this for https://github.com/jupyterhub/binderhub/issues/794 to limit number of projects per user and inform users if they try to have more projects than allowed.

minrk:
consecutive_failure_limit was added specifically to solve issues faced on mybinder.org (a self-diagnosis of a truly unhealthy Hub), so I agree that setting it to 0 would not be a good plan.

I think we also do indeed want slow_spawn_timeout to be 0 for BinderHub (and I think will probably go away as a Hub option in the future, always using the 0 behavior). This ensures prompt replies. Setting it to anything else gives quite nondeterministic behavior because failures could happen while we're waiting, in which case we get the error, or not, in which case we don't.

What I think we want to do us use the JupyterHub progress API (which is in part inspired by BinderHub), which should let us hook up an event stream and relay those messages up through the BinderHub pipe. That ought to get us errors that we want.

bitnik:
Closing this PR in favor of https://github.com/jupyterhub/binderhub/pull/950

PR #820, opened 2019-04-08
[MRG] Generate coverage while running tests

betatim:
I found this inconsistency between how we run our tests while investigating why codecov doesn't show me which lines have coverage and which don't.

I was expecting https://codecov.io/gh/jupyterhub/binderhub/src/master/binderhub/build.py to show me which lines are covered. We seem to be computing some kind of coverage number because for the whole file it knows, but for some reason it doesn't show which lines are missing. If you look at code from (say) repo2docker https://codecov.io/gh/jupyter/repo2docker/src/master/repo2docker/app.py you see the green blobs in front of lines that are covered as well as background shading.

### DIFF
--- a/ci/test-helm
+++ b/ci/test-helm
@@ -81,5 +81,5 @@ mark="remote"
 if [[ -z "$GITHUB_ACCESS_TOKEN" ]]; then
   mark="$mark and not github_api"
 fi
-pytest -vsx -m "$mark"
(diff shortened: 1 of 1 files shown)


betatim:
At least inn the codecov pages linked from this PR it seems like we get the big green dots in front of each line back.

PR #822, opened 2019-04-09
bump jupyterhub chart c83896f...03215dd

minrk:
gets jupyterhub 1.0b2

it's been a while since we bumped this chart

https://github.com/jupyterhub/zero-to-jupyterhub/compare/c83896f...03215dd


betatim:
Let's merge this on Monday (or so) and then deploy it straight away to see what happens?

minrk:
@betatim good plan. I'm going to merge & deploy this one now, so we've got a whole EU day to keep an eye on it.

PR #824, opened 2019-04-14
[MRG] Update the contributer guide

betatim:
This changes the contributer documentation around a bit. I was coming back to the project after a bit away wanting to use the contributor guide to quickly figure out how to run tests and got lost.

I think this way of organising them also makes it less intimidating to get started because the first thing isn't "now you need to install minikube and helm". A lot of work can be done without ever touching all that stuff.

It also gets less and less verbose as we progress down the rabbit hole. The assumption is that if you aren't scared of installing minikube and helm you have worked out how to clone a git repository (or can do so). This fixes a pet peeve of mine that a lot of guides are so verbose that you can't find anything if you are using them as a reference. Maybe the solution is to have a guide/howto as part of the documentation and a reference in `CONTRIBUTING.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,18 +1,140 @@
-# Contributing
+# Contributing to BinderHub
 
-Welcome! As a [Jupyter](https://jupyter.org) project, we follow the [Jupyter contributor guide](https://jupyter.readthedocs.io/en/latest/contributor/content-contributor.html).
+Welcome! As a [Jupyter](https://jupyter.org) project, we follow the
+[Jupyter contributor guide](https://jupyter.readthedocs.io/en/latest/contributor/content-contributor.html).
 
-To develop binderhub, you can use a local installation of JupyterHub on minikube,
-and run binderhub on the host system.  Note that you will quickly hit your API limit
-on GitHub if you don't have a token.
+There are several different setups for developing BinderHub, depending on which
+parts of it you want to change: the [documentation](#documentation-changes),
+the [user interface](#user-innterface-changes), or the
+[kubernetes integration](#Kubernetes-integration-changes).
 
-## Installation
+
+## Documentation changes
+
+Work on the documentation requires the least amount of setup.  You will need
+to have a modern version of [Python](https://www.python.org/).
+
+1. Clone the BinderHub repository to your local computer and ```cd``` into it.
+   ```bash
+   git clone https://github.com/jupyterhub/binderhub
+   cd binderhub
+   ```
+1. 1. Install BinderHub and the documentation tools:
+
+   ```bash
+   python3 -m pip install -r doc/doc-requirements.txt
+   ```
+
+1. The documentation is located in the `doc/` sub-directory, `cd` into it:
+
+   ```bash
+   cd ./doc
+   ```
+
+1. To build the documentation run:
+
+   ```bash
+   make html
+   ```
+
+1. Open the main documentation page in your browser, it is located at
+   `_build/html/index.html`. On a Mac you can open it directly from the
+   terminal with `open _build/html/index.html`.
+
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,18 +1,149 @@
...
-and run binderhub on the host system.  Note that you will quickly hit your API limit
-on GitHub if you don't have a token.
+There are several different setups for developing BinderHub, depending on which
+parts of it you want to change: the [documentation](#documentation-changes),
+the [user interface](#user-innterface-changes), or the
+[kubernetes integration](#Kubernetes-integration-changes).
 
-## Installation
+
+## Documentation changes
```
sgibson91: In the docs section, I'd mention that you use RST and link to a tutorial if people aren't familiar

CONTRIBUTING.md
```diff
@@ -1,18 +1,149 @@
...
+output and take a few minutes to run. The tests will attempt to clean up after
+themselves on your minikube cluster.
+
+
+### One-time installation
+
+To run the full BinderHub it needs to have access to a kubernetes cluster
+with a JupyterHub installed on it. This is what we will setup in this section.
 
 Before you begin, there are a few utilities that need to be installed:
```
sgibson91: May be worth mentioning that `apt` is Linux based and provide alternatives for Mac users?

CONTRIBUTING.md
```diff
@@ -25,131 +156,92 @@ sudo apt install socat
...
    ```
 
-1. [Install Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) to run Kubernetes locally.
+1. [Install Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/)
+   to run Kubernetes locally.
 
-   For MacOS, you may find installing from https://github.com/kubernetes/minikube/releases may be
-   more stable than using Homebrew.
+   For MacOS, you may find installing from https://github.com/kubernetes/minikube/releases
+   may be more stable than using Homebrew.
```
sgibson91: I'm not certain that this statement is true anymore. Certainly when I first tried the contributing guidelines I had a lot of minikube-related issues that were fixed by installing via Homebrew.

CONTRIBUTING.md
betatim: Let me know what you think of the sentence I added above. When we originally wrote the guide we discussed about using conda or system python, OSX commands etc and the conclusion then was that we would give the Ubuntu commands that are as simple as possible and assume that if you dislike them or want to deviate because you have opinions you'd be able to do so.

For example virtualenvs/conda envs don't feature but are a good idea IMHO. However writing instructions for that opinionated stack (and the not opinionated version) would be a lot of work.

What do you think of this approach and an we make a nice one sentence version of all of the above to explain this to people?

CONTRIBUTING.md
betatim: Deleted.

betatim:
Thanks for the ideas. I updated the doc based on your comments. They all got implemented. There is one more thing to discuss (or punt to a later issue): https://github.com/jupyterhub/binderhub/pull/824#discussion_r276079020

CONTRIBUTING.md
sgibson91: Looks great! I think the clarification that `apt` is Ubuntu based is enough. In my naivety, I was very confused that I couldn't execute the first steps until it was explained to me that the command wasn't _supposed_ to work for me :)

betatim: Merci.

betatim:
Ready to merge.

meeseeksmachine:

https://discourse.jupyter.org/t/should-jupyterhub-and-binderhub-come-in-more-flavors/832/2

PR #826, opened 2019-04-15
[MRG] No more beta for BinderHub

betatim:
Is it time to remove the beta badge?

### DIFF
--- a/binderhub/templates/about.html
+++ b/binderhub/templates/about.html
@@ -4,7 +4,6 @@
 {% block main %}
 <div id="main" class="container">
   <div class="row">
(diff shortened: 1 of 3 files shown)


minrk:
Agreed. Do we want to make an "out of beta" announcement?

betatim:
> Agreed. Do we want to make an "out of beta" announcement?

Sounds like a good idea. Right now I'm thinking "let's write a super short blog post". Something like "no more beta because this has been running stable for a while now, seeing others deploy and operate BinderHubs as well (point to pangeo, gesis and Turing Way). Thanks for all the help and support!"

PR #835, opened 2019-04-29
Update to Docs: Clarify where in secret.yaml GitHub Personal Access Token should be added

sgibson91:
Hi,

This is a _tiny_ PR to update the docs surrounding adding the Personal Access Token for the GitHub API limit to `secret.yaml`. It wasn't entirely clear that it needs to go under `jupyterhub` and I got caught out this morning.

Thanks!

### DIFF
--- a/doc/setup-binderhub.rst
+++ b/doc/setup-binderhub.rst
@@ -306,9 +306,10 @@ an API access token to raise your API limit to 5000 requests an hour.
 
 3. Update ``secret.yaml`` by entering the following::
 
-    config:
-      GitHubRepoProvider:
-        access_token: <insert_token_value_here>
+    jupyterhub:
+      config:
+        GitHubRepoProvider:
+          access_token: <insert_token_value_here>
 
 This value will be loaded into `GITHUB_ACCESS_TOKEN` environment variable and
 BinderHub will automatically use the token stored in this variable when making


sgibson91:
I'm now questioning whether this is correct. My Hub is becoming very problematic when I have this in my `secret.yaml` 🤔Slow to start after periods of inactivity and often not spinning up user pods. Hub becomes much more responsive when I remove the access token.

sgibson91:
I think I've managed to convinced myself that this is incorrect. Going to revert the PR.

betatim:
It seems weird that the hub would become unresponsive with this change. Even if it is "the wrong" way to configure it. We should investigate that.

Whether or not you need the extra level of "indentation" depends on whether you setup your own meta chart (like we do for https://github.com/jupyterhub/mybinder.org-deploy/) or if you just use the binderhub chart.

sgibson91:
Some kind of check for indentations during `helm upgrade` would be nice. So many of my problems would have been fixed if I had a message like "this key isn't where I expected it to be so I won't implement it this time".

sgibson91:
"P.S. This may cause weird Hub behaviour"!

consideRatio:
@sgibson91 I've been thinking about these pain points as well without coming up with a concrete idea on how to resolve the pain in a sustainable manner =/

betatim:
Can we do something with helm linting or specifying a flag to the `helm` executable to say "please shout if there are things defined in `values.yaml` that aren't consumed in the chart(s)? I think this is super tricky to do but you never know :-/

sgibson91:
Yeah, other than checking the parents of each key I'm not sure of a good way either 🤔

PR #837, opened 2019-04-29
Removing note about breaking changes

choldgraf:
Since we've now advertised BinderHub as stable + ready for production, shall we remove this message about breaking changes etc?


betatim:
Abso-merge-a-lutely!

PR #838, opened 2019-05-01
DinD documentation

### LINKED ISSUE #832: Better document how builds are done and when dind should be used
I ran into #761 a few weeks ago and got stuck there until @minrk suggest configuring it to use dind.
This is not mentioned in the documentation, so perhaps we should mention it there 😄 .

jhamman:
closes #832


consideRatio:
Wow thanks @jhamman ! That made it a lot clearer to me. I deployed a BinderHub for neurips once and that is all my BinderHub deployment experience, but at that time I didn't understand this but now I grasp this thanks to your documentation!

:heart: :tada: LGTM

betatim:
Question for Yuvi and/or Min: are there also security benefits for running builds via docker-in-docker? In my mind one reason to run in DIND mode is that you have better control over what the `docker build...` command executed by repo2docker has access to in terms of files and resources. But it is a fuzzy concept in my head so I can't write a sentence to include in the docs :(

consideRatio:
@betatim I find this very challenging, and while working with gitlab's helm chart I've considered related security concerns about Docker overall to build and run things for users in CI pipelines.

Here are some of these concerns written down by GitLab's team:
https://docs.gitlab.com/runner/executors/kubernetes.html#using-docker-in-your-builds

jhamman:
Bumping this so we don't loose too much momentum. @betatim / @consideRatio - should we add some comments about DIND and security? I would need input from you on what to write.

consideRatio:
@jhamman I'm not knowledgeable enough about these aspects yet to suggest something =/

doc/setup-binderhub.rst
```diff
@@ -378,3 +383,32 @@ Then update ``secret.yaml`` with the following::
...
+By default, BinderHub will build pods with the host Docker installation.
+This often means you are stuck with whatever version of Docker provided by your
+cloud provider. BinderHub supports an alternative that uses `Docker-in-Docker
+(DinD) <https://hub.docker.com/_/docker>`_. To turn `dind` on, you'll need to set
+the following configuration in your ``config.yaml`` file::
+
+    dind:
+      enabled: true
+      daemonset:
+      image:
```
betatim: this and the next two lines need to be indented. Though they are the same as the default so we probably don't need them at all?

doc/setup-binderhub.rst
```diff
@@ -287,8 +287,15 @@ You now have a functioning BinderHub at the above IP address.
 
 .. _api-limit:
 
+Customizing your Deployment
+---------------------------
+
+The Helm chart used to install your BinderHub deployemnt has a lot of options
+for you to tweak. Below is a few pointers for how to configure some of the most
```
betatim:
```suggestion
for you to tweak. Below a few pointers for how to configure some of the most
```

betatim:
When PRs go stale feel free to bump them by following up on the questions/discussion points by suggesting a resolution that seems to follow the consensus. I think a frequent reason for why things go stale is that there is no immediate actionable thing and no one is trying to get to one. I think it is best if the PR proposer tries to do the driving by suggesting a next actionable item.

For now I would skip the security question and maybe @jhamman can create a new PR addressing/adding thoughts on that. That way we can get this PR merged.

doc/setup-binderhub.rst
jhamman: I've fixed the indentation. I think it is helpful to show the default the image specification here though as this is the sort of thing someone may want to customize.

jhamman:
@betatim - thanks for the review. I've updated things a bit and this should be good to go in.

> should we add some comments about DIND and security

I don't know. I'm not familiar with this issue so I'll leave this for the next contributor.

doc/setup-binderhub.rst
```diff
@@ -419,3 +424,32 @@ Then update ``secret.yaml`` with the following::
...
+By default, BinderHub will build pods with the host Docker installation.
+This often means you are stuck with whatever version of Docker provided by your
+cloud provider. BinderHub supports an alternative that uses `Docker-in-Docker
+(DinD) <https://hub.docker.com/_/docker>`_. To turn `dind` on, you'll need to set
+the following configuration in your ``config.yaml`` file::
+
+    dind:
+      enabled: true
+      daemonset:
+        image:
```
betatim:
```suggestion
        image:
          # this lets you set the docker version
```

PR #853, opened 2019-05-23
add default toleration to build pods

### LINKED ISSUE #851: [Feature proposal] Add toleration "hub.jupyter.org/dedicated=user:NoSchedule" to build pods
I'd like to propose adding a default toleration to binderhub build pods. In most cases, build pods are intended to be scheduled on the same nodes as user notebook sessions so I think the easiest path forward would be to match the jupyterhub taint/toleration for users: 

```
hub.jupyter.org/dedicated=user:NoSchedule
```

My motivation is that I'd like to restrict scheduling on my user node-pool to only _build_ and _notebook_ pods.

Adding a toleration does not require a matching taint be set, rather it allows the cluster manager to set taints that certain pods can tolerate. For this reason, I don't immediately see any downside to this feature being added. That said, I'm still learning in this space so I'm open to redirection.

jhamman:
This implements a minimally invasive addition to the build pod spec, adding a default tolerations to build pods that match [zero2jupyterhub](https://zero-to-jupyterhub.readthedocs.io/en/latest/optimization.html?highlight=taint#using-a-dedicated-node-pool-for-users).

```
hub.jupyter.org/dedicated=user:NoSchedule
```

cc @consideRatio 

closes #851

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -192,6 +192,22 @@ class Build:
                         env=env
                     )
                 ],
+                tolerations=[
+                    client.V1Toleration(
+                        key='hub.jupyter.org/dedicated',
+                        operator='Equal',
+                        value='user',
+                        effect='NoSchedule',
+                    ),
+                    # GKE currently does not permit creating taints on a node pool
+                    # with a `/` in the key field
+                    client.V1Toleration(
+                        key='hub.jupyter.org_dedicated',
+                        operator='Equal',
+                        value='user',
+                        effect='NoSchedule',
+                    ),
+                ],
                 node_selector=self.node_selector,
                 volumes=volumes,
                 restart_policy="Never",


betatim:
LGTM and tests seem happy. Merging now, if this turns out to be controversial we can always revert it.

jhamman:
Thanks @betatim - I'll push this to Pangeo's binder and report any problems.

consideRatio:
Is a build pods task to build a single image? Then i think this is really good!

consideRatio:
Hmmmm, one may consider using the custom user scheduler as well then btw, if not, they will stop scale downs unless there is 0 build pods for ten full minutes.

jhamman:
> Is a build pods task to build a single image? Then i think this is really good!

Yes, the build pod builds a single image. In my view, this is ideally done on the same node the notebook will be run, as it provides a higher likelihood that we can avoid a pull of the recently built image. I think there is still an element of chance here so some additional scheduling could help. 

> one may consider using the custom user scheduler as well then btw

This sounds like a great idea.

PR #856, opened 2019-05-23
add user tolerations to dind daemonset

jhamman:
this is a followup to #853. We need to allow the dind daemonset to run on user pods as well.


jhamman:
Tests passed now so I think this can be merged.

consideRatio:
Due to #853 is already done, I consider this only a technical fix, so I'll go ahead and merge it.

PR #858, opened 2019-05-25
[MRG] Tweak loading messages

betatim:
Instead of always saying "30s or longer" I think it is better to have a "starting" and "taking longer than usual" message. The change here is to show the "longer than usual" message earlier. I chose the cutoff at 17s because the [median launch time](https://grafana.mybinder.org/d/3SpLQinmk/1-overview?refresh=1m&panelId=28&fullscreen&orgId=1) is around 15s. 90% of launches complete in under a minute and 99% in around 4minutes. Maybe we can use these stats to build a "progress bar" for launches.

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -130,10 +130,10 @@ function build(providerSpec, log, path, pathType) {
 
   // Update the text of the loading page if it exists
   if ($('div#loader-text').length > 0) {
-    $('div#loader-text p').text("Loading repository (can take 30s or more to load): " + spec);
+    $('div#loader-text p').text("Starting repository: " + spec);
     window.setTimeout(function() {
-      $('div#loader-text p').html("Repository " + spec + " is taking longer than usual to load, hang tight!")
-    }, 120000);
+      $('div#loader-text p').html("Repository " + spec + " is taking longer than usual to start, hang tight!")
+    }, 17000);
   }
 
   $('#build-progress .progress-bar').addClass('hidden');


choldgraf:
I think this is a good idea and I like the language. What do you think about also adding a link to the Binder FAQ about "what affects the launch time" section: https://mybinder.readthedocs.io/en/latest/faq.html#what-factors-influence-how-long-it-takes-a-binder-session-to-start

Something like appending this at the end:

"Curious what affects the length of launches? Check out the [Binder documentation](https://mybinder.readthedocs.io/en/latest/faq.html#what-factors-influence-how-long-it-takes-a-binder-session-to-start)." 

Could also give something to click and read while they wait.

betatim:
I'd advocate iterating because I think the page is already far too busy with links, unfoldable content (the build log) and potentially a preview (or a 404 because preview doesn't work), etc.

My current suggestion would be to replace the "New to Binder ..." line with a more specific one like you suggested "Curious how to make this faster? ...click here..." when the launch takes more than 20seconds or so:

Screen Shot 2019-05-28 at 18 13 54

But I'd do that in a second PR because I don't know how to do it off the top of my head :)

consideRatio:
I think also we should not have a mentioning of a specific number of seconds, but I think we could perhaps do something like kubespawners progressbar listening on events, but instead we are a bit time dependent.

One thing we could do is utilize the size of the image perhaps.

betatim:
Shall we merge this and then see how we can improve? I think this PR gets us to a state that is better than what we currently have.

PR #859, opened 2019-05-25
[MRG] Make the URL used to generate launch badges configurable

betatim:
This lets you set the base URL used for generating badges. This is useful when you have a group of hubs at different hostnames but want badges generated on all of them to point to a common hostname. I think we should use this for operating the GKE and OVH cluster as a federation so that all clusters generate https://mybinder.org/ badges and launch URLs.

### DIFF
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -15,6 +15,7 @@ class MainHandler(BaseHandler):
     def get(self):
         self.render_template(
             "index.html",
+            badge_base_url=self.settings['badge_base_url'],
             base_url=self.settings['base_url'],
             submit=False,
             google_analytics_code=self.settings['google_analytics_code'],
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -123,12 +123,23 @@ class BinderHub(Application):
         '/',
         help="The base URL of the entire application",
         config=True)
-
     @validate('base_url')
     def _valid_base_url(self, proposal):
         if not proposal.value.startswith('/'):
             proposal.value = '/' + proposal.value
         if not proposal.value.endswith('/'):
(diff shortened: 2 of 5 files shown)


binderhub/main.py
```diff
@@ -15,6 +15,7 @@ class MainHandler(BaseHandler):
     def get(self):
         self.render_template(
             "index.html",
+            badge_base_url=self.settings['badge_base_url'],
```
bitnik: `badge_base_url=self.settings['badge_base_url'] or f"{self.request.protocol}://{self.request.host}{self.settings['base_url']}",` -> maybe this is better. so you can remove `if else`s in js code and it becomes cleaner.

betatim: I am not sure which one I prefer. I agree the JS has a few extra if statements, but I think I prefer it that way because it leaves the choice to the JS code. I dunno, it is just a feeling :-/

choldgraf:
This looks great to me - the one thing that's missing is any documentation about this feature 📚 . I guess it would show up here: https://binderhub.readthedocs.io/en/latest/reference/app.html maybe that's enough?

betatim:
Nods that is where you'd find an entry.

We could write a short how to section on setting up a federation? I think that is the only context in which you'd want to use this feature.

choldgraf:
@betatim just to double check are you planning on putting off the docs for this for another PR? (in which case I can open an issue about it so we don't forget)

betatim:
It should appear automatically in the reference section of the docs. Not sure any other documentation needs writing.

I'd make writing "howto setup a federation" a new issue as that will require a bit of thinking (for example what redirector should people use).

choldgraf:
then thanks for the contribution @betatim !

PR #863, opened 2019-05-28
Adding a simple age checker for a BinderHub

choldgraf:
This adds a page in the BinderHub documentation with a mini form that does the following (checked boxes means it is implemented in this PR, unchecked means I'd like to implement it, but not sure how to)

- [x] Takes as input the URL of a BinderHub
- [x] Press a button to grab the latest versions of software on the BinderHub
- [x] Display the versions w/ success or fail note
- [ ] List the number of **commits** of these versions relative to the latest published versions
- [ ] List the number of **days** of these versions relative to the latest published versions 

I've got the versions reading in properly (using `mybinderhub.org/versions`) but I can't figure out how to compare these versions to the "latest" versions (which I assume we should pull from here: https://jupyterhub.github.io/helm-chart/). Anybody have ideas for how this could happen?

Here's what the age checker looks like: https://302-89419368-gh.circle-artifacts.com/0/html/age-checker.html

### DIFF
--- a/doc/age-checker.rst
+++ b/doc/age-checker.rst
@@ -51,6 +51,37 @@ provides a quick interface to check how up-to-date each of them is.
    <div class="version_text" id="repo2dockerversiontext">repo2docker version: <span class="empty">Fill in the box then click "check age"</span></div>
 
    <script>
+   // commit data
+   var response_data = {};
+   var bhub_resp = $.get("https://api.github.com/repos/jupyterhub/binderhub/commits", (data) => {
+     response_data['bhub_json'] = data;
+     console.log("got binderhub commit data");
+   });
+   var r2d_resp = $.getJSON("https://api.github.com/repos/jupyter/repo2docker/commits", (data) => {
+     response_data['r2d_json'] = data;
+     console.log("got repo2docker commit data");
+   });
+   
+   function findCommitFromSHA(commits, sha) {
+    for (ii=0; ii < commits.length; ii++) {
+      var this_commit = commits[ii]
+      var this_sha = this_commit['sha'];
+      search = this_sha.indexOf(sha.substr(1));  // A hack because there seems to be a bug w/ the SHA
+      if (search != -1) {
+          var chosen_commit = this_commit
+          return chosen_commit
+      } 
+    }
+   };
+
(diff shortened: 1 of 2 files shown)


betatim:
https://mybinder.org/versions shows the version as reported by versionneer for BinderHub itself and the value set in the helm chart for repo2docker:

https://github.com/jupyterhub/binderhub/blob/5544f704cb0dd9b8d3d67442a0c744e1ddcb6089/binderhub/base.py#L111-L120

This means that the repo2docker one should be directly comparable to what is in the helm chart. The BinderHub one is a bit more tricky. It consist of the last released version (0.1.0), how many commits were made since then (218), the revision checked out (g5544f70), and indicates that the repo has been modified since that commit (dirty): "0.1.0+218.g5544f70.dirty"

choldgraf:
Was playing around with this again. I'm a bit closer, but now I'm running into CORS requests errors when trying to access `versions/` of some BinderHubs. For some reason this doesn't happen with the mybinder.org binderhub, but it does with the pangeo binderhub. Any idea how get around this? :-/

betatim:
The Pangeo BinderHub doesn't allow cross origin requests. For mybinder.org we do allow them. More details on that https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

choldgraf:
ahh I see - so approach in this PR (get versions from `yourbinderhub.org/versions`) is going to be fundamentally limited by whether those binderhubs allow this kind of traffic. Perhaps this is not the right place for this kind of information to live.

betatim:
We could change the versions handler so it always allows people to access (only) this endpoint. Or you have to implement the fetching of versions in the server of the "age display" instead of fetching it browser side.

choldgraf:
it seems reasonable to me that `versions/` should be accessible from anywhere (a main point is to provide machine-readable access)...maybe something that's "on by default" but that people can turn off if they really want to ?

betatim:
Closing this as we haven't worked on it for a while and there is `/versions` now which thanks to the version string updates we've made tells you how many commits since the last release have passed as well as the git SHA in use.

If you want to keep working on this reopen it :)

PR #865, opened 2019-06-02
[MRG] Make base badge URL available in the loading screen as well

betatim:
We also need to add the base URL data to the loading page. This was missed in #859, noticed after deploying on mybinder.org


betatim:
Self-merge as this is a fix to a bug from a previous PR and should be uncontroversial.

PR #868, opened 2019-06-06
adding federation page

choldgraf:
This adds a page to list the federated BinderHubs that are powering mybinder.org, with the idea that it can be expanded over time. It should be a way for us to highlight how Binder is a network of deployments rather than a single one, as well as a way for others to highlight their contributions to the BinderHub network.

Eventually we may want this page in a different place than the binderhub docs, but this is a start to iterate on!

Here's how it looks: https://314-89419368-gh.circle-artifacts.com/0/html/federation/federation.html


doc/script/gen_federation.py
```diff
@@ -0,0 +1,51 @@
...
+import pandas as pd
+import os
+import os.path as op
+from ruamel import yaml
+
+# Variables
+N_PER_ROW = 4
+
+# Code to generate the HTML grid
+template_binderhub = '<td align="center" class="contrib_entry"><a href="{URL_BINDERHUB}"><img src="{LOGO}" class="fed_logo" alt="{RUN_BY}" /></a><br /><p class="name">{BINDERHUB_SUBDOMAIN}</p><br /><p class="name">Run by</p><p class="name"><b><a href={URL_INFO}>{RUN_BY}</a></b></p><br /><p class="name">Funded by</p><p class="name"><b>{FUNDED_BY}</b></p>'
```
betatim: We should make the "funded by" a link to the website of those that pay.

doc/index.rst
```diff
@@ -70,6 +70,7 @@ in the community have done.
    customizing
    authentication
    known-deployments
+   federation/federation
```
betatim: should lines 70, 71, 72, 73 have different indentations?

choldgraf: do they? I can't see a difference

choldgraf:
finally getting back to this after a week of solid straight meetings :-)

thanks for the comments @betatim!

This push adds more content to the page after some conversations I had with folks who had more questions about this. It also updates the links in the federation list so that they go to different "run by" and "funded by" pages!

doc/index.rst
betatim: I could swear they were indented by an extra character but not any more.

doc/federation/federation.rst
```diff
@@ -0,0 +1,87 @@
...
+===================================
+
+Here is a list of the current members of the BinderHub federation:
+
+.. include:: data-federation.txt
+
+
+.. _federation/joining:
+
+Joining the BinderHub Federation
```
betatim: This title and the next one are "the same", at least they are too similar to make it clear what the difference between them is. Maybe we could call this a FAQ section or "Points to consider when thinking about joining" (needs a shorter version).

doc/federation/federation.rst
```diff
@@ -0,0 +1,87 @@
...
+================================
+
+The Binder Project is an open community of developers as well as deployers
+of BinderHub. BinderHub is designed to be used by any organization for
+their own purposes. However, many members of the community also run
+the BinderHub deployments at ``mybinder.org`` as a large, free public service
+to the community.
```
betatim: I'd remove this paragraph and instead put some of the info in the next one which I think should be the lead one as it answers the question "what is this federation you speak of?"

doc/federation/federation.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Behind ``mybinder.org`` is a **federation of BinderHubs** - these are
```
betatim: How about the following:

Behind mybinder.org is a federation of BInderHubs. This means that there are several independent hubs that each serve a fraction of the traffic created by people clicking links pointing to mybinder.org. Anyone (a company, university or individual) is welcome to deploy a BinderHub that forms part of the federation.'

Adding a new hub to the federation requires a mix of two kinds of resources: compute and human power to operate the hub. The two extremes of this mixture are:
* you donate compute power that the mybinder.org team has full control over which means you don't have to be involved in day to day operations, or
* you donate compute power over which the mybinder.org team does not have full control which means you are also responsible for day to day operations.

doc/federation/federation.rst
```diff
@@ -0,0 +1,87 @@
...
+   any member of the BinderHub federation will be committed to keeping their
+   BinderHub running with a reasonable uptime, but we don't have any legal
+   framework to enforce this. Use your best judgment when deciding if you'd
+   like to join the BinderHub federation - if you can confidently say your
+   BinderHub will be up the large majority of the time, then that's fine.
+3. **What kind of cloud resources would I need?** This depends on how many
+   you have :-)  We can increase or decrease the percentage of ``mybinder.org``
+   traffic that goes to your BinderHub based on what you can handle. If you
+   can only handle a few dozen user pods (remember users get ~ 2GB of RAM and
+   1 CPU) then we can make sure you stay under that amount.
```
betatim: We can't actually make sure we stay below this so I''d remove the sentence leaving the sentence before which mentions that you will get traffic proportional to what you can handle.

doc/federation/federation.rst
```diff
@@ -0,0 +1,87 @@
...
+The BinderHub Federation
+========================
+
+While it may seem like ``mybinder.org`` is a single website, it is
+in fact a **federation** of teams that deploy public BinderHubs to serve
+the community. This page lists the BinderHubs that currently help
+power ``mybinder.org``.
+
+Visiting ``mybinder.org`` will randomly redirect you to one
+of the following BinderHubs. You may also directly access a BinderHub
```
betatim: I'd remove the sentence encouraging people to directly link to a particular cluster. I treat the fact that you can as an implementation detail people should not rely on but can use if they know what they are doing. A bit like Python classes let me poke around the insides of instances if I want to but it is clear that I am looking at the internals and that things might change without notice.

choldgraf:
sorry for dropping off on this one, I got sucked into a black hole of talks and tutorials :-)

I think I addressed each of @betatim's comments - lemme know what you think, or if other changes should be made!

PR #872, opened 2019-06-12
Add 'origin' field to launch events

yuvipanda:
Lets us differentiate launches from different BinderHub
installations

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -413,6 +413,12 @@ class BinderHub(Application):
         config=True,
     )
 
+    normalized_origin = Unicode(
+        '',
+        config=True,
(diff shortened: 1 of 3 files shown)


betatim:
Looks good.

Do you think we need some normalising capabilities or will need them? For example the OVH cluster can be reached via multiple hostnames, should we let the operator of the hub configure how to normalise those (into one hostname?)? I'm tending towards "no".

yuvipanda:
@betatim hmm, that's a good point actually. I think it might be a good idea to allow that - especially if we want to not expose the hostnames to end users. That'll futz with `self.request.host`.

yuvipanda:
@betatim updated

PR #876, opened 2019-06-13
document launch schema version 3

bitnik:
follow-up https://github.com/jupyterhub/binderhub/pull/872


betatim:
Thank you for catching up the docs.

PR #878, opened 2019-06-13
[MRG] Add instructions for setting up an Azure Container Registry

### LINKED ISSUE #869: How-to guide for setting up a private container repository
Hi!

I was wondering, is a step-by-step guide for deploying and connecting a private docker container repository (from Google or Azure, etc.) within the scope of the BinderHub docs? Or should we just include links to tutorials? (Happy to write the Azure docs if yes.)

sgibson91:
### Summary

This PR will add documentation around setting up and connecting an Azure Container Registry (ACR) to a BinderHub.

fixes #869 

### Changes Proposed in this PR

* `doc/setup-registry.rst` - Azure command line instructions for creating an ACR, assigning the correct permissions and collecting relevant information have been added
* `doc/setup-binderhub.rst` - templates for `secret.yaml` and `config.yaml` have been added to parse the information to BinderHub
* An "Important" note box is added to the custom registry section of `doc/setup-binderhub.rst` to point out the hard dependency on the `gcr.io` image name structure.

### DIFF
--- a/doc/setup-binderhub.rst
+++ b/doc/setup-binderhub.rst
@@ -188,7 +188,7 @@ where:
 * `<ACR_NAME>` is the name you gave to your ACR,
 * `<SERVICE_PRINCIPAL_ID>` is the AppID of the Service Principal with AcrPush role assignment,
 * `<SERVICE_PRINCIPAL_PASSWORD>` is the password for the Service Principal,
-* `<project-name>` is an arbitrary name that is required due to BinderHub's hard dependency on the `gcr.io` image structure above.
+* `<project-name>` is an arbitrary name that is required due to BinderHub assuming that the image_prefix contains an extra level for the project name.
   See `this issue <https://github.com/jupyterhub/binderhub/issues/800>`_ for futher discussion.
   If this is not provided, you may find BinderHub rebuilds images every launch instead of pulling them from the ACR.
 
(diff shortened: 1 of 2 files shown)


doc/setup-binderhub.rst
```diff
@@ -148,6 +163,25 @@ Update ``config.yaml`` by entering the following::
...
+      BinderHub:
+        use_registry: true
+        image_prefix: <ACR_NAME>.azurecr.io/<project-name>/<prefix>-
+      DockerRegistry:
+        token_url: "https://<ACR_NAME>.azurecr.io/oauth2/token?service=<ACR_NAME>.azurecr.io"
+
+where:
+
+* `<ACR_NAME>` is the name you gave to your ACR,
+* `<project-name>` is an arbitrary name that is required due to BinderHub's hard dependency on the `gcr.io` image structure above.
```
betatim: Can we replace "hard dependency" with "assumes that the `image_prefix` contains an extra level for the project name". We could fix this after all but so far haven't gotten round to it.

Do you think it would help people if we made a recommendation here as to what to use as a name? My recommendation would be to reuse `<ACR_NAME>`.

betatim:
Looks excellent to me. I left one comment about "hard dependencies".

I didn't try and follow the instructions (no AZR account) but read the commands and didn't spot any obvious typos.

doc/setup-binderhub.rst
sgibson91: Changing the language :+1:

It's tough to suggest a name as it has to be unique across _all_ of Azure. So I ended up with `testregistry1234sgibson91THISHADBETTERWORK` before my test registry was deployed into a unique namespace 🙂 This is why I suggested the `check-name` function in the setup registry docs.

betatim: AH, I didn't mean to suggest a actual name name for ACR_NAME. I was thinking we should say that "a good value to use for `<project-name>` is the same value as you use for your `<ACR_PROJECT>`".

sgibson91: Ah, I see! Yeah that would work. I've personally gone with the name I gave to the BinderHub, so if there's multiple Hubs attached, you can see what came from where.

PR #879, opened 2019-06-17
[MRG] Manually list BinderHub dependencies in docs requirements

betatim:
Our docs build broken when pycurl was added as a dependency. It has some non Python package requirements that aren't fulfilled on RTD. This PR copies the requirements to the docs requirements manually. I think we change our dependencies rarely enough that we can get away having to synnc them manually.

### DIFF
--- a/doc/doc-requirements.txt
+++ b/doc/doc-requirements.txt
@@ -4,7 +4,17 @@ recommonmark==0.4.0
 sphinx-copybutton
 alabaster_jupyterhub
 traitlets
-# install BinderHub from source but because `pip install` is executed in the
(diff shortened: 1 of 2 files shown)


doc/doc-requirements.txt
```diff
@@ -4,7 +4,17 @@ recommonmark==0.4.0
...
+escapism
+tornado
+traitlets
+docker
+jinja2
+prometheus_client
+python-json-logger
+jupyterhub
+jsonschema
+#pycurl is not installed for docs
```
willingc:
```suggestion
#pycurl Do not install for docs as it breaks the RTD build. Its primary use is for mocks in testing .
```

PR #881, opened 2019-06-17
[MRG] Add basic news header to the HTML templates

### LINKED ISSUE #877: Banner to display information/messages to users
Currently there is no good way to add a "sponsored by" banner to the pages served directly by BinderHub. One has to modify the HTML templates directly which is not ideal.

This issue is about adding a config option similar to https://github.com/jupyterhub/binderhub/blob/0df2fc82e2af42398ae1103a433695dde4dd6b0a/binderhub/app.py#L97-L106 which lets users display an additional message on the about page but is used to show users information about the hub they are on.

Uses for this could be either status messages, information from the admins or who is paying for the compute.

In my mind something like this banner at the top of the main page as well as the landing page:

Screen Shot 2019-06-13 at 10 49 12

where the user can configure the text and maybe the colour (or choose from a set of background and text colour combos to make sure contrast is high enough).

How or if to display similar information once the user is in their binder is something I'd punt to a second issue. Last time we discussed it there was no clear consensus on how to do this.

betatim:
This PR adds the ability to have a basic "news" banner at the top of all pages. You can set the HTML to be displayed inside the banner via a config option.

Closes #877

When the message is not an empty string a banner with grey background (like the form) will appear at the top of each page (landing, loading, about, error, etc).

Screenshot_2019-06-17 Binder

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -107,7 +107,7 @@ class BinderHub(Application):
         config=True
     )
 
-    news_message = Unicode(
+    banner_message = Unicode(
         '',
         help="""
         Message to display in a banner on all pages.
@@ -532,7 +532,7 @@ class BinderHub(Application):
             'google_analytics_code': self.google_analytics_code,
             'google_analytics_domain': self.google_analytics_domain,
             'about_message': self.about_message,
-            'news_message': self.news_message,
+            'banner_message': self.banner_message,
             'extra_footer_scripts': self.extra_footer_scripts,
             'jinja2_env': jinja_env,
             'build_memory_limit': self.build_memory_limit,
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -112,8 +112,9 @@ class BinderHub(Application):
         help="""
         Message to display in a banner on all pages.
 
-        Will be directly inserted into the about page's source so you can use
-        raw HTML.
+        The value will be inserted "as is" into a HTML <div> element
+        with grey background, located at the top of the BinderHub pages. Raw
+        HTML is supported.
         """,
         config=True
     )
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -107,6 +107,17 @@ class BinderHub(Application):
         config=True
     )
 
+    news_message = Unicode(
+        '',
+        help="""
+        Message to display in a banner on all pages.
+
+        Will be directly inserted into the about page's source so you can use
+        raw HTML.
+        """,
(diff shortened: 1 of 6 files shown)


binderhub/app.py
```diff
@@ -107,6 +107,17 @@ def _log_level(self):
         config=True
     )
 
+    news_message = Unicode(
```
choldgraf: Could we call this "banner message" instead of "news message"? I feel like that's a bit more descriptive and it follows terminology that lots of other sites use (I'm thinking here what a person would google for to learn how to do this, "How to add a banner to my BinderHub" sounds more likely to me than "How to add a news message to my BinderHub").

betatim: Good point! Done.

binderhub/static/index.css
```diff
@@ -50,6 +50,14 @@ p > a:hover {
 }
 
 
+#msg-container {
```
choldgraf:
```suggestion
#banner-container {
```

binderhub/templates/page.html
```diff
@@ -11,6 +11,12 @@
 <body>
   {% block body %}
 
+  {% if banner %}
+  <div id="msg-container">
```
choldgraf:
```suggestion
  <div id="banner-container">
```

betatim:
Renamed the ID used.

I was wondering about bolder/bigger font or colours and though that because the banner can't be dismissed keeping it somewhat "not in your face" would be good. Speculating that the fact that it normally isn't there is enough to make people notice it.

binderhub/app.py
```diff
@@ -107,6 +107,17 @@ def _log_level(self):
         config=True
     )
 
+    banner_message = Unicode(
+        '',
+        help="""
+        Message to display in a banner on all pages.
+
+        Will be directly inserted into the about page's source so you can use
+        raw HTML.
```
consideRatio: I suggest the following rephrasing as I got surprised about "directly inserted" in practice meant a container with grey background would wrap the content.

"The content of `banner_message` will be inserted as it is within a `<div>` with grey background, located at the top of the BinderHub pages. Raw HTML is supported."

betatim: Good point, also noticed the docstring was still referring to "about message" so fixed that as well.

betatim:
As two people gave a positive review I am merging it now instead of waiting till California wakes up. This way we can start using it for the user survey.

PR #883, opened 2019-06-26
adding whitelisted specs

### LINKED ISSUE #882: Add a repository quota white list
It would be helpful if we had the ability to give some repositories special status relative to others. It seems that a good start (and one that would be immediately useful) would be to set up a **white list** configuration for repositories.

I looked into a little bit and don't have time to work on it more today, but putting my thoughts here in case it helps point someone in the right direction (maybe @sgibson91 can also riff off of this for creating a "whitelist" of non-banned repositories too?)

## Potential implementation

My guess is that this would follow a similar pattern as the **banned_specs** configuration. E.g. here's how it looks in the configuration for mybinder.org

https://github.com/jupyterhub/mybinder.org-deploy/blob/master/mybinder/values.yaml#L46

It seems like this should be done at the RepoProvider level, although one question is where the quote whitelist would actually be checked. The banning check is done at the `get` event here: https://github.com/jupyterhub/binderhub/blob/908c4439f63b08744e2074e76dabb9883b1632e9/binderhub/builder.py#L233

While the quota checking is done under the `launch` event here: https://github.com/jupyterhub/binderhub/blob/908c4439f63b08744e2074e76dabb9883b1632e9/binderhub/builder.py#L462

The trick is that the `get` event has access to the repoprovider instance, while launch does not...so that's the main question I've got for implementing this.

## Other questions

* How would we test something like this? I maybe in the test infrastructure we can set the per_repo_quota to be really small (like, 1) and then try to launch a second test pod and make sure it works?
* More generally, is there a way to create async launch requests with python?

choldgraf:
This PR adds the ability to **whitelist** specifications with the GitHub repoprovider.

it closes #882 though doesn't quite follow the pattern there (which suggested a more general config architecture for specific specs).

Still not sure how to test it, but it seems to me that this would work for the whitelist usecase. @betatim what do you think? Happy to iterate on this a bit to try and get it working.

### DIFF
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -71,6 +71,15 @@ class RepoProvider(LoggingConfigurable):
         config=True
     )
 
+    whitelisted_specs = List(
+        help="""
+        List of specs to whitelist for quota limits.
+
+        Should be a list of regexes (not regex objects) that match specs which should be whitelisted
+        """,
+        config=True
+    )
+
     unresolved_ref = Unicode()
 
     git_credentials = Unicode(
@@ -88,6 +97,17 @@ class RepoProvider(LoggingConfigurable):
             # Ignore case, because most git providers do not
             # count DS-100/textbook as different from ds-100/textbook
             if re.match(banned, self.spec, re.IGNORECASE):
+                return True
+        return False
+
+    def is_whitelisted(self):
+        """
+        Return true if the given spec is whitelisted
+        """
+        for whitelisted in self.repository_whitelist:
+            # Ignore case, because most git providers do not
+            # count DS-100/textbook as different from ds-100/textbook
+            if re.match(whitelisted, self.spec, re.IGNORECASE):
                 return True
         return False
 
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -71,11 +71,11 @@ class RepoProvider(LoggingConfigurable):
         config=True
     )
 
-    whitelisted_specs = List(
+    high_quota_specs = List(
         help="""
-        List of specs to whitelist for quota limits.
+        List of specs to assign a higher quota limit.
 
-        Should be a list of regexes (not regex objects) that match specs which should be whitelisted
+        Should be a list of regexes (not regex objects) that match specs which should have a higher quota
         """,
         config=True
     )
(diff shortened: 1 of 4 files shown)


betatim:
I liked the suggestion in the issue :-/

An important use case for the quota is to protect people from accidentally launching too many instances and then "taking over" all the resources in the cluster. This means we should not operate without a cap. A bit like our auto-scaling node pool. It has a limit we set to a value that is so high that we don't expect ever to reach it, but there is a limit in case unexpected things happen.

If we want to go with the exemption setup first as a stepping stone towards a full "per repo quota config" I'd rename it from whitelist to "increased quota" or "high quota specs"? Mostly because whitelist could mean anything.

choldgraf:
👍 let me put some cycles into the generalized approach then - I was doing it this way in the name of "good is better than perfect" but at the same time it'll get harder to use a more generalized thing if we keep doing one-by-one options in the way we're doing them now.

betatim:
Do we know if the Jupyterlab release is happening any time soon?

I'd be happy to merge something that introduces a second (higher) quota level. So instead of whitelist we could call it `high_quota_specs` and use it specifically for this week/JLab launch.

By having a "high quota" thing we also circumnavigate the issue of how to transition the `banned_specs` config to the new system. And I think we learn something about how to build the per repo quota by doing this. 

Things we'd need to do:
* [ ] rename config key to `high_quota_specs`
* [ ] find a name for and add config key to set the high quota
* [ ] change the quota check in `launch()` to fetch the right limit
* [ ] some tests (I left a comment on the code with an idea)

binderhub/repoproviders.py
```diff
@@ -91,6 +100,17 @@ def is_banned(self):
                 return True
         return False
 
+    def is_whitelisted(self):
```
betatim: I think this is the method that we should test first. I'd instantiate the repo provider class, set the `spec` attribute by hand (if it isn't easy to do the normal way because repo provider makes network calls) and then test if `is_whitelisted()` gives the right answer for different spec/whitelist combos.

betatim: We can reuse a lot of the code from https://github.com/jupyterhub/binderhub/blob/b5ccb6c6b5259d53bc0f2a062cfbe290afe71544/binderhub/tests/test_repoproviders.py#L66-L92 for this I think

choldgraf:
OK, the latest push renames the spec, adds in a check to choose whether use the "low" or "high" quota limit, and adds a test for this (though this is a little bit more complex than the banned test so we're not testing some new functionality) . Lemme know what you think.

binderhub/tests/test_repoproviders.py
```diff
@@ -83,6 +83,26 @@ def test_banned():
...
+    provider = GitHubRepoProvider(
+        spec='jupyterhub/zero-to-jupyterhub-k8s/v0.4',
+        high_quota_specs=[
+            '^yuvipanda.*'
+        ]
+    )
+    assert not provider.has_higher_quota()
+
+
+def test_not_higher_quota():
```
betatim:
```suggestion
def test_higher_quota():
```

binderhub/tests/test_repoproviders.py
```diff
@@ -83,6 +83,26 @@ def test_banned():
     assert provider.is_banned()
 
 
+def test_higher_quota():
```
betatim:
```suggestion
def test_not_higher_quota():
```

betatim:
Ready to merge from your side (title says still WIP).

Looks good to me.

choldgraf:
I'm fine as is, if you wanna merge!

Issue #887, opened 2019-07-02
Add regex-based repository configuration


choldgraf:
In #882 @betatim and I discussed exposing configuration for specific key:value pairs using regexes (similar to how we ban repositories, but with an extra level that would let you add custom key/values)

I had suggested:

> so in this case the config file would look something like:

```
  config:
    GitHubRepoProvider:
      # Add banned repositories to the list below
      # They should be strings that will match "^<org-name>/<repo-name>.*"
      spec_configuration_override:
        ^ines/spacy-binder.*:
		    quota: 1000
        ^binder-examples/requirements.*:
   			quota: 10     
```

> You could imagine in the future extending that if you wanted other repo-specific config keys. Does that make sense?

and @betatim replied:

> I think we should consolidate banned_specs with this new key so we don't end up with two places that do similar stuff. It would be polite to provide some amount of time where people can keep using banned_specs (internally we translate it to a repo with a quota of zero?!) and get a warning that they should transition to the new way. Seems like the polite thing to do and makes mybinder.org ops team's life a little easier too.

> Not a fan of the new name spec_configuration_override. I had to think a good moment about what it meant, what was likely to be the config found within it, etc. I'll suggest some names in the PR.

> What do you think of using this as a way to configure limits for whole organisations (say all repos in the binder-examples org)? It is possible right now, I like the idea of being able to do this and maybe it gives us an idea for how to name this config option.

Then the issue got closed...this issue is just keeping that thread alive :-)

betatim:
Except for trying to come up with a more intuitive name I'd merge a PR implementing this in no time. It would let us set limits per repo which would be nice for heavy hitters and a kind of "soft ban" and one day maybe even per spec resources.

betatim:
Time to close this?

choldgraf:
I think so? Do we have a definitive test to know that it works yet?

betatim:
Kind of as it is used on mybinder.org. Could transition some of the high spec repos to it to make really sure.

PR #888, opened 2019-07-03
[mrg] Per spec configuration

choldgraf:
In https://github.com/jupyterhub/binderhub/issues/887 we talked about how it'd be cool to have per-repo configuration. This is a first attempt at implementing it.

The basic idea is that we operate similar to how "banned" and "quota increase" functionality works, with one difference:

Instead of a **list** of specification regexes to match to "quota increase" or "banned", we have a **dictionary** where keys are specification regexes, and values are dictionaries of "key:value" pairs that can over-ride configuration on a repo-specific basis.

For example, you could do something like:

```
config:
    GitHubRepoProvider:
      # Add banned repositories to the list below
      # They should be strings that will match "^<org-name>/<repo-name>.*"
      spec_configuration_override:
        ^ines/spacy-binder.*:
		    quota: 1337
        ^binder-examples/requirements.*:
   			banned: true     
```

## Questions to answer

1. I agree w/ @betatim that "spec_configuration_override" isn't great, any suggestions for other names?
2. What's a clean way to document what key:value configuration pairs are meaningful in this case? We'll need to add checks for them one-by-one in the code, and for now I guess we'll just start with a `quota: <int>` key/value. In the future, should we just expose this with good documentation for what the accepted key/values are? (we could also imagine some key/values being repoprovider-specific)

## Need to add

- [x] Docs
- [x] Tests
- [x] Better naming

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -426,15 +426,6 @@ class BuildHandler(BaseHandler):
         """Ask JupyterHub to launch the image."""
         # Load the spec-specific configuration if it has been overridden
         repo_config = provider.repo_config(self.settings)
-        spec_config = provider.spec_configuration_override
-
-        # check quota first
-        if provider.has_higher_quota():
-            quota = self.settings.get('per_repo_quota_higher')
-        elif "per_repo_quota" in spec_config:
-            quota = spec_config.get('per_repo_quota')
-        else:
-            quota = self.settings.get('per_repo_quota')
 
         # the image name (without tag) is unique per repo
         # use this to count the number of pods running with a given repo
@@ -468,6 +459,7 @@ class BuildHandler(BaseHandler):
 
         # TODO: put busy users in a queue rather than fail?
         # That would be hard to do without in-memory state.
+        quota = repo_config.get('quota')
         if quota and matching_pods >= quota:
             app_log.error("%s has exceeded quota: %s/%s (%s total)",
                 self.repo_url, matching_pods, quota, total_pods)
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -425,6 +425,7 @@ class BuildHandler(BaseHandler):
     async def launch(self, kube, provider):
         """Ask JupyterHub to launch the image."""
         # Load the spec-specific configuration if it has been overridden
+        repo_config = provider.repo_config(self.settings)
         spec_config = provider.spec_configuration_override
 
         # check quota first
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -424,9 +424,14 @@ class BuildHandler(BaseHandler):
 
     async def launch(self, kube, provider):
         """Ask JupyterHub to launch the image."""
+        # Load the spec-specific configuration if it has been overridden
+        spec_config = provider.spec_configuration_override
+
         # check quota first
         if provider.has_higher_quota():
             quota = self.settings.get('per_repo_quota_higher')
+        elif "per_repo_quota" in spec_config:
+            quota = spec_config.get('per_repo_quota')
         else:
             quota = self.settings.get('per_repo_quota')
 
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -80,6 +80,16 @@ class RepoProvider(LoggingConfigurable):
         config=True
     )
 
+    spec_configuration_override = Dict(
+        help="""
+        Dictionary that defines per-repository configuration overrides.
+
+        Should be a Dictionary. Keys are regexes (not regex objects) that
+        match specs. Values are dictionaries of "config_name: config_value" pairs.
+        """,
+        config=True
+    )
+
     unresolved_ref = Unicode()
 
     git_credentials = Unicode(
@@ -110,6 +120,21 @@ class RepoProvider(LoggingConfigurable):
             if re.match(higher_quota, self.spec, re.IGNORECASE):
                 return True
         return False
+
+    def spec_configuration_override(self):
+        """
+        Return the spec configuration overrides if the given spec has them.
+        """
+        if not isinstance(self.spec_configuration_override, dict):
+            raise ValueError(
+                "Specification-specific override is not a "
+                "dictionary, found type %s" % type(self.spec_configuration_override))
+        for spec, config in self.spec_configuration_override:
+            # Ignore case, because most git providers do not
+            # count DS-100/textbook as different from ds-100/textbook
+            if re.match(spec, self.spec, re.IGNORECASE):
+                return config
+        return {}
 
     @gen.coroutine
     def get_resolved_ref(self):
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -140,8 +140,8 @@ class RepoProvider(LoggingConfigurable):
 
         # Spec regex-based configuration
         for item in self.spec_config:
-            pattern = item.pop('pattern', None)
-            config = item.pop('config', None)
+            pattern = item.get('pattern', None)
+            config = item.get('config', None)
             if not isinstance(pattern, str):
                 raise ValueError(
                     "Spec-pattern configuration expected "
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -80,14 +80,17 @@ class RepoProvider(LoggingConfigurable):
         config=True
     )
 
-    spec_config = Dict(
+    spec_config = List(
         help="""
-        List of tuples that defines per-repository configuration.
+        List of dictionaries that define per-repository configuration.
 
-        Each item in the list should have two values.
+        Each item in the list is a dictionary with two keys:
 
-        The first is a regex (not a regex object) that matches specs.
-        The second is a dictionary of "config_name: config_value" pairs.
+            pattern : string
+                defines a regex pattern (not a regex object) that matches specs.
+            config : dict
+                a dictionary of "config_name: config_value" pairs that will be
+                applied to any repository that matches `pattern`
         """,
         config=True
     )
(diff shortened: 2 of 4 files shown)


binderhub/builder.py
```diff
@@ -424,9 +424,14 @@ def escape(s):
 
     async def launch(self, kube, provider):
         """Ask JupyterHub to launch the image."""
+        # Load the spec-specific configuration if it has been overridden
+        spec_config = provider.spec_configuration_override
+
         # check quota first
         if provider.has_higher_quota():
```
betatim: Should we retire this at the same time as we introduce the new functionality?

binderhub/builder.py
```diff
@@ -424,9 +424,14 @@ def escape(s):
...
             quota = self.settings.get('per_repo_quota_higher')
+        elif "per_repo_quota" in spec_config:
+            quota = spec_config.get('per_repo_quota')
```
betatim:
```suggestion
            quota = spec_config.get('quota')
```
(also on the line above)

Though you could write this as:

```python
quota = spec_config.get('quota', self.settings.get('per_repo_quota'))
```
which is a well used pattern in Python for saying "get this from the dictionary and if it doesn't exist use this thing as default"

binderhub/repoproviders.py
```diff
@@ -111,6 +121,21 @@ def has_higher_quota(self):
                 return True
         return False
 
+    def spec_configuration_override(self):
+        """
+        Return the spec configuration overrides if the given spec has them.
+        """
+        if not isinstance(self.spec_configuration_override, dict):
```
betatim: I think because you [specified `Dict` for the option](https://github.com/jupyterhub/binderhub/pull/888/files#diff-c5688934f1e6dc3e932b6c84c1bbbd5dR83) we should never get here/have to check this as traitlets checks for us.

betatim:
How about `spec_config` as name? With a comment in the doc string of the traitlet to say that you only have to use this to override settings per spec/if no entry is found defaults are used.

For documentation of which keys/options are valid I'd put a pointer in the traitlets' doc that says "repo providers are free to define their own options, see each repo provider for a list of valid keys and their meaning". Then add appropriate docs to the repo providers/base repo provider.

binderhub/repoproviders.py
```diff
@@ -111,6 +121,21 @@ def has_higher_quota(self):
                 return True
         return False
 
+    def spec_configuration_override(self):
```
betatim: What do you think of naming this `spec_config()` and taking care of finding and populating all the default values in this function? Then in the builder we wouldn't need to have any checks for "do we override this or not, what is the default". We have one place that is already ready to use.

The reason to call it `spec_config` is that it would be for one particular spec and not a whole bunch any more. Maybe we could even call it `repo_config`? Depends a bit on what words the builder uses (does it refer to building a spec or a repo). But this might be getting a bit OCD on naming :)

TL;DR: should we compute the configuration (combining over rides and defaults) for the spec here?

binderhub/builder.py
minrk: Probably good to have a deprecation transition, in which case include a warning if the old way is used.

binderhub/repoproviders.py
```diff
@@ -111,6 +121,21 @@ def has_higher_quota(self):
...
+        Return the spec configuration overrides if the given spec has them.
+        """
+        if not isinstance(self.spec_configuration_override, dict):
+            raise ValueError(
+                "Specification-specific override is not a "
+                "dictionary, found type %s" % type(self.spec_configuration_override))
+        for spec, config in self.spec_configuration_override:
+            # Ignore case, because most git providers do not
+            # count DS-100/textbook as different from ds-100/textbook
+            if re.match(spec, self.spec, re.IGNORECASE):
```
minrk: FWIW, using `re.match` compiles the regex on each call, which means we are compiling every regex here on every repo provider. Compiling them on application load would make this a bit more efficient.

Related to this: dicts are in a random order, but overrides might want to have a priority if more than one pattern matches the repo. With a dict, the behavior is undefined, with a list it would be consistent. So perhaps a list of tuples is more appropriate than a dict?

binderhub/builder.py
choldgraf: I'm fine w/ a transition (though we've still only ever had one release of BinderHub lol)

binderhub/repoproviders.py
choldgraf: ah ok - so it'll raise an error if somebody passes a list and not a dict in the helm config?

binderhub/repoproviders.py
choldgraf: I like it - that seems reasonable to me, lemme give it a stab and see how it looks

binderhub/repoproviders.py
choldgraf: mmmm, I am not sure how to do this :-) feel free to suggest the code that would still allow for config patterns like the one in https://github.com/jupyterhub/binderhub/pull/888#issue-294300289 and I'll give it a shot

binderhub/repoproviders.py
choldgraf: see the latest push for an attempt at what this would look like, is this the kinda thing you had in mind? 

https://github.com/jupyterhub/binderhub/pull/888/commits/1399d58986da19c77139b8a9b8cedadc25c13c91#diff-c5688934f1e6dc3e932b6c84c1bbbd5dR133

In this case, `spec_config` would be defined at the helm chart config level, and repo_config is a method of RepoProvider that returns a dictionary of configuration values for that *repository* (which might have been updated from the spec_config setting)

binderhub/builder.py
betatim: True on paper and in practice we release several times a week and people actually use BinderHub so I think we should be nice to people. The "higher quota" thing has only existed for a week or two so :-/

```python
warnings.warn(
    "XXX is deprecated, use YYY instead",
    DeprecationWarning
)
```
is the generic Python warnings thing. It isn't ideal (most people won't see deprecation warnings except in their tests). Does traitlets have a tool for deprecations?

binderhub/repoproviders.py
betatim: Good point about the random order! I had pondered that one might have multiple patterns and if they should be additive or "last one wins" or what, then decided that for a first pass we should ignore that. However the random order thing is something we need to address (or inform the user that they have more than one pattern that matches and that this is an error (for now)).

I think what we want is a list like:
```python
specs = [{"pattern": "some-pattern", "quota": 10}, {"pattern": "some-other-pattern", "quota": 12}]
```
in YAML it would look like:
```yaml
# not super sure about the first one but the second two should work
- {pattern: "yet another pattern", quota: 33}
- pattern: some-pattern
  quota: 10
- pattern: some-other-pattern
  quota: 12
```

choldgraf: Ah I like this pattern! @minrk does it look good to you as well?

binderhub/builder.py
minrk: Due to the invisible-by-default deprecation warnings, I think that application-level deprecation warnings should actually *not* use DeprecationWarning, so they are visible by default. It's a tricky choice since deprecation warnings are always safely ignored "for now" but if folks never see them, they may as well not exist. As for a transition, I was mostly thinking of ourselves at mybinder.org, since it would allow us to decouple "adopt the new way" from "bump binderhub" PRs without config changes.

FWIW, I don't think the deprecation needs to be a blocker here, as the old way still works. Cleanly deprecating the old way can be a follow-up PR.

> Does traitlets have a tool for deprecations?

Traitlets doesn't have any helpers for deprecations, but the way traitlets works can make writing deprecations handy since deprecations can go in observers, and completely removed from the bodies of methods, etc. for example:

```python
@observe('old_way')
def _old_way_changed(self, change):
    warnings.warn("old way is deprecated as of...new way is...")
    self.new_way = modified_form(change.new)
    # never need to look directly at self.old_way after this

...

def method(self):
    only_use(self.new_way)
```

The simplest version of this can lead to undefined behavior (which wins?) if both the old and new way are specified in the same config, but I think that's not a big deal if the old-way warning is visible.

binderhub/repoproviders.py
minrk: yes, that looks great to me! The only question is that 'pattern' is in the same namespace as the config overrides. This is nice since it's more concise, but could allow collisions or confusion. The more rigorous, but slightly more tedious would be to explicitly separate the match from the override config:

```yaml
- pattern: some-pattern
  config:
    quota: 10
    other_config: "x"
```

This would allow adding other siblings to `pattern` if we e.g. had some other options that should influence the selection rather than the overridden config. I don't have strong feelings, but I've occasionally regretted not separating things like this in JupyterHub in the past (see spawner options in the REST API).

You can see the difference in code where 'pattern' needs to be handled specially:

```python
# copy because we need to modify the dict
update_config = config.copy()
# remove pattern before updating to config because pattern is not part of the config:
pattern = config.pop('pattern')
if pattern.match(...):
    config.update(update_config)
```

vs:

```python
pattern = item['pattern']
update_config = item['config']
if pattern.match(spec):
    config.update(update_config)
```

binderhub/builder.py
betatim: I think using the traitlets `observe` together with printing a warning (via `warnings.warn()`) is a good way to do this.

betatim:
> FWIW, I don't think the deprecation needs to be a blocker here, as the old way still works. Cleanly deprecating the old way can be a follow-up PR.

That is a good point. Let's follow that suggestion. Being able to bump binderhub on mybinder.org without having to also change the config makes life easier (and lets us use henchbot!).

choldgraf:
See the latest push for the following updates:

- The API should now follow what @minrk suggested, how does that look?
- A buncha tests added (let's see if they pass!)
- Some [minor docs added](https://369-89419368-gh.circle-artifacts.com/0/html/customizing.html#custom-configuration-for-specific-repositories)

binderhub/repoproviders.py
```diff
@@ -111,6 +126,38 @@ def has_higher_quota(self):
                 return True
         return False
 
+    def repo_config(self, settings):
```
choldgraf: should we make this a property or keep an explicit method?

minrk: I think it can't be a property if it's taking arguments.

binderhub/repoproviders.py
```diff
@@ -111,6 +126,38 @@ def has_higher_quota(self):
...
+
+        # Defaults and simple overrides
+        if self.has_higher_quota():
+            repo_config['quota'] = settings.get('per_repo_quota_higher')
+        else:
+            repo_config['quota'] = settings.get('per_repo_quota')
+
+        # Spec regex-based configuration
+        for item in self.spec_config:
+            pattern = item.pop('pattern', None)
```
minrk: I think we want `get` not `pop`, to avoid modifying the item.

binderhub/tests/test_repoproviders.py
```diff
@@ -92,6 +92,65 @@ def test_higher_quota():
...
+        "pattern": '^yuvipanda.*',
+        "config": {
+            "key1": "val1",
+            "quota": 999
+        }
+    }
+
+    provider = GitHubRepoProvider(
+        spec='jupyterhub/zero-to-jupyterhub-k8s/v0.4',
+        config=[base_config]
```
minrk: spec_config?

betatim:
I'd got with option (1) for the moment in terms of priority/when to stop.

binderhub/repoproviders.py
betatim: I am happy with it being a method.

choldgraf:
thanks for the feedback on tests etc

The latest commit:

* Fixes the tests so they pass
* Adds a test for default behavior w/o a matching spec
* Adds a note to documentation about configuration override

betatim:
Nice work! Tests, docs and a new feature!

---

What is the etiquette for marking conversations as "resolved"? Can I do it in a PR that isn't mine? Should the author do it? When I am the author I tend to mark things are solved (if they don't auto collapse) when I have implemented the feedback or the discussion has somehow ended. In this PR there were several people involved so it felt weird to do that but I also found myself scrolling through each discussion several times trying to not miss new comments.

choldgraf:
Re: "resolved" that's a good question, I'm generally happy for people to resolve things even if I'm the author on the PR, as it often means that my to-do list has shrunk :-)

as with the other team compass guidelines, I trust the judgment of other folks on the project so I'm happy for them to take the initiative there

Issue #675, opened 2018-09-28
Resolving branch names for repo providers that aren't github.com or gitlab.com


betatim:
For repoproviders that are not GitHub.com or Gitlab.com users have to specify the commit SHA by hand. Leaving the SHA field in the form empty (default) or writing `master` doesn't work.

This is because BinderHub talks to the GitHub API to resolve `master` or the branch name to a SHA. However we currently don't do this for repoproviders that aren't GitHub.com or GitLab.com.

Explaining to new comers why it works for one GitLab instance but not a different one is hard.

@minrk suggested we try and detect if a source is a gitlab instance and then talk to the API without authorisation token. This would get users the same behaviour as for an instance we "know" about. The downside is we might run into API rate limits.

So we need some good UI around this.

A stopgap might be to edit the form we show to make it more obvious which fields are required and which optional in the different cases.

minrk:
Additionally, if there's a way to resolve refs with the *git* API without cloning, that would be great. I'm not aware of a way to do this, though.

djhoese:
I'd like to +1 this feature request. I have a gitlab instance at my work where the repository is public and would like to point binder to the master branch. Having to point it to a commit hash causes a weird chicken/egg problem where I can't include the URL to binder in a README until I already have a working configuration commit. If I make a change I have to update the URL in the README every time *after* I commit.

sauerburger:
I had the same chick-egg problem. I created a website with a quick workaround until this is supported natively: https://b4gl.sauerburger.io/

The site uses the gitlab api to resolve the ref and then redirects to mybinder.org using the resolved git commit. Feel free to use it.

betatim:
Closed by #895

Issue #843, opened 2019-05-07
Resolve remote git references using provider agnostic method


manics:
Follow up from https://discourse.jupyter.org/t/how-to-use-bitbucket-repositories-on-mybinder-org/1000/4

### Proposed change
- Resolve branch names using `git ls-remote` in the bare git provider https://github.com/jupyterhub/binderhub/blob/0b8f99af0a23a7828810b03f010edcefc9a46195/binderhub/repoproviders.py#L126-L160
- Potentially also replace provider specific API calls to resolve branch names

E.g.:
```
git ls-remote https://bitbucket.org/ueacomputervision/image-labelling-tool
f3ddb33e4839f8a0fe73c168993b405adc13daf0        HEAD
8116559e4030d1d4f557617780410cbca7896663        refs/heads/canvas
d50105bacbb74ca139c5bf38ef1657190a006929        refs/heads/group
f3ddb33e4839f8a0fe73c168993b405adc13daf0        refs/heads/master
0b92e511ce8e04f26131efddde9b4a2f5ef9e1e1        refs/heads/multi-user
bc53e6aa12b6d7988de514c8c92e29877be5cc61        refs/heads/separate-modules
a1b2c3bcc2e563afbd335013805d19c95d67224d        refs/heads/server-side-refactor
```
```
$ git ls-remote https://bitbucket.org/ueacomputervision/image-labelling-tool master
f3ddb33e4839f8a0fe73c168993b405adc13daf0        refs/heads/master
```

As indicated in the linked discussion currently the resolution of branch-names to sha1s is done using provider specific API calls, for anything unsupported users must figure out the commit hash themselves.

### Who would use this feature?
Anyone not using GitHub or GitLab.

betatim:
You think you know a lot about Git and then something like this comes along. Every day is a school day! We should definitely add this to the "bare git repo provider". For GitHub, Gitlab.com and others where we have an API we can call I'd leave things as is. It works, it is cheap and we have a good caching setup.

For bare git repositories this will be a major usability improvement!

betatim:
Closed by #895

PR #895, opened 2019-07-11
Use git ls-remote to resolve refs for git provider

hugokerstens:
Implements the approach specified in https://github.com/jupyterhub/binderhub/issues/843.

### DIFF
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -13,6 +13,7 @@ import os
 import time
 import urllib.parse
 import re
+import subprocess
 
 from prometheus_client import Gauge
 
@@ -173,7 +174,7 @@ class GitRepoProvider(RepoProvider):
 
     Users must provide a spec of the following form.
 
-    <url-escaped-namespace>/<resolved_ref>
+    <url-escaped-namespace>/<unresolved_ref>
 
     eg:
     https%3A%2F%2Fgithub.com%2Fjupyterhub%2Fzero-to-jupyterhub-k8s/f7f3ff6d1bf708bdc12e5f10e18b2a90a4795603
@@ -186,16 +187,32 @@ class GitRepoProvider(RepoProvider):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        url, resolved_ref = self.spec.rsplit('/', 1)
+        url, unresolved_ref = self.spec.rsplit('/', 1)
         self.repo = urllib.parse.unquote(url)
-        if not resolved_ref:
-            raise ValueError("`resolved_ref` must be specified as a query parameter for the basic git provider")
-        self.sha1_validate(resolved_ref)
-        self.resolved_ref = resolved_ref
-        self.unresolved_ref = resolved_ref
+        self.unresolved_ref = urllib.parse.unquote(unresolved_ref)
+        if not self.unresolved_ref:
+            raise ValueError("`unresolved_ref` must be specified as a query parameter for the basic git provider")
+        try:
+            self.sha1_validate(unresolved_ref)
+            self.resolved_ref = unresolved_ref
+        except ValueError:
+            pass
 
     @gen.coroutine
     def get_resolved_ref(self):
+        if hasattr(self, 'resolved_ref'):
+            return self.resolved_ref
+
+        command = ["git", "ls-remote", self.repo, self.unresolved_ref]
+        result = subprocess.run(command, text=True, stdout=subprocess.PIPE)
+        if result.returncode:
+            raise Exception("Unable to run git ls-remote to get the `resolved_ref`")
+        if not result.stdout:
+            raise ValueError("The specified branch, tag or commit SHA was not found on the remote.")
+        resolved_ref = result.stdout.split(None, 1)[0]
+
+        self.sha1_validate(resolved_ref)
+        self.resolved_ref = resolved_ref
         return self.resolved_ref
 
     def get_repo_url(self):
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -187,7 +187,7 @@ class GitRepoProvider(RepoProvider):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        url, unresolved_ref = self.spec.rsplit('/', 1)
+        url, unresolved_ref = self.spec.split('/', 1)
         self.repo = urllib.parse.unquote(url)
         self.unresolved_ref = urllib.parse.unquote(unresolved_ref)
         if not self.unresolved_ref:
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -206,7 +206,7 @@ class GitRepoProvider(RepoProvider):
         command = ["git", "ls-remote", self.repo, self.unresolved_ref]
         result = subprocess.run(command, text=True, stdout=subprocess.PIPE)
         if result.returncode:
-            raise Exception("Unable to run git ls-remote to get the `resolved_ref`")
+            raise RuntimeError("Unable to run git ls-remote to get the `resolved_ref`")
         if not result.stdout:
             raise ValueError("The specified branch, tag or commit SHA ('{}') was not found on the remote repository."
                              .format(self.unresolved_ref))
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -208,7 +208,7 @@ class GitRepoProvider(RepoProvider):
         if result.returncode:
             raise Exception("Unable to run git ls-remote to get the `resolved_ref`")
         if not result.stdout:
-            raise ValueError("The specified branch, tag or commit SHA was not found on the remote.")
+            raise ValueError("The specified branch, tag or commit SHA ('{}') was not found on the remote repository.".format(self.unresolved_ref)
         resolved_ref = result.stdout.split(None, 1)[0]
 
         self.sha1_validate(resolved_ref)
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -208,7 +208,8 @@ class GitRepoProvider(RepoProvider):
         if result.returncode:
             raise Exception("Unable to run git ls-remote to get the `resolved_ref`")
         if not result.stdout:
-            raise ValueError("The specified branch, tag or commit SHA ('{}') was not found on the remote repository.".format(self.unresolved_ref)
+            raise ValueError("The specified branch, tag or commit SHA ('{}') was not found on the remote repository."
+                             .format(self.unresolved_ref))
         resolved_ref = result.stdout.split(None, 1)[0]
 
         self.sha1_validate(resolved_ref)
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -175,8 +175,10 @@ class GitRepoProvider(RepoProvider):
     Users must provide a spec of the following form.
 
     <url-escaped-namespace>/<unresolved_ref>
+    <url-escaped-namespace>/<resolved_ref>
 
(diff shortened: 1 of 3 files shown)


akhmerov:
Shouldn't unresolved ref default to `master`?

binderhub/repoproviders.py
```diff
@@ -186,16 +187,32 @@ class GitRepoProvider(RepoProvider):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        url, resolved_ref = self.spec.rsplit('/', 1)
+        url, unresolved_ref = self.spec.rsplit('/', 1)
```
akhmerov: While this is correct if we only accept SHA, it stops being correct if we need a branch name:

image

hugokerstens:
It already defaults to master in the front-end javascript if no ref is specified.

binderhub/repoproviders.py
hugokerstens: Good catch. There is no way to know which parts belong to the repository URL and which belong to the ref, so this is a fundamental issue with how the build URL is passed to the repo provider. There is actually no easy way to fix this, except for guessing by trying to split at each `/` and calling `git ls-remote`.

akhmerov: I see other providers, e.g. `GithubRepoProvider` change what they mean with the ref. Could we try the same approach here?

hugokerstens: I just realized the namespec is url escaped, so using the regular `split` should work. Disregard my comment above.

hugokerstens: Fixed in 92ad34c

binderhub/static/js/index.js
```diff
@@ -76,7 +76,7 @@ function updateRepoText() {
   }
   else if (provider === "git") {
     text = "Arbitrary git repository URL (http://git.example.com/repo)";
-    tag_text = "Git commit SHA";
+    tag_text = "Git branch, tag, or commit SHA";
```
akhmerov: Should be just `"Git branch, tag, or commit"` to match other similar providers exactly.

akhmerov:
Test failures seem relevant; also we need to test the new cases.

binderhub/static/js/index.js
hugokerstens: It doesn't work for partial commit strings, that's why I explicitly added 'SHA'.

akhmerov: :+1:

minrk:
If it works as advertised, this is awesome! In the future, it could also mean that we could greatly reduce our needed git provider API support, since all we'd need to do is turn provider-specific-url into git url, which should generally be doable purely in our code and without making any API requests.

hugokerstens:
The [test](https://travis-ci.org/jupyterhub/binderhub/jobs/557352266) for 073dba6 failed because Jupyterhub is not available for some reason.

binderhub/repoproviders.py
```diff
@@ -186,16 +187,32 @@ class GitRepoProvider(RepoProvider):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        url, resolved_ref = self.spec.rsplit('/', 1)
+        url, unresolved_ref = self.spec.split('/', 1)
```
betatim: Why the change from `rsplit` to `split`?

akhmerov: See the discussion above: a branch name may contain `/`. Is it possible to obtain url and the ref separately at this stage?

binderhub/repoproviders.py
```diff
@@ -186,16 +187,32 @@ class GitRepoProvider(RepoProvider):
...
-        self.resolved_ref = resolved_ref
-        self.unresolved_ref = resolved_ref
+        self.unresolved_ref = urllib.parse.unquote(unresolved_ref)
+        if not self.unresolved_ref:
+            raise ValueError("`unresolved_ref` must be specified as a query parameter for the basic git provider")
+        try:
+            self.sha1_validate(unresolved_ref)
+            self.resolved_ref = unresolved_ref
+        except ValueError:
+            pass
```
betatim: There are a few things here which aren't obvious from reading it a few times so maybe we can add some comments.

Why are we now swallowing exceptions here?

On L196 calling `self.sha1_validate(unresolved_ref)` should fail a lot because users can pass in a value like `master` so why call it now and not after we resolve the reference?

It seems weird to set `self.resolved_ref` to the value of `unresolved_ref`. I think we should delay that to when we have a resolved ref (and it it to `None` in the meantime).

binderhub/repoproviders.py
```diff
@@ -186,16 +187,32 @@ class GitRepoProvider(RepoProvider):
...
 
     @gen.coroutine
     def get_resolved_ref(self):
+        if hasattr(self, 'resolved_ref'):
+            return self.resolved_ref
+
+        command = ["git", "ls-remote", self.repo, self.unresolved_ref]
+        result = subprocess.run(command, text=True, stdout=subprocess.PIPE)
+        if result.returncode:
+            raise Exception("Unable to run git ls-remote to get the `resolved_ref`")
```
betatim: Let's make this a `RuntimeError` instead of an `Exception` which is hard to catch for users of this code.

binderhub/repoproviders.py
```diff
@@ -186,16 +187,32 @@ class GitRepoProvider(RepoProvider):
...
+        if not result.stdout:
+            raise ValueError("The specified branch, tag or commit SHA was not found on the remote.")
```
betatim:
```suggestion
            raise ValueError("The specified branch, tag or commit SHA ('{}') was not found on the remote repository.".format(self.unresolved_ref)
```
Showing the user the value that arrived here can help them figure out why it isn't working.

binderhub/repoproviders.py
hugokerstens: I was looking for a more appropriate exception but couldn't think of any, `RuntimeError` seems more at its place here indeed.

binderhub/tests/test_build.py
```diff
@@ -16,7 +16,7 @@
 @pytest.mark.parametrize("slug", [
     "gh/binderhub-ci-repos/requirements/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd",
     "git/{}/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd".format(
-        quote("https://github.com/binderhub-ci-repos/requirements")
+        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
```
betatim: Why this change to the way the URL is quoted?

hugokerstens: In the real situation the URL is encoded that way (every special character, including `/`). By default the `quote` function does not url encode `/`, so that's why it needs to specified this way.

binderhub/repoproviders.py
hugokerstens: We pass on an error since we postpone resolving the ref to the getter of `resolved_ref` if the ref is not a commit SHA. It makes sense to move this sha1 check to the getter for the `resolved_ref` and not set the `resolved_ref` at all in the constructor. This is in line with the other repo providers.

binderhub/repoproviders.py
hugokerstens: The method used above is the only way to get them separately.

In the current implementation the reference is not url encoded like the url, but I think it's better if we DO encode it. Then the `/` is the true divider between the url and ref, instead of the first `/` like now.

betatim: The discussion is [here](https://github.com/jupyterhub/binderhub/pull/895#discussion_r302481788).

hugokerstens:
> ## 1 - `unresolved_ref` renamed for clarity?
> So, perhaps we can name this variable to indicate it not as an unresolved reference, but simply a _reference_ or _unparsed reference_, _provided reference_ or _input reference_ etc?

That's indeed a lot more readable. If we move the sha1 check to the getter ([as discussed here](https://github.com/jupyterhub/binderhub/pull/895#discussion_r303006606)), renaming it is not needed anymore. We can just speak of an unresolved reference in the constructor, and only bother with validating in the getter for the resolved ref.

betatim:
I think switching to `unresolved_ref` as a variable name is a good idea and consistent with how other providers name it. Maybe the most precise name would be `potentially_unresolved_ref` but that is quite long :-/

Right now when you enter a git repo URL in the UI you end up with the following link `https://mybinder.org/v2/git/http%3A%2F%2Fgit.example.com%2Frepo/123124124`
Screen Shot 2019-07-12 at 16 44 47

so we should make sure that format continues to work. It makes it a bit tricky to deal with branches that contain `/` though. From a quick look at the gitlab.com provider it seems to do what you discussed in the comment and encode the `/` in the URL of the repo and leave the ones in the branch/tag name alone. So that would be a good example to follow.

We should extend the tests to handle the new possibilities (as well as some invalid ones?). For this to work we need to mock the calls to `git ls-remote` as we don't want to actually run commands that need networking. To test the resolving via `git ls-remote` we can setup a local "throw away" repo, clone it, and then use the original as the "remote". I think that would work.

consideRatio:
It is this change that led me to consider the variable name in the first place I guess. Previously, a resolved reference was required, but now it cannot be a resolved reference? Well the code is currently meant to support that, but I got confused due to that and figured any kind of reference actually works.

```diff
    Users must provide a spec of the following form.

-    <url-escaped-namespace>/<resolved_ref>
+    <url-escaped-namespace>/<unresolved_ref>
```

binderhub/repoproviders.py
consideRatio: I also appreciate managing all logic within the getter as there is one. It was confusing to me the first time I read this through that some of it was found there but not all of it.

binderhub/repoproviders.py
```diff
@@ -186,16 +189,36 @@ class GitRepoProvider(RepoProvider):
...
+            # The ref is a head/tag and we resolve it using `git ls-remote`
+            command = ["git", "ls-remote", self.repo, self.unresolved_ref]
+            result = subprocess.run(command, text=True, stdout=subprocess.PIPE)
+            if result.returncode:
+                raise RuntimeError("Unable to run git ls-remote to get the `resolved_ref`")
+            if not result.stdout:
+                raise ValueError("The specified branch, tag or commit SHA ('{}') was not found on the remote repository."
+                                .format(self.unresolved_ref))
+            resolved_ref = result.stdout.split(None, 1)[0]
+            self.sha1_validate(resolved_ref)
```
hugokerstens: Is validating the SHA that is returned by `git ls-remote` too much? I think this safe-guard is nice to have for unexpected output of `git ls-remote`.

binderhub/repoproviders.py
hugokerstens: Is the current implementation satisfactory? 

The only thing I'm not 100% sure about is setting the `resolved_ref` in the `else` block. It is a good practice to only have code in the `try` block that produces the error you want to catch. However, it could be more readable if the `resolved_ref` is set directly after validating it in the `try` block. I tried to make it more readable by adding some comments, but still assign it in the `else` block.

hugokerstens:
I added some extra documentation for the `GitRepoProvider` that explicitly states that both resolved and unresolved references are possible to address the concerns by @consideRatio. However, the variable is still called `unresolved_ref` in the code, which is confusing as well. The problem here is the ambiguity of 'unresolved'. It could either mean unresolved as in yet to be resolved to a commit hash, OR unresolved as in not validated at all.

chicocvenancio:
This fixes #675 if I'm not misunderstanding something. A great improvement.

betatim:
Thanks for bumping this @chicocvenancio. Deployed this locally and tested it with some GitHub and GitLab repositories (pretending they were just git repos). Seems to work 🎉.

Merging! Thanks for this cool feature!

PR #901, opened 2019-07-17
Bump JupyterHub chart

chicocvenancio:
Bump JupyterHub to latest chart. `03215dd` is from April and does not capture several improvements. I noticed the need for this with CHP version and the redirection to https port.


betatim:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/compare/03215dd...dcde99a is the diff between the two versions. I'll take a quick look at it before hitting merge as I don't know exactly all the things that have changed and whether we can deploy this straight to mybinder.org or if we need to do some adjustments to BinderHub or other dependencies.

PR #904, opened 2019-07-26
[MRG] Add a /health endpoint to BinderHub

betatim:
This reports the health of BinderHub and the services it needs to run.

It is quite primitive but maybe good enough to get going? In particular the request to `/health` will take a long time if we have to retry connections to the hub or the registry. A smarter way would be to monitor those independently of someone calling the end point and having the most recent results available. Not quite sure what the best way to do that would be (where would the results be stored so they are "globally available"?)

### DIFF
--- a/binderhub/health.py
+++ b/binderhub/health.py
@@ -87,6 +87,8 @@ class HealthHandler(BaseHandler):
         res = await self.check_jupyterhub_api(self.hub_url)
         checks.append({"service": "JupyterHub API", "ok": res})
 
-        self.write(
-            json.dumps({"ok": all(check["ok"] for check in checks), "checks": checks})
-        )
+        overall = all(check["ok"] for check in checks)
+        if not overall:
+            self.set_status(503)
+
+        self.write(json.dumps({"ok": overall, "checks": checks}))
--- a/binderhub/health.py
+++ b/binderhub/health.py
@@ -1,0 +1,92 @@
+import asyncio
+import json
+
+from functools import wraps
+
+from tornado.httpclient import AsyncHTTPClient
+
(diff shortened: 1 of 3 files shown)


betatim:
Another question: what status code should this return when something is not working but we still want to send the JSON body (so that the requester can see what is broken)? I'd suggest 503 but that is nothing more than a hunch.

binderhub/health.py
```diff
@@ -0,0 +1,92 @@
...
+    async def check_jupyterhub_api(self, hub_url):
+        """Check JupyterHub API health"""
+        await AsyncHTTPClient().fetch(hub_url + "hub/health", request_timeout=2)
+
+        return True
+
+    @false_if_raises
+    @retry
+    async def check_docker_registry(self):
+        """Check docker registry health"""
```
betatim: Note to self: maybe docker registries have a `/health` end point we can call instead?

betatim: `curl "https://dlqpwel7.gra5.container-registry.ovh.net/api/health" -H "accept: application/json"`

betatim: Maybe https://dlqpwel7.gra5.container-registry.ovh.net/v2/ and https://gcr.io/v2/ are the better choices as they are part of the docker registry. https://github.com/docker/distribution/issues/629 suggests that you should get a status 401 here if everything is well.

betatim:
Ready for some review. I don't know a good way to test the docker registry part :-/ Planning on testing that once it is deployed.

betatim:
I'll merge it now. I think there are bound to be a few snags but I can't find any more right now -> needs some real world traffic/usage.

PR #906, opened 2019-07-27
adding social meta tags to binder pages

### LINKED ISSUE #903: Add social media image to page "head"
I've noticed that Google Collaboratory links include the Colab logo when they're shared on social media. e.g.:

https://twitter.com/GalOshri/status/1105962349310046208

I looked into it, and it seems like there are a few "meta" tags that we can add to our HTML that will cause twitter/facebook/etc to render images rather than "just" the links:

https://css-tricks.com/essential-meta-tags-social-media/

I think if we added those meta tags to the template here:

https://github.com/jupyterhub/binderhub/blob/master/binderhub/templates/page.html#L8

then we should get the same effect but for the Binder logo.

What do people think about:

1. Creating an image w/ the Binder logo and "Binder" underneath it that's fit for web viewing.
2. Added "meta" tags so that this image was displayed when people shared binder links on twitter?

I think it could be a good way to make links more discoverable and build up the brand recognition for the logo

choldgraf:
closes #903 

I believe that this will cause Binder links that are shared on social media to display a "Binder" logo on them. I followed [the guidelines here](https://css-tricks.com/essential-meta-tags-social-media/) in the dimensions and `meta` tags for the logo.

I'm not sure how to test this out without actually deploying it to a BinderHub :-/  anybody have ideas on that one? Otherwise we can give a merge and test it out on the [twitter card validator](https://cards-dev.twitter.com/validator) when it's live


betatim:
Merging and deploying.

PR #907, opened 2019-07-28
absolute URL for social image

choldgraf:
From looking at the [twitter card validator](https://cards-dev.twitter.com/validator) it seems like we *are* correctly pulling some information from the binder link, but it looks from [this SO post](https://stackoverflow.com/questions/9858577/open-graph-can-resolve-relative-url) that the open graph images will only work with absolute URLs.

So let's try coding an absolute URL for mybinder.org in there

https://stackoverflow.com/questions/9858577/open-graph-can-resolve-relative-url

In the future perhaps we can expose a template variable like [base_url](https://github.com/jupyterhub/binderhub/blob/a52925198dca15f69a23869a861529334391f078/helm-chart/images/binderhub/binderhub_config.py#L89) though I'm not sure how to actually set that variable

### DIFF
--- a/binderhub/templates/page.html
+++ b/binderhub/templates/page.html
@@ -10,6 +10,10 @@
   <meta property="og:title" content="The Binder Project">
   <meta property="og:description" content="Reproducible, sharable, interactive computing environments.">
   <meta property="og:image" content="https://mybinder.org/static/images/logo_social.png">
+  <meta property="og:image:width" content="1334">
+  <meta property="og:image:height" content="700">
+  <meta property="og:image:alt" content="The Binder Project Logo" />
+
   <meta name="twitter:card" content="summary_large_image">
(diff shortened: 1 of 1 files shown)


choldgraf:
I'm gonna merge and deploy once the tests pass because this is pretty minor + doesn't work now anyway :-)

choldgraf:
merging since first tests passed and the next commit just added a couple meta tags

betatim:
To include the host checkout https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.static_url though we probably want to use 

https://github.com/jupyterhub/binderhub/blob/master/binderhub/app.py#L148-L158

as the "base" instead of the host on which BinderHub is actually running. In the case of our federation it would mean you'd get different URLs depending on which cluster you get sent to. I think.

PR #909, opened 2019-07-29
disable continuous-image-puller

bitnik:
disables `continuous-image-puller`. Before it was disabled by default.

ref: https://github.com/jupyterhub/binderhub/pull/901


betatim:
This change seems reasonable to me as the benefit of pre-pulling the networking tools is minimal. They will be pulled as soon as a pod starts on a node and most of the time a new node appears because the node pool is scaling up to start a new user pod.

Issue #908, opened 2019-07-28
Improvements to the social media images


choldgraf:
Now that social media images work, a few things we could do to improve them:

- [x] Make the background of the image white so that it pops better
- [ ] Add a square image as well (I think that you can have both a rectangular and a square image and sites will choose the best one)
- [x] Should we make any changes to the tagline that is shared as well? I was thinking it’d be good to have the word “open” in there
- [ ] Get the full URL of the image automatically(copypasted from @betatim):

   > To include the host checkout 
   > https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.static_url 
   > though we probably want to use 

   > https://github.com/jupyterhub/binderhub/blob/master/binderhub/app.py#L148-L158

   > as the "base" instead of the host on which BinderHub is actually running. In the case 
   > of our federation it would mean you'd get different URLs depending on which cluster 
   > you get sent to. I think.

betatim:
My comment in Point 4 was about how to make the URL of the image absolute and reusing some config we already have. Tried to make the text better by editing your comment.

---

Another idea would be to generate the content of the image "on the fly" to include the name of the repo that is being launched. A bit like shields.io let's you generate custom shields by specifying things in the URL. I think that would be nice to make it easier to see where the link is going. At least on twitter I now just see a t.co link which doesn't tell me much about what link I am about to click.

In another iteration we could think about how to combine the binder logo with the social media image of a repository (on GitHub).

Thse would be a great task to help out with if you know design but are new to programming as the first steps would be about mock ups and finding a good design.

---

Another thing to discuss is should we show the large card or the small card? So card type `summary` vs `summary_large_image`.

---

Should there be more margin on the left and right side of our card image? Right now it feels like it is an image that got auto-cropped to me :-/

choldgraf:
It seems like the margin point is an easy one to address...I can make a PR that shrinks this down a little bit and adds a white background.

re: summary vs. large summary, it sounded like most blogs recommended defaulting to the "larger" image because it is more common across other platforms. but happy to think about other perspectives there.

re: auto-generating the name, that sounds like a great idea although more complex than I can tackle in a weekend afternoon hack :-)

betatim:
> auto generating images

Nods, this isn't a Sunday morning hack project, would make a nice first contribution though :) Probably wise to tackle it in two steps even: figure out what the image should look like (design) and then figure out the code to generate it on the fly.

> large vs small

[...]

PR #910, opened 2019-07-29
tweaks to social images

choldgraf:
Per https://github.com/jupyterhub/binderhub/issues/908 I think these are some improvements to the social images.

In this PR:

* Adds margins to the Binder logo
* Adds logic to choose the link title based on if we're an index template vs. a loading template (and adds `spec` information to the loading title)

## Todo

Discuss whether the way I've chosen to add spec information to the loading title makes sense. A few thoughts to consider:

* Do we want a different structure than just `gh/org/repo/ref`?
  * We could always add a little dictionary of `spec`: `human readable spec` pairs, so `gh` resolves to `GitHub` etc.
  * Do we wanna get clever about parsing the spec itself or leave it as `item/item/item`?
* Should the spec be in the description instead of the title, etc. This might give us more space to work with, and then we could keep "The Binder Project" as the title.

Open to ideas here! I just realized that it might be straightforward to do so wanted to open the proof of concept here.


betatim:
I think on all pages that are related to a repo we should make the description about what it is the user will get if they click on the page, not a description of the binder project. I'd aim for `...GitHub repository <org>/<repo>...` as the blurb to identify the `spec` instead of putting the raw spec. Would need a dictionary to map provider shortcode to a "human readable" thing (or maybe we make a template sentence that can be different for all providers?).

For the pages that aren't tied to a specific repo we should put a snappy sentence about Binder. On the landing page we use "Turn a Git repo into a collection of interactive notebooks" as the snappy tag line. Should we make that snappier and then use it on the card as well?

choldgraf:
OK, the latest push adds a quick "provider: human readable text" mapping and then uses that for the social media description. That should work well enough (though perhaps we want to cap the total number of characters? maybe something to keep an eye on)

binderhub/main.py
```diff
@@ -46,6 +53,7 @@ def get(self, provider_prefix, _unescaped_spec):
             raise HTTPError(400, str(e))
 
         provider_spec = f'{provider_prefix}/{spec}'
+        social_desc = ": ".join([provider_prefix, SPEC_NAMES[provider_prefix]])
```
betatim: This will render to something like `gh: GitHub repository` no? Is that what you meant to do? I was expecting it to render to something like `GitHub repository: <org>/<repo>`

binderhub/templates/about.html
```diff
@@ -1,5 +1,9 @@
 {% extends "page.html" %}
 
+{% block head %}
+<meta property="og:description" content="Reproducible, sharable, open, interactive computing environments.">
```
betatim: Not a huge fan of the "open" here. Do we need it? The previous version had a better "feel" or rhythm to it when reading. Because the rhythm isn't so nice it makes me as a reader think we are just stuffing adjectives in for SEO purposes or so we can tick of buzzwords.

Another point is that I think outside of the bubble of open science not soo many people know what to do with "open" as a description. What is an open computing environment? Anyone can see what I am doing? I think a word that is better is "public" (as in only public repos) but that is confusing too because the env you get isn't public, it is sourced from a public repo.

IMHO the goal should be to let the user know what will happen to them if they click the link (where will I get taken?) and entice them to click it. We don't need to explain what Binder is, people will look that up if they care to know.

This is why I'd go with the three adjectives version.

binderhub/main.py
choldgraf: oops, yep, will fix

binderhub/templates/about.html
choldgraf: I think three adjectives sounds better too. The reason I'm trying to add "open" is because I see it as one of the main distinguishing factors compared w/ other cloud services. E.g., it'd be somewhat trivial for Colab to add some kind of "share an environment" feature. Where does that leave Binder? I was trying to highlight things that are uniquely Binder Project.

That said, I think it's awkward enough and we don't have consensus enough that it shouldn't be in this PR

choldgraf:
@betatim see latest commit, I think I've addressed your two comments

binderhub/templates/about.html
willingc: While "open" is a convenient term, it can have multiple interpretations: open, pseudo-open, open requirements, etc. Some of the distinguishing factors are transparency and inspection/introspection of builds and dependencies. It would be great to at some point break down a reproducible build into the key blocks where inspection is important (requirements, container build, versions).

willingc: Which is my long winded way of saying that we need to capture and publicize the distinguishing factors. Perhaps not in the title but yes keeping focus on it for those who will care.

choldgraf: @willingc I think that's a good point - it actually reminds me [of a recent blog post from tal yarkoni](http://www.talyarkoni.org/blog/2019/07/13/i-hate-open-science/) that you might like if you haven't seen it yet. I agree that we can and should try to dig down to the specifics and why they're useful in BinderHub! Happy to iterate on that in another PR

choldgraf:
@betatim if you give this a merge I'll be around the rest of the afternoon to spot-check any errors that might come up once it's deployed (since this one is hard to test w/o a public binderhub)

binderhub/templates/about.html
betatim: Good points. Investing effort into explaining why Binder is the way it is, why we think that this is a good trade off, etc is a good idea. I keep thinking back to "Binder, so advanced it might contain magic" (advance tech being indistinguishable from magic) and the fact that most magic tricks are actually very simple to understand once they are explained to you. So I am pondering how we can explain Binder like that :) "Binder performs a magic trick" (building your env from nothing), let me show you how it does it. See, there is no magic here, you could do it yourself! *ramblingrambling*

binderhub/main.py
```diff
@@ -46,6 +53,7 @@ def get(self, provider_prefix, _unescaped_spec):
             raise HTTPError(400, str(e))
 
         provider_spec = f'{provider_prefix}/{spec}'
+        social_desc = f"{SPEC_NAMES[provider_prefix]}: {spec}"
```
betatim: I'll merge this now but I think we need to chop off the first part of the `spec` so that we get "GitHub repository: <org>/<repo>" instead of what the current code will give us: "GitHub repository: gh/<org>/<repo>". At least that is what my neural network python parser says ;)

betatim: ☝️ @choldgraf

choldgraf: I don't *think* that's true (at least, I think that the `provider_spec` is the thing that includes `gh`, while `spec` should just be the stuff after, but I just deployed it to mybinder.org and let's see :-)

betatim: From what we see on twitter: you were right :D

PR #913, opened 2019-08-03
updating social tweaks

choldgraf:
This does a couple things for the social images:

* Makes it so that we have full social meta tags on both the index and loading templates, no more shared tags
* index page uses long binder logo (so if people only link to `mybinder.org` and that's it, you'll see the big logo
* loading page now loads a compact image on twitter (not sure how it behaves on facebook but that's probably way less often). The title is the spec, the description is "binder project" general stuff.

Curious what @betatim thinks about it - we can iterate a bit, this is just a first pass


betatim:
Merged because it is tricky to imagine how it will look and easy to see if we deploy it.

I don't know enough about cards and FB. Is it possible to define several "cards" and each page (FB, twitter, etc) will use the "last one it understands"?

betatim:
Should we copy what GitHub does? This is a screenshot of the `<head>` on https://github.com/jupyterhub/binderhub

Screen Shot 2019-08-03 at 16 12 01

they seem to define one image for twitter and then again using the opengraph method. Maybe that is how you get different sized images for twitter (small/compact) and FB (large)?

choldgraf:
I do think we can define a twitter-specific image...that should be pretty straightforward. Shall we make that a PR?

choldgraf:
(fyi it'll be later today or tomorrow for me to implement this, I'm in seattle!)

PR #916, opened 2019-08-07
Add extraArgs to dind

### LINKED ISSUE #914: Allow further options in dind
The binderhub helm chart (when dind enabled) creates pods with fixed arguments:

```
        args:
          - dockerd
          - --storage-driver={{ .Values.dind.storageDriver }}
          - -H unix://{{ .Values.dind.hostSocketDir }}/docker.sock
```

For some deployments it may be necessary to add/change some of the options. Specifically in my case I need to specify a --mtu option for setting a smaller than the default docker MTU. It would be good to be able to specify extra options there.

enolfc:
Allows to specify extra arguments to the docker daemon for dind. The list of args will be added to the `arguments` of the container without further processing.

Fixes #914

### DIFF
--- a/helm-chart/binderhub/templates/dind/daemonset.yaml
+++ b/helm-chart/binderhub/templates/dind/daemonset.yaml
@@ -41,6 +41,9 @@ spec:
           - dockerd
           - --storage-driver={{ .Values.dind.storageDriver }}
           - -H unix://{{ .Values.dind.hostSocketDir }}/docker.sock
+          {{- if .Values.dind.daemonset.extraArgs }}
+{{ toYaml .Values.dind.daemonset.extraArgs | indent 10 }}
+          {{- end }}
         securityContext:
           privileged: true
         volumeMounts:
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -131,6 +131,8 @@ dind:
     image:
       name: docker
       tag: 18.09.2-dind
+    # Additional command line arguments to pass to dockerd
+    extraArgs: []
   storageDriver: overlay2
   resources: {}
   hostSocketDir: /var/run/dind


helm-chart/binderhub/templates/dind/daemonset.yaml
```diff
@@ -41,6 +41,9 @@ spec:
           - dockerd
           - --storage-driver={{ .Values.dind.storageDriver }}
           - -H unix://{{ .Values.dind.hostSocketDir }}/docker.sock
+          {{- if .Values.dind.daemonset.extraArgs }}
+{{ toYaml .Values.dind.daemonset.extraArgs | indent 10 }}
```
betatim: Should the indent by 10 or 12? Or does it depend on if we use `{{- if .Values.dind.daemonset.extraArgs }}` vs `{{ if .Values.dind.daemonset.extraArgs -}}` (used above on L31). I can't hold all the templating rules in my head :-( maybe the easiest is to run the chart and double check how it is rendered.

enolfc: I was also struggling to understand what's the right indent, but I did run the chart to check:

```
helm template \
   --set jupyterhub.proxy.secretToken="$(openssl rand -hex 32)"  \
   --set jupyterhub.hub.services.binder.apiToken=$(openssl rand -hex 32) \
   --set dind.enabled=true \
   --set 'dind.daemonset.extraArgs={--mtu=1234,--arg2,--arg3=foobar}'  \
   binderhub  | grep -A 5 dockerd
          - dockerd
          - --storage-driver=overlay2
          - -H unix:///var/run/dind/docker.sock
          - --mtu=1234
          - --arg2
          - --arg3=foobar
```

enolfc: btw, if using the `{{ if -}}` it will break the indent as the first line will already be indented so it will show up as:

```
                     - --mtu=1234
          - --arg2
```

betatim: Cool! And thanks for double checking. Templating languages seem to be non intuitive in 90% of cases ;)

betatim:
Merged and congratulations on your first code contribution!! 🎉

enolfc:
Thanks! it's a pleasure to contribute :)

PR #920, opened 2019-08-08
Add AWS ECR support

### LINKED ISSUE #705: AWS ECR registry for BinderHub deployment
I am trying to use ECR as the docker image registry for a BinderHub deployment. I was looking at the different settings in the helm chart and I am not sure how they map from the GCR examples provided in the docs to ECR. 

When explicitly pushing images to a registry, this is the command I use for ECR:
```
docker push 1234567890.dkr.ecr.region.amazonaws.com/repo:tag
```

How to translate this into the various components requested in the values file. Specifically the following in the values that are being set in `configmap.yaml`.
```
registry:
  enabled: true
  prefix: binderhub-local/
  host: https://gcr.io
  authHost:
  authTokenUrl: https://gcr.io/v2/token?service=gcr.io
  username: _json_key
  password:
```
For authentication, the nodes in our Kubernetes cluster are allowed to pull from ECR as they have IAM roles configured. Can I leave the authentication options blank here.

Really appreciate the guidance.

chicocvenancio:
Starting the PR, I'll push a few more commits before it is ready for review. 

FIX #705


betatim:
Conflicts need resolving and the test failures checking out. I restarted the build and it failed again so I think this might be a real failure.

Is there a way to avoid having to expand the RBAC rights? If we can avoid giving more permissions that would be a "good thing (tm)". So I thought I'd raise it to see what we can come up with.

What does it mean for people upgrading if we change the RBAC permissions? I have a vague memory that there is a difference between creating and editing permission.

chicocvenancio:
I'll try to make the config more robust, at this point it is expecting AWS specific config to be there, and since it is not for the CI tests it is failing.

> Is there a way to avoid having to expand the RBAC rights? If we can avoid giving more permissions that would be a "good thing (tm)". So I thought I'd raise it to see what we can come up with.
I can't think of a straightforward way of dropping added rights entirely. ECR login tokens expire in 12 hours, right now the tokens are set both in `DockerRegistry.password` and as a secret `binder-push-secret`, both need to be set for pushing and pulling images to work. One option to restrict RBAC a little further is to specify the `binder-push-secret` in the rule, so it only applies to it. 

> What does it mean for people upgrading if we change the RBAC permissions? I have a vague memory that there is a difference between creating and editing permission.
We're using Pulumi to deploy Binderhub, so I'm not completely sure if helm+tiller does something weird here. I wouldn't expect it to, but I can take a look and test it out.

ivan-gomes:
@chicocvenancio anything I can help with to get this to the finish line?

meeseeksmachine:

https://discourse.jupyter.org/t/binder-jupyterhub-activity-round-up-week-1/2704/1

consideRatio:
Hey @chicocvenancio, @ivan-gomes and @betatim! I'm going through various PRs hoping to help them resolve one way or another or perhaps get an updated description about the current status of things.

I'm closing this in favor of the PR building on this with additional commits: #1055.

PR #923, opened 2019-08-11
[MRG] Automatically open the log while building a repository

betatim:
Automatically open the log if we are building a repo. The goal is to make it obvious to people visiting a Binder for the first time that stuff is happening and progress is made when they have to wait a long time.


betatim:
cc @choldgraf

PR #924, opened 2019-08-11
[MRG] Add more loading messages

betatim:
Adds a few more "waiting for launch" messages.

I tried to add a full stop at the end of each message as I think it looks weird to not end a sentence

cc @choldgraf

### DIFF
--- a/binderhub/static/js/src/loading.js
+++ b/binderhub/static/js/src/loading.js
@@ -1,12 +1,16 @@
 // Cycle through helpful messages on the loading page
 const help_messages = [
-    'New to Binder? Check out the <a target="_blank" href="https://mybinder.readthedocs.io/en/latest/">Binder Documentation</a> for more information',
-    'If a Binder takes a long time to launch, it is usually because Binder needs to pull the environment onto a computer for the first time',
-    'The tool that powers this page is called <a target="_blank" href="https://binderhub.readthedocs.io">BinderHub</a>. It is an open source tool that anybody can deploy',
-    'The Binder team has <a target="_blank" href="https://mybinder-sre.readthedocs.io/en/latest/">a site reliability guide </a> that talks about what it is like to run a BinderHub',
-    'You can connect with the Binder community in the <a target="_blank" href="https://discourse.jupyter.org/c/binder">Jupyter community forum</a>',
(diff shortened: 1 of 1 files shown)


binderhub/static/js/src/loading.js
```diff
@@ -1,12 +1,16 @@
...
-    'You can connect with the Binder community in the <a target="_blank" href="https://discourse.jupyter.org/c/binder">Jupyter community forum</a>',
-    'Your launch may take longer the first few times that this environment has been run. This is because the machine needs to load your environment.',
-    'You can learn more about building your own Binder repositories in <a target="_blank" href="https://docs.mybinder.org">the Binder community documentation</a>.'
+    'New to Binder? Check out the <a target="_blank" href="https://mybinder.readthedocs.io/en/latest/">Binder Documentation</a> for more information.',
+    'You can learn more about building your own Binder repositories in <a target="_blank" href="https://docs.mybinder.org">the Binder community documentation</a>.',
+    'We use the <a target="_blank" href="https://repo2docker.readthedocs.io/">repo2docker</a> tool to automatically build the environment in which to run your code.',
+    'Take a look at the <a target="_blank" href="https://repo2docker.readthedocs.io/en/latest/config_files.html">full list of configuration files supported by repo2docker.',
```
choldgraf:
```suggestion
    'Take a look at the <a target="_blank" href="https://repo2docker.readthedocs.io/en/latest/config_files.html">full list of configuration files supported by repo2docker.</a>',
```

choldgraf:
lgtm other than that closed link tag. thanks! i wonder if at some point we should add fun ones in there...

choldgraf:
for some reason github says I dont have push permissions on the inline commit, so once you add it, then Ill merge

betatim:
> for some reason github says I dont have push permissions on the inline commit, so once you add it, then Ill merge

Thanks for spotting this! The same happens to me as well with suggestions and not being allowed to commit them. But not on all and every repo so I haven't yet been able to work out what we are doing wrong. My guess was that this happens when the "Allow edits from maintainers" box wasn't ticked but for this PR it is ticked ... so I am fresh outta ideas :-/

PR #925, opened 2019-08-12
adding jupyterlab file paths for preview

### LINKED ISSUE #645: Preview not working for Jupyter Lab link
Found the following yesterday:

Opening a repo in binder using the notebook using this link 
https://mybinder.org/v2/gh/fangohr/introduction-to-python-for-computational-science-and-engineering/master?filepath=index.ipynb 

works fine: a preview of the index.ipynb file is displayed, while the container is provisioned. See jupyter-notebook-file-preview-works.png (attached).

jupyter-notebook-file-preview-works

Opening the same repo in binder using Jupyter Lab also works, but the preview that is displayed is for the table of contents, not the file specified: 
https://mybinder.org/v2/gh/fangohr/introduction-to-python-for-computational-science-and-engineering/master?urlpath=lab/tree/index.ipynb
See jupyterlab-file-preview-not-working.png (attached).

jupyterlab-file-preview-not-working

If it helps: both links are available on [this page](https://github.com/fangohr/introduction-to-python-for-computational-science-and-engineering/blob/master/Readme.md) which tries to give users a choice to either use classic notebook or JupyterLab. 

There is another recent issue that may be related (https://github.com/jupyterhub/binderhub/issues/644) but I found the link not so clear, so link that I created a new one).

choldgraf:
closes #645 

I think that this should add support for JupyterLab paths in the nbviewer preview.

It is pretty simple, basically just adds a few lines to:

* Check whether `urlpath` is provided in the URL
* If so, check whether `lab/tree/` exists in the `urlpath`
* If so, set `filepath` to whatever is on the right side of `lab/tree/`.

I think this should work for at least vanilla jupyterlab paths. It might break in unpredicted ways if people start using jupyterlab but also other kinds of URL patterns, but we can beef up this function when we get to those situations.

### DIFF
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -64,8 +64,8 @@ class ParameterizedMainHandler(BaseHandler):
            
             # Check if we have a JupyterLab + file path, if so then use it for the filepath
             urlpath = self.get_argument('urlpath', '').lstrip('/')
-            if 'lab/tree/' in urlpath:
-                filepath = urlpath.split('lab/tree/')[-1]
+            if urlpath.startswith("lab") and "/tree/" in urlpath:
+                filepath = urlpath.split('tree/', 1)[-1]
             
             blob_or_tree = 'blob' if filepath else 'tree'
             nbviewer_url = f'{nbviewer_url}/{org}/{repo_name}/{blob_or_tree}/{ref}/{filepath}'
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -61,6 +61,12 @@ class ParameterizedMainHandler(BaseHandler):
             org, repo_name, ref = spec.split('/', 2)
             # NOTE: tornado unquotes query arguments too -> notebooks%2Findex.ipynb becomes notebooks/index.ipynb
             filepath = self.get_argument('filepath', '').lstrip('/')
+           
+            # Check if we have a JupyterLab + file path, if so then use it for the filepath
+            urlpath = self.get_argument('urlpath', '').lstrip('/')
+            if 'lab/tree/' in urlpath:
+                filepath = urlpath.split('lab/tree/')[-1]
+            
             blob_or_tree = 'blob' if filepath else 'tree'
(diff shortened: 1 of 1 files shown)


binderhub/main.py
```diff
@@ -61,6 +61,12 @@ def get(self, provider_prefix, _unescaped_spec):
             org, repo_name, ref = spec.split('/', 2)
             # NOTE: tornado unquotes query arguments too -> notebooks%2Findex.ipynb becomes notebooks/index.ipynb
             filepath = self.get_argument('filepath', '').lstrip('/')
+           
+            # Check if we have a JupyterLab + file path, if so then use it for the filepath
+            urlpath = self.get_argument('urlpath', '').lstrip('/')
+            if 'lab/tree/' in urlpath:
```
betatim:
```suggestion
            if urlpath.startswith("lab/tree"):
```
would this be an improvement/less prone to matching accidentally?

choldgraf: I was originally doing it this way to allow for [workspaces in the url too](https://jupyterlab.readthedocs.io/en/stable/user/urls.html#combining-url-functions), but upon looking it seems like that'd actually be a more complicated rule. What if instead we did something like:

```
if urlpath.startswith("lab") and `/tree/` in urlpath:
```

and then do the split on `/tree/` and take whatever is to the right of the first instance of `/tree/`?

choldgraf: this is in the latest push, it seemed to work for me locally (both regular lab URLs and workspace URLs)

betatim: What different kinds of URLs could we get here? I am not really familiar with what Jupyter Lab does/what workspaces are :-/

choldgraf: See the link I sent above, that's what I was using to figure out what urls to expect

betatim:
Merge? If you've tested it locally I'd be happy to press merge and see what happens. Investigating and learning about URL schemes and how they could go wrong seems like a high effort thing compared to the worst that could happen if we just deploy it.

choldgraf:
I'm +1 on merging, the only thing that can go wrong is that it doesn't work, which would be our current situation :-)

Issue #919, opened 2019-08-08
helm chart binderhub-0.2.0-409f200 contains bhub image with tag 0.2.0-9174211


bitnik:
Hi, I just deployed binderhub-0.2.0-409f200 on GESIS Binder, but I realised that it uses binderhub image with tag 0.2.0-9174211 instead of 0.2.0-409f200. I think something went wrong when https://github.com/jupyterhub/binderhub/pull/838 is merged or am I wrong? Actually it is only documentation update. So it shouldn't create a new chart actually?

If you point me the code where chart generation is done, I can also look at what is wrong.

betatim:
We use https://github.com/jupyterhub/chartpress to make the charts and it is called in https://github.com/jupyterhub/binderhub/blob/10a3deeab94bdf584094bdcdb674f5c59a14d38b/ci/test-helm#L47 For some runs `$PUSH` contains the arguments that ask `chartpress` to publish stuff.

bitnik:
I just ran `chartpress` locally

```bash
git checkout 9174211
chartpress
```

Output
```bash
$> git log -n 1 --pretty=format:%h -- . ../setup.py ../binderhub
$> git log -n 1 --pretty=format:%h -- images/image-cleaner chartpress.yaml
Skipping build for jupyterhub/k8s-image-cleaner:0.2.0-c56fb9e, it already exists
$> git log -n 1 --pretty=format:%h -- ../setup.py ../binderhub .. chartpress.yaml
Skipping build for jupyterhub/k8s-binderhub:0.2.0-9174211, it already exists
Updating binderhub/values.yaml: imageCleaner.image: {'repository': 'jupyterhub/k8s-image-cleaner', 'tag': '0.2.0-c56fb9e'}
Updating binderhub/values.yaml: image: {'repository': 'jupyterhub/k8s-binderhub', 'tag': '0.2.0-9174211'}
```

For helm chart tag it runs `git log -n 1 --pretty=format:%h -- . ../setup.py ../binderhub` which outputs `409f200` on commit `9174211` because last change is done in doc folder. But for image tag it runs `git log -n 1 --pretty=format:%h -- ../setup.py ../binderhub .. chartpress.yaml` with extra paths `.. chartpress.yaml` and it outputs `9174211`. That is the reason why image and helm chart tags are different. Those extra paths are added in `build_imagesmethod (https://github.com/jupyterhub/chartpress/blob/master/chartpress.py#L221-L224).

[...]

PR #927, opened 2019-08-14
determine chart version with last change in project root

bitnik:
for https://github.com/jupyterhub/binderhub/issues/919


betatim:
Let's try it!

PR #929, opened 2019-08-16
[MRG] Try to resolve an image name to determine registry health

betatim:
A docker registry doesn't have an explicit health endpoint. The recommendation I could find in the docker registry repo was to use the root end point and if it returns a 200 or 401 the registry is healthy. However for the OVH registry we frequently have the problem that we think the registry is healthy but we can't resolve images.


betatim:
Self merge to see this in action.

PR #932, opened 2019-08-16
[MRG] Fix docker registry health check

betatim:
The image name being passed was wrong and the check didn't return `True` when it succeeded.

Also added a throttle so that we only check every 15s

### DIFF
--- a/binderhub/health.py
+++ b/binderhub/health.py
@@ -119,15 +119,19 @@ class HealthHandler(BaseHandler):
         return True
 
     @false_if_raises
+    @at_most_every(interval=10)
     @retry
     async def check_docker_registry(self):
         """Check docker registry health"""
+        app_log.info("Checkinng registry status")
         registry = self.settings["registry"]
         # we are only interested in getting a response from the registry, we
         # don't care if the image actually exists or not
+        image_name = self.settings["image_prefix"] + "some-image-name:12345"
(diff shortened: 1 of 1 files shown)


binderhub/health.py
```diff
@@ -119,15 +119,19 @@ def initialize(self, hub_url=None):
         return True
 
     @false_if_raises
+    @at_most_every(interval=10)
     @retry
     async def check_docker_registry(self):
         """Check docker registry health"""
+        app_log.info("Checkinng registry status")
```
choldgraf:
```suggestion
        app_log.info("Checking registry status")
```

binderhub/health.py
```diff
@@ -119,15 +119,19 @@ def initialize(self, hub_url=None):
         return True
 
     @false_if_raises
+    @at_most_every(interval=10)
```
choldgraf:
```suggestion
    @at_most_every(interval=15)
```

(assuming since you said "15 seconds" in your comment

binderhub/health.py
betatim: I need to take my laptop to the Apple store to get the keyboard fixed. Even the software fix to stop repeated keys ain't working well enough any more :-/

betatim:
Thanks for catching those. Merged and will deploy it once its built.

Issue #816, opened 2019-03-31
Allow for disabling of nbviewer 


stklik:
Currently, all notebooks from Github show a preview using nbviewer.
In some situations this causes a "404: Not Found" message, even though everythign is correct (e.g. when trying to launch a file that is `wrapped` inside a Docker image).

For example: https://mybinder.org/v2/gh/crestdsl/crestdsl-demo-binder/master?filepath=GettingStarted.ipynb launches a notebook that is not in the repo, but in the Docker image that is launched by the repo.

I believe that providing an url-argument e.g. `&preview=false` could be useful in these situations to disable the preview (so users aren't confused).

The concerned code places (as far as I see) are the `ParameterizedMainHandler` in [main.py](https://github.com/jupyterhub/binderhub/blob/01d643d7ae98f560f08f4fb5e43bfbf34f159863/binderhub/main.py#L47), where it would be necessary to extract the argument and pass it to the `loading.html` template. Code to add should be something like `preview=self.get_argument('preview', True)`.

The other part is in the template itself (e.g. near [Line 27](https://github.com/jupyterhub/binderhub/blob/01d643d7ae98f560f08f4fb5e43bfbf34f159863/binderhub/templates/loading.html#L27) ) where a `{% if preview %}`
can be added.  (Line 30 already contains a condition that could also be easily extended...)

This issue is to allow feedback / opposition before starting implementing. Thus:
   1. What do you think?
   2. Is my current though process correct? or is my "very simple" solution not as simple after all?

PS: lacking time right now, so PR will come in a few weeks earliest.

betatim:
I think there are two ways this could go:
* we add a URL parameter like this (see #712 for naming)
* or we make our preview smarter so that it hides itself if it can't show something useful

I don't know how much harder the second one is but it would be cool as it doesn't rely on the creator of the link knowing what they are doing and keeps the URLs shorter.

Is it feasible for BinderHub to make a `HEAD` request to nbviewer to find out if we would show a 404 or other error page and then decide on hiding the preview?

stklik:
> I think there are two ways this could go:
> 
> * we add a URL parameter like this (see #712 for naming)
> * or we make our preview smarter so that it hides itself if it can't show something useful

Does one exclude the other? An URL param would certainly allow people to control the preview (i.e. even if there was something to useful to show?)

It appears to me that the former is the MVP solution. The more sophisticated one is the "nice-to-have".

As for 
[...]

PR #934, opened 2019-08-20
[MRG] Check if nbviewer URL would show an error

### LINKED ISSUE #922: 404 page shown during build when pointing to a repo
I have a binder repo [here](https://github.com/davidanthoff/teaching-2019-aere-workshop/tree/v1.0.1). When people click on the "Launch binder" button, they see this:

image

Which is really not ideal because then they think that something went wrong and that there is some bug, when in fact everything is fine.

I think the reason it looks that way is that I'm actually not linking to a specific notebook, but instead just to the whole repo. I think in that case it would be better to not show a 404 link during build, but something that looks less scary and that doesn't look like something went wrong.

EDIT: This was originally posted over in the repo2docker repo as https://github.com/jupyter/repo2docker/issues/761, but @betatim asked me to move it here. He also pointed out a workaround over there, which I have since implemented in my repo. So the link in this issue here now points to an old version of the repo that has the old type of link. While the workaround is great, I think it would be great if things worked without it.

betatim:
If the nbviewer widget would show a 404 or other error we don't want to
display it. We check the status before rendering the template by making
a HEAD request to nbviewer, if the status code is >=400 we don't show
the preview.

closes #922


betatim:
@lesteve I only saw [your comment](https://github.com/jupyterhub/binderhub/issues/816#issuecomment-522847777) after making this PR. Sorry :-/

This implements your idea using the tornado HTTP client instead of urllib. The code has to work async/non-blocking and I don't think urllib supports that (yet).

It isn't quite the feature requested in #816 because it doesn't give the choice of disabling or not to the user. However this is a quick win, we can still add the "disable even if the preview works" options if we decide we want that.

lesteve:
Very nice to see this, thanks a lot!

No worries about superseding my naive attempt. I got binder setup locally using minikube and learned a few useful things along the way!

lesteve:
Very nice to see this merged! Naive question: I assume I have to monitor the commits in https://github.com/jupyterhub/mybinder.org-deploy in particular the changes in [values.yaml](https://github.com/jupyterhub/mybinder.org-deploy/blob/master/mybinder/values.yaml) (I have read the blog post https://blog.jupyter.org/automating-mybinder-org-dependency-upgrades-in-10-steps-bb5e38542059).

Are there additional steps I need to be aware of until this change reach mybinder.org.

Note: I am not in a hurry at all, I am asking out of curiosity and to have some very rough estimate when I can give this a whirl on mybinder.org.

betatim:
We have a bot (henchbot, just referenced this PR) that opens a PR on the deploy repo. The bot runs about once an hour and usually(depends on where people are, if it is night, weekend, etc) someone clicks merge an hour or two after the bot opened the deploy PR. Then it is live!

lesteve:
OK that matches what I understood from the blog post, thanks for confirming this!

Issue #946, opened 2019-09-08
Optimizing scheduling of build pods


consideRatio:
This issue is about how to optimize the scheduling of BinderHub specific _Build Pods_ (BPs).

Out of scope of this issue is the discussion on how to schedule the user pods, which could be done with *image locality* (`ImageLocalityPriority` configuration) in mind.

# Scheduling goals
1. **Historic repo build node locality (Historic locality)**
  It is believed that often a builder pod rebuilding a repo would work faster if it had previously built a repository even though the repo has changed, for example in the README.md file of the repo. So, we want to schedule build pods on nodes where it has previously been scheduled if possible.
2. **Not blocking scale down in autoscaling clusters**
  We would also like to avoid schedule pods on nodes that may want to scale down.

### What is our desired scheduling practice?
We need to answer how we actually want the build-pods to schedule, it is not a obvious way to do it is typically hard to optimize both for performance and auto-scaling viability for example.

### Boilerplate desired scheduling practice
I'll now provide a boilerplate idea to start out from on how to schedule the build pods.

1. We make the scheduler informed about if this is an autoscaling cluster or not. If it is, we greatly reduce the score of almost empty nodes that we could scale down.
2. We schedule the build pod on nodes with history of building this.

# Technical solution implementation details
We must utilize a non-default scheduler. We could use the default [kube-scheduler binary](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/) and customize its behavior through configuration, or we could make our own. I think we add too much complexity if we are to make our own though, even though [making your own scheduler is certainly possible](https://www.youtube.com/watch?v=q8MFm2jwXpA).

- We utilize a custom scheduler, but like [z2jh's scheduler](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/master/jupyterhub/templates/scheduling/user-scheduler) we deploy a [official kube-scheduler binary](gcr.io/google_containers/kube-scheduler-amd64) with a customized configuration that we can reference from the build pods specification using the `spec.schedulerName` field.

   > \[`spec.schedulerName`\] If specified, the pod will be dispatched by specified scheduler. If not specified, the pod will be dispatched by default scheduler. --- [Kubernetes PodSpec documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#podspec-v1-core).

[...]

PR #949, opened 2019-09-12
[MRG] add "sticky builds" functionality

betatim:
The ranking is computed using the Rendez-vous hashing algorithm.

In the abstract this can be used for all sorts of stateless ranking, however this PR is part of https://github.com/jupyterhub/binderhub/issues/946#issuecomment-529778393 and provides a method for ranking possible nodes to schedule build pods on.

The goal is to eventually have a setup where `rendezvous_rank()` is called just before submitting the build pod spec to compute a preferred node for the pod:

```python
ranking = rendezvous_rank(['node1', 'node2', 'node3'], 'https://github.com/org/repo')
# now attach ranking[0][1] as the preferred node affinity to the build pod
```

Let me know what you think of building this in small PRs. I'd propose to review and merge this without having to build out the rest.

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -568,6 +568,7 @@ class BinderHub(Application):
             "build_image": self.build_image,
             'build_node_selector': self.build_node_selector,
             'build_pool': self.build_pool,
+            "sticky_builds": self.sticky_builds,
             'log_tail_lines': self.log_tail_lines,
             'pod_quota': self.pod_quota,
             'per_repo_quota': self.per_repo_quota,
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -196,6 +196,19 @@ class BinderHub(Application):
         config=True,
     )
 
+    sticky_builds = Bool(
+        False,
+        help="""
+        Attempt to assign builds for the same repository to the same node.
+
+        In order to speed up re-builds of a repository all its builds will
+        be assigned to the same node in the cluster.
+
+        Note: This feature only works if you also enable docker-in-docker support.
+        """,
+        config=True,
+    )
+
     use_registry = Bool(
         True,
         help="""
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -12,6 +12,8 @@ from kubernetes import client, watch
 from tornado.ioloop import IOLoop
 from tornado.log import app_log
 
+from .utils import rendezvous_rank
+
 
 class Build:
     """Represents a build of a git repository into a docker image.
@@ -36,7 +38,7 @@ class Build:
     """
     def __init__(self, q, api, name, namespace, repo_url, ref, git_credentials, build_image,
                  image_name, push_secret, memory_limit, docker_host, node_selector,
-                 appendix='', log_tail_lines=100):
+                 appendix='', log_tail_lines=100, sticky_builds=False):
         self.q = q
         self.api = api
         self.repo_url = repo_url
@@ -55,6 +57,10 @@ class Build:
 
         self.stop_event = threading.Event()
         self.git_credentials = git_credentials
+
+        self.sticky_builds = sticky_builds
+
+        self._component_label = "binderhub-build"
 
     def get_cmd(self):
         """Get the cmd to run to build the image"""
@@ -144,8 +150,59 @@ class Build:
         """Put the current action item into the queue for execution."""
         self.main_loop.add_callback(self.q.put, {'kind': kind, 'payload': obj})
 
+    def get_affinity(self):
+        """Determine the affinity term for the build pod.
+
+        There are a total of two affinity strategies, which one is used depends
+        on how the BinderHub is configured.
+
+        In the default setup the affinity of each build pod is an "anti-affinity"
+        which causes pods to prefer to schedule on separate nodes.
+
+        In a setup with docker-in-docker pods for a particular repository
+        prefer to schedule on the same node in order to reuse the build cache
+        of previous builds.
+        """
+        dind_pods = self.api.list_namespaced_pod(
+            self.namespace,
+            label_selector="component=dind,app=binder",
+        )
+        if self.sticky_builds and dind_pods:
+            nodes = dict((item.spec.node_name, item.spec)
+                         for item in dind_pods.items)
+            ranked_nodes = rendezvous_rank(list(nodes.keys), self.repo_url)
+            best_node_name = ranked_nodes[0][1]
+            # we reuse the affinity term of the DIND pod instead of
+            # constructing our own. But we want a soft preference instead of
+            # a hard preference in case that node is full or has gone away
+            # in the meantime
+            affinity = nodes[best_node_name].spec.affinity
+            affinity.node_affinity.preferred_during_scheduling_ignored_during_execution = affinity.node_affinity.required_during_scheduling_ignored_during_execution
+            affinity.node_affinity.required_during_scheduling_ignored_during_execution = None
+
+        else:
+            affinity = client.V1Affinity(
+                pod_anti_affinity=client.V1PodAntiAffinity(
+                    preferred_during_scheduling_ignored_during_execution=[
+                        client.V1WeightedPodAffinityTerm(
+                            weight=100,
+                            pod_affinity_term=client.V1PodAffinityTerm(
+                                topology_key="kubernetes.io/hostname",
+                                label_selector=client.V1LabelSelector(
+                                    match_labels=dict(
+                                        component=self._component_label
+                                    )
+                                )
+                            )
+                        )
+                    ]
+                )
+            )
+
+        return affinity
+
     def submit(self):
-        """Submit a image spec to openshift's s2i and wait for completion """
+        """Submit a build pod to create the image for the repository."""
         volume_mounts = [
             client.V1VolumeMount(mount_path="/var/run/docker.sock", name="docker-socket")
         ]
@@ -166,13 +223,12 @@ class Build:
         if self.git_credentials:
             env.append(client.V1EnvVar(name='GIT_CREDENTIAL_ENV', value=self.git_credentials))
 
-        component_label = "binderhub-build"
         self.pod = client.V1Pod(
             metadata=client.V1ObjectMeta(
                 name=self.name,
                 labels={
                     "name": self.name,
-                    "component": component_label,
+                    "component": self._component_label,
                 },
                 annotations={
                     "binder-repo": self.repo_url,
@@ -211,23 +267,7 @@ class Build:
                 node_selector=self.node_selector,
                 volumes=volumes,
                 restart_policy="Never",
-                affinity=client.V1Affinity(
-                    pod_anti_affinity=client.V1PodAntiAffinity(
-                        preferred_during_scheduling_ignored_during_execution=[
-                            client.V1WeightedPodAffinityTerm(
-                                weight=100,
-                                pod_affinity_term=client.V1PodAffinityTerm(
-                                    topology_key="kubernetes.io/hostname",
-                                    label_selector=client.V1LabelSelector(
-                                        match_labels=dict(
-                                            component=component_label
-                                        )
-                                    )
-                                )
-                            )
-                        ]
-                    )
-                )
+                affinity=self.get_affinity()
             )
         )
 
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -153,32 +153,43 @@ class Build:
     def get_affinity(self):
         """Determine the affinity term for the build pod.
 
-        There are a total of two affinity strategies, which one is used depends
-        on how the BinderHub is configured.
+        There are a two affinity strategies, which one is used depends on how
+        the BinderHub is configured.
 
         In the default setup the affinity of each build pod is an "anti-affinity"
-        which causes pods to prefer to schedule on separate nodes.
+        which causes the pods to prefer to schedule on separate nodes.
 
-        In a setup with docker-in-docker pods for a particular repository
-        prefer to schedule on the same node in order to reuse the build cache
-        of previous builds.
+        In a setup with docker-in-docker enabled pods for a particular
+        repository prefer to schedule on the same node in order to reuse the
+        docker layer cache of previous builds.
         """
         dind_pods = self.api.list_namespaced_pod(
             self.namespace,
             label_selector="component=dind,app=binder",
         )
         if self.sticky_builds and dind_pods:
-            nodes = dict((item.spec.node_name, item.spec)
-                         for item in dind_pods.items)
-            ranked_nodes = rendezvous_rank(list(nodes.keys), self.repo_url)
-            best_node_name = ranked_nodes[0][1]
-            # we reuse the affinity term of the DIND pod instead of
-            # constructing our own. But we want a soft preference instead of
-            # a hard preference in case that node is full or has gone away
-            # in the meantime
-            affinity = nodes[best_node_name].spec.affinity
-            affinity.node_affinity.preferred_during_scheduling_ignored_during_execution = affinity.node_affinity.required_during_scheduling_ignored_during_execution
-            affinity.node_affinity.required_during_scheduling_ignored_during_execution = None
+            node_names = [pod.spec.node_name for pod in dind_pods.items]
+            ranked_nodes = rendezvous_rank(node_names, self.repo_url)
+            best_node_name = ranked_nodes[0]
+
+            affinity = client.V1Affinity(
+                node_affinity=client.V1NodeAffinity(
+                    preferred_during_scheduling_ignored_during_execution=[
+                        client.V1PreferredSchedulingTerm(
+                            weight=100,
+                            preference=client.V1NodeSelectorTerm(
+                                match_fields=[
+                                    client.V1NodeSelectorRequirement(
+                                        key="metadata.name",
+                                        operator="In",
+                                        values=[best_node_name],
+                                    )
+                                ]
+                            ),
+                        )
+                    ]
+                )
+            )
 
         else:
             affinity = client.V1Affinity(
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -167,6 +167,7 @@ class Build:
             self.namespace,
             label_selector="component=dind,app=binder",
         )
+
         if self.sticky_builds and dind_pods:
             node_names = [pod.spec.node_name for pod in dind_pods.items]
             ranked_nodes = rendezvous_rank(node_names, self.repo_url)
--- a/binderhub/tests/test_utils.py
+++ b/binderhub/tests/test_utils.py
@@ -6,20 +6,18 @@ def test_rendezvous_rank():
     # other buckets are removed
     key = "k1"
     first_round = utils.rendezvous_rank(["b1", "b2", "b3"], key)
-    second_round = utils.rendezvous_rank([first_round[0][1], first_round[1][1]], key)
-
-    print(first_round)
+    second_round = utils.rendezvous_rank([first_round[0], first_round[1]], key)
 
     assert first_round[0] == second_round[0], key
 
     key = "sdsdggdddddd"
     first_round = utils.rendezvous_rank(["b1", "b2", "b3"], key)
-    second_round = utils.rendezvous_rank([first_round[0][1], first_round[1][1]], key)
+    second_round = utils.rendezvous_rank([first_round[0], first_round[1]], key)
 
     assert first_round[0] == second_round[0], key
 
     key = "crazy frog is a crazy key"
     first_round = utils.rendezvous_rank(["b1", "b2", "b3"], key)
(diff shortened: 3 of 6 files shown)


binderhub/tests/test_utils.py
```diff
@@ -0,0 +1,25 @@
+from binderhub import utils
+
+
+def test_rendezvous_rank():
+    # check that a key doesn't move if its assigned bucket remains but the
+    # other buckets are removed
```
betatim: I think this isn't strictly a guarantee made by the algorithm. It only promises to move as few keys as possible. However that is quite tricky to test, ideas welcome.

binderhub/build.py
```diff
@@ -144,8 +150,59 @@ def progress(self, kind, obj):
...
+        prefer to schedule on the same node in order to reuse the build cache
+        of previous builds.
+        """
+        dind_pods = self.api.list_namespaced_pod(
+            self.namespace,
+            label_selector="component=dind,app=binder",
+        )
+        if self.sticky_builds and dind_pods:
+            nodes = dict((item.spec.node_name, item.spec)
+                         for item in dind_pods.items)
```
consideRatio: First time i see dictionary comprehension like this, and perhaps a different naming of the variable?

```python
dind_pods_node_affinity = {
    item.spec.node_name: item.spec.affinity.node_affinity
    for item in dind_pods.items()
}
```

But hmmm... I don't understand how we can reuse the dind pods affinity term? Oh!!! Is it because it is a daemonset and has an automatic hard affinity for the specific hostname? Clever! That should be clarified though!

Hmmmm but hmmm... there could be additional hard affinities set on these pods. Also, required and preferred has a different schema doesn't they? Yes, see: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#nodeaffinity-v1-core

Both will share a NodeSelectorTerm object type though. I suggest we search for the NodeSelectorTerm object within `required_during_scheduling_ignored_during_execution.nodeSelectorTerms` that references a kubernetes.io/hostname label with some value. A helper function could be in place for this. Then, we make two entries for the top two ranked nodes where the highest ranked node gets a weight of 100 and the other a weight of 50.

binderhub/app.py
```diff
@@ -196,6 +196,19 @@ def _valid_badge_base_url(self, proposal):
...
 
+    sticky_builds = Bool(
+        False,
+        help="""
+        Attempt to assign builds for the same repository to the same node.
+
+        In order to speed up re-builds of a repository all its builds will
+        be assigned to the same node in the cluster.
+
+        Note: This feature only works if you also enable docker-in-docker support.
```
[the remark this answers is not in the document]
betatim: Right now because we use the list of DIND pods to find the list of nodes. To support it without DIND pods we'd need a way to evaluate the node selector expression, which needs more access rights (listing all nodes).

Where would be a good place to document this thinking? The docs for the end user don't seem like the right place (I think it would be too much detail). Maybe in the code that implements the feature?

binderhub/utils.py
```diff
@@ -1,8 +1,37 @@
...
+    """Rank the buckets for a given key using Rendez-vous hashing
+
+    Each bucket is scored for the key and the best match is assigned the
+    highest score. The return value is a list of tuples of (score, bucket),
+    sorted in decreasing order.
+    """
+    ranking = []
+    for bucket in buckets:
+        h = blake2b_as_float(b"%s-%s" % (str(key).encode(), str(bucket).encode()))
+        score = 1 / -log(h)
```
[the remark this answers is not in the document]
betatim: Agreed. I'll remove the conversion and `log(x)`.

It came from directly implementing the algorithm shown on http://www.snia.org/sites/default/files/SDC15_presentations/dist_sys/Jason_Resch_New_Consistent_Hashings_Rev.pdf (slide 26++). I think the reason to use it is if you want to be able to assign weights, which we don't need.

binderhub/build.py
betatim: I switched to explicitly creating the affinity spec.

After your comments and digging into things I realised trying to be clever here was not worth it. Constructing the affinity spec is tedious to type once (now) but easier to understand in the future.

binderhub/tests/test_utils.py
```diff
@@ -0,0 +1,25 @@
...
+    key = "k1"
+    first_round = utils.rendezvous_rank(["b1", "b2", "b3"], key)
+    second_round = utils.rendezvous_rank([first_round[0][1], first_round[1][1]], key)
```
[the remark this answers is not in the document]
betatim: I don't think that keys have to remain associated to the same bucket if a new one is added.

Imagine you have three buckets that split the space like this: `--->b1------->b2------>b3--`. Let's assume that the key `k` is placed like this `--->b1--k---->b2------>b3--`. This means it will get placed in `b2`. Now we add `b4` and it just happens to get placed between `k` and `b2` like so `--->b1--k-->b4->b2------>b3--`. Now `k` will be placed on `b4`, not `b2` any more.

I think it makes sense that this is true because if it wasn't then no keys would ever get assigned to a new bucket when one is added. Starting with a situation with only one bucket `b1`, then any possible key `kX` would get assigned to `b1`. Now we add `b2`. Now we want about half of all the keys `kX` to be assigned to `b2` and half to `b1`. So some keys have to move.

binderhub/utils.py
```diff
@@ -1,8 +1,37 @@
 """Miscellaneous utilities"""
 from collections import OrderedDict
+from hashlib import blake2b
+from math import log
+
 from traitlets import Integer, TraitError
 
 
+def blake2b_as_float(b):
+    """Compute Blake2 digest and convert it to a float.
```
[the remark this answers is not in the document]
betatim: Described that the input is bytes and renamed it to `blake2b_as_int`. An alternative name could be `blake2b_digest_as_int` but that feels a bit long :-/

binderhub/utils.py
```diff
@@ -1,8 +1,37 @@
...
+
+    Use this to create a float [0, 1) based on the Blake2 hash function.
```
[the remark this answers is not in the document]
betatim: I don't think we need to link to the docs for things which are in the standard library of Python. If we start doing this then in `utils.py` along we'd have several links to the docs :-/

binderhub/app.py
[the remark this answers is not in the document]
consideRatio: I don't mind seeing it documented here, but if not, we could open an issue to represent this shortcoming could be overcome.

binderhub/utils.py
```diff
@@ -1,8 +1,37 @@
...
+def rendezvous_rank(buckets, key):
+    """Rank the buckets for a given key using Rendez-vous hashing
+
+    Each bucket is scored for the key and the best match is assigned the
+    highest score. The return value is a list of tuples of (score, bucket),
+    sorted in decreasing order.
+    """
+    ranking = []
+    for bucket in buckets:
+        h = blake2b_as_float(b"%s-%s" % (str(key).encode(), str(bucket).encode()))
```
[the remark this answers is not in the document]
betatim: This is still missing here. I was hoping we could get this merged and tested in production before adding that.

The reason to have multiple "entries" for each bucket is to smooth out the clumpyness of random numbers as well as spread out assignments when a bucket is removed.

With (say) four buckets there is a high chance that one bucket receives a lot more than 25% of keys due to "clumpyness".

If we remove one of the four buckets then all of its keys will end up being assigned to one of the three remaining buckets, which might overload it.

The fix for both these problems is to represent each bucket with (say) 100 "pseudo buckets".

Code for the ranking would look something like:
```python
def rank(buckets, key, N=100):
    ranking = []
    for bucket in buckets:
        for i in range(N):
            score = blake2b_as_int(b"%s-%s-%s" % (str(key).encode(), str(bucket).encode(), str(i).encode()))
            ranking.append((score, bucket))

    return [b for (s, b) in sorted(ranking, reverse=True)]

File: binderhub/utils.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (consideRatio) [reply binderhub/utils.py]: I was thinking more in lines of get_hash_as_int or similarly as a function name, as I figure it is irrelevant to know the hash implementation algorithm and it increase readability of the code reading get_hash_as_int as compared to get_blake2b_as_int for example as it would require the code-reader to read up on blake2b, and perhaps start wondering if it was important to use blake2b specifically, etc.

File: binderhub/tests/test_utils.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (consideRatio) [reply binderhub/tests/test_utils.py]: We agree, I don't think that keys will remain either if a new bucket is added.

I think there is symmetry still that allows us to conclude what the comment suggested - that we actually successfully test both removing and adding. If we would have started calculating 90 buckets and then calculated 90+10 and slimmed the second ranking, we would have tested for adding, but now we calculate 90+10 first and then for 90 etc.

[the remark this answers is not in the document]
Comment (consideRatio) [reply binderhub/tests/test_utils.py]:
```python
    # verify ranking order is retained after buckets
    # are removed, but also if they are added, which
    # follows from symmetry

    key = "repo1"
    buckets = ["bucket_"+str(i) for i in range(90)]
    extra_buckets = ["extra_bucket_"+str(i) for i in range(10)]

    ranking_1 = utils.rendezvous_rank(buckets + extra_buckets, key)
    for extra_bucket in extra_buckets:
        ranking_1.remove(extra_bucket)

    ranking_2 = utils.rendezvous_rank(buckets, key)

    assert ranking_1 == ranking_2
```

consideRatio:
# Notes
- [A good learning resource about the algorithms](http://www.snia.org/sites/default/files/SDC15_presentations/dist_sys/Jason_Resch_New_Consistent_Hashings_Rev.pdf).
- We could support sticky build pods even without dind enabled, but our current way of finding the valid nodes to schedule on (which isn't obvious because of custom required affinities etc) is made by identifying the dind pods within the namespace.

binderhub/utils.py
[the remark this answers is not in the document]
betatim: I'd be hesitant to use the word `hash` because there is a builtin function in Python called `hash()` that returns integers. The problem with `hash()` is that we can not use it because it is a carefully constructed hash function for the use cases of CPython itself. It doesn't behave like a hash function where two similar inputs produce wildly different outputs.

For example `hash(i) == i` for most (small) integers and similarly `hash(-3) == -3` and other "weird" stuff. So it matters that we don't use `hash()` but it doesn't matter so much which of the hash functions from `hashlib` we use. Some of them might be more expensive to compute or produce results which aren't easy to turn into 64bit integers.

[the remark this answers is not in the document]
consideRatio: For those that has this python insight could verify that that was not a concern perhaps. I like the word hash as it is what I've come to associate with the idea of taking something big and generating something small.

I felt the idea was valid as we would be fine to swap blake2b for some other hash function later if we felt like it and that would been fine, as the way we use the function remains - we want to get a hash from binary data as a number.

binderhub/utils.py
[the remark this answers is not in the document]
consideRatio: @betatim it is missing and pointless with this implementation, as we only generate bucket hashes with key as as a salt, as compared to generating hashes for the buckets in a standalone manner. We won't get the issues of clumping of many repos, because each bucket-repo combo will cause a new ranking where each bucket has an entirely different ranking order with entirely different scores.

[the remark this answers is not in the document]
consideRatio: The issue where clumping arise due to the standalone hashes of individual bucket names, but won't arise if we always hash a "bucket name+repo name" string. Note that the "consistent hashing" algorithm was described like this and had that shortcoming, but the "rendezvous algorithm" in the linked slides you provided doesn't have this as it was described it was hashing "bucket name+repo name" as compared to consistent hashing that only hashed "bucket name".

binderhub/tests/test_utils.py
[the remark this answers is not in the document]
betatim: Resolving this conversation as I added a new test implementing your suggestion. If we need to lets start a new conversation connected to a line of that test.

binderhub/tests/test_utils.py
consideRatio: I think this follows, and that this follows specifically due to proof of the "perfectly stable" algorithm classification.

binderhub/utils.py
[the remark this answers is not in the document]
betatim: I renamed it `blake2b_hash_as_int` as a way to help the user who doesn't know that Blake2 is a hash. I added a comment in `rendezvous_rank()` to explain some of the constraints on what hash function can be used here.

Because `blake2b_hash_as_int()` lives in a public namespace, to be used by who ever needs it, I think it should have a name that tells the caller what it does. In `rendezvous_rank()` we don't care so much so I added a comment explaining that.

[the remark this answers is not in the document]
consideRatio: resolved with "blake2b hash as int"

binderhub/tests/test_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    first_round = utils.rendezvous_rank(["b1", "b2", "b3"], key)
+    second_round = utils.rendezvous_rank([first_round[0], first_round[1]], key)
+
+    assert first_round[0] == second_round[0], key
+
+    key = "crazy frog is a crazy key"
+    first_round = utils.rendezvous_rank(["b1", "b2", "b3"], key)
+    second_round = utils.rendezvous_rank([first_round[0], first_round[1]], key)
+
+    assert first_round[0] == second_round[0], key
```
consideRatio: ### First test
- [ ] It test the same thing for 3 different keys, I'd like for it to be clear that this is a repetition by comments or using a for loop or reducing it to a single test.
- [ ] Perhaps write something more concrete in the assertion message

binderhub/tests/test_utils.py
```diff
@@ -0,0 +1,35 @@
...
+def test_rendezvous_independence():
+    # check that the relative ranking of two buckets doesn't depend on the
+    # presence of a third bucket
+    key = "k1"
+    two_buckets = utils.rendezvous_rank(["b1", "b3"], key)
+    three_buckets = utils.rendezvous_rank(["b1", "b2", "b3"], key)
+    # remove "b2" and check ranking
+    three_buckets.remove("b2")
+
+    assert two_buckets == three_buckets
```
consideRatio: ### Second test
It does the exact same thing as the first test, so it really just adds on test repetition. It does check the full ordering of the items in the list as compared to the first that only checks the first element, but checking the full ordering is very similar to checking the first element only if the list is only two in length.

- [ ] Make the lists contain some more entries to not only test that the list are the same in size as before.

consideRatio:
```suggestion
    assert two_buckets == three_buckets
    
def test_rendezvous_rank_perfect_stability():
    # Verify perfect stability of the algorithm experimentally. This means,
    # for a given key, to have a consistent ranking order internally among
    # a common set of buckets.

    key = "repo1"
    common_buckets = ["common_bucket_" + str(i) for i in range(80)]
    extra_buckets = ["extra_bucket_" + str(i) for i in range(20)]

    common_buckets_ranked = utils.rendezvous_rank(common_buckets, key)

    common_and_extra_buckets_ranked = utils.rendezvous_rank(common_buckets + extra_buckets, key)
    for extra_bucket in extra_buckets:
        common_and_buckets_ranked.remove(extra_bucket)

    assert common_buckets_ranked == common_and_buckets_ranked, "Adding or removing a bucket reordered the remaining buckets!"
```

binderhub/tests/test_utils.py
betatim: I removed the repeated tests.

binderhub/tests/test_utils.py
betatim: I agree that this tests essentially the same as the previous test but I thought you wanted to have a test that compares the total ranking not just the top choice which is why I added this.

consideRatio: In the case with three buckets and removing one, you end up with bucket rankings with two elements each, picking out the first element determines the second anyhow, so it is pretty much the same test as the first except you verify that dimensionality isn't wierd or that you ended up with duplication of elements in the array etc.

betatim: Aha! Thanks for the pointer to the comment on the comment. Now I think I understand better what you mean. I added a test with a large number of buckets like you suggested.

In retrospect that comment-on-the-comment was because I was confusing the case of removing and adding buckets. Since then I had become less confused and forgotten about it. I now think the comment-on-comment only applies to the adding buckets case.

I added the test you suggested as well as one for the case of adding a bucket. I'll mark this as resolved and we can start a new thread on the newest code.

binderhub/tests/test_utils.py
```diff
@@ -0,0 +1,56 @@
...
+    key = "crazy frog is a crazy key"
+    first_round = utils.rendezvous_rank(["b1", "b2", "b3"], key)
+    second_round = utils.rendezvous_rank([first_round[0], first_round[1]], key)
+
+    assert first_round[0] == second_round[0], key
+
+
+def test_rendezvous_independence():
+    # check that the relative ranking of 80 buckets doesn't depend on the
+    # presence of 20 extra buckets
```
betatim: The stability test you suggested.

binderhub/tests/test_utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+    for i in range(80, 100):
+        hundred_buckets.remove("b%i" % i)
+
+    assert eighty_buckets == hundred_buckets
+
+
+def test_rendezvous_redistribution():
+    # check that approximately a third of keys move to the new bucket
+    # when one is added
```
betatim: An idea for how to test that keys move when a new bucket becomes available and that the pattern of movement is right. I think this is how it should be but not sure. WDYT?

binderhub/tests/test_utils.py
```diff
@@ -12,12 +12,45 @@ def test_rendezvous_rank():
...
+                from_b2 += 1
+            # should always move to the newly added bucket
+            assert three_buckets[0] == "b3"
+
+    # because of statistical fluctuations we have to leave some room when
+    # making this comparison
+    assert 0.31 < n_moved / n_keys < 0.35
+    # keys should move from the two original buckets with approximately
+    # equal probability
+    assert abs(from_b1 - from_b2) < 10
```
consideRatio: Nice that you add a redistribution test, it makes me happy to grasp that this really works as we expect!

Assuming 1000 keys moved from either b1 or b2 and this is considered a coin flip, then the average difference will be sqrt(1000) which is about 31.6.

binderhub/tests/test_utils.py
consideRatio: :heart: thanks! Symmetry makes it possible to phrase this as

```
    # check that the relative ranking of 80 remaining buckets doesn't depend on
    # the removal of 20 buckets from the original 100 buckets
``

File: binderhub/tests/test_utils.py (hunk 12 above)
Comment (consideRatio) [reply binderhub/tests/test_utils.py]: Excellent!

Since we actually hash "nodes-key", this test doesn't need to be run multiple times, but if we would hash node separately from key and as consistent hashing was described to do as compared to our rendezvous hashing, then we could by fluke have two node hashes that are spaced luckily allow for the new node to to catch 1/3 of the stuff.

But hmmm, could you position nodes like pointers on a clock to initially have a fair share and then also after have a fair share?

Thats a clean test as well to have I think, to check that we have a 1/2 distribution initially and then get a 1/3 distribution after, combined with the previous test about perfectly stable we capture all kinds of logic. You are already doing this to some degree but it is captures mostly by the `abs(from_b1 - from_b2) < 10` statement.

File: binderhub/tests/test_utils.py (hunk 13 above)
Comment (betatim) [reply binderhub/tests/test_utils.py]: I rounded it to 30. My feeling is that we just need to get the right order of magnitude here that fails the test when something really weird is happening and minimises false alarms.

File: binderhub/tests/test_utils.py [hunk 14]
```diff
@@ -52,9 +52,8 @@ def test_rendezvous_redistribution():
...
-    assert abs(from_bucket["b1"] - from_bucket["b2"]) < 10
-
-    # use the standard deviation of the expected number of entries in each
-    # bucket to get an idea for a reasonable (order of magnitude) difference
-    assert abs(start_in["b1"] - start_in["b2"]) < (n_keys/2)**0.5
+    # equal probability. We pick 30 because it is "about right"
+    assert abs(from_bucket["b1"] - from_bucket["b2"]) < 30
+    # the initial distribution of keys should be roughly the same
+    # We pick 30 because it is "about right"
+    assert abs(start_in["b1"] - start_in["b2"]) < 30
```
consideRatio: The scale of the about right is different between the from_bucket and start_in difference, on average, the random walk distance if you flip a +1 or -1 over and over, is `sqrt(N)`.

For the from_bucket case it is `sqrt(~1000) == ~32` and for the start_in case it is `sqrt(3000) == ~55`.

consideRatio:
Keeps LGTM if you want to self-merge, I don't want to do it as you may want to finish up some detail though.

betatim:
Thanks for the ideas and patience!!

PR #954, opened 2019-09-16
added  Authorization header if access_token provided

kaseyhackspace:
This fixes issue https://github.com/jupyterhub/binderhub/issues/952 by adding an `Authorization: token <access_token>` header to Github API requests from Binderhub if access_token is passed by either `binderhub_config.py` or `GithubRepoProvider` in `secret.yaml`.

PR #951, opened 2019-09-12
[MRG] Add Figshare to UI

nuest:
! This depends on an unreleased version of repo2docker !

This closes #938.

But it's not ready yet :-) Just would like to get some feedback :-), and need help:

With the UI, I get a `/master` at the end of the URL that I don't want:

image

`http://localhost:8585/v2/figshare/10.6084/m9.figshare.9782777.v1/master`

Where does that get added? I can't find it :-/.

OTOH, the local test URL without `/master`

http://localhost:8585/v2/figshare/10.6084/m9.figshare.9782777.v2

gives me "Waiting for build to start" and in the logs there is just 

```
[..]
[D 190912 13:05:18 build:94] 2 build pods
[D 190912 13:05:18 build:141] Build phase summary: {
     "Pending": 2
    }
```

For several minutes now... shouldn't there be more logs?

-------

### Related questions

- With
```bash
python3 -m pip install -e .
./testing/minikube/install-hub
```
I should have the latest code running in my local kube, right?
- _Is it worth writing down in the docs all the places where something has to be added for a new provider?_ Though this list might be outdated with a new UI:
  - `binderhub/main.py`
    - `SPEC_NAMES`
    - `from .repoproviders import `
  - `repo_providers = Dict(`
  - `binderhub/event-schemas/launch.json` (though unsure what that is for)
  - `binderhub/repoproviders.py` the actual `class FigshareProvider(RepoProvider):`
  - `binderhub/static/js/index.js` the example text
  - `binderhub/templates/index.html` the item in the dropdown
- Have you considered generating some bits, e.g.  build `repro_providers` at runtime by asking all the imported instances of `RepoProvidder` how they want to be added to that list ?
- Is it worth attempting to generate the UI parts from a service endpoint for the next ()?
- Is is possible to run the BinderHub outside of Kubernetes to debug the code?

PR #956, opened 2019-09-18
[MRG] GitLab CE repository was moved to GitLab Foss

betatim:
Change the URL for the repository to the new name after the GitLab CE repository got moved.

Will unblock #951 and #954


kaseyhackspace:
This is all good @betatim 😄

PR #966, opened 2019-09-27
Add support for storing build images in gitlab registries

### LINKED ISSUE #965: Add support for gitlab as build registry
We're trying to use a gitlab registry as backend for build images.

The setup is done using a helm chart, and passing in the configuration:
```yaml
config:
  DockerRegistry:
    token_url: "https://gitlab.cern.ch/jwt/auth"
    username: <username>
    password: <password>
  BinderHub:
    image_prefix: gitlab-registry.cern.ch/binder/images/build-
registry:
  url: "gitlab-registry.cern.ch"
```

After debugging the multiple authz errors, it seems to boil down to:
1. missing "service": "container_registry" when fetching a valid token using the token_url (this is [required by gitlab](https://docs.gitlab.com/ee/administration/packages/container_registry.html#enable-the-container-registry))
2. the build image name logic assumes a dockerhub like structure (up to two levels in the path). The gitlab registry allows more and we rely on this (see above with binder/images/actual-image-name)

I'll push a PR soon, just figuring out the best way to make 2. generic.

rochaporto:
Add support for gitlab token_url requests and accept gitlab registry structures, where any number of levels in the image name is possible.

Example:
binder/images/build-<slug>:tag

Where binder is the group, images the project and the rest the image name and tag in the 'images' registry. Subgroups can be added creating additional path levels. The previous logic assumed two levels only, which matches the docker.io behavior.

This change implies the image_prefix will *only* contain the image name, not a repo dns name. So 'rochaporto/myimage' is ok, but not 'docker.io/rochaporto/myimage'. If needed i could look into some additional logic for this, but given urllib.parse expects a scheme it won't be direct.

 Fixes #965

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -271,7 +271,7 @@ class BuildHandler(BaseHandler):
         ).replace('_', '-').lower()
 
         if self.settings['use_registry']:
-            image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
+            image_manifest = await self.registry.get_image_manifest(*image_name.split(':', 1))
             image_found = bool(image_manifest)
         else:
             # Check if the image exists locally!


betatim:
Thanks for digging into this. Every time I end up looking at this code I think we should refactor it a bit and add some tests because I find it hard to "run it in my head" to check the various scenarios of different registries (and then which cases we already handle via what kind of workaround...). So below some code that I think reproduces the values we'd see in a setup like mybinder.org:

```python
In [1]: image_prefix = "gcr.io/binder-prod/r2d-f18835fd-"

In [3]: '{prefix}{build_slug}:{ref}'.format(prefix=image_prefix, build_slug="some-escaped-url", ref="1234")
Out[3]: 'gcr.io/binder-prod/r2d-f18835fd-some-escaped-url:1234'

In [5]: image_name = '{prefix}{build_slug}:{ref}'.format(prefix=image_prefix, build_slug="some-escaped-url", ref="1234")

In [6]: '/'.join(image_name.split('/')[-2:]).split(':', 1)
Out[6]: ['binder-prod/r2d-f18835fd-some-escaped-url', '1234']

In [7]: image_name.split(':', 1)
Out[7]: ['gcr.io/binder-prod/r2d-f18835fd-some-escaped-url', '1234']
```
This means your hunch that what is passed to `get_image_manifest()` is indeed just the name like `binder-prod/r2d-f18835fd-some-escaped-url`. But the new code doesn't produce the same result. I'm trying to remember why we aren't using `image_name.split('/', 1)[1].split(':', 1)`.

I think this would work for the gcr.io case and also for the GitLab registry case.

binderhub/registry.py
```diff
@@ -192,7 +192,7 @@ def get_image_manifest(self, image, tag):
         # first, get a token to perform the manifest request
         if self.token_url:
             auth_req = httpclient.HTTPRequest(
-                url_concat(self.token_url, {"scope": "repository:{}:pull".format(image)}),
+                url_concat(self.token_url, {"scope": "repository:{}:pull".format(image), "service": "container_registry"}),
```
betatim: For one of the clusters on mybinder.org we [use a Harbor registry](
https://github.com/jupyterhub/mybinder.org-deploy/blob/d91b634ccb8f1fe5fa2bb02551ad576f7f7e516d/config/ovh.yaml#L12) and have a URL like: https://dlqpwel7.gra5.container-registry.ovh.net/service/token?service=harbor-registry as the `token_url`. What happens if the token URL already specifies a service parameter and we try and set it again from here?

We should try and not break existing users, in which case maybe we should add a section [below here](https://github.com/jupyterhub/binderhub/blob/master/doc/setup-binderhub.rst#if-you-are-using-azure-container-registry) specifically for using a GitLab registry.

rochaporto: You're right of course. Not sure why i didn't just think of passing it as a query arg. I'll drop that commit/

rochaporto:
> Thanks for digging into this. Every time I end up looking at this code I think we should refactor it a bit and add some tests because I find it hard to "run it in my head" to check the various scenarios of different registries (and then which cases we already handle via what kind of workaround...). So below some code that I think reproduces the values we'd see in a setup like mybinder.org:
> 
> ```python
> In [1]: image_prefix = "gcr.io/binder-prod/r2d-f18835fd-"
> 
> In [3]: '{prefix}{build_slug}:{ref}'.format(prefix=image_prefix, build_slug="some-escaped-url", ref="1234")
> Out[3]: 'gcr.io/binder-prod/r2d-f18835fd-some-escaped-url:1234'
> 
> In [5]: image_name = '{prefix}{build_slug}:{ref}'.format(prefix=image_prefix, build_slug="some-escaped-url", ref="1234")
> 
> In [6]: '/'.join(image_name.split('/')[-2:]).split(':', 1)
> Out[6]: ['binder-prod/r2d-f18835fd-some-escaped-url', '1234']
> 
> In [7]: image_name.split(':', 1)
> Out[7]: ['gcr.io/binder-prod/r2d-f18835fd-some-escaped-url', '1234']
> ```
> 
> This means your hunch that what is passed to `get_image_manifest()` is indeed just the name like `binder-prod/r2d-f18835fd-some-escaped-url`. But the new code doesn't produce the same result. I'm trying to remember why we aren't using `image_name.split('/', 1)[1].split(':', 1)`.
> 
> I think this would work for the gcr.io case and also for the GitLab registry case.

Do you have a set of inputs->outputs to be used for testing? I'm happy to extract this logic to a separate function and add a few tests.

betatim:
Not really :( The best I can think of is configs of existing deployments. For them we could feed in the current config, capture the output and then create a test based on this not changing.

The one posted above is what we use for the GKE cluster of mybinder.org.

For the OVH cluster we use this config https://github.com/jupyterhub/mybinder.org-deploy/blob/c7197ee387048d20838de3c2ab70135f4a83107c/config/ovh.yaml#L9-L14

This is the config of the GESIS cluster https://github.com/gesiscss/orc/blob/aa81bd432330547f76ee8ee8c5fcaf3c8735051e/gesisbinder/config.yaml#L32 (I think it uses DockerHub as storage)

The Pangeo BinderHub uses gcr.io as well https://github.com/pangeo-data/pangeo-binder/blob/7276aae9bf6077761255ef1f41243e6d391bf827/deploy/prod.yaml#L10 (so config wise a copy&paste from mybinder.org)

bdrian:
Hi! I also recently stumbled upon the image_prefix format issue and did not yet realize there's an open issue/PR for this :-) Here is how we solved it:
```diff
diff --git a/binderhub/builder.py b/binderhub/builder.py
index 85d7fd9..1ae560e 100644
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -271,7 +271,10 @@ class BuildHandler(BaseHandler):
         ).replace('_', '-').lower()
 
         if self.settings['use_registry']:
-            image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
+            image_name_splits = image_name.split('/')
+            if len(image_name_splits) > 2:
+                image_name_splits = image_name_splits[1:]
+            image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name_splits).split(':', 1))
             image_found = bool(image_manifest)
         else:
             # Check if the image exists locally!
```

This approach cuts the first part off if we have more than two slash-split components in the image_name. This should be fine when assuming:
 * The local image store, if used, will not be used with names that have more than two components (so binderhub/images/name:tag would not work)
 * Docker Hub is the only remote registry that can be referred to with a two-component image name (like `jupyterhub/singleuser:sometag`)
 * All other remote repositories have to be referred to with at least three components (like `myregistry.com/user/image:tag`, or `myregistry.com/user/something/image:tag`)
 * All remote repositories operate at their domain root (e.g. a repository base url like someserver.org/registry would not be possible)

This is still not perfect, but hopefully reasonable enough for most setups, and as a benefit there's no need to change configurations or other code. However, the solution proposed by @rochaporto would induce less headache overall, since one doesn't need to guess which parts are registry and which are the image name.

A third solution could be to always split the first part of image_name off, similar to what @betatim suggested.

> I'm trying to remember why we aren't using image_name.split('/', 1)[1].split(':', 1)

Possibly because of the magic Docker Hub names, for example if you used `myuser/binderhub-` as prefix instead of `registry-1.docker.io/myuser/binderhub-`, the above code would fail, while the original does not. One can make this work by requiring users to specifically prefix `registry-1.docker.io/` if using the Docker Hub as registry.

betatim:
@bdrian do you want to make a new PR with your suggested patch and some of the test cases? This PR has been quiet for a while so I think it would be fine to start a new one with fresh momentum (and a slightly different approach).

bdrian:
@betatim I'll submit a similar PR soon, just have to finish the local testing. The patch above is dumb on second thought.

rochaporto:
Closing this one in favor of #986

PR #973, opened 2019-10-07
Zenodo, Figshare: return doi url instead of spec

bitnik:
Zenodo and Figshare providers returns spec (DOI) when `get_repo_url` is called. As much as I understand from the comment this is done because repo url is passed to repo2docker. But this is passed not only to repo2docker but also to appendix. And when I check the repo2docker tests ([zenodo](https://github.com/jupyter/repo2docker/blob/master/tests/unit/contentproviders/test_zenodo.py) and [figshare](https://github.com/jupyter/repo2docker/blob/master/tests/unit/contentproviders/test_figshare.py)), repo2docker also works with doi.org urls. In this PR I updated `get_repo_url` methods of zenodo and figshare, so they return doi.org url which is a real url, not a spec.

### DIFF
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -235,9 +235,7 @@ class ZenodoProvider(RepoProvider):
         return resolved_spec
 
     def get_repo_url(self):
-        # While called repo URL, the return value of this function is passed
-        # as argument to repo2docker, hence we return the spec as is.
-        return self.spec
+        return f"https://doi.org/{self.spec}"
 
     async def get_resolved_ref_url(self):
         resolved_spec = await self.get_resolved_spec()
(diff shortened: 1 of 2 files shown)


consideRatio:
Hey @bitnik! I'm going through various PRs hoping to help them resolve one way or another or perhaps get an updated description about the current status of things. What do you think about this PR?

To me, this PR looks very limited in scope and such. I think what's needed for this to get merged is:

1. A clarification of why a change is motivated, I didn't understand it from the description.
2. A clarification if this is likely to cause a breaking change or not.

The description could be quite clear as it is, but I fail to understand the "but also to appendix" part where "appendix" is entirely unknown to me.

consideRatio:
I'll go for a close of this as a stale PR that I struggle to review. Happy to reopen and help review if you wish to follow this up @bitnik!

PR #974, opened 2019-10-07
ref_url and resolved_spec methods for repo providers

bitnik:
Hi, I added 2 new methods for each repo provider:

- `get_resolved_spec`: returns spec with resolved ref
- `get_ref_url`: returns repo url on specific commit in history

and by using these 2 new methods 2 new urls are added into appendix

- `persistent_binder_url`: binder url with resolved ref. e.g., instead of name of the branch `master`, that url will contain commit hash of the master branch when image is built
- `ref_url`: repo url on specific commit in history

I think these 2 urls are useful, if someone wants to use appendix. For example to create a button to copy binder link.

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -313,16 +313,18 @@ class BuildHandler(BaseHandler):
             push_secret = None
 
         BuildClass = FakeBuild if self.settings.get('fake_build') else Build
-        binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
-            proto=self.request.protocol,
-            host=self.request.host,
-            base_url=self.settings['base_url'],
-            provider=provider_prefix,
-            spec=spec,
-        )
+
+        binder_url = f"{self.request.protocol}://{self.request.host}{self.settings['base_url']}v2/" \
+                     f"{provider_prefix}/{spec}"
+        resolved_spec = await provider.get_resolved_spec()
+        persistent_binder_url = f"{self.request.protocol}://{self.request.host}{self.settings['base_url']}v2/" \
+                                f"{provider_prefix}/{resolved_spec}"
+        ref_url = await provider.get_ref_url()
         appendix = self.settings['appendix'].format(
             binder_url=binder_url,
             repo_url=repo_url,
+            persistent_binder_url=persistent_binder_url,
+            ref_url=ref_url,
         )
 
         self.build = build = BuildClass(
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -319,7 +319,7 @@ class BuildHandler(BaseHandler):
         resolved_spec = await provider.get_resolved_spec()
         persistent_binder_url = f"{self.request.protocol}://{self.request.host}{self.settings['base_url']}v2/" \
                                 f"{provider_prefix}/{resolved_spec}"
-        ref_url = await provider.get_ref_url()
+        ref_url = await provider.get_resolved_ref_url()
         appendix = self.settings['appendix'].format(
             binder_url=binder_url,
             repo_url=repo_url,
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -314,11 +314,21 @@ class BuildHandler(BaseHandler):
 
         BuildClass = FakeBuild if self.settings.get('fake_build') else Build
 
-        binder_url = f"{self.request.protocol}://{self.request.host}{self.settings['base_url']}v2/" \
-                     f"{provider_prefix}/{spec}"
+        binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
+            proto=self.request.protocol,
+            host=self.request.host,
+            base_url=self.settings['base_url'],
+            provider=provider_prefix,
+            spec=spec,
+        )
         resolved_spec = await provider.get_resolved_spec()
-        persistent_binder_url = f"{self.request.protocol}://{self.request.host}{self.settings['base_url']}v2/" \
-                                f"{provider_prefix}/{resolved_spec}"
+        persistent_binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
+            proto=self.request.protocol,
+            host=self.request.host,
+            base_url=self.settings['base_url'],
+            provider=provider_prefix,
+            spec=resolved_spec,
+        )
         ref_url = await provider.get_resolved_ref_url()
         appendix = self.settings['appendix'].format(
             binder_url=binder_url,
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -227,9 +227,10 @@ class ZenodoProvider(RepoProvider):
     async def get_resolved_spec(self):
         if not hasattr(self, 'record_id'):
             self.record_id = await self.get_resolved_ref()
-        # zenodo registers a DOI which represents all versions.
+        # zenodo registers a DOI which represents all versions of a software package
         # and it always resolves to latest version
-        # for that case, we have to replace it with record id
+        # for that case, we have to replace the version number in DOIs with
+        # the specific (resolved) version (record_id)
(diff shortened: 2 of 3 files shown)


binderhub/builder.py
```diff
@@ -313,16 +313,18 @@ def escape(s):
...
         BuildClass = FakeBuild if self.settings.get('fake_build') else Build
-        binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
-            proto=self.request.protocol,
-            host=self.request.host,
-            base_url=self.settings['base_url'],
-            provider=provider_prefix,
-            spec=spec,
-        )
+
+        binder_url = f"{self.request.protocol}://{self.request.host}{self.settings['base_url']}v2/" \
```
betatim: nit: I find `{proto}://{host}{base_url}v2/{provider}/{spec}` easier to read than the longer version as f-string. If you don't have a strong preference I'd vote for sticking with the `.format()` style that lets us have shorter "templates".

binderhub/repoproviders.py
```diff
@@ -208,11 +224,17 @@ def get_resolved_ref(self):
         self.record_id = r.effective_url.rsplit("/", maxsplit=1)[1]
         return self.record_id
 
+    async def get_resolved_spec(self):
+        return self.spec
```
betatim: We should not return the DOI here because the DOI is like `master` or a tag. At different moments in time it can point to different things. In Zenodo this flexibility is used to allow versions. What we should return here is the Zenodo record ID (which is like the commit). However we can't use that spec and feed it back into a URL to launch a new binder (I think). (Probably also applies to the other DOI based providers)

betatim:
Looks nice. I had a quick scroll through the code and had a few conceptual questions that I think we can answer by changing the names of the newly added methods.

If I put a URL like https://github.com/binder-examples/requirements into BinderHub then `get_ref_url` returns https://github.com/binder-examples/requirements/tree/da202b4dd88edf6aef2b5ec516ba5b991806691d right? In which case I think we should help the reader by calling it `get_resolved_ref_url()`. That way all the methods that return resolved thing have resolved in their name.

For Zenodo we resolve the DOI to a Zenodo record ID so I think that is what we should be returning in the `get_resolved_ref_url()`. But this means BinderHub needs a way to start a Binder from a Zenodo record ID. For r2d you can already run `repo2docker https://zenodo.org/record/3232985` so we can pass a DOI or a "resolved URL". https://mybinder.org/v2/zenodo/3232985/ doesn't work though. We'd need to extend the Zenodo provider to understand this different kind of spec.

binderhub/repoproviders.py
bitnik: Here I am confused. Zenodo DOI contains the record ID, this makes it is different than branch or tag name or no?

bitnik:
> If I put a URL like https://github.com/binder-examples/requirements into BinderHub then get_ref_url returns https://github.com/binder-examples/requirements/tree/da202b4dd88edf6aef2b5ec516ba5b991806691d right? In which case I think we should help the reader by calling it get_resolved_ref_url(). That way all the methods that return resolved thing have resolved in their name.

Yes and you are right, it makes sense to rename it to `get_resolved_ref_url`.

bitnik:
> For Zenodo we resolve the DOI to a Zenodo record ID so I think that is what we should be returning in the get_resolved_ref_url(). But this means BinderHub needs a way to start a Binder from a Zenodo record ID. For r2d you can already run repo2docker https://zenodo.org/record/3232985 so we can pass a DOI or a "resolved URL". https://mybinder.org/v2/zenodo/3232985/ doesn't work though. We'd need to extend the Zenodo provider to understand this different kind of spec.

Did you mean here `get_resolved_spec()` instead of `get_resolved_ref_url()`? I think `get_resolved_spec()` can return DOI (spec), because it already contains the record ID for Zenodo or article id and version for Figshare. And they are not like branch or tag names as in git repos.

binderhub/repoproviders.py
betatim: I was thinking of this test https://github.com/jupyterhub/binderhub/blob/a168d069772012c52f9ac7056ec22d779927ae69/binderhub/tests/test_repoproviders.py#L40-L52 where the spec is `10.5281/zenodo.3242074`. This might (right now) point to a record ID of 3242074 but as I understand it that is an "accident". So in the future the same DOI could point to a different record ID. Most of my knowledge about versioning in Zenodo comes from https://help.zenodo.org/#versioning

Because the record ID that a DOI points to can change it made me think that a DOI is like a tag or branch name. The record ID is like the commit SHA.

bitnik: thanks a lot Tim, now I understand. I just made a new commit to fix this for zenodo and figshare. can you check it?

bitnik:
@betatim could you do another review?

binderhub/tests/test_repoproviders.py
```diff
@@ -37,45 +37,81 @@ def test_spec_processing(spec, raw_user, raw_repo, raw_ref):
...
-
+@pytest.mark.parametrize('spec,resolved_spec,resolved_ref,resolved_ref_url,build_slug', [
+    ['10.5281/zenodo.3242074',
+     '10.5281/zenodo.3242074',
+     '3242074',
+     'https://doi.org/10.5281/zenodo.3242074',
+     'zenodo-3242074'],
+    # 10.5281/zenodo.705645 -> This DOI represents all versions, and will always resolve to the latest one
+    # for now it is 3242074
+    ['10.5281/zenodo.3242073',
```
betatim:
```suggestion
    ['10.5281/zenodo.705645',
```
I think, based on your comment above.

bitnik: but https://doi.org/10.5281/zenodo.705645 and https://doi.org/10.5281/zenodo.3242073 are different repos. https://doi.org/10.5281/zenodo.3242073  is the url for all versions of the repo which is created by Binder Project and used in tests

bitnik: sorry, my comment is wrong :(

bitnik: i corrected the comment

binderhub/repoproviders.py
```diff
@@ -208,11 +224,24 @@ def get_resolved_ref(self):
         self.record_id = r.effective_url.rsplit("/", maxsplit=1)[1]
         return self.record_id
 
+    async def get_resolved_spec(self):
+        if not hasattr(self, 'record_id'):
+            self.record_id = await self.get_resolved_ref()
+        # zenodo registers a DOI which represents all versions.
+        # and it always resolves to latest version
+        # for that case, we have to replace it with record id
+        resolved_spec = self.spec.split("zenodo")[0] + "zenodo." + self.record_id
```
betatim: Ahh! This creates a new DOI from the record ID! Maybe it is too early on Friday morning but I had to read this a few times to realise it :)

LeoHBK:
Maybe we should RT that to Base, Over.

binderhub/builder.py
```diff
@@ -313,16 +313,28 @@ def escape(s):
...
         BuildClass = FakeBuild if self.settings.get('fake_build') else Build
+
         binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
             proto=self.request.protocol,
             host=self.request.host,
             base_url=self.settings['base_url'],
             provider=provider_prefix,
             spec=spec,
         )
+        resolved_spec = await provider.get_resolved_spec()
```
LeoHBK: [2067.pdf](https://github.com/jupyterhub/binderhub/files/3716063/2067.pdf)

LeoHBK: Conflict Resolved

binderhub/repoproviders.py
bitnik: i think my comment was also not clear. hopefully it is better now and feel free to change it :)

betatim:
Merged!

PR #981, opened 2019-10-21
helm chart: fix compatibility with helm >=2.15.0

### LINKED ISSUE #980: Helm 2.15.0 breaks BinderHub's chart
Hi! Helm 2.15.0 got release a couple of days ago, and it looks like they fixed an issue with integers being parsed as floats (cf. [changelog](https://github.com/helm/helm/releases/tag/v2.15.0), [possibly relevant PR](https://github.com/helm/helm/pull/6010)). This breaks the BinderHub helm chart right here:
https://github.com/jupyterhub/binderhub/blob/889380cf8bca5b3e461802c007a57f52e96dc6fa/helm-chart/binderhub/templates/deployment.yaml#L9

The error would be:
> Error: render error in "binderhub/templates/deployment.yaml": template: binderhub/templates/deployment.yaml:9:15: executing "binderhub/templates/deployment.yaml" at <eq .Values.replicas 1.0>: error calling eq: incompatible types for comparison

One can fix this by casting .Values.replicas to int and comparing it to an integer value; this preserves compatibility with older helm versions. PR is following promptly.

bdrian:
Helm 2.15.0 presents .Values.replicas as int, which is actually the correct behaviour, but this used to be float. This PR fixes compatibility with newer helm versions by always casting the replicas variable to int and then comparing it to an integer.

Closes #980

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -6,7 +6,7 @@ spec:
   replicas: {{ .Values.replicas }}
   strategy:
     rollingUpdate:
-        {{- if eq .Values.replicas 1.0 }}
+        {{- if eq (.Values.replicas | int) 1 }}
         maxSurge: 1
         maxUnavailable: 0
         {{- end }}


betatim:
LGTM. Merging and thanks for taking the time to contribute

betatim:
(apparently the actual "click merge" got lost in me switching from one train to another)

consideRatio:
For future reference, I think helm reverted in 2.15.1 the change that made this PR become needed in helm 2.15.0.

https://github.com/helm/helm/commit/41a697ed10ca21ec0d97e4b3bf7726e14c4957cc

PR #988, opened 2019-10-25
Set Binder ingress service port to 80

rochaporto:
The targetPort in the container is default 8585, but the service port is
hard-coded in the template to 80.


betatim:
I think we never noticed this because of https://kubernetes.github.io/ingress-nginx/user-guide/miscellaneous/#why-endpoints-and-not-services which suggests that nginx-ingress uses endpoints and not services.

PR #989, opened 2019-10-25
fix cleaning pods up after testing

bitnik:
I think something change in the definition of `delete_namespaced_pod` method in kubernetes package. When I run tests locally, I start to get: `TypeError: delete_namespaced_pod() takes 3 positional arguments but 4 were given`.  I think in that way because `kubernetes>=4.*` is in `requirements.txt` and it installs latest kubernetes package but in `binderhub/helm-chart/images/binderhub/requirements.txt` it is fixed to version 7 (`kubernetes==7.0.*`). So we don't see this error in travis.

So first I used keyword arguments in `delete_namespaced_pod` method call to fix that problem. And then I also updated `requirements.txt` and `doc/doc-requirements.txt`, so they install the same version of kubernetes packages as in docker image.

### DIFF
--- a/binderhub/tests/conftest.py
+++ b/binderhub/tests/conftest.py
@@ -222,9 +222,9 @@ def cleanup_pods(namespace, labels):
         print(f"deleting pod {pod.metadata.name}")
         try:
             kube.delete_namespaced_pod(
-                pod.metadata.name,
-                namespace,
-                kubernetes.client.V1DeleteOptions(grace_period_seconds=0),
+                name=pod.metadata.name,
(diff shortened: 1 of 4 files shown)


betatim:
Could we update all the way to kubernetes 9.x.x? In the z2jh chart this was just done and keeping them in sync would maybe(?) make our maintenance life easier.

bitnik:
It makes sense to me and looking at that table I see no problem with upgrading to 9.0.*. Locally I have also just run tests with python-kubernetes  9.0.* and there is no errors.

Issue #939, opened 2019-08-28
Proxy support for BinderHub deployments


bdrian:
Hi!

Our BinderHub lives in an on-prem Kubernetes cluster, which requires a proxy to access the internet. I didn't find an option to specify which proxy to use throughout all BinderHub-related pods, and the conventional HTTP[S]_PROXY, NO_PROXY or their lowercase variants don't seem to be honored. Without a proxy most builds fail either before starting a build (you cannot resolve GitHub-URLs, for example), or during it (because you cannot download any dependencies).

Right now, we solved this by uglily patching a few python files throughout the respective images. Changed were: 1) calls to tornado's httpclient in Binderhub itself, 2) the environment variables in build pod definitions and 3) the docker build arguments in repo2docker. The last one is obviously not part of BinderHub itself, but since the build pods are setup by BinderHub, it would be nice to have a way to pass a proxy configuration on to repo2docker. Perhaps one has to open another issue for the repo2docker-side of this integration.

Is there any work made being towards this, or did I perhaps miss a configuration option? I could contribute code for this, but without any promises about when it's finished.

betatim:
Thanks for posting and setting up a BinderHub behind a proxy!

> Is there any work made being towards this, or did I perhaps miss a configuration option?

I am not aware of any work, AFAIK you are the first person to stop by who has this need.

> I could contribute code for this, but without any promises about when it's finished.
I think that is par for the course in open-source. In light of this it might be worth starting by making a list of all the places that need changing and any dependencies between them. Basically trying to make a plan where each step can be tackled by one person and merged one at a time. This way it doesn't matter if the person who started has to switch to other work.

It would also be worth exploring what options kubernetes provides to redirect/proxy traffic on a "whole cluster" scale. I've never investigated this but it would be attractive because we need to modify less code and it could cover things like non-HTTPS traffic as well (`git clone git@hostname.example.com...`) and we already use https://kubernetes.io/docs/concepts/services-networking/network-policies/ to restrict traffic.

nuest:
(Just discussing this issue IRL with @betatim @minrk)

Here are the links to changes @bdrian needed to do, including some changes to try out kaniko:

im repo2docker-image:
https://zivgitlab.uni-muenster.de/a_broe10/jupyter.repo2docker/blob/master/patches/app.py.patch#L42

in binderhub-image:
[...]

PR #990, opened 2019-10-28
Proxy Support

bdrian:
This PR adds support for running BinderHub behind a proxy, at least once complementary changes are made in repo2docker. I'm not sure if this is the prettiest way to go, feedback is welcome.

See #939

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -7,6 +7,7 @@ import datetime
 import json
 import threading
 from urllib.parse import urlparse
+import os
 
 from kubernetes import client, watch
 from tornado.ioloop import IOLoop
@@ -235,6 +236,24 @@ class Build:
         if self.git_credentials:
             env.append(client.V1EnvVar(name='GIT_CREDENTIAL_ENV', value=self.git_credentials))
 
+        # copy additional variables from current environment to new pod environment
+        proxy_environment_variables = [
+            'http_proxy',
+            'https_proxy',
+            'HTTP_PROXY',
+            'HTTPS_PROXY',
+            'no_proxy',
+            'NO_PROXY',
+        ]
+
+        for env_var in proxy_environment_variables:
+            try:
+                env.append(client.V1EnvVar(name=env_var, value=os.environ[env_var]))
+            except KeyError:
+                # skip the environment variable if it isn't present
+                pass
+
+        component_label = "binderhub-build"
         self.pod = client.V1Pod(
             metadata=client.V1ObjectMeta(
                 name=self.name,
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -3,12 +3,13 @@ Main handler classes for requests
 """
 import urllib.parse
 
-from tornado.httpclient import AsyncHTTPClient, HTTPRequest
+from tornado.httpclient import HTTPRequest
 from tornado.web import HTTPError, authenticated
 from tornado.httputil import url_concat
 from tornado.log import app_log
 
 from .base import BaseHandler
+from .utils import ProxiedAsyncHTTPClient
 
 SPEC_NAMES = {
     "gh": "GitHub",
@@ -77,7 +78,7 @@ class ParameterizedMainHandler(BaseHandler):
 
             # Check if the nbviewer URL is valid and would display something
             # useful to the reader, if not we don't show it
-            client = AsyncHTTPClient()
+            client = ProxiedAsyncHTTPClient()
             # quote any unicode characters in the URL
             proto, rest = nbviewer_url.split("://")
             rest = urllib.parse.quote(rest)
--- a/binderhub/registry.py
+++ b/binderhub/registry.py
@@ -11,6 +11,8 @@ from tornado.httputil import url_concat
 from traitlets.config import LoggingConfigurable
 from traitlets import Dict, Unicode, default
 
+from .utils import ProxiedAsyncHTTPClient
+
 DEFAULT_DOCKER_REGISTRY_URL = "https://registry.hub.docker.com"
 DEFAULT_DOCKER_AUTH_URL = "https://index.docker.io/v1"
 
@@ -187,7 +189,7 @@ class DockerRegistry(LoggingConfigurable):
 
     @gen.coroutine
     def get_image_manifest(self, image, tag):
-        client = httpclient.AsyncHTTPClient()
+        client = ProxiedAsyncHTTPClient()
         url = "{}/v2/{}/manifests/{}".format(self.url, image, tag)
         # first, get a token to perform the manifest request
         if self.token_url:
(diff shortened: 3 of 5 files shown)


bdrian:
The problem above is fixed and Travis should be able to [build and test successfully](https://travis-ci.org/jupyterhub/binderhub/builds/604371653), however the pipeline is randomly failing for reasons unrelated to this PR ([one](https://travis-ci.org/jupyterhub/binderhub/builds/604349979), [two](https://travis-ci.org/jupyterhub/binderhub/builds/604391901)).

meeseeksmachine:

https://discourse.jupyter.org/t/binder-behind-outbound-proxy/7428/2

g-braeunlich:
Hey guys.
Just had a look at this PR. Does the commit to change the build pod work for you?
In my case, to get it working, I had to inject a `/root/.docker/config.json` file, containing the proxy vars into the build pod.
Just setting env vars in the build pod was not enough as the docker client inside the pod did not pick up the env vars.
See: https://medium.com/@bennyh/docker-and-proxy-88148a3f35f7

g-braeunlich:
This is my current changes (on top of current master), which made it work for me:

```
diff --git a/binderhub/build.py b/binderhub/build.py
index ae16d61..aa8a0c6 100644
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -7,6 +7,7 @@ import datetime
 import json
 import threading
+import os
 
 from kubernetes import client, watch
 from tornado.ioloop import IOLoop
@@ -256,6 +257,9 @@ class Build:
         if self.git_credentials:
             env.append(client.V1EnvVar(name='GIT_CREDENTIAL_ENV', value=self.git_credentials))
 
+        # If there are proxy_* env variables, inject a /root/.docker/config.json:
+        self.mount_docker_runtime_config(volumes, volume_mounts)
+        
         self.pod = client.V1Pod(
             metadata=client.V1ObjectMeta(
                 name=self.name,
@@ -404,6 +408,45 @@ class Build:
         """Stop watching a build"""
         self.stop_event.set()
 
+    def mount_docker_runtime_config(self, volumes, volume_mounts):
+        proxy_vars = {
+            var: os.environ.get(var, os.environ.get(var.upper()))
+            for var in ("http_proxy", "https_proxy", "no_proxy")
+        }
+
+        def camel_casify(var):
+            return var.replace("_p", "P")
+
+        # Filter None and map do camel case:
+        proxy_vars = {camel_casify(var): val for var, val in proxy_vars.items() if val}
+        if not proxy_vars:
+            return
+
+        config_map = client.V1ConfigMap(
+            api_version="v1",
+            kind="ConfigMap",
+            metadata=client.V1ObjectMeta(
+                name="https-build-docker-runtime-config",
+                labels={"component": self._component_label},
+            ),
+            data={"config.json": json.dumps({"proxies": {"default": proxy_vars}}, indent=2)},
+        )
+        try:
+            self.api.create_namespaced_config_map(self.namespace, config_map)
+        except client.rest.ApiException as e:
+            if e.status != 409: # 409 is ok: already created
+                raise
+
+        volume = client.V1Volume(
+            name="https-build-vol-docker-runtime-config",
+            config_map=client.V1ConfigMapVolumeSource(name=config_map.metadata.name),
+        )
+        volumes.append(volume)
+        volume_mounts.append(
+            client.V1VolumeMount(mount_path="/root/.docker/", name=volume.name)
+        )
+    
+
 class FakeBuild(Build):
     """
     Fake Building process to be able to work on the UI without a running Minikube.
diff --git a/helm-chart/binderhub/templates/rbac.yaml b/helm-chart/binderhub/templates/rbac.yaml
index 5716b5a..9a4dcd8 100644
--- a/helm-chart/binderhub/templates/rbac.yaml
+++ b/helm-chart/binderhub/templates/rbac.yaml
@@ -14,6 +14,9 @@ rules:
 - apiGroups: [""]
   resources: ["pods/log"]
   verbs: ["get"]
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["create"]
 ---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
```

g-braeunlich:
Alternatively, the config_map also could be templated with helm (either have to introduce new proxy values or somehow try to extract the proxy env vars out of the extraEnv values)

betatim:
I've not followed this PR but if you are interested in working on this @g-braeunlich: do you want to take over the commits in this PR and make a new PR where you add your changes on top? I think having someone who uses this feature and, as a result, is interested in pushing this over the finish line and able to test it would be great.

g-braeunlich:
@betatim Definitely interested, but I cannot promise that I will have time to also test the part outside the repo2docker build pod. We only connect to internal repos

manics:
Option 3 could be to add an entrypoint/command script in the [repo2docker container](https://github.com/jupyterhub/repo2docker/blob/master/Dockerfile) that creates the required `docker.json` file inside the container (perhaps by checking for `*_PROXY` environment variables?) before executing repo2docker?

g-braeunlich:
That makes a lot of sense. I will file a MR there.
Additionally this would require to make the env vars for the build pod configurable or pass the env vars from the binder pod to the build pod

manics:
Sounds good! I'm not 100% sure it'll work though. If it does then one advantage of doing it in the repo2docker Dockerfile is it should "just work" for anyone using that container outside of BinderHub.

betatim:
> I cannot promise that I will have time to also test the part outside the repo2docker build pod.

That is fine. I think as long as we are clear about what we are doing, can do, can test/not test, etc we can keep moving this forward with the resources/time each of us has.Basically aim for adding things that make the situation better but resist adding things where we aren't sure/can't test if they will make things better. No code is better than broken code :D

g-braeunlich:
> Sounds good! I'm not 100% sure it'll work though. If it does then one advantage of doing it in the repo2docker Dockerfile is it should "just work" for anyone using that container outside of BinderHub.

While preparing the PR, I had the following thoughts:
What if there are some users which have their own usecases and mount a custom /root/.docker/config.json?
The entrypoint then would overwrite their file.
Or if the file is mounted read only, it will break their workflow.

manics:
Good point. Maybe create the file if a `*_PROXY` environment variable is set, and otherwise do nothing?

g-braeunlich:
My current suggestion would be:
If no *_PROXY is set OR the file already exists -> do nothing
If the file exists and PROXY vars are set, display a warning message

g-braeunlich:
This is the PR: https://github.com/jupyterhub/repo2docker/pull/1003

manics:
Fixed by
- https://github.com/jupyterhub/repo2docker/pull/1003
- https://github.com/jupyterhub/binderhub/pull/1255

PR #994, opened 2019-11-02
add hub_url_local as hub_url currently represents both public and local

manics:
I don't know if this is useful so opening it for comments.

I just ran Binderhub on a misconfigured test system where external ingress is limited by IP range. Unfortunately the IP range didn't include itself 😀

BinderHub uses the external JupyterHub API instead of the internal K8s service. Obviously the firewall will be fixed, but is this potentially useful anyway?

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -98,7 +98,7 @@ spec:
               key: "binder.hub-token"
         {{- if .Values.config.BinderHub.auth_enabled }}
         - name: JUPYTERHUB_API_URL
-          value: {{ (print (.Values.config.BinderHub.hub_url | trimSuffix "/") "/hub/api/") }}
+          value: {{ (print (.Values.config.BinderHub.hub_url_local | default .Values.config.BinderHub.hub_url | trimSuffix "/") "/hub/api/") }}
         - name: JUPYTERHUB_BASE_URL
           value: {{ .Values.jupyterhub.hub.baseUrl | quote }}
         - name: JUPYTERHUB_CLIENT_ID
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -681,7 +681,7 @@ class BinderHub(Application):
                 tornado.web.StaticFileHandler,
                 {'path': os.path.join(self.tornado_settings['static_path'], 'images')}),
             (r'/about', AboutHandler),
-            (r'/health', HealthHandler, {'hub_url': self.hub_url}),
+            (r'/health', HealthHandler, {'hub_url': self.hub_url_local}),
             (r'/', MainHandler),
             (r'.*', Custom404),
         ]
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -379,14 +379,21 @@ class BinderHub(Application):
         config=True,
     )
 
-    hub_url_public = Unicode(
+    hub_url_local = Unicode(
         help="""
-        The public base URL of the JupyterHub instance where users will run if different from hub_url
+        The base URL of the JupyterHub instance for local/internal traffic
+
+        If local/internal network connections from the BinderHub process should access
+        JupyterHub using a different URL than public/external traffic set this, default
+        is hub_url
         """,
         config=True,
     )
+    @default('hub_url_local')
+    def _default_hub_url_local(self):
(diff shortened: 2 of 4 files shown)


manics:
Based on https://github.com/jupyterhub/binderhub/pull/1101 this problem is not unique to my test cluster!

helm-chart/images/binderhub/binderhub_config.py
```diff
@@ -84,7 +84,8 @@ def get_value(key, default=None):
     c.BinderHub.build_namespace = os.environ['BUILD_NAMESPACE']
 
 if c.BinderHub.auth_enabled:
-    hub_url = urlparse(c.BinderHub.hub_url)
+    hub_url = urlparse(c.BinderHub.hub_url_public if 'hub_url_public' in c.BinderHub
+                       else c.BinderHub.hub_url)
```
consideRatio, line 88: Perhaps we can make hub_url_public have a `@default` decorated function returning self.hub_url? I think this logic is too messy to use whenever we want to use hub_url_public, which happen to only be once at the moment but who knows.

consideRatio:
@manics I want to resurrect my work to make BinderHub support automatic HTTPS acquisition using z2jh, and I'll need this if I recall correctly. If you resolve the merge conflicts I'll review this and help it get merged.

consideRatio:
I think a key point of review for this PR regards if we want to assume hub_url is a local URL and add a hub_url_public option, or assume it is a public url, and add a hub_url_local option.

From what I see in #1139 and https://binderhub.readthedocs.io/en/latest/zero-to-binderhub/setup-binderhub.html#connect-binderhub-and-jupyterhub, I think perhaps it is better to add a hub_url_local and assume hub_url to be a public path. What do you think?

@bitnik perhaps you have some input about this?

manics:
@consideRatio I'm glad it's useful for more than just me 😄 . I'll resolve the conflicts and try your default suggestion. I don't have any preference on whether to add `hub_url_public` or `hub_url_local`.

consideRatio:
Lets go with `hub_url_local`, making `hub_url` become `_public` implicitly which it seems documented to be used as currently.

bitnik:
> Lets go with hub_url_local, making hub_url become _public implicitly which it seems documented to be used as currently.

I dont have anything to add on, and I agree with @consideRatio

manics:
I added one more commit. I've only partially tested this because k8s broke for unrelated reasons, but I think the PR is ready for review, especially since you already understand what's going on.

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -98,7 +98,7 @@ spec:
               key: "binder.hub-token"
         {{- if .Values.config.BinderHub.auth_enabled }}
         - name: JUPYTERHUB_API_URL
-          value: {{ (print (.Values.config.BinderHub.hub_url | trimSuffix "/") "/hub/api/") }}
+          value: {{ (print (.Values.config.BinderHub.hub_url_local | default .Values.config.BinderHub.hub_url | trimSuffix "/") "/hub/api/") }}
```
consideRatio, line 101: Nice catch adding this as well!

consideRatio: If auth_enabled, this URL is required to be mounted as an env var? Hmmm? Is this because the JupyterHub imported HubOAuth classes looks for this URL or similar?

Ah yes it seems so, this could been configured with `c.HubAuth.api_url = c.BinderHub.hub_url_local` someplace instead, because JUPYTERHUB_API_URL serves the purpose of a default value for the c.HubAuth.api_url traitlet.

It seems like...

- JUPYTERHUB_API_URL can be configured with c.HubAuth.api_url
- JUPYTERHUB_BASE_URL can be configured with c.HubAuth.hub_prefix
- JUPYTERHUB_CLIENT_ID can be configured with c.HubOAuth.oauth_client_id
- JUPYTERHUB_OAUTH_CALLBACK_URL can be configured with c.HubOAuth.oauth_redirect_uri

Well, okay, let's configure them through environment variables.

manics: @consideRatio I was confused by the apparent duplication between config and env-vars too. I only discovered the problem when deploying it to a test system.

PR #1001, opened 2019-11-14
Document automatic HTTPS

xhochy:
There is a lot of unofficial, mostly outdated information in Github issues on how to setup HTTPS for a binderhub. This is the way I got it working and would love to have it included in the documentation for clarity.

### DIFF
--- a/doc/https.rst
+++ b/doc/https.rst
@@ -81,7 +81,7 @@ file ``nginx-ingress.yaml``:
 Afterwards we install the ingress proxy using
 ``helm install stable/nginx-ingress --name binderhub-proxy --namespace <same-namespace-as-binderhub> -f nginx-ingress.yaml``.
 Then wait until it is ready and showing the correct IP when looking at the output of
-``kubectl --namespace qc-binder-dev get services binderhub-proxy-nginx-ingress-controller``.
+``kubectl --namespace <same-namespace-as-binderhub> get services binderhub-proxy-nginx-ingress-controller``.
 
 Adjust BinderHub config to serve via HTTPS
 ------------------------------------------
--- a/doc/https.rst
+++ b/doc/https.rst
@@ -87,7 +87,7 @@ Adjust BinderHub config to serve via HTTPS
 ------------------------------------------
 
 With the static IP, DNS records and ingress proxy setup, we can now change our
-BinderHub configuration to serve via HTTPS. Therefore adjust your ``config.yaml``
+BinderHub configuration to serve traffic via HTTPS. Therefore adjust your ``config.yaml``
 with the following sections and apply it using ``helm upgrade ...``.
 
 .. code:: yaml
--- a/doc/https.rst
+++ b/doc/https.rst
@@ -17,7 +17,10 @@ Setup IP & domain
 2. Buy a domain name from a registrar. Pick whichever one you want.
 3. Set A records to your above retrieved external IP, one for Binder and
    one for JupyterHub. We need two distinct subdomains for the routing to
-   the two different services as they will be served by the same proxy.
+   the two different services as they will be served by the same ingress proxy.
+   We suggest you use ``hub.binder.`` for JupyterHub and ``binder.`` for your
+   BinderHub. Once you are done your BinderHub will be available at
+   ``https://binder.``.
 4. Wait some minutes for the DNS A records to propagate.
 
 cert-manager for automatic TLS certificate provisioning
@@ -110,7 +113,7 @@ with the following sections and apply it using ``helm upgrade ...``.
             enabled: true
             type: nginx
         tls:
-           - secretName: <jupyterhub-URL>-tls
+           - secretName: <jupyterhub-URL-with-dashes-instead-of-dots>-tls
              hosts:
               - <jupyterhub-URL>
 
@@ -126,7 +129,7 @@ with the following sections and apply it using ``helm upgrade ...``.
           enabled: true
           type: nginx
       tls:
-        - secretName: <binderhub-URL>-tls
+        - secretName: <binderhub-URL-with-dashes-instead-of-dots>-tls
           hosts:
             - <binderhub-URL>
 
--- a/doc/https.rst
+++ b/doc/https.rst
@@ -1,0 +1,134 @@
+Secure with HTTPS
+=================
+
+To enable HTTPS on your BinderHub you can setup an ingress proxy and configure
+it to serve both, the Binder and JupyterHub interface, using TLS. You can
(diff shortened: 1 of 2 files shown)


betatim:
Thanks a lot for doing this! It has been a neglected area of the docs.

I like that the steps build on the existing docs. For mybinder.org we've created a chart that depends on the binderhub, kubelego and nginx-ingress charts. This means everything there is slightly different from what is described in the guide here.

Should we add a note about using helm or not to install certmanager? I don't know which way is better but if we have some information on how to choose it would be useful for readers. My naive assumption when starting to read was that the instructions would be following https://docs.cert-manager.io/en/latest/getting-started/install/kubernetes.html#installing-with-helm and that guess turned out to be wrong :D

Other than this question there are a few nits but I think this can be merged (restarted the CI builds and will check why they are unhappy if they fail again).

doc/https.rst
```diff
@@ -0,0 +1,134 @@
...
+-----------------
+
+1. Get a static IP(v4) address that you will assign to your ingress proxy
+   later. For example, on Google Cloud this can be done using
+   ``gcloud compute addresses create <alias-name-for-ip> --region <region>``
+   and retrieve the assigned IP using ``gcloud compute addresses list``.
+2. Buy a domain name from a registrar. Pick whichever one you want.
+3. Set A records to your above retrieved external IP, one for Binder and
+   one for JupyterHub. We need two distinct subdomains for the routing to
+   the two different services as they will be served by the same proxy.
```
betatim: Should we use ingress here just to use a word that will be used again later? I think the super mega correct and jargon'y word would be reverse proxy (not sure this would add to anything but confusion).

doc/https.rst
```diff
@@ -0,0 +1,134 @@
...
+Setup IP & domain
+-----------------
+
+1. Get a static IP(v4) address that you will assign to your ingress proxy
+   later. For example, on Google Cloud this can be done using
+   ``gcloud compute addresses create <alias-name-for-ip> --region <region>``
+   and retrieve the assigned IP using ``gcloud compute addresses list``.
+2. Buy a domain name from a registrar. Pick whichever one you want.
+3. Set A records to your above retrieved external IP, one for Binder and
+   one for JupyterHub. We need two distinct subdomains for the routing to
```
betatim: What about adding something like: "We suggest you use hub.binder.<domain> for JupyterHub and binder.<domain> for your BinderHub. Once you are done your BinderHub will be available at https://binder.<domain>"? To give people an idea what to name stuff and which of these URLs users will end up having to type.

doc/https.rst
```diff
@@ -0,0 +1,134 @@
...
+        email: <your-contact-mail>
+        server: https://acme-v02.api.letsencrypt.org/directory
+        privateKeySecretRef:
+          # Secret resource used to store the account's private key.
+          name: letsencrypt-production
+        solvers:
+        - http01:
+            ingress:
+              class: nginx
+
```
betatim:
```suggestion

See the documentation for `more details on configuring the issuer <https://docs.cert-manager.io/en/latest/tasks/issuers/setup-acme/index.html>`_.
```

doc/https.rst
```diff
@@ -0,0 +1,134 @@
...
+.. code:: yaml
+
+    controller:
+      service:
+        loadBalancerIP: <STATIC-IP>
+
+Afterwards we install the ingress proxy using
+``helm install stable/nginx-ingress --name binderhub-proxy --namespace <same-namespace-as-binderhub> -f nginx-ingress.yaml``.
+Then wait until it is ready and showing the correct IP when looking at the output of
+``kubectl --namespace qc-binder-dev get services binderhub-proxy-nginx-ingress-controller``.
```
betatim:
```suggestion
``kubectl --namespace <same-namespace-as-binderhub> get services binderhub-proxy-nginx-ingress-controller``.
```

doc/https.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Adjust BinderHub config to serve via HTTPS
+------------------------------------------
+
+With the static IP, DNS records and ingress proxy setup, we can now change our
+BinderHub configuration to serve via HTTPS. Therefore adjust your ``config.yaml``
```
betatim:
```suggestion
BinderHub configuration to serve traffic via HTTPS. Therefore adjust your ``config.yaml``
```

doc/https.rst
```diff
@@ -0,0 +1,134 @@
...
+          - <jupyterhub-URL>
+        annotations:
+          kubernetes.io/ingress.class: nginx
+          kubernetes.io/tls-acme: "true"
+          cert-manager.io/issuer: letsencrypt-production
+          https:
+            enabled: true
+            type: nginx
+        tls:
+           - secretName: <jupyterhub-URL>-tls
```
betatim: Can you use dots `.` in secret names?

xhochy: You can but I think it is not seen as a good practice. Would the following, longer version work better?
```suggestion
           - secretName: <jupyterhub-URL-with-dashes-instead-of-dots>-tls
```

xhochy:
> Thanks a lot for doing this! It has been a neglected area of the docs.
> 
> I like that the steps build on the existing docs. For mybinder.org we've created a chart that depends on the binderhub, kubelego and nginx-ingress charts. This means everything there is slightly different from what is described in the guide here.

It would probably make sense to have this as a public helm chart; maybe exchange kube-lego with cert-manager as the latter is considered the successor of the first.

> 
> Should we add a note about using helm or not to install certmanager? I don't know which way is better but if we have some information on how to choose it would be useful for readers. My naive assumption when starting to read was that the instructions would be following https://docs.cert-manager.io/en/latest/getting-started/install/kubernetes.html#installing-with-helm and that guess turned out to be wrong :D

I documented the way I got it working, we can work on improving that. I will have a look at using the helm chart also for cert-manager.

doc/https.rst
betatim: Looks good to me.

betatim:
Sounds good.

Do you want to work on the two non-resolved comments or should we merge now?

xhochy:
> Sounds good.
> 
> Do you want to work on the two non-resolved comments or should we merge now?

I worked on the comments and this should be ready for merge once CI passes.

doc/https.rst
xhochy: I've put `ingress proxy` in the text, reverse would be the correct naming probably but this is not occurring in any other line yet.

betatim:
Thanks a lot for writing down how you did it and the nice PR!

PR #1004, opened 2019-11-18
Bump and update for fresh chartpressing

consideRatio:
I bumped chartpress but also clarified various things etc in docs.

- removed use of deprecated --commit-range that is now handled
automatically.
- trimmed paths in chartpress.yaml to what mattered and documented some
reasons for that.
- added notes to the Dockerfile which I spent time understanding more in
detail while working on this PR.
- added resetTag and resetVersion to the chartpress.yaml configuration,
allowing the use of the --reset flag to reset values.yaml and Chart.yaml
to the state it was in before chartpress was run locally.

---

## Potentially breaking change
Note [the Helm chart versions for JupyterHub](https://jupyterhub.github.io/helm-chart/) with names like `jupyterhub-0.9.0-alpha.1.033.fd3e470`? With chartpress 0.4.3, this is what you will get for BinderHub as well. The pattern is `<chartname>-<version>[-<pre-release-tag>[.nnn.sha]]`, this could influence henchbot etc. perhaps.

### DIFF
--- a/helm-chart/chartpress.yaml
+++ b/helm-chart/chartpress.yaml
@@ -1,23 +1,28 @@
+# For a reference on this configuration, see the chartpress README file.
+# ref: https://github.com/jupyterhub/chartpress
 charts:
   - name: binderhub
     imagePrefix: jupyterhub/k8s-
     repo:
       git: jupyterhub/helm-chart
       published: https://jupyterhub.github.io/helm-chart
-    paths:
-      - ..
+    resetTag: local
+    resetVersion: 0.2.0
+    # NOTE: All paths will be set relative to this file's location, which is in the
+    # helm-chart folder.
     images:
       image-cleaner:
         valuesPath: imageCleaner.image
       binderhub:
-        # Context to send to docker build for use by the Dockerfile
+        # Context to send to docker build for use by the Dockerfile. We pass the
+        # root folder in order to allow the image to access and build the python
+        # package. Since we do that, chartpress will always react to changes in
+        # documentation and other things, and always consider the chart version
+        # to change along with the image version.
         contextPath: ..
-        # Dockerfile path relative to chartpress.yaml
+        # Since we changed the contextPath, we must also change the
+        # dockerfilePath. This is because chartpress assume the Dockerfile will
+        # reside in the contextPath folder, and since we overrode the default of
+        # images/binderhub it will be the wrong folder.
         dockerfilePath: images/binderhub/Dockerfile
         valuesPath: image
-        # additional paths relevant to the image
-        # in addition to image directory itself
-        # paths are relative to chartpress.yaml
-        paths:
-          - ../setup.py
-          - ../binderhub
--- a/ci/test-helm
+++ b/ci/test-helm
@@ -34,21 +34,39 @@ if [[ ! -z "$GITHUB_ACCESS_TOKEN" ]]; then
 EOF
 fi
 
+# FIXME: Move to a dedicated build stage in travis like in z2jh
+#
+# ref: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/.travis.yml
 if [[ "$TRAVIS_PULL_REQUEST" == "false" && "$TRAVIS_BRANCH" == "master" ]]; then
   openssl aes-256-cbc -K $encrypted_d8355cc3d845_key -iv $encrypted_d8355cc3d845_iv -in travis.enc -out travis -d
   chmod 0400 travis
   export GIT_SSH_COMMAND="ssh -i ${PWD}/travis"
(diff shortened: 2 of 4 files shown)


helm-chart/chartpress.yaml
```diff
@@ -1,23 +1,28 @@
+# For a reference on this configuration, see the chartpress README file.
+# ref: https://github.com/jupyterhub/chartpress
 charts:
   - name: binderhub
     imagePrefix: jupyterhub/k8s-
     repo:
       git: jupyterhub/helm-chart
       published: https://jupyterhub.github.io/helm-chart
-    paths:
-      - ..
```
betatim: https://github.com/jupyterhub/binderhub/pull/927 is the PR that added this. Check out the issue linked in it for some details on the bug that was fixed by adding this here. What has changed that we don't need this any more?

helm-chart/chartpress.yaml
```diff
@@ -1,23 +1,28 @@
...
+        # reside in the contextPath folder, and since we overrode the default of
+        # images/binderhub it will be the wrong folder.
         dockerfilePath: images/binderhub/Dockerfile
         valuesPath: image
-        # additional paths relevant to the image
-        # in addition to image directory itself
-        # paths are relative to chartpress.yaml
-        paths:
-          - ../setup.py
-          - ../binderhub
```
betatim: I assume these are being proposed for removal because the `contextPath` already covers them? Is there a way to check that? I have a vague memory that `contextPath` is the build context sent to dockerd but that the `paths` are used to determine if an image needs to be built at all.

betatim:
@henchc what do you think about the version tag changing and henchbot?

In general I'd be conservative with changing the build infrastructure in this repository. "Never touch a running system." It works well, there are several automations built on top of this that help keep mybinder.org fresh as well as some other public BinderHubs.

Some of the content of the chartpress config were arrived at after debugging subtle bugs that don't always happen (and had gone unnoticed for a long time). So we should take our time when changing things to make sure they don't resurface.

Not against modernisation but at a slow pace please 🐌

betatim:
@choldgraf could you take a look at https://circleci.com/gh/jupyterhub/binderhub/578?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link which seems to be failing because it can't find a tag/reference. Is this a bit of the docs that is being "copy&pasted" from a different repo? It worked a few days ago when `master` was last built which makes me wonder if the recent changes (pandas theme re-org) in a different repo broke something:-/

betatim:
My guess is that [this code](https://github.com/jupyterhub/binderhub/blob/c80b26cafe73f253428a3bebfe4292dfb76796e3/doc/conf.py#L198-L208) is what is now failing @choldgraf. A new PR to fix it would be great.

helm-chart/chartpress.yaml
consideRatio: Before, chartpress did not consider changes to the images it depended on so this was a hack to resolve it. But, now chartpress does this in [this code section](https://github.com/jupyterhub/chartpress/blob/ac2fd8a1f9cb7505c0e215c673683f90a11d329d/chartpress.py#L689-L695), so we don't need this workaround.

helm-chart/chartpress.yaml
consideRatio: Correct, but the build context is used as one foundational path alongside those, so these don't add anything as they are part of the build context.

consideRatio:
@betatim does this kind of error happen often?

https://travis-ci.org/jupyterhub/binderhub/jobs/613744946?utm_medium=notification&utm_source=github_status

I'm confident it is something I've tracked down and fixed in z2jh + jupyterhub in some PRs, but they require a bump of z2jh + jupyterhub. Anyhow, I assume we can also press restart build job on that to make things succeed.

meeseeksmachine:

https://discourse.jupyter.org/t/binder-jupyterhub-activity-round-up-week-1/2704/1

helm-chart/chartpress.yaml
bitnik: I think that code section is not in chartpress version 0.4.3, that's must be the reason of https://github.com/jupyterhub/binderhub/issues/919#issuecomment-559370276

consideRatio: It was meant to use the info about the file-requirements of the images alongside its own, so it should not had mattered. The issue is that this may still be required but wasnt meant to still be required.

There are tests etc setup in chartpress, but ive apparently missed this test case.

bitnik: Sorry, I think I was not clear or I couldn't understand you. 

What I mean is PR https://github.com/jupyterhub/chartpress/pull/68 is not included in v0.4.3. And that PR fixes the problem we have in binderhub right now. I think we need a version upgrade in chartpress.

PR #1005, opened 2019-11-18
Fix broken doc link

consideRatio:
ping @betatim @choldgraf - solves https://github.com/jupyterhub/binderhub/pull/1004#issuecomment-555226799!

It does not fix an issue relating to installing kubernetes==11.0.0b1, which is a pre-release. Does conda do it when it is referenced from a environment file but not if using pip directly? Hm... Confused, but that is what the circle-ci error about not finding kubernetes.api.client is about.

---

I don't understand why we install binderhub stuff in doc-requirements.txt, why do we do this? https://github.com/jupyterhub/binderhub/commit/73072c94f5d6b1b3ed8fb45897bc16e1810e8f18

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -196,7 +196,7 @@ texinfo_documents = [
 # -- Custom scripts -------------------------------------------
 
 # Grab the latest version of the k8s and helm install instructions.
-helm_instructions = "https://raw.githubusercontent.com/jupyterhub/zero-to-jupyterhub-k8s/master/doc/source/setup-helm.rst"
+helm_instructions = "https://raw.githubusercontent.com/jupyterhub/zero-to-jupyterhub-k8s/master/doc/source/setup-jupyterhub/setup-helm.rst"
 
 resp = requests.get(helm_instructions)
 with open('./helm.txt', 'w') as ff:


choldgraf:
hmmmmm, maybe for the API documentation? (seems like this is also breaking in circleci right now?)

consideRatio:
To solve this build error, we should ensure that we don't install kubernetes==11.0.0b1, which is avoided as far as I know by not passing `--pre` to pip install -r requirements.txt, but we are using an environments.yml file now etc, perhaps that does --pre by default? Eh...

betatim:
We don't use conda to install things when building the docs.

The CircleCI config is https://github.com/jupyterhub/binderhub/blob/master/.circleci/config.yml which says it runs `pip install -r docs/doc-requirements.txt`. This `requirements.txt` lists the dependencies of BinderHub explicitly because the full list of dependencies installs things that can't be installed on RTD (pycurl for example). The confusing thing is that this requirements file explicitly lists `kubernetes==9.*` and yet 11.0.0b1 gets installed. I assume this is because `pip` doesn't compute an overall "set of dependencies" (the 'solving" step of conda). My hunch is that something else that we install causes the new version of k8s to be pulled in, but I am not sure what (or how really because to get a prerelease version you need a `--pre` somewhere ;-/)

betatim:
After `ssh`ing into the build I see the following:
```
$ cat doc/doc-requirements.txt
# documentation specific extra packages
sphinx>=1.7
recommonmark==0.4.0
sphinx-copybutton
alabaster_jupyterhub
traitlets
pandas
ruamel.yaml
# install BinderHub dependencies. We manually list them here because some
# dependencies (like pycurl) can't be installed on ReadTheDocs and aren't
# needed to build the docs.
kubernetes>=4.*
escapism
tornado
traitlets
docker
jinja2
prometheus_client
python-json-logger
jupyterhub
jsonschema
#pycurl Do not install for docs as it breaks the RTD build. Its primary use is for mocks in testing .
```
which is weird because https://github.com/jupyterhub/binderhub/blob/c80b26cafe73f253428a3bebfe4292dfb76796e3/doc/doc-requirements.txt is what is currently on master. `git log -2` shows that this branch started from a commit in September. Could it be that you started from a old master or something?

consideRatio:
Ah but why do the doc requirement need binderhub installed at all?

betatim:
Because we run autodoc to generate API docs from the code directly.

betatim:
I rebaseed your branch on the current `master`, let's see what happens.

consideRatio:
@betatim wieee!

PR #1006, opened 2019-11-19
Bump Z2JH and fix CI tests (Running != Ready)

consideRatio:
There was [a test failure](https://travis-ci.org/jupyterhub/binderhub/jobs/613744946?utm_medium=notification&utm_source=github_status) because the hub became running but later
crashed, these kinds of crashes can happen but the hub will not become
ready until the risk of it happening has gone away. So, if we wait for
the hub to be Ready as compared to only become Running, we are good.

For this to make a difference, Running must be distinguished from Ready,
which it will become in a Z2JH release where we introduced a Readiness
probe. What version of z2jh are we using?

The answer is we are using an too old one to benefit from those fixes...

Hmm... So then I had to bump z2jh, and that means a bump of JupyterHub to beyond 1.0.0.

### DIFF
--- a/helm-chart/binderhub/requirements.yaml
+++ b/helm-chart/binderhub/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: jupyterhub
-  version: "0.9-5eb48bc"
+  version: "0.9.0-alpha.1.033.fd3e470"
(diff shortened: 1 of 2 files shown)


betatim:
Rebased this onto the latest master. Lets see what happens.

PR #1009, opened 2019-11-25
add pre_launch_hook config to Launcher

bitnik:
This PR adds a callable config `pre_launch_hook` into Launcher. This hook is executed before starting the user server. I think this is useful when someone deploys a non-standard BinderHub. For example last days I am (again) working on a BinderHub deployment with persistent storage. And I need to do some checks and raise errors if necessary before starting the user server. Right now I do it with a monkey-patch which I don't like.

`pre_launch_hook` is by default None, so if not set, BinderHub works as before.


manics:
How should this be configured in the helm chart? I've tried
```
extraConfig:
  spawntest: |
    def pre_launch_hook(launcher, image, username, server_name, repo_url):
        print('pre_launch_hook', launcher, image, username, server_name, repo_url)
        assert False
    c.binderhub.launcher.Launcher.pre_launch_hook = pre_launch_hook
```
but binderhub fails:
```
AttributeError: 'LazyConfigValue' object has no attribute 'launcher'
```

bitnik:
You should `c.Launcher.pre_launch_hook = pre_launch_hook` instead of `c.binderhub.launcher.Launcher.pre_launch_hook = pre_launch_hook`.

As far as I know, any class inherits from `traitlets.config.Configurable` can be cofigured as `c.<classname>.<configname>`.

binderhub/traitlets.py
```diff
@@ -0,0 +1,18 @@
+from traitlets import TraitType
```
betatim: Is it a good idea to create an internal package with the same name as an external package we import?

I'm worried that one day we will spend forever trying to see the difference between `.traitlets` and `traitlets` :-/

However I don't have a good alternative name as a suggestion.

binderhub/launcher.py
```diff
@@ -51,6 +53,18 @@ class Launcher(LoggingConfigurable):
         Time is scaled exponentially by the retry attempt (i.e. 2, 4, 8, 16 seconds)
         """
     )
+    pre_launch_hook = Callable(
+        None,
+        config=True,
+        allow_none=True,
+        help="""
+        An optional hook function that you can implement some checks before starting user server.
```
betatim:
```suggestion
        An optional hook function that you can use to implement checks before starting a user's server. 
```

binderhub/utils.py
```diff
@@ -136,6 +139,35 @@ def url_path_join(*pieces):
     return result
 
 
+def maybe_future(obj):
```
betatim: BinderHub depends on JupyterHub so could we import this utility from there instead of copying?

binderhub/traitlets.py
bitnik: To be honest I didn't think about it, I just copied it from jupyterhub repo (https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/traitlets.py).

But when you say it, I understand. Should we keep it same with jupyterhub or I can rename it to `extra_traitlets.py`?

binderhub/utils.py
bitnik: That's a good idea. I try it now. Actually we can do the same for `Callable` traitlet.

binderhub/traitlets.py
betatim: If JupyterHub does it then maybe it isn't as much of a trap as I expect it to be. Let's keep it this way. Then at least it is a consistent trap :D

manics:
@bitnik Thanks! Functionally this works as expected:
```yaml
extraConfig:
  spawntest: |
    from tornado.httpclient import AsyncHTTPClient, HTTPRequest
    import json
    async def pre_launch_hook(launcher, image, username, server_name, repo_url):
        print('pre_launch_hook', launcher, image, username, server_name, repo_url)
        client = AsyncHTTPClient()
        token = 'SECRET_API_TOKEN'
        req = HTTPRequest(
            'https://mastodon.social/api/v1/statuses',
            'POST',
            body=json.dumps({
                "status": "{} launched by {}".format(image, username)
            }),
            headers={
                'Authorization': 'Bearer ' + token,
                'Content-Type': 'application/json',
            }
        )
        rsp = await client.fetch(req)
        print(rsp, rsp.body)
    c.Launcher.pre_launch_hook = pre_launch_hook
```
https://mastodon.social/@yetanotheraccount
yac-bh

betatim:
I think this looks nice.

One thing I am pondering and have no good idea on: should this pre launch hook be allowed to modify things (and if yes what)? It seems in JupterHub a lot of use cases for these hooks is to allow them to modify the things being passed to them/somehow communicate with their caller. If we don't have a good idea this might just be overthinking it though.

bitnik:
> should this pre launch hook be allowed to modify things (and if yes what)?

I don't understand what you mean here. Isn't this hook already allowed to modify the things passed and also to communicate back to UI?

bitnik:
@betatim what do you think after the last commit?

betatim:
Merged! I was thinking of having the callable return objects that it modified instead of modifying arguments. But lets think about that when we get there.

PR #1010, opened 2019-11-26
[MRG] Pass BINDER_REPO_URL and other vars to spawned container at runtime

manics:
Follow-up to https://github.com/jupyterhub/mybinder.org-deploy/pull/1202

`repo_url` is already passed to to the spawner:
https://github.com/jupyterhub/binderhub/blob/e0d720d8b663f2078040c973876f656207e37817/binderhub/builder.py#L512-L513

https://github.com/jupyterhub/binderhub/blob/e0d720d8b663f2078040c973876f656207e37817/binderhub/launcher.py#L155-L159

So instead of setting it at build-time (only applies to new builds) it can be set at run-time.

I've called it `BINDER_REPO_URL` instead of `REPO_URL` used in https://github.com/jupyterhub/mybinder.org-deploy/pull/1202 to make it more specific.

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -255,7 +255,23 @@ class BuildHandler(BaseHandler):
             await self.fail("Could not resolve ref for %s. Double check your URL." % key)
             return
 
-        ref_url = self.ref_url = await provider.get_resolved_ref_url()
+        self.ref_url = await provider.get_resolved_ref_url()
+        resolved_spec = await provider.get_resolved_spec()
+
+        self.binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
+            proto=self.request.protocol,
+            host=self.request.host,
+            base_url=self.settings['base_url'],
+            provider=provider_prefix,
+            spec=spec,
+        )
+        self.persistent_binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
+            proto=self.request.protocol,
+            host=self.request.host,
+            base_url=self.settings['base_url'],
+            provider=provider_prefix,
+            spec=resolved_spec,
+        )
 
         # generate a complete build name (for GitHub: `build-{user}-{repo}-{ref}`)
 
@@ -316,26 +332,11 @@ class BuildHandler(BaseHandler):
 
         BuildClass = FakeBuild if self.settings.get('fake_build') else Build
 
-        binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
-            proto=self.request.protocol,
-            host=self.request.host,
-            base_url=self.settings['base_url'],
-            provider=provider_prefix,
-            spec=spec,
-        )
-        resolved_spec = await provider.get_resolved_spec()
-        persistent_binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
-            proto=self.request.protocol,
-            host=self.request.host,
-            base_url=self.settings['base_url'],
-            provider=provider_prefix,
-            spec=resolved_spec,
-        )
         appendix = self.settings['appendix'].format(
-            binder_url=binder_url,
+            binder_url=self.binder_url,
             repo_url=repo_url,
-            persistent_binder_url=persistent_binder_url,
-            ref_url=ref_url,
+            persistent_binder_url=self.persistent_binder_url,
+            ref_url=self.ref_url,
         )
 
         self.build = build = BuildClass(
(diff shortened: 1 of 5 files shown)


bitnik:
Nice! Should we pass all those URLs to spawner and do the same in `get_env`? Then we can finally delete `appendix` completely.

manics:
@bitnik I just had the same thought! I'm investigating.

manics:
```
jovyan@jupyter-test-2dipython-2dipython-2din-2ddepth-2d7u57im2e:~$ env | grep BINDER_
BINDER_PORT=tcp://10.x.x.x:80
BINDER_PORT_80_TCP_PROTO=tcp
BINDER_SERVICE_PORT=80
BINDER_REPO_URL=https://github.com/ipython/ipython-in-depth
BINDER_PORT_80_TCP_ADDR=10.x.x.x
BINDER_PORT_80_TCP_PORT=80
BINDER_REF_URL=https://github.com/ipython/ipython-in-depth/tree/7e5ce96cc9251083979efdfc393425f1229a4a68
BINDER_PERSISTENT_URL=http://x.example.org/v2/gh/ipython/ipython-in-depth/7e5ce96cc9251083979efdfc393425f1229a4a68
BINDER_SERVICE_HOST=10.x.x.x
BINDER_PORT_80_TCP=tcp://10.x.x.x:80
BINDER_URL=http://x.example.org/v2/gh/ipython/ipython-in-depth/master
```
Added environment variables are:
- `BINDER_REPO_URL=https://github.com/ipython/ipython-in-depth`
- `BINDER_REF_URL=https://github.com/ipython/ipython-in-depth/tree/7e5ce96cc9251083979efdfc393425f1229a4a68`
- `BINDER_URL=http://x.example.org/v2/gh/ipython/ipython-in-depth/master`
- `BINDER_PERSISTENT_URL=http://x.example.org/v2/gh/ipython/ipython-in-depth/7e5ce96cc9251083979efdfc393425f1229a4a68`

@bitnik Do the value of these variables look correct to you?

bitnik:
Yes. And code change also looks good to me.

manics:
Updated the docs and the minikube config to include `get_env()` in BinderSpawner.

BinderSpawner could be a real class, e.g. in a new file `binderhub/spawner.py`. Could discuss separately?

bitnik:
> Perhaps BinderSpawner could be a real class, e.g. in a new file binderhub/spawner.py.

I think this is a good idea, so we have cleaner values.yaml and when we update the `BinderSpawner`, it is less error-prone. But as you said, I also think we should discuss it separately.

binderhub/builder.py
```diff
@@ -255,6 +255,24 @@ def escape(s):
             await self.fail("Could not resolve ref for %s. Double check your URL." % key)
             return
 
+        self.ref_url = await provider.get_resolved_ref_url()
+        resolved_spec = await provider.get_resolved_spec()
+
+        self.binder_url = '{proto}://{host}{base_url}v2/{provider}/{spec}'.format(
```
betatim: We should check if we need to/should use https://github.com/jupyterhub/binderhub/blob/2a7145f40dfc9d9face452d3f3f4502eeb4392a2/binderhub/app.py#L149-L153 instead of the request host. `badge_base_url` is used by hubs in the federation to always generate links with `https://mybinder.org/...` instead of the individual cluster's host name.

betatim: The current "Copy Binder Link" button already seems to only generate links that start with `https://mybinder.org/...`. I think this is because we try and do some clever stuff in the JS. Maybe we can simplify that if we generate the "right" URL here already?

manics:
I've split the URLs into separate variables for the Binder host and the API requests. The should help reduce the coupling between the backend and frontend.

```
BINDER_LAUNCH_HOST=http://x.example.org/
BINDER_PERSISTENT_REQUEST=v2/gh/ipython/ipython-in-depth/7e5ce96cc9251083979efdfc393425f1229a4a68
BINDER_REF_URL=https://github.com/ipython/ipython-in-depth/tree/7e5ce96cc9251083979efdfc393425f1229a4a68
BINDER_REPO_URL=https://github.com/ipython/ipython-in-depth
BINDER_REQUEST=v2/gh/ipython/ipython-in-depth/master
```

betatim:
Nitpick: Can we make the URIs start with a `/`? Would make it consistent with things like `JUPYTERHUB_OAUTH_CALLBACK_URL=/user/binder-examples-requirements-693pt67r/oauth_callback` and `JUPYTERHUB_SERVICE_PREFIX`.

doc/authentication.rst
```diff
@@ -34,6 +34,19 @@ you need to add the following into ``config.yaml``:
                     # binder service sets the image spec via user options
                     self.image = self.user_options['image']
                   return super().start()
+
+              def get_env(self):
```
bitnik: I think you forget to update here?

manics:
@betatim Can do. The reason I didn't was so that `BINDER_LAUNCH_HOST` (which includes a trailing `/` from `baseurl`) + `BINDER_PERSISTENT_REQUEST` would work without needing to remove one of the `/`

betatim:
Nods, it is annoying/tedious to have to remove an extra `/`. Does Python's `urljoin` do the right thing?

bitnik:
> Does Python's urljoin do the right thing?

No, I did that mistake recently: https://github.com/jupyterhub/mybinder.org-deploy/pull/1206

manics:
@betatim After reading up on [`urljoin`](https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin) I think these _shouldn't_ have a leading `/`. `/` indicates an absolute URL, but since this deliberately excludes the baseurl it's relative and shouldn't start with a `/`.

`JUPYTERHUB_SERVICE_PREFIX` _is_ an absolute URL, for example if you run JupyterHub under a prefix it's included: `JUPYTERHUB_SERVICE_PREFIX=/jupyter/user/test/ipython-ipython-in-depth-815uaxyt/`

betatim:
Every day is a school day!

PR #1012, opened 2019-11-27
Add "selector" to BinderHub Deployment object for k8s 1.16

### LINKED ISSUE #1011: K8s 1.16 deprecates extensions/v1beta1 for apiVersion, breaking the Helm charts
The current Helm charts don't work with [Kubernetes 1.16](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/), apparently because of the `apiVersion` deprecation:
> The v1.16 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:
> ...
> - DaemonSet, Deployment, StatefulSet, and ReplicaSet (in the **extensions/v1beta1** and **apps/v1beta2** API groups)
>   - Migrate to use the **apps/v1** API, available since v1.9. Existing persisted data can be retrieved/updated via the **apps/v1** API.

The initial error I get is:
```
% microk8s.helm install jupyterhub/binderhub --version=0.2.0-c04966f --name=binder-test --namespace=binder-test -f secret.yaml -f config.yaml
Error: validation failed: [unable to recognize "": no matches for kind "DaemonSet" in version "extensions/v1beta1", unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"]
```

I'm new to K8s, Helm and Binder. I tried replacing the versions with `apps/v1`, after downloading the helm charts, but that gives another error:
```
% wget https://jupyterhub.github.io/helm-chart/binderhub-0.2.0-c04966f.tgz
% tar zxf binderhub-0.2.0-c04966f.tgz
% grep extensions/v1beta1 binderhub -RlZ | xargs -0 grep -lZ -e DaemonSet -e Deployment -e StatefulSet -e ReplicaSet | xargs -0 sed -i '/^apiVersion/s:extensions/v1beta1:apps/v1:'
% helm install --name=binder-test --namespace=binder-test -f secret.yaml -f config.yaml ./binderhub
Error: release binder-test failed: Deployment.apps "binder" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: map[string]string{"app":"binder", "component":"binder", "heritage":"Tiller", "name":"binder", "release":"binder-test"}: `selector` does not match template `labels`]
```

I'm new to K8s, Helm and Binder. Not sure what else to do except downgrade K8s. I'm using [microk8s](https://microk8s.io/docs/), so that should be relatively straightforward.

consideRatio:
We also need to bump to a k8s 1.16 compateble z2jh helm chart release,
as the user-scheduler isn't working currently with k8s 1.16 for example.

Closes #1011

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -1,9 +1,14 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: binder
 spec:
   replicas: {{ .Values.replicas }}
+  selector:
+    matchLabels:
+      app: binder
+      component: binder
+      release: {{ .Release.Name }}
   strategy:
     rollingUpdate:
         {{- if eq (.Values.replicas | int) 1 }}
--- a/helm-chart/binderhub/templates/dind/daemonset.yaml
+++ b/helm-chart/binderhub/templates/dind/daemonset.yaml
@@ -1,5 +1,5 @@
 {{ if .Values.dind.enabled -}}
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: {{ .Release.Name }}-dind
--- a/helm-chart/binderhub/templates/image-cleaner.yaml
+++ b/helm-chart/binderhub/templates/image-cleaner.yaml
@@ -1,6 +1,6 @@
 {{ if .Values.imageCleaner.enabled -}}
 {{- $Values := .Values -}}
-apiVersion: extensions/v1beta1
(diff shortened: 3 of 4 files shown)


helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -1,9 +1,14 @@
...
 kind: Deployment
 metadata:
   name: binder
 spec:
   replicas: {{ .Values.replicas }}
+  selector:
+    matchLabels:
+      app: binder
+      component: binder
+      release: {{ .Release.Name }}
```
consideRatio: note btw that I left out some labels here, it is not considered good practice selecting on some, such as `heritage: tiller` which is good because it is gone in helm 3 :D

PR #1016, opened 2019-11-30
[MRG] Update dockerd version used for DIND

betatim:
Upgrade from 18.x to 19.x for the dockerd we use for DIND.

Keeping current is probably a good thing.

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -145,7 +145,7 @@ dind:
   daemonset:
     image:
       name: docker
-      tag: 18.09.2-dind
+      tag: 19.03.5-dind
     # Additional command line arguments to pass to dockerd
     extraArgs: []
   storageDriver: overlay2


betatim:
I kind-of tested this on a local deployment with minikube and DIND. It successfully builds the minimal dockerfile example from binder-examples but for the "requirements" example it fails on the  third step with `E: You don't have enough free space in /var/cache/apt/archives/.`. I think this is an artefact of using DIND in minikube.

Unless someone knows a reason why this shouldn't work I'd suggest we merge and deploy it and see what happens.

consideRatio:
I recall an increased security requirement or something like that regarding certificates in communication, if it works at all then im +1 because if that was an issue we would know quickly i think.

betatim:
I think we side step the certificates-by-default (which is indeed in the release notes/I saw it somewhere) topic because we explicitly set the docker socket to be a socket file, not TCP. If it is broken at such a fundamental level I think even a basic build wouldn't have worked on minikube (and will fail on staging).

choldgraf:
LGTM - thanks for the bump!

PR #969, opened 2019-10-01
Add Dataverse to UI. Fixes #900

Xarthisius:
Note: This PR depends on an unreleased version of repo2docker

This adds Dataverse DOI as a valid target for BinderHub UI. Tests are expected to fail right now as I developed it against unreleased version of WT, where there's a slight change to DOI format returned by the lookup endpoint ("doi:10..." vs "10..."). However, it doesn't affect functionality and this PR in any way. Everything should be working as is.

PR #967, opened 2019-09-28
[MRG] add hydroshare as a repo provider

sblack-usu:
repo2docker changes:
https://github.com/jupyter/repo2docker/pull/800

PR #1017, opened 2019-11-30
adding repoproviders documentation

choldgraf:
This is a documentation update for repository providers, it does these main things:

1. Updates our table of repoprovider specs to include recent additions
2. Adds a new page dedicated to repository providers, and moves a section of the API docs there (with the goal of making this information more discoverable)
3. Adds a section on adding a new repository provider, so it's easier to
   provide guideance on PRs like #969 

What do folks think?


doc/developer/repoproviders.rst
```diff
@@ -0,0 +1,60 @@
+.. _providers-section:
+
+====================
+Repository Providers
+====================
+
+Repository Providers (or RepoProviders) are
+locations where git repositories are stored (e.g.,
```
betatim:
```suggestion
locations where repositories are stored (e.g.,
```
The Zenodo or FigShare providers don't (need to) give us something that is a Git repository (contains a `.git` subdirectory). I think calling the thing we start from a "repository" is a good compromise between using something specific (c.f. "directory" is super vague I think, even though it is strictly true) and trying to move away from the idea that it has to be a git repository.

doc/developer/repoproviders.rst
```diff
@@ -0,0 +1,60 @@
...
+| Figshare   | ``figshare``       | ``<figshare-DOI>``                                          |                            |
++------------+--------------------+-------------------------------------------------------------+----------------------------+
+| Git        | ``git``            | ``<url-escaped-url>/<commit-sha>``                          | arbitrary HTTP git repos   |
++------------+--------------------+-------------------------------------------------------------+----------------------------+
+
+Adding a new repository provider
+================================
+
+It is possible to add new repository providers to BinderHub, allowing
+a BinderHub deployment to fetch git repositories from new locations
```
betatim:
```suggestion
a BinderHub deployment to fetch repositories from new locations
```

doc/developer/repoproviders.rst
```diff
@@ -0,0 +1,60 @@
...
+on the web. Doing so involves defining your own RepoProvider sub-class
+and modifying a set of methods/attributes to interface with the online
+provider to which you are providing access.
+
+In order to extend the supported repository providers,
```
betatim: Should we add a step that it is highly likely that your first step would be to add a new content provider to repo2docker?

doc/developer/repoproviders.rst
choldgraf: good point, totally agree

choldgraf:
accepted your comments and added a note about content providers. I also realized we don't have a similar section for how to add a new content provider in the repo2docker docs, so I opened https://github.com/jupyter/repo2docker/issues/823 too

nuest:
I re-visited a repo provider I've added (see #951) and found that I had to make changes to `binderhub/static/js/index.js` for the example text and `binderhub/templates/index.html` for the dropdown entry. AFAICS that's missing from the steps.

[ The JS file reminds me of the need for a proper API to generate the UI... still very good to document the current state though! ]

choldgraf:
@nuest aren't both of those addressed here: https://github.com/jupyterhub/binderhub/pull/1017/files#diff-cbeb18fbf4548b84e49c010cf02f439dR57 ? Let me know if I'm missing something!

(and I totally agree about the need to do less hard-coding etc)

nuest:
@choldgraf My bad, those files are indeed included in that changeset!

choldgraf:
anything else to do here? If we can get this in then we'll ask #967 and #969 to document themselves and then we can get those merged too

choldgraf:
I thought about it a bit more and decided to add repository provider information to the table, the new version is here:

https://643-89419368-gh.circle-artifacts.com/0/html/developer/repoproviders.html#supported-repoproviders

betatim:
Merging this now. We can keep tweaking it in a new PR.

PR #1018, opened 2019-12-01
Add instructions for running the helm chart on minikube

betatim:
Some notes I would have found useful to remind myself on how to deploy the whole BinderHub chart onto minikube to test changes to the chart itself.

As part of this I tried running with DIND enabled but that turned out to be tricky because the dockerd in the DIND pod kept running out of disk space. See #1016


choldgraf:
LGTM - thanks!

PR #1019, opened 2019-12-01
[MRG] Add a "absolute" size mode to the image cleaner

betatim:
On some deployments with DIND enabled the docker images from our builds
and the k8s dockerd are on the same partition. This means the GC of DIND
and k8s get in each others way. The "absolute" mode computes the size of
the docker directory and acts if it is too big in bytes.

The problem arises because our image cleaner and the k8s garbage collection both look at the fractional fullness of the disk. What ends up happening is that one of the docker image stores gets emptied completely because even removing all the images from it doesn't drop the disk usage low enough.

With this "absolute" mode we can configure a maximum size in GB per docker image store.

The downside is that computing the size of a tree of directories is slower than asking for the fullness of a partition.

Merging this doesn't change the default behaviour, it is an option you can opt-in to. The reason I added this is so we can use it on the OVH cluster where we have exactly the problem I described above.

### DIFF
--- a/helm-chart/binderhub/templates/image-cleaner.yaml
+++ b/helm-chart/binderhub/templates/image-cleaner.yaml
@@ -54,6 +54,8 @@ spec:
           value: /var/lib/docker
         - name: IMAGE_GC_DELAY
           value: {{ $Values.imageCleaner.delay | quote }}
+        - name: IMAGE_GC_THRESHOLD_TYPE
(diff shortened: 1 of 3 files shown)


choldgraf:
Nice - could you add a quick bit of comment context to the place where "get_used" happens so a reader know what the implication is? I'd add it myself but am on mobile

PR #1020, opened 2019-12-01
Bump chartpress to 0.5.0

consideRatio:
Chartpress 0.5.0 contains a bugfix that makes a chart's images' context
paths be considered when determening the version of the chart as well as
the chart folder. This was not done before and caused issues, that made
us require https://github.com/jupyterhub/binderhub/pull/1014 which can
now be reverted in this commit.

PR #1021, opened 2019-12-02
[Do not merge] Bump chartpress 0.5.0 - A verification

consideRatio:
This is meant to verify that the chart version is bumped along with the image version in an additional commit added to this PR over the one in #1020.

To verify the chart version is bumped, we will inspect the CI build logs about what image tag and chart version is used. We want it to be the same and not to see the hub image tag to use a later version than the chart.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,5 +6,5 @@ pytest-asyncio
 pytest-cov
 requests
 ruamel.yaml>=0.15
-chartpress>=0.4.3,<0.5
+chartpress==0.5.*
 jupyterhub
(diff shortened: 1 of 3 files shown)


consideRatio:
Note that the TravisCI builds are triggered, but this PR does isn't updated as it should with the travis CI status but only contains OK from CircleCI, something is up with that, which is unrelated to this PR.

Here is a build along with logs allowing us to verify if it works as it should or not.
https://travis-ci.org/jupyterhub/binderhub/jobs/619578001

consideRatio:
I tested this locally and conclude that all is good.

```
Updating binderhub/Chart.yaml: version: 0.2.0-023.15040e3
Updating binderhub/values.yaml: imageCleaner.image.tag: {'repository': 'jupyterhub/k8s-image-cleaner', 'tag': '0.2.0-022.21260bb'}
Updating binderhub/values.yaml: image.tag: {'repository': 'jupyterhub/k8s-binderhub', 'tag': '0.2.0-023.15040e3'}
```

consideRatio:
I

PR #1023, opened 2019-12-04
[MRG] Reduce repeated code in BinderSpawner

### LINKED ISSUE #1015: Make BinderSpawner a real class
As discovered in https://github.com/jupyterhub/binderhub/pull/1010 `BinderSpawner` is defined in-line in three different places:
- https://github.com/jupyterhub/binderhub/blob/9756642c9b2789695af3edf2ba6802765dbf9531/helm-chart/binderhub/values.yaml#L79
- https://github.com/jupyterhub/binderhub/blob/9756642c9b2789695af3edf2ba6802765dbf9531/testing/minikube/jupyterhub-helm-config.yaml#L23
- https://github.com/jupyterhub/binderhub/blob/9756642c9b2789695af3edf2ba6802765dbf9531/doc/authentication.rst#enabling-authentication

We could reduce duplication in this repo by making it a class, e.g. in `binder/spawner.py`. As part of this a parameter `enable_auth: bool` could be added so it's easier to configure

manics:
Closes #1015 

There should be no change in behaviour for the default `BinderSpawner` or for the `BinderSpawner` described in https://binderhub.readthedocs.io/en/latest/authentication.html

### DIFF
--- a/doc/customizing.rst
+++ b/doc/customizing.rst
@@ -27,10 +27,20 @@ For example::
 
   binderhub:
     jupyterhub:
-      extraConfig:
-        10-binder-customisations:
-          class MyCustomBinderSpawner(BinderSpawner):
-              ...
+      hub:
+        extraConfig:
+          10-binder-customisations: |
+            class MyCustomBinderSpawner(BinderSpawner):
+                ...
+
+            c.JupyterHub.spawner_class = MyCustomBinderSpawner
+
+BinderHub uses the `jupyterhub.hub.extraConfig setting
+<https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/advanced.html#hub-extraconfig>`_
+to customise JupyterHub.
+For example, ``BinderSpawner`` is defined under the ``00-binder`` key.
+Keys are evaluated in alphanumeric order, so later keys such as
+``10-binder-customisations`` can use objects defined in earlier keys.
 
 About page customization
 ------------------------
--- a/doc/customizing.rst
+++ b/doc/customizing.rst
@@ -22,6 +22,15 @@ For information on how to configure your JupyterHub deployment, see the
 `JupyterHub for Kubernetes Customization Guide
 <https://zero-to-jupyterhub.readthedocs.io/en/latest/#customization-guide>`_.
 
+If you want to customise the spawner you can subclass it in ``extraConfig``.
+For example::
+
+  binderhub:
+    jupyterhub:
+      extraConfig:
+        10-binder-customisations:
+          class MyBinderSpawner(BinderSpawner):
+              ...
 
 About page customization
 ------------------------
--- a/doc/customizing.rst
+++ b/doc/customizing.rst
@@ -29,7 +29,7 @@ For example::
     jupyterhub:
       extraConfig:
         10-binder-customisations:
-          class MyBinderSpawner(BinderSpawner):
+          class MyCustomBinderSpawner(BinderSpawner):
               ...
 
 About page customization
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -65,9 +65,6 @@ jupyterhub:
   hub:
     extraConfig:
       00-binder: |
-        import os
-        import sys
-        import yaml
         from tornado import web
 
         # get custom config from values.custom
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -59,58 +59,65 @@ jupyterhub:
     users: true
   custom:
     cors: *cors
+    binderauth_enabled: false
   rbac:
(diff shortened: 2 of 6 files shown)


doc/customizing.rst
```diff
@@ -22,6 +22,15 @@ For information on how to configure your JupyterHub deployment, see the
 `JupyterHub for Kubernetes Customization Guide
 <https://zero-to-jupyterhub.readthedocs.io/en/latest/#customization-guide>`_.
 
+If you want to customise the spawner you can subclass it in ``extraConfig``.
```
betatim: I tried to write a short explanation of what the reason for having 10-, 20- etc is. What do you think?

The keys (like `10-binder-customisations`) in the `extraConfig` dictionary will be merged
in alphabetical order. This means that you can use configuration from earlier keys in later ones.
In this case our custom spawner inherits from `BinderSpawner` which is defined in `00-binder`

Something like that?

doc/customizing.rst
```diff
@@ -22,6 +22,15 @@ For information on how to configure your JupyterHub deployment, see the
...
+For example::
+
+  binderhub:
+    jupyterhub:
+      extraConfig:
+        10-binder-customisations:
+          class MyBinderSpawner(BinderSpawner):
```
betatim:
```suggestion
          class MyCustomBinderSpawner(BinderSpawner):
```
just to avoid any chance of confusing this with the spawner that is used on mybinder.org :D

testing/minikube/jupyterhub-helm-config.yaml
```diff
@@ -3,80 +3,23 @@
...
+  cors:
+    allowOrigin: "*"
+
 hub:
   cookieSecret: "36091e950f98f033aeae2520e2fe4c8599bc391598f910e510ad670765fbc1ff"
   db:
     type: "sqlite-memory"
   services:
     binder:
-      admin: true
```
betatim: Isn't this what makes the binder service an "admin" in the eyes of JupyterHub?

testing/minikube/install-hub
```diff
@@ -43,11 +45,21 @@ jupyterhub = get_hub_chart_dependency()
 check_call(['helm', 'repo', 'add', 'jupyterhub', jupyterhub['repository']])
 check_call(['helm', 'repo', 'update', 'jupyterhub'])
 
+# Extract the JupyterHub config from the helm chart values.yaml
```
betatim: Could we extend this comment to explain why we need to do this? It would help me (right now) because I can't quite work out why we need to extract it from the chart and then load it via an argument.

(I think it is because we don't actually install the BinderHub chart when we run binderhub on minikube in "dev mode")

betatim:
I left a few comments but nothing major. Let's see what others think as well.

testing/minikube/jupyterhub-helm-config.yaml
bitnik: I think, with the changes in `install-hub` file (https://github.com/jupyterhub/binderhub/pull/1023/files#diff-971e2aef3f7ade4d8bed25e7e6d1c858) now it is taken from values.yaml. So this line can be deleted

helm-chart/binderhub/values.yaml
```diff
@@ -59,58 +59,65 @@ jupyterhub:
...
+    binderauth_enabled: false
   rbac:
     enabled: true
   hub:
     extraConfig:
-      binder: |
+      00-binder: |
         import os
         import sys
         import yaml
```
bitnik: these 3 packages are not used

testing/minikube/jupyterhub-helm-auth-config.yaml
```diff
@@ -5,6 +5,8 @@ hub:
     binder:
       oauth_redirect_uri: "http://127.0.0.1:8585/oauth_callback"
       oauth_client_id: "binder-oauth-client-test"
+custom:
+  binderauth_enabled: true
```
bitnik: should we add here `singleuser.cmd`?
```yaml
singleuser:
  cmd: jupyterhub-singleuser
```

testing/minikube/jupyterhub-helm-config.yaml
manics: That's right, it's merged with the default config so only the overrides are needed.

testing/minikube/install-hub
manics: Done. As for _why_ it's deployed separately, I've no idea

manics:
Thanks for reviewing, I think that's all requested changes made.

meeseeksmachine:

https://discourse.jupyter.org/t/a-persistent-binderhub-deployment/2865/7

meeseeksmachine:

https://discourse.jupyter.org/t/a-persistent-binderhub-deployment/2865/9

PR #1024, opened 2019-12-04
[MRG] travis: test all branches

manics:
This is useful when developing on forks. The push/publish functionality in this chart is already protected by a conditional so it should be safe to enable testing of all branches:
https://github.com/jupyterhub/binderhub/blob/bcadc594b970cb75f084564b4843001552e62c8a/ci/test-helm#L40

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ git:
 services:
   - docker
 sudo: required
-branches:
-  only:
-    - master
 
 before_install:
   - nvm install 8; nvm use 8


betatim:
When do we need this? For travis runs in a fork?

I think we added this because sometimes people do create a new branch in the "main" repository to fix a small typo or so and then make a PR. That PR used to run all tests twice.

manics:
YesCorrect, it's so anyone developing in their own fork/branch can have their changes tested by Travis. This is how I normally work.

Currently the only way to have Travis test your changes on a fork are:
- push to your own `master`: not great if you want to keep your `master` branch clean. Also means you can only test one branch at a time
- open a PR against `jupyterhub/binderhub`: adds unnecessary noise to the repo, also takes up Travis slots on this repo
- change `.travis.yml` on your branch, revert the change if a PR is opened.

Equivalent change on Z2JH: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/323#pullrequestreview-81528714

betatim:
Let's merge this and see how we do. I think branches-in-upstream is an anti-pattern so discouraging it is Ok with me :) As long as nothing bad happens when someone does do it I am happy.

I agree with you that I think the `if` statement should protect us.

ps. opening a WIP PR should always be ok, even if it is just to get CI services. I don't think of it as noise. If we are too eager to comment on WIP PRs and create noise that way we should try and reduce that.

manics:
> I think branches-in-upstream is an anti-pattern so discouraging it is Ok with me :)

I completely agree!

PR #1026, opened 2019-12-04
reduce configuration of authentication with named servers

bitnik:
This PR removes `use_named_servers` config from binderhub and makes `Launcher` to get it from an env variable which is set by `jupyterhub.hub.allowNamedServers`.

### DIFF
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -35,6 +35,12 @@ class Launcher(LoggingConfigurable):
     hub_api_token = Unicode(help="The API token for the Hub")
     hub_url = Unicode(help="The URL of the Hub")
     create_user = Bool(True, help="Create a new Hub user")
+    allow_named_servers = Bool(
+        os.getenv('JUPYTERHUB_ALLOW_NAMED_SERVERS', False),
+        config=True,
+        help="Named user servers are allowed. This is used only when authentication is enabled and "
+             "to set unique names for user servers."
+    )
     retries = Integer(
         4,
         config=True,
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -36,7 +36,7 @@ class Launcher(LoggingConfigurable):
     hub_url = Unicode(help="The URL of the Hub")
     create_user = Bool(True, help="Create a new Hub user")
     allow_named_servers = Bool(
-        os.getenv('JUPYTERHUB_ALLOW_NAMED_SERVERS', False),
+        os.getenv('JUPYTERHUB_ALLOW_NAMED_SERVERS', "false") == "true",
         config=True,
         help="Named user servers are allowed. This is used only when authentication is enabled and "
              "to set unique names for user servers."
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -103,7 +103,7 @@ spec:
         - name: JUPYTERHUB_OAUTH_CALLBACK_URL
           value: {{ .Values.jupyterhub.hub.services.binder.oauth_redirect_uri | quote}}
         - name: JUPYTERHUB_ALLOW_NAMED_SERVERS
-          value: {{ .Values.jupyterhub.hub.allowNamedServers }}
+          value: {{ .Values.jupyterhub.hub.allowNamedServers | quote }}
         {{- end }}
         {{ if .Values.extraEnv }}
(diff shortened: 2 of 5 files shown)


helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -102,6 +102,8 @@ spec:
           value: {{ .Values.jupyterhub.hub.services.binder.oauth_client_id | quote}}
         - name: JUPYTERHUB_OAUTH_CALLBACK_URL
           value: {{ .Values.jupyterhub.hub.services.binder.oauth_redirect_uri | quote}}
+        - name: JUPYTERHUB_ALLOW_NAMED_SERVERS
+          value: {{ .Values.jupyterhub.hub.allowNamedServers }}
```
manics:
```suggestion
          value: {{ .Values.jupyterhub.hub.allowNamedServers | quote }}
```
Needs to be quoted otherwise `true` is converted to a boolean, but it needs to be kept as a string since it's an env-var.

manics:
Spoke too soon... ran into another problem:
```
traitlets.traitlets.TraitError: The 'allow_named_servers' trait of a Launcher instance must be a boolean, but a value of 'true' <class 'str'> was specified.

`jupyterhub` config:
```yaml
jupyterhub:
  custom:
    binderauth_enabled: true
  hub:
    baseUrl: /jupyter/
    redirectToServer: false
    services:
      binder:
        oauth_redirect_uri: "http://example.org/oauth_callback"
    # Allow up to 2 servers per authenticated user
    allowNamedServers: true
    namedServerLimitPerUser: 2
```

bitnik:
@manics sorry, I forgot about that. Now I tested this locally and it works.

And you are right we have to find a way to add tests for these cases.

binderhub/launcher.py
```diff
@@ -35,6 +35,12 @@ class Launcher(LoggingConfigurable):
     hub_api_token = Unicode(help="The API token for the Hub")
     hub_url = Unicode(help="The URL of the Hub")
     create_user = Bool(True, help="Create a new Hub user")
+    allow_named_servers = Bool(
+        os.getenv('JUPYTERHUB_ALLOW_NAMED_SERVERS', "false") == "true",
```
manics: It feels like there should be a standard way of converting an env-var to a bool, for instance to handle `False|True|0|1` etc which might be an issue if binderhub is run independently of the helm chart. 

I'm not aware of anything though, and since the helm chart should always pass `true|false` this is fine for now unless anyone else has an quick fix.

bitnik:
@betatim what do you think? can we merge this?

binderhub/launcher.py
betatim: I wish for a solution to this as well (in particular in other projects at work) but I've never been able to find anything ;(

betatim:
I've not deployed a BinderHub with auth/named servers in a long time and won't get around to trying (and thinking about) this before I leave for my holiday. Maybe @minrk has a moment to do so.

If the layout/format of the yaml files changes (where you define something) we should figure out a way to emit a helpful error message to users who upgrade and wonder why everything breaks. Also, it should not quietly (but in a broken state) keep working if someone uses their old helm chart. (I've not looked closely enough to know if you changed how the helm chart values are organised so this is a "shot into the dark", ignore it if it is irrelevant :))

Otherwise, if you think this is ready to go: merge it. When we find bugs we can always fix them later.

manics:
> If the layout/format of the yaml files changes (where you define something) we should figure out a way to emit a helpful error message to users who upgrade and wonder why everything breaks.

I think this is one example where having defined releases helps, otherwise there's ambiguity about whether something that's changed in one PR and reverted/modified in another PR a week later is "breaking".

betatim:
> I think this is one example where having defined releases helps

Agreed. In this case I was thinking "are we changing something that has been that way for a long time"? What is a long time? Months?

If we have releases we need to come up with a policy for deprecating features (marked as deprecated in release N, emits a warning if you turn on warnings, release N+1 always emits a warning, and then in release N+2 using it is "broken") and giving people advice on how to transition. We could build that infrastructure pretending that each merge to master is a release and then maybe do time based releases? I don't think just doing the semver thing of bumping the version numbering appropriately is enough because it doesn't solve the problem that an update breaks people's setup and not everyone can update their setup "over night".

(Maybe this conversation should be in its own issue though)

PR #1027, opened 2019-12-11
test repo provider methods

bitnik:
Hi, I added a test to check if a repo provider have all required methods defined. I thought this would be helpful while adding new repo providers.

### DIFF
--- a/binderhub/tests/test_repoproviders.py
+++ b/binderhub/tests/test_repoproviders.py
@@ -11,6 +11,23 @@ from binderhub.repoproviders import (
 )
 
 
+@pytest.mark.parametrize(
+    'required_method', [
+        "get_build_slug",
+        "get_repo_url",
+        "get_resolved_spec",
+        "get_resolved_ref",
+        "get_resolved_ref_url",
+    ]
+)
+async def test_repo_provider_methods(app, required_method):
+    """
+    Test if each repo provider has required methods defined
+    """
+    for rp in app.repo_providers.values():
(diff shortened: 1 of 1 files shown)


binderhub/tests/test_repoproviders.py
```diff
@@ -11,6 +11,23 @@
...
+        "get_resolved_ref",
+        "get_resolved_ref_url",
+    ]
+)
+async def test_repo_provider_methods(app, required_method):
+    """
+    Test if each repo provider has required methods defined
+    """
+    for rp in app.repo_providers.values():
+        assert callable(getattr(rp, required_method, None)), f"{rp.__name__} doesn't have method {required_method}"
```
manics: I think these tests will always pass because the methods are defined in the base class:
https://github.com/jupyterhub/binderhub/blob/8c51534a9517d40f82fa2546e99660e88d94f5e7/binderhub/repoproviders.py#L164-L184

An alternative might be make `RepoProvider` an abstract base class: https://docs.python.org/3/library/abc.html#abc.abstractmethod

I haven't used them before, but the docs suggests `abc.ABC` can be used as a mixin.

bitnik: ah I thought it would raise `NotImplementedError` but they are actually not called. It looks like I didn't think about this enough, sorry. I will do it, thanks @manics !

consideRatio:
This seems like a PR that doesn't accomplish what was intended, and has been left open for a long time. I'll go for a close here!

Issue #1022, opened 2019-12-03
If an authenticated user has namedServerLimitPerUser servers you see a generic "Failed to launch image"


manics:
```yaml
config:
  BinderHub:
    use_registry: true
    image_prefix: xxx/yyy-
    hub_url_public: http://public.example.org/jupyter/
    hub_url: http://hub:8081/jupyter/
    log_level: 0
    debug: true
    use_named_servers: true
    auth_enabled: true

jupyterhub:
  hub:
    baseUrl: /jupyter/
    redirectToServer: false
    services:
      binder:
        oauth_redirect_uri: "http://public.example.org/oauth_callback"
        oauth_client_id: "binder-oauth-zzz"
    extraConfig:
      binder: |
        from kubespawner import KubeSpawner

        class BinderSpawner(KubeSpawner):

          def start(self):
              if 'image' in self.user_options:
                # binder service sets the image spec via user options
                self.image = self.user_options['image']
              return super().start()

          def get_env(self):
              env = super(BinderSpawner, self).get_env()
              if 'repo_url' in self.user_options:
                  env['BINDER_REPO_URL'] = self.user_options['repo_url']
              for key in (
                      'binder_ref_url',
                      'binder_launch_host',
                      'binder_persistent_request',
                      'binder_request'):
                  if key in self.user_options:
                      env[key.upper()] = self.user_options[key]
              return env

        c.JupyterHub.spawner_class = BinderSpawner
    # Allow up to 2 servers per authenticated user
    allowNamedServers: true
    namedServerLimitPerUser: 2
```

If you already have two servers running you see the generic message
```
Found built image, launching...
Launching server...
Launch attempt 1 failed, retrying...
Launch attempt 2 failed, retrying...
Launch attempt 3 failed, retrying...
Failed to launch image xxx/yyy-zzz:tag
```
which gives the impression there was a problem running the image, and may lead the user to waste time rebuilding the image.

The actual problem is shown in the binderhub logs:
```
    b'{"status": 400, "message": "User test already has the maximum of 2 named servers.  One must be deleted before a new server can be created"}'
```
This should be indicated to the user.

PR #1029, opened 2019-12-12
show an explanatory message if an authenticated user exceeds the limit of named servers

bitnik:
closes https://github.com/jupyterhub/binderhub/issues/1022

### DIFF
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -41,6 +41,11 @@ class Launcher(LoggingConfigurable):
         help="Named user servers are allowed. This is used only when authentication is enabled and "
              "to set unique names for user servers."
     )
+    named_server_limit_per_user = Integer(
+        os.getenv('JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER', 0),
+        config=True,
+        help="""Maximum number of concurrent named servers that can be created by a user at a time."""
+    )
     retries = Integer(
         4,
         config=True,
@@ -172,6 +177,19 @@ class Launcher(LoggingConfigurable):
             user_data = await self.get_user_data(username)
             if server_name in user_data['servers']:
                 raise web.HTTPError(409, "User %s already has a running server." % username)
+        elif self.named_server_limit_per_user > 0:
+            # authentication is enabled with named servers
+            # check if user has already reached to the limit of named servers
+            user_data = await self.get_user_data(username)
+            len_named_spawners = len([s for s in user_data['servers'] if s != ''])
+            if self.named_server_limit_per_user <= len_named_spawners:
+                raise web.HTTPError(
+                    409,
+                    "User {} already has the maximum of {} named servers."
+                    "  One must be deleted before a new server can be created".format(
+                        username, self.named_server_limit_per_user
+                    ),
+                )
 
         if self.pre_launch_hook:
             await maybe_future(self.pre_launch_hook(self, image, username, server_name, repo_url))
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -42,7 +42,7 @@ class Launcher(LoggingConfigurable):
              "to set unique names for user servers."
     )
     named_server_limit_per_user = Integer(
-        os.getenv('JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER', 0),
+        int(os.getenv('JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER', 0)),
         config=True,
         help="""Maximum number of concurrent named servers that can be created by a user at a time."""
     )
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -44,7 +44,7 @@ class Launcher(LoggingConfigurable):
     named_server_limit_per_user = Integer(
         int(os.getenv('JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER', 0)),
         config=True,
-        help="""Maximum number of concurrent named servers that can be created by a user at a time."""
+        help="""Maximum number of concurrent named servers that can be created by a user."""
     )
     retries = Integer(
         4,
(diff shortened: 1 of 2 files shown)


helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -102,8 +102,12 @@ spec:
           value: {{ .Values.jupyterhub.hub.services.binder.oauth_client_id | quote}}
         - name: JUPYTERHUB_OAUTH_CALLBACK_URL
           value: {{ .Values.jupyterhub.hub.services.binder.oauth_redirect_uri | quote}}
+        {{- if .Values.jupyterhub.hub.allowNamedServers }}
         - name: JUPYTERHUB_ALLOW_NAMED_SERVERS
-          value: {{ .Values.jupyterhub.hub.allowNamedServers | quote }}
+          value: true
```
manics:
```suggestion
          value: "true"
```
Environment variables must always be strings so `true` needs to be quoted, otherwise you get errors like:
```
Error: UPGRADE FAILED: cannot patch "binder" with kind Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found t, error found in #10 byte of ...|,"value":true},{"nam|..., bigger context ...|{"name":"JUPYTERHUB_ALLOW_NAMED_SERVERS","value":true},{"name":"JUPYTERHUB_NAMED_SERVER_LIMIT_PER_US|...
```
FYI I'm using Helm 3, does it work on Helm 2?

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -102,8 +102,12 @@ spec:
...
+        - name: JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER
+          value: {{ .Values.jupyterhub.hub.namedServerLimitPerUser }}
```
manics:
```suggestion
          value: {{ .Values.jupyterhub.hub.namedServerLimitPerUser | quote  }}
```
As before this needs to be quoted other I get an error. However after adding `| quote` I run into an error in BinderHub:
```
traitlets.traitlets.TraitError: The 'named_server_limit_per_user' trait of a Launcher instance must be an int, but a value of '2' <class 'str'> was specified.
```

betatim: This is weird :-/ Are you sure you tried this with the old way of specifying the value? My guess would be that it didn't use to work previously either. `os.getenv("JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER", 0)` should always return a string (if the variable is set). So we should always have had this problem.

Maybe the code to fetch the variable should be `int(os.getenv("JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER", "0"))` instead. Which makes me wonder: can't the traitlet take care of fetching values from the env for us and dealing with the conversion? (I suspect not)

manics: I might've been unclear, "as before" refers to my [earlier comment in this PR on quoting](https://github.com/jupyterhub/binderhub/pull/1029#discussion_r372679000). `JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER` is new in this PR.

I agree it would be nice if traitlets could optionally cast values, but they don't.

helm-chart/binderhub/templates/deployment.yaml
bitnik: no, it doesn't work with Helm 2 neither

bitnik:
@manics @betatim thanks, I made changes that you suggested. Before I didn't have time to test the chart, but now I did and it works.

binderhub/launcher.py
```diff
@@ -41,6 +41,11 @@ class Launcher(LoggingConfigurable):
         help="Named user servers are allowed. This is used only when authentication is enabled and "
              "to set unique names for user servers."
     )
+    named_server_limit_per_user = Integer(
+        int(os.getenv('JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER', 0)),
+        config=True,
+        help="""Maximum number of concurrent named servers that can be created by a user at a time."""
```
betatim:
```suggestion
        help="""Maximum number of concurrent named servers that can be created by a user."""
```

binderhub/launcher.py
```diff
@@ -172,6 +177,19 @@ def unique_name_from_repo(self, repo_url):
             user_data = await self.get_user_data(username)
             if server_name in user_data['servers']:
                 raise web.HTTPError(409, "User %s already has a running server." % username)
+        elif self.named_server_limit_per_user > 0:
+            # authentication is enabled with named servers
+            # check if user has already reached to the limit of named servers
+            user_data = await self.get_user_data(username)
+            len_named_spawners = len([s for s in user_data['servers'] if s != ''])
```
betatim: For my education: when does it happen that a list item is an empty string here? Was expecting `user_data['servers']` to contain the list of server names, which should always be a none empty string.

```suggestion
            len_named_spawners = len([s for s in user_data['servers'] if s])
```
For even more Pythonicness (in a follow up PR if you  like)

betatim:
LGTM!

Agree with manics that it would be nice to  show this error earlier but after a quick read of this PR I am not sure why it doesn't happen earlier :-/ Something for a new pr.

binderhub/launcher.py
bitnik: > For my education: when does it happen that a list item is an empty string here?

Default server's name is an empty string by default. In auth bhub with named servers, a user can always go to control panel and do "Start My Server" and this is also included in `user_data['servers']`.

meeseeksmachine:

https://discourse.jupyter.org/t/display-error-message-to-the-user-when-max-number-of-named-servers-is-reached/10479/2

PR #1031, opened 2019-12-14
[MRG]: badge_base_url can be a function that gets the parameter from the HTTP RequestHandler

manics:
This allows the `badge_base_url` setting and therefore the `BINDER_LAUNCH_HOST` environment variable to be set from a HTTP request, and is intended to allow a BinderHub to have different values of the variable depending on the original source of the request.

For example, Gesis currently has a mix of `https://mybinder.org` and `https://notebooks.gesis.org/binder` in different places. By changing `badge_base_url` to check for a parameter in the request it should be possible to have the URLs match consistently depending on whether the user was redirected from mybinder (e.g. the redirector could set a query parameter  `binder_launch_host=https://mybinder.org/`) or directly (no parameter, default behaviour).

For example:
```
    def badge_base_url(handler):
        launch_host = (
            handler.request.headers.get('X-Binder-Launch-Host', '') or
            handler.get_argument('binder_launch_host', ''))
        print('launch_host', launch_host)
        if launch_host and not launch_host.endswith('/'):
            launch_host += '/'
        if launch_host == 'https://mybinder.org/':
            return launch_host
        return ''
    c.BinderHub.badge_base_url = badge_base_url
```

See discussion starting from https://github.com/jupyterhub/mybinder.org-deploy/pull/1284#issuecomment-560273078

I've tested this locally using curl:
`curl http://binder-test-host/build/gh/manics/jupyter-offlinenotebook/7ba3520553c414d0b55e80b96c0f6df85f903078?binder_launch_host=https://mybinder.org`
this will spit out a URL and token that can be copied into a browser.

To work with the mybinder federation this will require the parameter to be passed in proxy requests made by the redirector:
- https://github.com/jupyterhub/mybinder.org-deploy/blob/f92e921b031876ad7dd9f5ac05b33eba793c12a6/images/federation-redirect/app.py#L117
- https://github.com/jupyterhub/mybinder.org-deploy/blob/f92e921b031876ad7dd9f5ac05b33eba793c12a6/images/federation-redirect/app.py#L166

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -258,7 +258,14 @@ class BuildHandler(BaseHandler):
         self.ref_url = await provider.get_resolved_ref_url()
         resolved_spec = await provider.get_resolved_spec()
 
-        self.binder_launch_host = self.settings['badge_base_url'] or '{proto}://{host}{base_url}'.format(
+        launch_host = None
+        if self.settings['launch_host_allowed']:
+            launch_host_hdr = self.request.headers.get('X-Binder-Launch-Host', '')
+            if self.settings['launch_host_allowed'](launch_host_hdr):
+                launch_host = launch_host_hdr
+                if launch_host and not launch_host.endswith('/'):
+                    launch_host += '/'
+        self.binder_launch_host = launch_host or self.settings['badge_base_url'] or '{proto}://{host}{base_url}'.format(
             proto=self.request.protocol,
             host=self.request.host,
             base_url=self.settings['base_url'],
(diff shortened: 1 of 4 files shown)


binderhub/builder.py
```diff
@@ -258,7 +258,10 @@ def escape(s):
         self.ref_url = await provider.get_resolved_ref_url()
         resolved_spec = await provider.get_resolved_spec()
 
-        self.binder_launch_host = self.settings['badge_base_url'] or '{proto}://{host}{base_url}'.format(
+        badge_base_url = self.settings['badge_base_url']
+        if callable(badge_base_url):
+            badge_base_url = badge_base_url(self)
```
bitnik: I think it would be better to put these 3 lines into a method/property `badge_base_url` of `BaseHandler` and use it

manics: Done!

PR #1033, opened 2019-12-18
re-organize requirements files

bitnik:
I tried to re-organize requirements files and reduce them

1. added instruction to install `helm-chart/images/binderhub/requirements.txt` when installing BinderHub locally with minikube. this is also what [travis does](https://github.com/jupyterhub/binderhub/blob/fb977f4642638da69f184fdc7a7fd3c26a157a9f/.travis.yml#L19-L22) and I think they should be same f we want to have same results from tests.

2. removed `pycurl` from `requirements.txt` because it is already included in `helm-chart/images/binderhub/requirements.txt`. with this we don't have to manually update `doc-requirements.txt` each time `requirements.txt` is updated.

3. removed `jupyterhub` from `dev-requirements.txt` because it is not required for building docs, it is already included in `helm-chart/images/binderhub/requirements.txt` and they (`dev-requirements.txt` and `helm-chart/images/binderhub/requirements.txt`) are installed together.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,7 +110,7 @@ every day development.
 * Install JupyterHub in minikube with helm `./testing/minikube/install-hub`
 * Setup `docker` to use the same Docker daemon as your minikube cluster `eval $(minikube docker-env)`
 * Start BinderHub `python3 -m binderhub -f testing/minikube/binderhub_config.py`
-* Visit your BinderHub at[http://localhost:8585](http://localhost:8585)
+* Visit your BinderHub at [http://localhost:8585](http://localhost:8585)
 
 To execute most of our test suite you need a running minikube cluster.
 It does not need to have anything installed on it though:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -217,7 +217,7 @@ sudo apt install socat
 1. Install BinderHub and its development requirements:
 
     ```bash
-    python3 -m pip install -e . -r dev-requirements.txt -r helm-chart/images/binderhub/requirements.txt
+    python3 -m pip install -e . -r dev-requirements.txt
     ```
 
 1. Install JupyterHub in minikube with helm
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,7 +141,7 @@ all images involved and deploy it locally. Steps to do this:
 1. setup docker to user the minikube dockerd `eval $(minikube docker-env)`
 1. build the helm chart `cd helm-chart && chartpress && cd ..`
 1. install the BinderHub chart with
-```
+```bash
 helm install \
   --name binder-test \
   --namespace binder-test-helm \
@@ -217,7 +217,7 @@ sudo apt install socat
 1. Install BinderHub and its development requirements:
 
     ```bash
-    python3 -m pip install -e . -r dev-requirements.txt
+    python3 -m pip install -e . -r dev-requirements.txt -r helm-chart/images/binderhub/requirements.txt
     ```
 
 1. Install JupyterHub in minikube with helm
--- a/doc/doc-requirements.txt
+++ b/doc/doc-requirements.txt
@@ -6,17 +6,7 @@ traitlets
 pandas
 ruamel.yaml
 git+https://github.com/pandas-dev/pandas-sphinx-theme.git@master
-# install BinderHub dependencies. We manually list them here because some
-# dependencies (like pycurl) can't be installed on ReadTheDocs and aren't
-# needed to build the docs.
-kubernetes==9.0.*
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -217,7 +217,7 @@ sudo apt install socat
 1. Install BinderHub and its development requirements:
 
     ```bash
-    python3 -m pip install -e . -r dev-requirements.txt
+    python3 -m pip install -e . -r dev-requirements.txt -r helm-chart/images/binderhub/requirements.txt
```
manics: It should be possible to nest requirements files, try adding
```
-r helm-chart/images/binderhub/requirements.txt
```
to the end of `dev-requirements.txt`?

bitnik: @manics I did it, but couldn't decide if I should update [.travis.yaml](https://github.com/jupyterhub/binderhub/blob/0b5080ade21dbda5ad159e675bacf25c61d8e422/.travis.yml#L19-L22) (by removing line 22 and replacing line 19 with `pip install --upgrade . -r dev-requirements.txt`) or leave it as it is.

manics:
Hmm, this might require some investigation:
https://github.com/jupyterhub/binderhub/blob/0b5080ade21dbda5ad159e675bacf25c61d8e422/.travis.yml#L19-L22
The `npm` commands above are also in `setup.py`:
https://github.com/jupyterhub/binderhub/blob/0b5080ade21dbda5ad159e675bacf25c61d8e422/setup.py#L21-L26

You could try removing all four travis lines and see if `pip install --upgrade -r dev-requirements.txt` will automatically run the required `npm` commands?

If it doesn't then you could do what you suggested, remove line 19 and update 22? If `npm` is only generating Javascript/CSS assets it shouldn't matter that the python dependencies are missing.

doc/doc-requirements.txt
```diff
@@ -6,17 +6,7 @@ traitlets
 pandas
 ruamel.yaml
 git+https://github.com/pandas-dev/pandas-sphinx-theme.git@master
-# install BinderHub dependencies. We manually list them here because some
-# dependencies (like pycurl) can't be installed on ReadTheDocs and aren't
-# needed to build the docs.
-kubernetes==9.0.*
```
betatim: How will these dependencies now get installed on read the docs?

requirements.txt
```diff
@@ -10,4 +8,3 @@ prometheus_client
 python-json-logger
 jupyterhub
 jsonschema
-pycurl
```
betatim: I was expecting to see this line be replaced with an include/reference to another requirements file to make sure pycurl continues to be installed. What did I miss?

bitnik:
> You could try removing all four travis lines and see if pip install --upgrade -r dev-requirements.txt will automatically run the required npm commands?

No it doesn't run npm commands in that case. It does when `-e/--editable` flag is added.

doc/doc-requirements.txt
bitnik: with `-e .`, setup.py installs `requirements.txt` https://github.com/jupyterhub/binderhub/blob/0b5080ade21dbda5ad159e675bacf25c61d8e422/setup.py#L50

Actually now I am thinking we even don't need here `-e/--editable` flag, only `.` should be enough. what do you think?

requirements.txt
bitnik: right, I would expect that too and now I am confused.

`pycurl` is installed in `helm-chart/images/binderhub/requirements.txt`
https://github.com/jupyterhub/binderhub/blob/0b5080ade21dbda5ad159e675bacf25c61d8e422/helm-chart/images/binderhub/requirements.txt#L1
And when I go though the docs, ["User interface changes"](https://github.com/jupyterhub/binderhub/blob/master/CONTRIBUTING.md#user-interface-changes) is the only place that we only install `requirements.txt` and there we don't need to have pycurl. For example, when we run tests locally, we have to install `dev-requirements.txt` which installs also `helm-chart/images/binderhub/requirements.txt`.

bitnik:
after learning that `pip install -e .` also runs npm commands in setup.py, I think npm instructions in [User interface changes](https://github.com/jupyterhub/binderhub/blob/master/CONTRIBUTING.md#user-interface-changes) section in docs is redundant. But I think it is nice to have it explicitly, so people now how these js and css files are generated.

yuvipanda:
Are you interested in rebasing this, @bitnik? Am happy to review and get it merged :)

bitnik:
hey :) I will try to work on this and the other PR tomorrow.

bitnik:
While rebasing, I realised that almost all changes in this PR are already done. So I am closing this PR.

yuvipanda:
Thanks for taking a look :)

PR #1036, opened 2019-12-20
[DOC] Fix gitlab spec format

rprimet:
Copy the gitlab spec format from repoproviders.py as it seems more accurate


betatim:
Thanks for catching this!

https://mybinder.org/v2/gl/lfortran%2Fweb%2Flfortran-binder/master works (as your doc update suggests) and https://mybinder.org/v2/gl/lfortran/web/lfortran-binder/master doesn't.

PR #1037, opened 2020-01-12
Bump chartpress for helm 3 compatebility of dev releases

consideRatio:
This is important for Helm 3 compatebility, see the chartpress changelog: https://github.com/jupyterhub/chartpress/blob/master/CHANGELOG.md#060---2020-01-12


bitnik:
Thanks Erik! I think this is important to merge before we merge any other PRs. So we don't generate charts which are incompatible with helm3.

Issue #1025, opened 2019-12-04
Serve UI config from the backend


betatim:
Following up on https://github.com/jupyterhub/binderhub/pull/1017#issuecomment-561082718. In particular the comment about "reduce duplication".

A possible way to do that which I've seen in other projects is to add an endpoint to the backend from which the UI loads the config. We'd have an endpoint like `binder.example.com/_config` which returns a JSON which is then used to configure the dropdown and labels on the main page.

The endpoint at `/_config` has access to the whole config of the binder hub like for example which repo providers are configured etc.

Things to do if you'd like to work on this or even just make a small prototype to see how it would work:
* add a new endpoint at something like `/_config` that serves up a static config
* add new code  to our JS which loads the config
* transfer the labels currently configured in the JS to the repo provider classes
* modify the `/_config` endpoint to dynamically generate the JSON it serves from the repo provider classes

Addressing the problem that we have some duplication between repo2docker and BinderHub is something I'd tackle in a new PR and not as part of this one.

choldgraf:
I think it's a great idea - you could imagine a BinderHub provider that wants to have a drop-down item that is institution-specific. This could be a way that they can easily configure their hub to do so without needing to fork the BinderHub codebase

consideRatio:
Closing as resolved by #1038!

PR #1038, opened 2020-01-15
Serve UI config from the backend

GeorgianaElena:
In my quest of learning more about the binderhub repo, I decided to try out the steps described in https://github.com/jupyterhub/binderhub/issues/1025, towards serving the config from the backend :sun_with_face:

- [x]  add a new endpoint at something like /_config that serves up a static config
- [x] add new code to our JS which loads the config
- [x] transfer the labels currently configured in the JS to the repo provider classes
- [x] modify the /_config endpoint to dynamically generate the JSON it serves from the repo provider classes

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -729,7 +729,7 @@ class BinderHub(Application):
                 {'path': os.path.join(self.tornado_settings['static_path'], 'images')}),
             (r'/about', AboutHandler),
             (r'/health', HealthHandler, {'hub_url': self.hub_url_local}),
-            (r'/_config', ConfigHandler, {'repo_providers': self.repo_providers}),
+            (r'/_config', ConfigHandler),
             (r'/', MainHandler),
             (r'.*', Custom404),
         ]
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -729,7 +729,7 @@ class BinderHub(Application):
                 {'path': os.path.join(self.tornado_settings['static_path'], 'images')}),
             (r'/about', AboutHandler),
             (r'/health', HealthHandler, {'hub_url': self.hub_url_local}),
-            (r'/_config', ConfigHandler),
+            (r'/_config', ConfigHandler, {'repo_providers': self.repo_providers}),
             (r'/', MainHandler),
             (r'.*', Custom404),
         ]
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -39,6 +39,7 @@ from jupyterhub.traitlets import Callable
 from .base import AboutHandler, Custom404, VersionHandler
 from .build import Build
 from .builder import BuildHandler
+from .config import ConfigHandler
 from .health import HealthHandler
 from .launcher import Launcher
 from .log import log_request
@@ -728,6 +729,7 @@ class BinderHub(Application):
                 {'path': os.path.join(self.tornado_settings['static_path'], 'images')}),
             (r'/about', AboutHandler),
             (r'/health', HealthHandler, {'hub_url': self.hub_url_local}),
+            (r'/_config', ConfigHandler),
             (r'/', MainHandler),
             (r'.*', Custom404),
         ]
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -35,6 +35,7 @@ class MainHandler(BaseHandler):
             google_analytics_code=self.settings['google_analytics_code'],
             google_analytics_domain=self.settings['google_analytics_domain'],
             extra_footer_scripts=self.settings['extra_footer_scripts'],
+            repo_providers=SPEC_NAMES,
         )
 
 
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -35,7 +35,7 @@ class MainHandler(BaseHandler):
             google_analytics_code=self.settings['google_analytics_code'],
             google_analytics_domain=self.settings['google_analytics_domain'],
             extra_footer_scripts=self.settings['extra_footer_scripts'],
-            repo_providers=SPEC_NAMES,
+            repo_providers=self.settings['repo_providers'],
         )
 
 
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -372,6 +372,13 @@ class HydroshareProvider(RepoProvider):
     name = Unicode("Hydroshare")
     url_regex = re.compile(r".*([0-9a-f]{32}).*")
 
+    labels = {
+        "text": "Hydroshare resource id or URL",
+        "tag_text": "Git branch, tag, or commit SHA",
+        "ref_prop_disabled": True,
+        "label_prop_disabled": True,
+    }
+
     def _parse_resource_id(self, spec):
         match = self.url_regex.match(spec)
         if not match:
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -582,7 +582,6 @@ class GitLabRepoProvider(RepoProvider):
     }
 
     def __init__(self, *args, **kwargs):
-        print("GITLAB contructor")
         super().__init__(*args, **kwargs)
         self.quoted_namespace, unresolved_ref = self.spec.split('/', 1)
         self.namespace = urllib.parse.unquote(self.quoted_namespace)
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -212,6 +212,8 @@ class ZenodoProvider(RepoProvider):
     """
     name = Unicode("Zenodo")
 
+    display_name = "Zenodo DOI"
+
     labels = {
         "text": "Zenodo DOI (10.5281/zenodo.3242074)",
         "tag_text": "Git ref (branch, tag, or commit)",
@@ -256,6 +258,9 @@ class FigshareProvider(RepoProvider):
     Users must provide a spec consisting of the Figshare DOI.
     """
     name = Unicode("Figshare")
+
+    display_name = "Figshare DOI"
+
     url_regex = re.compile(r"(.*)/articles/([^/]+)/([^/]+)/(\d+)(/)?(\d+)?")
 
     labels = {
@@ -305,6 +310,8 @@ class FigshareProvider(RepoProvider):
 
 class DataverseProvider(RepoProvider):
     name = Unicode("Dataverse")
+
+    display_name = "Dataverse DOI"
 
     labels = {
         "text": "Dataverse DOI (10.7910/DVN/TJCLKP)",
@@ -370,6 +377,9 @@ class HydroshareProvider(RepoProvider):
     Users must provide a spec consisting of the Hydroshare resource id.
     """
     name = Unicode("Hydroshare")
+
+    display_name = "Hydroshare resource"
+
     url_regex = re.compile(r".*([0-9a-f]{32}).*")
 
     labels = {
@@ -442,6 +452,8 @@ class GitRepoProvider(RepoProvider):
 
     name = Unicode("Git")
 
+    display_name = "Git repository"
+
     labels = {
         "text": "Arbitrary git repository URL (http://git.example.com/repo)",
         "tag_text": "Git ref (branch, tag, or commit) SHA",
@@ -511,6 +523,8 @@ class GitLabRepoProvider(RepoProvider):
 
     name = Unicode('GitLab')
 
+    display_name = "GitLab.com"
+
     hostname = Unicode('gitlab.com', config=True,
         help="""The host of the GitLab instance
 
@@ -568,6 +582,7 @@ class GitLabRepoProvider(RepoProvider):
     }
 
     def __init__(self, *args, **kwargs):
+        print("GITLAB contructor")
         super().__init__(*args, **kwargs)
         self.quoted_namespace, unresolved_ref = self.spec.split('/', 1)
         self.namespace = urllib.parse.unquote(self.quoted_namespace)
@@ -625,6 +640,8 @@ class GitLabRepoProvider(RepoProvider):
 class GitHubRepoProvider(RepoProvider):
     """Repo provider for the GitHub service"""
     name = Unicode('GitHub')
+
+    display_name = 'GitHub'
 
     # shared cache for resolved refs
     cache = Cache(1024)
@@ -872,8 +889,11 @@ class GistRepoProvider(GitHubRepoProvider):
     If master or no ref is specified the latest revision will be used.
     """
 
-    name = Unicode('Gist')
-    hostname = Unicode('gist.github.com')
+    name = Unicode("Gist")
+
+    display_name = "Gist"
+
+    hostname = Unicode("gist.github.com")
 
     allow_secret_gist = Bool(
         default_value=False,
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -384,7 +384,7 @@ class HydroshareProvider(RepoProvider):
 
     labels = {
         "text": "Hydroshare resource id or URL",
-        "tag_text": "Git branch, tag, or commit SHA",
+        "tag_text": "Git ref (branch, tag, or commit)",
         "ref_prop_disabled": True,
         "label_prop_disabled": True,
     }
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -575,7 +575,7 @@ class GitLabRepoProvider(RepoProvider):
         return ""
 
     labels = {
-        "text": "GitLab repository name or URL",
+        "text": "GitLab.com repository or URL",
         "tag_text": "Git ref (branch, tag, or commit)",
         "ref_prop_disabled": False,
         "label_prop_disabled": False,
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -456,7 +456,7 @@ class GitRepoProvider(RepoProvider):
 
     labels = {
         "text": "Arbitrary git repository URL (http://git.example.com/repo)",
-        "tag_text": "Git ref (branch, tag, or commit) SHA",
+        "tag_text": "Git ref (branch, tag, or commit)",
         "ref_prop_disabled": False,
         "label_prop_disabled": False,
     }
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -212,6 +212,13 @@ class ZenodoProvider(RepoProvider):
     """
     name = Unicode("Zenodo")
 
+    labels = {
+        "text": "Zenodo DOI (10.5281/zenodo.3242074)",
+        "tag_text": "Git ref (branch, tag, or commit)",
+        "ref_prop_disabled": True,
+        "label_prop_disabled": True,
+    }
+
     async def get_resolved_ref(self):
         client = AsyncHTTPClient()
         req = HTTPRequest("https://doi.org/{}".format(self.spec),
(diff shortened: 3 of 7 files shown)


consideRatio:
Oh also eager to learn more about binderhub, perhaps I can activate some _retrieval practice_ by asking a question?

> Retrieval practice is a strategy in which bringing information to mind enhances and boosts learning. https://www.retrievalpractice.org/why-it-works

What config? There is configuration for a lot of different things, what does this configuration influence? What has previously provided it and how is it now provided?

Actually, it became a flurry of questions as I failed to know were to start in my confusion about __what config this is and what its function is__! Ah, this sums up my confusion better I think!

GeorgianaElena:
> What config? There is configuration for a lot of different things, what does this configuration influence? What has previously provided it and how is it now provided?

The same questions (+ dozens more "hows") I asked myself when I first read the issue. Given that this is my first interaction with this repo, I re-read the steps 20-30 times before I got an idea about what I'm supposed to be doing (and I'm still not sure I got it right).

So here my takeaways:
- The *config* in this context is every configurable used to construct the dropdown and labels on the main page.
- The *dropdown and labels on the main page* I believe are the ones related to the repo providers:
Screenshot from 2020-01-16 11-17-32
*This assumption might not be complete, but I wanted to simplify a bit the problem and receive an early feedback, otherwise I would have risked to drown into questions*
- Previously, the labels were configured through [this js function](https://github.com/jupyterhub/binderhub/blob/master/binderhub/static/js/index.js#L62).
- This PR moves the label configuration to the `_config` endpoint. Right now, the label config is hard-coded, but the end goal is to have this info posted to the `_config` endpoint by each of the providers (i.e. the backend).
- It is my understanding that another benefit of moving the configuration to the backend would ease the process of adding a new repo provider by reducing the number of parts that should be changed during this process.

Thanks for asking these questions @consideRatio! Answering them helped me put a bit of order in my head. :sun_with_face:

consideRatio:
Thanks for an excellent answer, very suitable for a person like me with little prerequisite knowledge about how these things worked!

Now i can understand the benefit of this PR to couple the frontend behavior to the repo providers more directly!

Hmmm, i write from mobile and this question pops up, that ill ask without inspecting details of current implementation more first:
Is it so, that we have fixed JS config for the webfrontend, rendered on browsers far away from the server, that in turn then asks binderhub server about how to render?

Do we perhaps want to instead want to inject that dynamically into the JS? It would add complexity of ensuring the clients doesnt have a cached JS script if we have changed it, but reduce the need for another webserver enspoint and having the browser do an additional webrequest which may be trivial though so perhaps no real benefit, hmmm. Do we need such endpoint for this reason only? Can we make the JS dynamically generates at all btw? Perhaps that is adding much work? Hmmmm.. What do you think regarding this?

Thank you @GeorgianaElena for onboarding me so well into this PR!

manics:
A potential benefit of having a separate `_config` endpoint is you can decouple the backend binderhub service from the frontend UI, or even support alternative front-ends.

GeorgianaElena:
@consideRatio, unfortunately I have very little experience with JS and frontend altogether and I really don't know how to answer. Although I could do some research and come up with an opinion, it would be based on other people's experience and not mine. So, my honest answer is *I don't know*

@manics has a really good point about the other benefits this endpoint can provide.

betatim:
I think making the config (or information needed to allow a frontend to configure itself) available via an endpoint is better than baking it into the JS. For one you'd have to understand how to bake it into the JS when it is compiled and minified. And it would not allow other frontends to configure themselves.

A downside for the user could be (to be tested) that the UI "flashes" (when it goes from defaults/no config to configured state) or has to be disabled until the config has been loaded.

Not sure there is a definitive "better way" of doing it. It is trade-offs all the way down :-/

GeorgianaElena:
Friendly ping @betatim :sun_with_face: 
Not sure how to continue with this PR  :eyes:

binderhub/app.py
```diff
@@ -646,6 +647,7 @@ def initialize(self, *args, **kwargs):
                 {'path': os.path.join(self.tornado_settings['static_path'], 'images')}),
             (r'/about', AboutHandler),
             (r'/health', HealthHandler, {'hub_url': self.hub_url}),
+            (r'/_config', ConfigHandler, {'repo_providers': self.repo_providers}),
```
bitnik: Actually, you don't need to pass repo_providers here. Because `repo_providers` config is passed to tornado app settings:

https://github.com/jupyterhub/binderhub/blob/207ae704752cde82f33fc771a3b58aa1ecc04b6d/binderhub/app.py#L590

So you can have this line as `(r'/_config', ConfigHandler),` and then in `ConfigHandler.generate_config` you can directly use `self.settings["repo_providers"]` to get value of `repo_providers` config and also you can delete `ConfigHandler.initialize` method.

bitnik:
I think one important missing thing is generating the provider select box in html. Right now config is generated with `repo_providers` config and I think this is how it should be, but html contains everything by default:

https://github.com/jupyterhub/binderhub/blob/207ae704752cde82f33fc771a3b58aa1ecc04b6d/binderhub/templates/index.html#L51-L60

binderhub/app.py
GeorgianaElena: Thanks a lot for reviewing this @bitnik :sparkles: 
I'll rebase and address your feedback asap.

binderhub/main.py
```diff
@@ -35,6 +35,7 @@ def get(self):
             google_analytics_code=self.settings['google_analytics_code'],
             google_analytics_domain=self.settings['google_analytics_domain'],
             extra_footer_scripts=self.settings['extra_footer_scripts'],
+            repo_providers=SPEC_NAMES,
```
bitnik: I think we should at least filter it by `repo_providers` config:
```python
            repo_providers={prefix: SPEC_NAMES[prefix] for prefix in SPEC_NAMES if prefix in self.settings['repo_providers']},
```
But probably better to use `self.settings['repo_providers']` itself directly:

```python
            repo_providers=self.settings['repo_providers'],
```
Together with this we can add `display_name` attribute in repo provider classes and use it in dropdown menu.

binderhub/static/js/index.js
```diff
@@ -59,50 +59,33 @@ function v2url(providerPrefix, repository, ref, path, pathType) {
...
 
+function loadConfig(callback) {
+  var req = new XMLHttpRequest();
+  req.onreadystatechange = function() {
+    if (req.readyState == 4 && req.status == 200)
+      callback(req.responseText)
+  };
+  req.open('GET', BASE_URL + "_config", true);
+  req.send(null);
+}
```
bitnik: maybe we should do this request only 1 time in `indexMain` function and pass config value into a global variable, then `updateRepoText` function can use that global variable. what do you think?

binderhub/static/js/index.js
```diff
@@ -59,50 +59,33 @@ function v2url(providerPrefix, repository, ref, path, pathType) {
...
+  loadConfig(function(res) {
+    var cfg = JSON.parse(res);
+    var provider = $("#provider_prefix").val();
+
+    var text = cfg[provider]["text"]
+    var tag_text = cfg[provider]["tag_text"]
+    var ref_prop_disabled = cfg[provider]["ref_prop_disabled"]
+    var label_prop_disabled = cfg[provider]["label_prop_disabled"]
+
+    $("#ref").prop("disabled", ref_prop_disabled);
```
bitnik:
```suggestion
    $("#ref").attr('placeholder', tag_text).prop("disabled", ref_prop_disabled);
```

binderhub/static/js/index.js
```diff
@@ -59,50 +59,33 @@ function v2url(providerPrefix, repository, ref, path, pathType) {
...
+    $("label[for=ref]").prop("disabled", label_prop_disabled);
```
bitnik:
```suggestion
    $("label[for=ref]").text(tag_text).prop("disabled", label_prop_disabled);
```

binderhub/static/js/index.js
```diff
@@ -59,50 +59,33 @@ function v2url(providerPrefix, repository, ref, path, pathType) {
...
+    $("#repository").attr('placeholder', text);
+    $("label[for=repository]").text(text);
+    $("#ref").attr('placeholder', tag_text);
```
bitnik:
```suggestion
```

binderhub/static/js/index.js
```diff
@@ -59,50 +59,33 @@ function v2url(providerPrefix, repository, ref, path, pathType) {
...
+    $("label[for=ref]").text(tag_text);
```
bitnik:
```suggestion
```

binderhub/static/js/index.js
GeorgianaElena: That's a good idea. Thanks @bitnik!

binderhub/static/js/index.js
```diff
@@ -272,6 +250,10 @@ function indexMain() {
     // setup badge dropdown and default values.
     updateUrls();
 
+    loadConfig(function(res) {
+      config_dict = JSON.parse(res);
```
bitnik: Because `loadConfig` is done asynchronously, when `updateRepoText()` is called first time during page load, `config_dict` has the empty value. I think we have to call  `updateRepoText()` here after config_dict is loaded and delete the other call in `indexMain()`.

GeorgianaElena: You're right @bitnik, thanks!
I have one question related to storing the config in a global dict though. Are there any reasons for the config to change while binderhub is running (though POST requests to the _config endpoint)? Will this be useful?

GeorgianaElena: @bitnik, I ended up checking in `updateRepoText()` if  `config_dict` is empty because I think otherwise there would still have been cases when `updateRepoText()` might have ended up with an empty config dict.
For example, if in a worst case scenario the request to `_config` would have taken a very long time (though I don't see this very likely) and during this time someone would have triggered the event associated with the provider selection (`provider_prefix_sel`), `updateRepoText()` would get in this case an empty dict. What do you think?

bitnik: > Are there any reasons for the config to change while binderhub is running (though POST requests to the _config endpoint)? Will this be useful?
I think we don't need sth like this now. The config is fetched each time the page is loaded and also I think this config will change rarely.

binderhub/static/js/index.js
```diff
@@ -59,50 +60,32 @@ function v2url(providerPrefix, repository, ref, path, pathType) {
...
-    $("#ref").prop("disabled", true);
-    $("label[for=ref]").prop("disabled", true);
-  }
-  else if (provider === "dataverse") {
-    text = "Dataverse DOI (10.7910/DVN/TJCLKP)";
-    $("#ref").prop("disabled", true);
-    $("label[for=ref]").prop("disabled", true);
+  if (Object.keys(config_dict).length === 0){
+    loadConfig(function(res) {
+      config_dict = JSON.parse(res);
```
bitnik: This is good but there is still a problem when page is loaded first time, I still see the default labels from html template:
image

This should be fixed if we update labels after this line too. We could put
```javascript
	var provider = $("#provider_prefix").val();
	var text = config_dict[provider]["text"]
	var tag_text = config_dict[provider]["tag_text"]
	var ref_prop_disabled = config_dict[provider]["ref_prop_disabled"]
	var label_prop_disabled = config_dict[provider]["label_prop_disabled"]

	$("#ref").attr('placeholder', tag_text).prop("disabled", ref_prop_disabled);
	$("label[for=ref]").text(tag_text).prop("disabled", label_prop_disabled);
	$("#repository").attr('placeholder', text);
	$("label[for=repository]").text(text);
```
into another function. And call it in the callback, after `config_dict = JSON.parse(res);` and also after if statement.

bitnik:
I have a suggestion which can be done in a separate PR if we agree on it.  

In "index.html" GitHub is the default selected repo provider and there are some default text for provider and ref:

https://github.com/jupyterhub/binderhub/blob/b6446b12b30f741d9e82b7aec1498ede4776cd79/binderhub/templates/index.html#L37-L48

But `repo_providers` is configurable, so maybe GitHub is not included or not in the first place. My suggestion is that we could accept first repo provider in `repo_providers` as default provider, pass it to template and fill provider and ref text with its attributes. This way we don't need to fetch labels from config endpoint and update text, when page is first loaded.

GeorgianaElena:
Great idea @bitnik! Will open a separate PR for it.

binderhub/static/js/index.js
GeorgianaElena: I also noticed that the url doesn't get updated when the page first loads either. Only after explicitly selecting a provider:

bitnik: When `updateRepoText` is called first time, js fails because `config_dict` is empty. My suggestion should solve this problem too.

GeorgianaElena: @bitnik, I just realized that I don't think it can be done a separate PR because a repo provider class doesn't have the needed attributes, only the name. Should I make the change here instead?

bitnik: I guess you replied to wrong place? :) I think that PR can be opened when this one is done. I am fine with adding those changes also into this PR, but it would make it harder for others to review this PR.

GeorgianaElena: Yup, wrong place, sorry. For the sake of having a continuation, I'm replying here still :)
I agree it will make it harder to review, but without those changes, this PR doesn't do what it should :(

bitnik: did try this (https://github.com/jupyterhub/binderhub/pull/1038#discussion_r386889545) out?

GeorgianaElena: Yup, I tried it and it solves the problem with the default labels from the html template, but not the issue with the url :(

The gif above is with that change.

bitnik: > And call it in the callback, after config_dict = JSON.parse(res); and also after if statement.

Sorry, actually we should call in the callback in if statement and then also in else statement.

bitnik:
I just tested this locally and for me it works very well. There is very minor change request. After that is done, I think we can merge this!

binderhub/repoproviders.py
```diff
@@ -534,7 +583,15 @@ def _default_git_credentials(self):
...
 
+    labels = {
+        "text": "GitLab repository name or URL",
+        "tag_text": "Git branch, tag, or commit",
+        "ref_prop_disabled": False,
+        "label_prop_disabled": False,
+    }
+
     def __init__(self, *args, **kwargs):
+        print("GITLAB contructor")
```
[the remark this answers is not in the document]
GeorgianaElena: :smile: sorry, I missed this print. Was curious to see when the repo provider classes get instantiated.

meeseeksmachine:

https://discourse.jupyter.org/t/custom-repoprovider-ui/7006/2

adriendelsalle:
Hi, thanks for this!

I tried to help with few things here https://github.com/GeorgianaElena/binderhub/pull/1 and would be happy to make it mergeable!

choldgraf:
@GeorgianaElena is there anything else to do on this one? I think that @adriendelsalle would appreciate getting this one merged!

consideRatio:
@bitnik does this LGTY?

choldgraf:
I assumed @bitnik was OK w/ it because of:

image

GeorgianaElena:
> @GeorgianaElena is there anything else to do on this one? I think that @adriendelsalle would appreciate getting this one merged!

@choldgraf, I don't think so. I'm not super familiar with `binderhub`, so I don't feel confident enough self-merging this. If anyone wants to give it one last quick review and merge it, I would be super grateful.

cc @betatim, who opened the initial issue

binderhub/tests/test_app.py
```diff
@@ -2,10 +2,33 @@
...
     check_output([sys.executable, '-m', 'binderhub', '--help-all'])
 
+def test_repo_providers():
+    b = BinderHub()
+
+    class Provider(RepoProvider):
+        pass
+
+    b.repo_providers = dict(gh=GitHubRepoProvider, gl=GitLabRepoProvider)
+    b.repo_providers = dict(p=Provider)
```
manics, line 26: This doesn't look right, `b.repo_providers` is set twice.

manics: And as soon as I posted that comment I realised what it's doing- it's checking the `repo_providers` property is validated. Maybe add a comment to help others?

GeorgianaElena: Thanks @manics!

choldgraf:
It sounds like comments are addressed, think we can merge this one? (@manics anything else you caught?)

manics (approved):
I think it's fine to merge.

Would be nice to have some Javascript or UI tests in place but not a blocker, maybe we should add an issue?

choldgraf:
+1 to adding an issue and getting this in. I'm not sure that I have the right information to open a useful issue, would you mind doing this @manics ? Or perhaps @adriendelsalle would be willing to do so, since this seems to be a feature he plans on using?

manics:
I've opened https://github.com/jupyterhub/binderhub/issues/1279

Since I haven't tested this could either @bitnik or @adriendelsalle please confirm this is still working as expected (it's been several months since @bitnik's approval so I'd like confirmation in case anything's changed in the mainline since), then I'll merge.

binderhub/repoproviders.py
```diff
@@ -349,8 +377,18 @@ class HydroshareProvider(RepoProvider):
...
     """
     name = Unicode("Hydroshare")
+
+    display_name = "Hydroshare resource"
+
     url_regex = re.compile(r".*([0-9a-f]{32}).*")
 
+    labels = {
+        "text": "Hydroshare resource id or URL",
+        "tag_text": "Git branch, tag, or commit SHA",
```
bitnik, line 387:
```suggestion
        "tag_text": "Git ref (branch, tag, or commit)",
```

binderhub/repoproviders.py
```diff
@@ -414,6 +452,15 @@ class GitRepoProvider(RepoProvider):
 
     name = Unicode("Git")
 
+    display_name = "Git repository"
+
+    labels = {
+        "text": "Arbitrary git repository URL (http://git.example.com/repo)",
+        "tag_text": "Git ref (branch, tag, or commit) SHA",
```
bitnik, line 459:
```suggestion
        "tag_text": "Git ref (branch, tag, or commit)",
```

binderhub/repoproviders.py
```diff
@@ -525,6 +574,13 @@ def _default_git_credentials(self):
             return r'username=binderhub\npassword={token}'.format(token=self.private_token)
         return ""
 
+    labels = {
+        "text": "GitLab repository name or URL",
```
bitnik, line 578:
```suggestion
        "text": "GitLab.com repository or URL",
```

bitnik (changes requested):
I just tested this locally and yes, it is still working as expected as described in https://github.com/jupyterhub/binderhub/issues/1025! My request is minor, it is just about updating labels.

bitnik (approved):

manics:
Thanks all 🥳

choldgraf:
wohoo!

GeorgianaElena:
yaaaay :rocket: ! :heart_eyes:

betatim:
Y'all sure this will work? If yes we will deploy this to mybinder.org :D

manics:
@bitnik has tested it so we know it works on a plain BinderHub. Since this adds a new endpoint I guess there's a small chance the mybinder redirector might not behave as expected, so best to deploy when someone's watching.

minrk:
This is being deployed now. I've tested it and it's working on staging, so all seems great. Thanks everyone!

adriendelsalle:
Thanks a lot!

PR #1044, opened 2020-01-28
Add contributor badge

GeorgianaElena:
This makes the contributing guide one-click away from possible contributors :rocket:
(P.S. The badge has already been added to TLJH, check it out: [Contribute](https://the-littlest-jupyterhub.readthedocs.io/en/latest/contributing/index.html))

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,6 +25,10 @@
    :target: https://gitter.im/jupyterhub/binder
    :alt: Gitter
 
+.. image:: https://img.shields.io/badge/I_want_to_contribute!-grey?logo=jupyter
+  :target: https://github.com/jupyterhub/binderhub/blob/master/CONTRIBUTING.md
+  :alt: Contribute
+
 What is BinderHub?
 ------------------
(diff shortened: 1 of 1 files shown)


betatim:
I like it, both how the badge looks and that it makes it super obvious where to click. It is another way to direct people towards our docs for contributing. I don't think we can have too many ways of making it easy to discover them :)

Let's see if there is anyone who dislikes this, otherwise I will merge it in a bit.

GeorgianaElena:
@consideRatio, what do you think about adding this badge to z2jh also (and other repos)?

consideRatio:
@GeorgianaElena I like it! In my mind this badge will increase the need for even better contributing documentation as it increases the amount of people arriving to the documentation and probably also includes more types of people. This in turn, makes me want to avoid disappointing anyone clicking such link with either a too complicated text or too vague text.

I think it makes great sense to add this to all our repositories, assuming we also have decent contributing documentation.

PR #1047, opened 2020-02-04
bump versions in chart images

minrk:
- bump binderhub dist to current stable (buster), Python to 3.7
- make $DIST a variable, for easier future updates
- bump node to current LTS, 12.x
- bump image-cleaner requirements and Python version to current stable

### DIFF
--- a/helm-chart/images/image-cleaner/requirements.txt
+++ b/helm-chart/images/image-cleaner/requirements.txt
@@ -1,2 +1,2 @@
 docker==4.1.0
-kubernetes==10.0.0
+kubernetes==10.0.1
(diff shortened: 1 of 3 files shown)


helm-chart/images/image-cleaner/requirements.txt
```diff
@@ -1,2 +1,2 @@
-docker==3.2.1
-kubernetes==3.0.0
+docker==4.1.0
+kubernetes==10.0.0
```
consideRatio: 10.0.1 contains a security fix

```suggestion
kubernetes==10.0.1
```

PR #1048, opened 2020-02-04
[WIP] sketching API v2

### LINKED ISSUE #13: API Proposal
This needs a versioned, simple RESTful API that allows for a wide variety of use cases. 

Some of the use cases it should cater for:

1. Working as an on-demand backend for ipywidgets and related toolkits that want to talk to a kernel
2. mybinder.org like use-cases with various levels of authentication

We should collect other use cases too before finalizing the API design.

Some API design guidelines I like (and should re-read before doing the design):

1. Google's API Design guidelines https://cloud.google.com/apis/design/
2. Heroku's API Design guidelines https://geemus.gitbooks.io/http-api-design/content/en/

### LINKED ISSUE #15: Stable URL scheme for badges + links to launch binder
This is different from #13 since it focuses solely on what links people will use when clicking 'launch binder' links, rather than a HTTP API that code can also call. Requires different design and guarantees.

We also need to have a compatibility layer so current launch binder links continue to work.

### Current URL structure ###

Only supported URL structure now is:

```
/repo/<username>/<repo>/<path-to-start>
```

While a great start, it has some extensibility problems:

1. Assumes it is using GitHub, and doesn't provide an easy way to use other providers.
2. Unclear how it can handle arbitrary git URLs (that aren't hosted in a 'provider' as such)
3. Not specified how you can provide other arbitrary run-time parameters in jupyterhubs that support it (such as memory limits, extra data to be provisioned, etc)

Whatever we do, we'll make sure these links continue to work for the foreseeable future. We should find and send PRs to people tho to change them.

### Proposed new URL structure ###

```
/v2/repo/<git-clone-url>/ref/<ref>/?<runtime-params>
```

where:

1. `git-clone-url` is URL encoded URL pointing to a git repository (such as a github URL). We will interpret this pretty generously.
2. `ref` points to a commit hash, branch or tag. If it's a branch or tag it'll be redirected to a commit hash - permanently for a tag and temporarily for a branch.
3. `runtime-params` is query parameters used for *all* runtime parameters - including the path to launch, and in the future other parameters too (extra data to load, RAM requirements, etc). These will be formally defined too.

This separates the build parameters (git url + ref) from the runtime parameters. Also establishes that the canonical URL is the one with the commit hash, rather than branch or tag info.

Possible modifications:

1. <s>Version this API too, with a `/v2/` prefix</s> Decided to actually do this.
2. Special case 'path' runtime parameter, include that in the path. Not sure how necessary that is.

### LINKED ISSUE #304: Version our API too!
Right now, our API urls are unversioned. We should version them too so we can change them without breaking too many people.

### LINKED ISSUE #358: API refactor process
Our current API has a bunch of problems:

1. Not versioned (see #304)
2. EventSource is maybe not the best fit (see #317)
3. Hard to extend for finding logs after building (see #75 and friends)

This issue would be to gather constraints and feature requests various audiences want so we can come up with an evolved new design for our API that enables the various use cases we want to support.

Some use cases off the top of my head:

1. The mybinder.org home page as it is now
2. Possible different loading page for when users just click a link from outside to a binder
3. External web applications using our API like nteract/play or thebelab
4. External non-web applications (like the iOS app, possible commandline applications) that use our API

/me puts operations hat on

From an operational perspective, I think the only constraint is:

1. Do not introduce state that needs to be shared across multiple running instances of binderhub. BinderHub should continue to be as stateless as possible

Others, feel free to chime in with workflows you would like to enable, current problems with the API and constraints you want to put on the API.

/cc @rgbkrk @minrk @betatim  since I've heard them talk about this before. Feel free to tag others too.

Note that we should continue to support the current API while building the new one, and make the new one the 'v2' API :)

minrk:
The event stream errors return, which prompt me to want to take a stab at the proposal in #358, which should significantly reduce the consequences of event stream issues (in theory, to zero).

The gist of the design from #358, by @yuvipanda:

- POST to start build, replies with URL to watch events
- event stream requests only get logs, never result in transactions
- POST to launch
- opaque JWT tokens used to pass stateful args between stages (build id, hub username, token, etc.)

Python code design choices so far:

- create v1, v2 subfolders for API handlers
- split `builder.py` into a Builder class which implements building and a v1 BuildHandler. The Builder will be used in both v1 and v2
- create base EventStreamHandler class, used in both v1 and v2

My plan is to refactor things based on v2 plans, while keeping v1 fully working

### implementation log

This is a very early WIP. I'll update with progress notes.

implementation challenges: lots of coupling between different layers that make separating these actions tricky, plus preserving the old handlers without tons of duplicate code. Separating them will be helpful in the long run, but will take some doing.

Couplings found so far:
- build/watch/launch logic is all in BuildHandler, much of which will need to be re-used in both v1 and v2 handlers (Builder object should help with this)
- Start-build and watch-build are tied together in the Build object, when ~only the build-id and namespace are required for watching builds.

closes #358
closes #304
closes #13 
closes #15

### DIFF
--- a/binderhub/base.py
+++ b/binderhub/base.py
@@ -1,9 +1,12 @@
 """Base classes for request handlers"""
 
+import asyncio
 import json
 
 from http.client import responses
 from tornado import web
+from tornado.log import app_log
+from tornado.iostream import StreamClosedError
 from jupyterhub.services.auth import HubOAuthenticated, HubOAuth
 
 from . import __version__ as binder_version
@@ -103,6 +106,7 @@ class Custom404(BaseHandler):
 
 class AboutHandler(BaseHandler):
     """Serve the about page"""
+
     async def get(self):
         self.render_template(
             "about.html",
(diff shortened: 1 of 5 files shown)


betatim:
Maybe off topic but seems worth pondering while working on this: #1038 to reduce the amount of duplicate information in our code base (each builder has to be defined in Python code and again in JS).

Another new endpoint we have discussed is one that would implement the logic needed for an "awesome bar". In my mind that is an endpoint to which you POST "what ever a user typed into the awesome bar" and get back structured information which can then be used to trigger a build.

I don't think these two points are directly relevant for this but might be worth keeping at the back of our minds so we can make it easier (in the future) to build them (vs harder because we made choices here without considering them).

minrk:
Definitely worth keeping in mind! I think both of those issues are a bit orthogonal to this change, though. Awesome bar, for example, would be a step 0 request, redirecting to the appropriate, stricter, versioned build API URL. And the server-side provider info would also happen at the same level as in #1038, I don't expect this change should affect that effort. Good to have thought that through, though.

PR #1049, opened 2020-02-04
Catch tornado timeout error when pulling image manifest

sgibson91:
This PR updates `binderhub/builder.py` to attempt to pull the image manifest from the registry 3 times and catch and log the tornado timeout HTTP 599 error we've been seeing on mybinder.org.

Related issue: https://github.com/jupyterhub/mybinder.org-deploy/issues/1323

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -290,8 +290,6 @@ class BuildHandler(BaseHandler):
         ).replace('_', '-').lower()
 
         if self.settings['use_registry']:
-            image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
-            image_found = bool(image_manifest)
             for _ in range(3):
                 try:
                     image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -296,7 +296,7 @@ class BuildHandler(BaseHandler):
                     image_found = bool(image_manifest)
                     break
                 except tornado.curl_httpclient.CurlError:
-                    app_log.warning("Tornado HTTP Timeout error: Failed to get image manifest for %s", image_name)
+                    app_log.exception("Tornado HTTP Timeout error: Failed to get image manifest for %s", image_name)
                     image_found = False
         else:
             # Check if the image exists locally!
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -296,6 +296,7 @@ class BuildHandler(BaseHandler):
                     image_found = bool(image_manifest)
                     break
                 except tornado.curl_httpclient.CurlError:
+                    app_log.warning("Tornado HTTP Timeout error: Failed to get image manifest for %s", image_name)
                     image_found = False
         else:
             # Check if the image exists locally!
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -297,7 +297,6 @@ class BuildHandler(BaseHandler):
                     break
                 except tornado.curl_httpclient.CurlError:
                     image_found = False
-                    pass
         else:
             # Check if the image exists locally!
             # Assume we're running in single-node mode or all binder pods are assigned to the same node!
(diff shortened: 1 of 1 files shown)


binderhub/builder.py
```diff
@@ -292,6 +292,14 @@ def escape(s):
         if self.settings['use_registry']:
             image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
             image_found = bool(image_manifest)
+            for _ in range(3):
+                try:
+                    image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
+                    image_found = bool(image_manifest)
+                    break
+                except tornado.curl_httpclient.CurlError:
```
manics: Have you thought about logging the error? It might be useful to see how often it happens, or to deal with other problems that get caught by this exception.

binderhub/builder.py
```diff
@@ -292,6 +292,14 @@ def escape(s):
...
+                    image_found = False
+                    pass
```
manics:
```suggestion
```
`pass` is only needed when there's no other statement

binderhub/builder.py
```diff
@@ -292,6 +292,14 @@ def escape(s):
         if self.settings['use_registry']:
             image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
             image_found = bool(image_manifest)
```
manics: You can delete the two above lines above since they're now in the try-loop

binderhub/builder.py
sgibson91: Can you explain a bit more? We can already see what the error in in the binder pod logs and I believe Tim's bot is giving us an estimate of how often it's happening

manics: Maybe I'm misunderstanding this, but isn't it logged because the exception isn't currently caught?

sgibson91: There seems to be a lot of logging stuff in [`app.py`](https://github.com/jupyterhub/binderhub/blob/master/binderhub/app.py) - I don't really know how that code interacts with this, if at all. Is it just a case of adding
`logging.warn(tornado.curl_httpclient.CurlError)` into the except block?

manics: I've been silent because I was hoping @betatim or someone else might answer 😀

- https://github.com/jupyterhub/binderhub/blob/531aad202c8a97db643da7c3034ba2c934512cc7/binderhub/app.py#L513-L515
- https://github.com/jupyterhub/binderhub/blob/531aad202c8a97db643da7c3034ba2c934512cc7/binderhub/app.py#L526

After looking through the code I can't tell from reading whether explicit logging is needed or not.

betatim: Sorry for being slow. We can estimate how often this happens for the current deployments based on how often the isthehubup bot posts to gitter. (It seems most of the errors it encounters are because of this problem).

Right now we also see this in the logs of the binderhub pod, because the exception is unhandled.

Once we catch the exception there will be no more automatic logging that it happened, because we caught the exception (from tornado's point of view nothing went wrong).

I'd take inspiration from https://github.com/jupyterhub/binderhub/blob/531aad202c8a97db643da7c3034ba2c934512cc7/binderhub/builder.py#L230 for how to add some explicit logging in the `except` clause.

sgibson91: Thanks @betatim! Logging added in https://github.com/jupyterhub/binderhub/pull/1049/commits/64d34d07ba843b4890aae8c86bf39700b50ec1f3

binderhub/builder.py
```diff
@@ -290,8 +290,14 @@ def escape(s):
...
         if self.settings['use_registry']:
-            image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
-            image_found = bool(image_manifest)
+            for _ in range(3):
+                try:
+                    image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
+                    image_found = bool(image_manifest)
+                    break
+                except tornado.curl_httpclient.CurlError:
+                    app_log.warning("Tornado HTTP Timeout error: Failed to get image manifest for %s", image_name)
```
sgibson91:
```suggestion
                    app_log.exception("Tornado HTTP Timeout error: Failed to get image manifest for %s", image_name)
```

betatim: The self-suggest :D

sgibson91: Sometimes it's just easier than pulling my changes and faffing with vscode :joy:

PR #1055, opened 2020-02-09
Add AWS ECR support #2

ivan-gomes:
Addresses https://github.com/jupyterhub/binderhub/issues/705 and follow-on/alternative to https://github.com/jupyterhub/binderhub/pull/920.

Includes updated user documentation on how to configure AWS IAM and BinderHub to use AWS ECR as the Docker registry. Modifications are additive, in that existing configurations should not be affected, and have been tested with ECR.


jhamman:
Cc @scotthq

betatim:
Thanks for this PR.

Has someone been running this code on AWS?

Having different registry classes is a good idea. Could also be used for GitLab like  registries (xref #986)

ivan-gomes:
Yes we are running this code on AWS with ECR and no further modifications.

binderhub/app.py
```diff
@@ -207,6 +207,14 @@ def _valid_badge_base_url(self, proposal):
         config=True,
     )
 
+    registry_class = Type(
+        DockerRegistry,
+        help="""
+        Registry class implementation, change to define your own
```
betatim:
```suggestion
        Change this to use different docker container registries.
        The default works with GCR, ACR and DockerHub. Use `AWSElasticContainerRegistry` on AWS.
```

How about something like this?

binderhub/app.py
```diff
@@ -207,6 +207,14 @@ def _valid_badge_base_url(self, proposal):
         config=True,
     )
 
+    registry_class = Type(
```
betatim: What do you think of renaming it to `docker_registry_class`? Just to be a bit more explicit.

binderhub/registry.py
```diff
@@ -224,3 +227,61 @@ def get_image_manifest(self, image, tag):
...
+        help="""
+        AWS region for ECR service
+        """,
+    )
+
+    ecr_client = Any()
+
+    @default("ecr_client")
+    def _get_ecr_client(self):
+        return boto3.client("ecr", region_name=self.aws_region)
```
betatim: Is boto3 async?

binderhub/app.py
ivan-gomes: Updated it in that direction.

binderhub/registry.py
```diff
@@ -224,3 +227,61 @@ def get_image_manifest(self, image, tag):
...
+
+    # TODO: cache auth if not expired - authorizationData[i]["expiresAt"]
+    def _get_ecr_auth(self):
+        auths = self.ecr_client.get_authorization_token()["authorizationData"]
+        auth = next(x for x in auths if x["proxyEndpoint"] == self.url)
+        self._patch_docker_config_secret(auth)
```
betatim: Could you explain a bit why we need to modify the push secret? All other registries work with a static push secret.

What happens when the build container is trying to push an image and the push secret is wrong/changed?

binderhub/registry.py
ivan-gomes: Not officially. There are third-party wrappers like aioboto3 - https://github.com/terrycain/aioboto3 - but we may not want to go there.

binderhub/app.py
ivan-gomes: Sure thing. I will refactor and re-test as soon as possible.

clovischedid:
Hi guys,

Any news about this PR? This is the missing point to me to adopt binderhub as my private notebooks provisioning pipeline.

Can I help in anything?

Best regards,
Clovis Chedid

binderhub/app.py
ivan-gomes: Done!

binderhub/registry.py
ivan-gomes: With AWS ECR, an IAM principal is used to generate an auth token that is valid for 12 hours - source: https://docs.aws.amazon.com/AmazonECR/latest/userguide/Registries.html

So what we're doing here is using the AWS boto3 client to get/generate an auth token and update the push secret with it, so the builder picks up a fresh, valid token in the case where it doesn't exist already. There is admittedly an edge case here if the builder takes longer than 12 hours to build, but I believe the likelihood in practice is acceptably low.

ivan-gomes:
Hi @clovischedid! Thanks for your interest and offer. I went ahead and responded to all the conversations and applied the naming change requested. I have also re-requested a review from @betatim. Hopefully it's up to snuff now. Otherwise I will try to be quicker on the iteration this time around. :)

binderhub/registry.py
betatim: BinderHub is written using tornado so we can't make calls over the network with a library which blocks. If we do then all of the BinderHub process will block while that network request is happening. We either need to use a threadpool to execute the boto3 calls or use a async library that you can `await`.

Depending on the complexity of the requests you are making another option would be to implement the HTTP call yourself using the `AsyncHTTPClient` class.

ivan-gomes: Ah I see. There is really only one request that needs to be made currently and it's relatively simple to do ourselves. Unless you see a value in integrating something like `aioboto3` we can implement the request ourselves using `AsyncHTTPClient`. Drawing inspiration from elsewhere, it seems this is the route that FargateSpawner for JupyterHub went as well when making AWS calls - ref: https://github.com/uktrade/fargatespawner/blob/c614a54ffd80d0fb8886d1ef9e8de2c938de7759/fargatespawner/fargatespawner.py#L322-L346.

If you approve of this path, I will start work on implementing and testing it.

betatim: Sounds good to me.

ivan-gomes: After biting into AWS's request signing process - ref: https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html - I realized heading down the request reimplementation may not be the best approach. Additionally, the kubernetes call is also not async, so I switched to using a threadpool as you suggested. It has now been implemented and tested.

ivan-gomes:
Status update: addressed all unresolved review comments. Awaiting re-review. :)

betatim:
Nice work!

@clovischedid do you have a moment to give this a spin on your deployment? Our tests in this area aren't mega so the knowledge of "someone actually used this feature and it worked" would be great to have.

clovischedid:
Hello guys, 

I'm sorry for my delay.

@betatim I will do my best, but currently, I do not have a production ready deployment. At this time, I'm focusing myself on make binderhub running with AWS managed Kubernets service EKS. I'm facing some network problems as newer version o EKS do not make bridge network as default.
By the way, I will try to make AWS ECR as my container registry, but, could you help me? I can't realize how to make helm deploy binderhub from a github branch instead of published version on helm repo.

Regards,
Clovis Chedid
> Nice work!
> 
> @clovischedid do you have a moment to give this a spin on your deployment? Our tests in this area aren't mega so the knowledge of "someone actually used this feature and it worked" would be great to have.

chicocvenancio:
> Hello guys,
> 
> I'm sorry for my delay.
> 
> @betatim I will do my best, but currently, I do not have a production ready deployment. At this time, I'm focusing myself on make binderhub running with AWS managed Kubernets service EKS. I'm facing some network problems as newer version o EKS do not make bridge network as default.
> By the way, I will try to make AWS ECR as my container registry, but, could you help me? I can't realize how to make helm deploy binderhub from a github branch instead of published version on helm repo.
> 
> Regards,
> Clovis Chedid
> 

Rebuild the BinderHub image from source and set the image to it with helm. 

I might be able to test this PR in a couple of weeks.

Thanks @ivan-gomes for the work!

ivan-gomes:
@clovischedid @chicocvenancio just checking in if you had a chance to test yet, no pressure.

ivan-gomes:
@betatim do you think it would be a good idea to merge so it would be easier for those interested like @clovischedid to try it out without having to build from source? For what it's worth, I have it deployed as documented with ECR and this addition doesn't affect default behavior. Hoping to avoid merge conflicts with master again.

zeyaddeeb:
I'd be happy to test and report back as well!

betatim:
It would be great to have someone test drive these instructions. Though given how long they've already been "in flight" it might be easier to merge them and then update them later as we discover things. With it being merged we might get more eyeballs on it.

ivan-gomes:
@betatim I totally agree, especially considering we just resolved another round of merge conflicts in documentation.

TomasBeuzen:
I'll be testing this out over the next few days - will report back

ivan-gomes:
@TomasBeuzen you can submit a PR to https://github.com/ivan-gomes/binderhub/tree/aws-ecr-support and I can merge it in. If it's minor and easier for you, you can also describe it here and I would be happy to incorporate the feedback into this PR.

ivan-gomes:
@TomasBeuzen any update for us? Thanks.

TomasBeuzen:
@ivan-gomes - so sorry I've been super busy preparing for the new school year starting (in Canada). This is on my list but probably not for another 2-3 weeks.

ivan-gomes:
Totally understand, best of luck!

ivan-gomes:
> Though given how long they've already been "in flight" it might be easier to merge them and then update them later as we discover things. With it being merged we might get more eyeballs on it.

@betatim I think we should go for it.

TomasBeuzen:
Apologies for delays on this, I got this up and running. I just [made a PR](https://github.com/ivan-gomes/binderhub/pull/2) for a few potentially helpful clarifications to the docs but otherwise thanks for all your work on this!

ivan-gomes:
That's great news @TomasBeuzen! I merged in your helpful changes. Thanks!

ivan-gomes:
@betatim just wanted to put this back on your radar. Hope you don't mind. :)

TomasBeuzen:
Pinging this as it would be nice to have it merged 🙂  Happy to help in any way I can!

btjones-me:
Hi folks, would love to see this functionality if it's still in the roadmap!

ivan-gomes:
I would be happy to resolve conflicts again once this PR has been tentatively / conditionally approved. :)

manics:
I think one issue to consider is how to test this. In the short term we can probably ask a few AWS users to test it for real and sign off on it. In the long term how do we ensure we don't inadvertently break it? It looks like the AWS registry works in a completely different way to a standard docker registry, so it's not going to be regularly tested. Do you have any ideas?

btjones-me:
Happy to test and feed back @manics

teticio (commented):
I tried this out on a minikube on EC2 and found that the `_patch_docker_secret` function failed. Firstly, there was no file at `/var/run/secrets/kubernetes.io/serviceaccount/namespace` and, secondly, the name of the secret was hard-coded to the wrong value (in my case it was `binder-build-docker-config` and, third, the call to `self.kube_client.patch_namespaced_secret` gave an error after fixing hese first two points. I managed to get the following code to work:

`import subprocess`

...

```
    def _patch_docker_config_secret(self, auth):
        """Patch binder-push-secret"""
        secret_data = {"auths": {self.url: {"auth": auth["authorizationToken"]}}}
        secret_data = base64.b64encode(json.dumps(secret_data).encode("utf8")).decode(
            "utf8"
        )
        subprocess.check_call(['kubectl',
             '--namespace', self.parent.build_namespace,
             'patch', 'secret', self.parent.push_secret,
             '-p', json.dumps({"data": {"config.json": secret_data}})
        ])
```

By the way, I found it very hard to get a test environment working that used a Docker registry. It seems to me that the way to do it would be to build a Docker image for the binder pod using the branched code, but the Dockerfiles used to build the release images don't appear to be available anywhere.

One more thing: using the `binderhub_config.py` file from the Helm repo with my `secret.yaml` and `config.yaml` files, I had to set the config differently to the instructions. I am not sure if this is a quirk of the test setup. Here is my `config.yaml` file prior to running an `envsubst` on it:

```
config:
  BinderHub:
    use_registry: true
    docker_registry_class: binderhub.registry.AWSElasticContainerRegistry
    image_prefix: $AWS_ACCOUNT.dkr.ecr.$AWS_REGION.amazonaws.com/binderhub-
    hub_url: $JUPYTERHUB_URL
  AWSElasticContainerRegistry:
    aws_region: $AWS_REGION
    url: https://$AWS_ACCOUNT.dkr.ecr.$AWS_REGION.amazonaws.com

```

manics:
A few thoughts after quickly going through this and related issues:
- The Registry class has been made configurable since this PR was written so this part will require some minor changes
- It looks like `boto3` is only used in the registry class. Although it's usually fine to install I've found it can occasionally cause conflicts with other dependencies. Not an issue at the moment, but might be a problem if we have to add other optional libraries in future. I'm therefore inclined to put the import inside the class instead.
- I'm uneasy about increasing the RBAC privileges for all deployments, maybe make it conditional? Could you also document the `_patch_docker_config_secret` method to include an explanation of what it does and why it's necessary? Or if you can think of a nicer way of doing this even better!
- Is there any way of getting some mock tests in, e.g. with http://docs.getmoto.org/en/latest/docs/services/ecr.html (it may not be possible). Otherwise there's a risk we'll accidentally break it, and you won't notice until someone deploys it in production.

thomas-bc:
This PR sparked a lot of interest in the past, and it would be really neat to allow Binder to run with AWS ECR.
I updated the PR by rebasing and making the changes that @manics and others requested. Those include:

- moving the import of boto3 within the registry class; 
- adding some explanation on the patching of secrets; 
- making the RBAC increase conditional (with associated docs in setting up Binder); 
- referencing the secret name instead of hardcoding it.

As for the mock-testing, I fully agree that this would certainly be nice to have. However the Moto library has not implemented all the calls that would be necessary to test here. I looked for other solutions but could not come up with much. 

Re-iterating here, but my organization has been using this fork for a while and it has been working perfectly. Maybe @teticio or @TomasBeuzen can confirm. I will be sure to test again with this updated version and report back.

Would love to hear some feedback!

manics:
Coming back to this after a while to take a fresh look, and after seeing https://github.com/jupyterhub/binderhub/issues/1506#issuecomment-1220103944, I was wondering if we could either move the ECR token request to the Builder and pass in the credentials as an environment variable of volume instead of overwriting the Kubernetes secret?

It's a bigger change, but I've just done some work to refactor the build classes https://github.com/jupyterhub/binderhub/pull/1518 and I think further changes to KubernetesBuildExecutor to support custom registries or passing tokens is reasonable, especially if the outcome is cleaner code. A few ideas:
- Split the `submit()` method to make it easier to override parts of the method, or add a hook that can be called https://github.com/jupyterhub/binderhub/blob/c4af713bb20ed90606e9795834d160bd8a332578/binderhub/build.py#L348-L379 and fetch the ECR token, then either:
  a. Convert the ECR token into login credentials, pass them in as an environment variable to the build pod instead of passing a secret. The build pod would need to accept a script/hook to run `docker login` or equivalent. We already use `GIT_CREDENTIAL_ENV` for private git credentials
    https://github.com/jupyterhub/binderhub/blob/c4af713bb20ed90606e9795834d160bd8a332578/binderhub/build.py#L78-L87
    so registry login credentials sounds reasonable?
  b. Use an service account (IRSA for AWS) as suggested in https://github.com/jupyterhub/binderhub/issues/1506 and handle the entire ECR login process in the build pod. This has the advantage of being cleaner, but means the repo2docker build pod needs to includes the AWS dependencies.

@yuvipanda what do you think? I'm leaning towards option (a), and modifying repo2docker to optionally login to the registry before pushing, but I'm very keen to hear other ideas.

manics:
This is the sort of repo2docker change I had in mind _if_ we go down the route of passing the credentials into the build pod:
https://github.com/jupyterhub/repo2docker/compare/main...manics:repo2docker:registry_login_kwargs?expand=1

omri-shilton:
any updates on this PR? we would really love to have that support!

manics:
@omri-shilton I'm working on an alternative approach:
https://github.com/jupyterhub/binderhub/issues/1623

PR #1062, opened 2020-02-11
use basic auth header to send client_id, client_secret

minrk:
sending client_id/secret in url params is [deprecated](https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/)

I initially thought it didn't work at all and we had to switch to access tokens, but I had messed something up in my testing.

We are already doing the right thing with access token in the Authorization header

### DIFF
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -648,21 +648,6 @@ class GitHubRepoProvider(RepoProvider):
     def _access_token_default(self):
         return os.getenv('GITHUB_ACCESS_TOKEN', '')
 
-    auth = Dict(
-        help="""Auth parameters for the GitHub API access
-
-        Populated from client_id, client_secret.
-    """
-    )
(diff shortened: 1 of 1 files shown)


binderhub/repoproviders.py
```diff
@@ -692,9 +677,12 @@ def get_repo_url(self):
...
     def github_api_request(self, api_url, etag=None):
         client = AsyncHTTPClient()
-        if self.auth:
-            # Add auth params. After logging!
-            api_url = url_concat(api_url, self.auth)
+
+        request_kwargs = {}
+        if self.client_id and self.client_secret:
+            request_kwargs.update(
+                dict(auth_username=self.client_id, auth_password=self.client_secret)
```
betatim: These keyword arguments are understood by tornado's HTTP client and cause it to do basic auth. leaving it here for the future because I had to double check :)

PR #1064, opened 2020-02-14
add more blocks into templates

bitnik:
The purpose of this PR is to make template customisation much easier. It also makes some minor changes.
 
- adds more blocks into templates
- re-organizes meta tags for social media preview. also fixes it in about.html
- removes styles.css from loading.html, it is already included in page.html

### DIFF
--- a/binderhub/templates/loading.html
+++ b/binderhub/templates/loading.html
@@ -1,9 +1,9 @@
 {% extends "page.html" %}
 
 {% block meta %}
-{{ super() }}
 <meta id="base-url" data-url="{{base_url}}">
 <meta id="badge-base-url" data-url="{{badge_base_url}}">
+{{ super() }}
 {% endblock meta %}
 
 {% block meta_social %}
--- a/binderhub/templates/loading.html
+++ b/binderhub/templates/loading.html
@@ -1,8 +1,11 @@
 {% extends "page.html" %}
 
-{% block head %}
+{% block meta %}
+{{ super() }}
 <meta id="base-url" data-url="{{base_url}}">
 <meta id="badge-base-url" data-url="{{badge_base_url}}">
+{% endblock meta %}
+
 <!-- Other OG information in page.html -->
 <meta property="og:title" content="{{social_desc}}">
 <meta property="og:image" content="https://mybinder.org/static/images/logo_square.png">
@@ -12,12 +15,11 @@
 <meta property="og:image:alt" content="The Binder Project Logo" />
 <meta name="twitter:card" content="summary" />
 
+{% block head %}
 {{ super() }}
 <script src="{{static_url("dist/bundle.js")}}"></script>
 <link href="{{static_url("dist/styles.css")}}" rel="stylesheet">
 <link href="{{static_url("loading.css")}}" rel="stylesheet">
-
-
 {% endblock head %}
 
 {% block main %}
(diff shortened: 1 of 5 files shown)


bitnik:
Hi, could one of you have a look at this? This contains very small changes in templates, but this changes let us simplify our custom templates which in the end makes binderhub version upgrades easier for us.

binderhub/templates/page.html
```diff
@@ -2,19 +2,41 @@
...
 <head>
   <title>{% block title %}Binder{% endblock %}</title>
-  {% block head %}
+
+  {% block meta %}
+  {% endblock meta %}
+
+  {# Social media previews #}
+  {% block meta_social %}
+  <meta property="og:title" content="The Binder Project">
```
minrk: Do we want this on every page, or just the loading page like before?

binderhub/templates/loading.html
```diff
@@ -1,23 +1,25 @@
...
 
-{% block head %}
+{% block meta %}
+{{ super() }}
 <meta id="base-url" data-url="{{base_url}}">
 <meta id="badge-base-url" data-url="{{badge_base_url}}">
-<!-- Other OG information in page.html -->
+{% endblock meta %}
+
+{% block meta_social %}
```
minrk: should meta_social be inside meta?

binderhub/templates/page.html
```diff
@@ -2,19 +2,41 @@
...
+  <meta property="og:image:height" content="700">
+  <meta property="og:image:alt" content="The Binder Project Logo" />
+  <meta name="twitter:card" content="summary_large_image">
+  {% endblock meta_social %}
+
+  {% block favicon %}
   <link id="favicon" rel="shortcut icon" type="image/png" href="{{static_url("favicon.ico")}}" />
-  <link href="{{static_url("dist/styles.css")}}" rel="stylesheet"></link>
+  {% endblock %}
+  {% block head %}
```
minrk: meta inside head?

binderhub/templates/page.html
bitnik: Before we had this also on index, loading and about pages too. On about page before it was partial though. Now with this PR, about page has all social tags, this is the difference. Or should we completely remove it from about page?

binderhub/templates/page.html
bitnik: this makes sense. but on the other hand nested blocks might be difficult when people want to customise. maybe better I should rename "head" block to "extra_head"?

bitnik: I changed my mind and made them nested in page.html.

binderhub/templates/page.html
choldgraf: I think we should double check that it degrades gracefully, but as a general rule I am +1 on it

binderhub/templates/loading.html
bitnik: if we put meta_social into meta here in loading.html, then we would have content of meta_social block 2 times in loading page; content of meta_social block in page.html (because of `{{ super() }}`) and also content of meta_social block in this template. actually that's why I didn't want to make blocks nested in the beginning, it is confusing.

choldgraf: ah that's a good point - some of the meta needs to be on all pages, while other meta is page-specific.

I wonder if we could use "if/then" statements to control behavior?

bitnik:
I want to rework on this issue and implement in smaller PRs.

PR #1066, opened 2020-02-20
fix catching of error checking registry

minrk:
`tornado` wasn't imported, causing a NameError here.

While here, switch to higher-level HTTPClientError (of which CurlError is a subclass)
so we catch more similar errors.

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -12,6 +12,7 @@ import escapism
 import docker
 from tornado.concurrent import chain_future, Future
 from tornado import gen
+from tornado.httpclient import HTTPClientError
 from tornado.web import Finish, authenticated
 from tornado.queues import Queue
 from tornado.iostream import StreamClosedError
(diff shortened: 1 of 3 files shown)


betatim:
Not sure I understand the error either. There doesn't seem to be anything relevant in https://docs.pytest.org/en/latest/changelog.html either which was my first idea.

minrk:
HTTPClientError was renamed in 5.1, but we have been running tests with 5.0. I've bumped the minimum version to 5.1, but I can also use the deprecated alias and stick with old tornado if we want.

betatim:
Version bump sounds good to me. We should try if we can go to 6.x at some point. BinderHub being shipped in a container should make us relatively free to upgrade (only constraint comes from people who heavily modify binderhub itself I'd say).

minrk:
> We should try if we can go to 6.x at some point

I was going to do that here, but figured it should be its own PR. There shouldn't be anything we need to update, I would guess.

PR #1070, opened 2020-02-26
[MRG] Pin the version of the sphinx theme used

### LINKED ISSUE #1069: Unwanted webpack popup on docs page
When I open the latest binderhub documentation (https://binderhub.readthedocs.io/en/latest/), first there is an alert saying "hi from webpack" and also style looks different:

image

Stable documentation (https://binderhub.readthedocs.io/en/stable/) has no problem.

betatim:
Closes #1069 

This should trigger a rebuild of our docs and pins us to a specific revision of the theme. We should move the pin forwards every once in a while to stay current. With a pinned revision we can explicitly check that things still work when we do update.

### DIFF
--- a/doc/doc-requirements.txt
+++ b/doc/doc-requirements.txt
@@ -5,7 +5,7 @@ sphinx-copybutton
 traitlets
 pandas
 ruamel.yaml
-git+https://github.com/pandas-dev/pandas-sphinx-theme.git@master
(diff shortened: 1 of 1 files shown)


betatim:
Preview: https://750-89419368-gh.circle-artifacts.com/0/html/index.html

choldgraf:
I was thinking the same thing, we should really start releasing versions on the pydata theme

PR #1076, opened 2020-03-23
Mounting binderhub_config.py instead of letting it be part of the image

iwilltry42:
As per this comment: https://github.com/jupyterhub/binderhub/blob/master/helm-chart/images/binderhub/Dockerfile#L42-L43

Now requires the user to mount the `binderhub_config.py` to `/etc/binderhub/config/binderhub_config.py`, making things more flexible especially on Kubernetes.

I have two open questions though:

1. should we leave a "default" `binderhub_config.py` file "hardcoded" in the image on the new path, which will then be overridden? To enable users without the configfile at hand to run it outside of Kubernetes.
2. should we change the CMD in the Dockerfile (like done here) or rather change the command + args in the Kubernetes manifest?


consideRatio:
> should we leave a "default" binderhub_config.py file "hardcoded" in the image on the new path, which will then be overridden? To enable users without the configfile at hand to run it outside of Kubernetes.

I'd say we should aim for a hard failure, and let users mount it with docker or similarly if they run it locally. I think a hard failure is better than confusion when there is a small difference.

> should we change the CMD in the Dockerfile (like done here) or rather change the command + args in the Kubernetes manifest?

I think both.

PR #1084, opened 2020-03-27
put meta tags for social media previews into block

bitnik:
This PR puts social media preview tags into block. This makes it easier to customize templates.

Currently on about page (https://mybinder.org/about), there is only description info for social media previews: `<meta property="og:description" content="Reproducible, sharable, open, interactive computing environments.">`. I also changed it, so about page has the same tags as index page for social media. I can also totally remove it from about page. What do you think?

### DIFF
--- a/binderhub/templates/about.html
+++ b/binderhub/templates/about.html
@@ -1,10 +1,5 @@
 {% extends "page.html" %}
 
-{% block head %}
-<meta property="og:description" content="Reproducible, sharable, open, interactive computing environments.">
-{{ super() }}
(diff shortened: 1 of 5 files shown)


yuvipanda:
Happy to merge this if you can fix the conflict, @bitnik

bitnik:
This is done :)

yuvipanda:
Thanks, @bitnik!!! I'll try deploy this soon too

yuvipanda:
This seems to work ok, @bitnik. Thank you!

PR #1093, opened 2020-04-26
[WIP] Move z2bhub docs to subdirectory

betatim:
This moves some of the source files which are in the top level docs directory to the sub-directory they belong to.

This cleans up our documentation sources a bit.


betatim:
Generating the redirects works locally but not on RTD

PR #1094, opened 2020-04-26
[MRG] Tweak generation of RTD redirect files

betatim:
Follow up to #1093

(is there a better way to debug RTD builds than doing lots of small PRs?)

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -193,7 +193,7 @@ internal_redirect_template = """
 <html>
   <head>
     <title>Going to {new_url}</title>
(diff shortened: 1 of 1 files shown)


betatim:
Nice. Now it seems to work.

The old link https://binderhub.readthedocs.io/en/latest/turn-off.html will now redirect you to the new location of the content. Advantage of using this over the redirects from the RTD GUI is that we can control it via `conf.py` (less clicking, etc) and we can take it with us when we change hoster (no lockin).

PR #1096, opened 2020-04-30
Add podAnnotations config for the binder deployment - will enable checksum change based restarts

bitnik:
This makes it possible to add additional annotations to binder deployment.

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -33,6 +33,9 @@ spec:
         # This lets us autorestart when the configmap changes!
         checksum/config-map: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
         checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+        {{- if .Values.annotations }}
+        {{- .Values.annotations | toYaml | trimSuffix "\n" | nindent 8 }}
+        {{- end }}
     spec:
       {{ if .Values.initContainers -}}
       initContainers:
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -33,8 +33,8 @@ spec:
         # This lets us autorestart when the configmap changes!
         checksum/config-map: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
         checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
-        {{- if .Values.annotations }}
-        {{- .Values.annotations | toYaml | trimSuffix "\n" | nindent 8 }}
+        {{- with .Values.podAnnotations }}
+        {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
     spec:
       {{ if .Values.initContainers -}}
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -197,4 +197,4 @@ initContainers: []
 extraVolumes: []
 extraVolumeMounts: []
 extraEnv: []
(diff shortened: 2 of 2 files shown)


bitnik:
Hi, could anyone have a look at this PR?

This will help us to improve our continuous deployment. Here is what we want to do:
- in values.yaml of our binderhub helm chart, we set an annotation `rollme` to binderhub with a default value
- CD script calculates the sha256 checksum of our custom templates, which are mounted as configmap
- CD script runs `helm upgrade` with `--set binderhub.annotations.rollme=<calculated_sha256_checksum>`

So whenever we update our templates and push that changes, a new binder pod is rolled.

choldgraf:
A quick thought- could you document this feature so others know what it could be used for etc?

bitnik:
I think https://github.com/jupyterhub/binderhub/pull/1096#issuecomment-639420359 is too specific to document in the repo, it is just a use case how we would use additional annotations. But this makes me think that probably we should create a post about how we setup our CD.

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -33,6 +33,9 @@ spec:
         # This lets us autorestart when the configmap changes!
         checksum/config-map: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
         checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+        {{- if .Values.annotations }}
+        {{- .Values.annotations | toYaml | trimSuffix "\n" | nindent 8 }}
+        {{- end }}
```
consideRatio, line 38:
```suggestion
        {{- with .Values.podAnnotations }}
        {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
        {{- end }}
```

helm-chart/binderhub/values.yaml
```diff
@@ -197,3 +197,4 @@ initContainers: []
 extraVolumes: []
 extraVolumeMounts: []
 extraEnv: []
+annotations: {}
```
consideRatio, line 200: I suggest this because I've seen Helm chart typically differentiate between pod annotations and deployment annotations, and you are changing podAnnotations in this PR.

```suggestion
podAnnotations: {}
```

consideRatio:
> A quick thought- could you document this feature so others know what it could be used for etc?

All our Helm charts use this kind of functionality already by having checksum annotations of our configuration, so @bitnik PR can be summarized as enabling Kenan's use of this kind of trick as a Helm chart user for additional custom configuration that should also trigger a restart of the binderhub pod.

helm-chart/binderhub/values.yaml
bitnik: thanks @consideRatio for the review!

> Helm chart typically differentiate between pod annotations and deployment annotations

Where can I read about this?

consideRatio: Here is output from `helm create` in Helm v3.4.0. There isn't much documentation about this as a best practice or similar so its quite hard to follow the Helm chart trends in my mind. I have contributed changes to the `helm create` templates, and that in my mind is the best way to verify a practice is worthy as a best practice, along with getting changes merged to their documentation about best practices.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ include "tmp.fullname" . }}
  labels:
    {{- include "tmp.labels" . | nindent 4 }}
spec:
  {{- if not .Values.autoscaling.enabled }}
  replicas: {{ .Values.replicaCount }}
  {{- end }}
  selector:
    matchLabels:
      {{- include "tmp.selectorLabels" . | nindent 6 }}
  template:
    metadata:
      {{- with .Values.podAnnotations }}
      annotations:
        {{- toYaml . | nindent 8 }}
      {{- end }}
# ...
```

bitnik: thanks!

PR #1102, opened 2020-05-21
[WIP] Experimenting with Stimulus JS

betatim:
This PR exists mostly so you can watch while I experiment with [StimulusJS](https://stimulusjs.org/). My goal is to understand how we'd integrate it into our webpack config and eventually how to have a single text input field ("awesome bar") which will then use code from https://github.com/Carreau/open-with-binder to generate the "launch link".

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,4 @@
+{
+  "presets": ["@babel/preset-env"],
+  "plugins": ["@babel/plugin-proposal-class-properties"]
+}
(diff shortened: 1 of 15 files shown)


betatim:
Kapture 2020-05-23 at 15 47 01

Made some progress with a new and simpler "launch a binder" page. It has just one form field into which you can paste a URL. It is using the code from https://github.com/Carreau/open-with-binder/blob/master/content_scripts/binderify.js to parse that URL and generate the "build" URL. Then it redirects people there.

Next to the launch button there is a "advanced options" link which takes you to the full launch form. The new page is at `/beta` so that nothing changes for unsuspecting visitors.

Next is handling things like DOIs/not links, adding back the badge copy&paste areas and making it more robust.

I made some small changes to the CSS for the new form but they should not change the old landing page, at least not on purpose.

betatim:
This PR is super outdated so closing it.

Issue #942, opened 2019-09-01
Unquote displayed link for "pure git" repository


betatim:
When a user launches a "pure git" repository we show the following on the build page:

<img width="895" alt="Screenshot 2019-09-01 at 16 31 22" src="https://user-images.githubusercontent.com/1448859/64077986-1641f600-ccd6-11e9-9c42-49c137802b30.png">

It should display something more like "Starting repository: https://gitlab.kwant-project.org...." instead. To reproduce this open https://mybinder.org/v2/git/https%3A%2F%2Fgitlab.kwant-project.org%2Fqt%2Flocalization_landscape.git/master?urlpath=lab/tree/index.ipynb

The place to change this is probably in https://github.com/jupyterhub/binderhub/blob/master/binderhub/repoproviders.py#L219 at the location where the value to display is passed to the template. Or maybe we need to do this in the template itself. Depends a bit on how the value passed into the template is used.

To work on this you need to know (or teach yourself) a little bit about URL quoting and Jinja templates.

OrthoDex:
Hi! Binder user here, can I take this up?

souvik3333:
Hi, Is anyone working on this?

PR #1103, opened 2020-05-21
decodeURIComponent the spec so it gets shown unquoted in the status message.

kpaschen:
The spec gets passed into the javascript (and the loading template)
quoted and is also used in build url created by BinderImage, so I did
not want to remove the url quoting for the spec everywhere, just in the
display string.
Intended to fix https://github.com/jupyterhub/binderhub/issues/942

If someone else is already working on this, please disregard my pull request, I'm just poking about to understand the codebase.

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -163,7 +163,7 @@ function build(providerSpec, log, path, pathType) {
   var spec = providerSpec.slice(providerSpec.indexOf('/') + 1);
   // Update the text of the loading page if it exists
   if ($('div#loader-text').length > 0) {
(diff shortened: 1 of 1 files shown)


kpaschen:
The Travis CI build failure looks unrelated, i just can't see how to trigger a re-run without adding another commit.

betatim:
Thanks for fixing this. In general people are encouraged to create a PR to signal "i'm working on this". This means if there isn't a PR linked to/from the issue already it should be safe to assume no one is working on it.

Your changes should be live on mybinder.org now (deployed in https://github.com/jupyterhub/mybinder.org-deploy/pull/1451)

PR #1109, opened 2020-06-04
Add a tip about ad blocker in the loading messages

1ucian0:
I was having problems with empty log and notebook not loading. In gitter, @sgibson91 and @betatim suggested that my ad blocker was interfering. Whitelisting mybinder.org did the trick.
 
This PR adds that tip into the loading messages.


betatim:
Merged! Should be live in a couple of hours :D

PR #1111, opened 2020-06-16
add resolved ref information into launch events

bitnik:
We have (unresolved) spec in launch events. I think having resolved ref info is also useful for later analysis.

### DIFF
--- a/binderhub/event-schemas/launch.json
+++ b/binderhub/event-schemas/launch.json
@@ -24,7 +24,7 @@
         },
         "ref": {
             "type": "string",
-            "description": "Resolved reference to the repo when it is being launched"
+            "description": "Resolved reference for the repo at the time of launch"
         },
         "status": {
             "enum": ["success", "failure"],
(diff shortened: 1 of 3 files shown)


binderhub/event-schemas/launch.json
```diff
@@ -22,6 +22,10 @@
             "type": "string",
             "description": "Provider specification for the repo being launched. Usually, <reponame>/<commit-specification>"
         },
+        "ref": {
+            "type": "string",
+            "description": "Resolved reference to the repo when it is being launched"
```
betatim, line 27:
```suggestion
            "description": "Resolved reference for the repo at the time of launch"
```
What do you think?

betatim:
To double check: a "unresolved ref" is for example `master` and a "resolved ref" would be `abc124123abcd` (a git commit hash). Right?

bitnik:
> To double check: a "unresolved ref" is for example `master` and a "resolved ref" would be `abc124123abcd` (a git commit hash). Right?

yes :)

betatim:
Merged :)

PR #1119, opened 2020-07-01
Add documentation for the launcher

### LINKED ISSUE #1118: Reference documentation for the launcher
### Proposed change

It would be nice to also have documentation for the launcher available here:

https://binderhub.readthedocs.io/en/latest/reference/ref-index.html

### Alternative options

Looking at the source code.

Users who would like to customize the behavior of the launcher, for example by providing a `pre_launch_hook`.

PR incoming.

jtpio:
Fixes #1118. 

image

### DIFF
--- a/doc/reference/launcher.rst
+++ b/doc/reference/launcher.rst
@@ -1,0 +1,16 @@
+launcher
+========
+
+
+Module: :mod:`binderhub.launcher`
+---------------------------------
+
+.. automodule:: binderhub.launcher
+
+.. currentmodule:: binderhub.launcher
+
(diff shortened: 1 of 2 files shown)


betatim:
Thank you!

betatim:
For a PR like this I always wonder: do we need the issue? Seems like we can reduce by one the number of notifications people get by just having the PR. Maybe the rule of thumb could be "if you are about to create a PR for something don't first create an issue, just make create the PR"?

jtpio:
Yes this one was very ad-hoc and could have been PR only :+1:

PR #1125, opened 2020-07-14
Fix possible causes of cordoned nodes in image-cleaner

minrk:
If image-cleaner crashed during cleaning, it could leave a node cordoned until the next time it needs to delete images. This PR should make this less likely with two changes:

- Uncordon prior to crashing due to unhandled errors. Right now, unhandled errors cleaning up an image would result in image-cleaner exiting and leaving the node cordoned, leading to the issue above.
- Check if cordoned at startup, to cleanup after crash of previous run. An annotation is added to track whether it was image-cleaner that was responsible for the current cordoned state). The first change should make this less likely, but it will never be impossible.

Additional minor code reorganization:

- move cordon event logging into cordon/uncordon functions, rather than prior to each call

We still haven't found logs for diagnosing the issue, since GKE appears to be the only deployment we have with accessible logs across pod restarts, but I have been able to confirm that yesterday, OVH cordoned node-1 without releasing it at 19:08 yesterday, and image-cleaner crashed at 19:09, making it pretty likely that this was what happened.

### DIFF
--- a/helm-chart/images/image-cleaner/image-cleaner.py
+++ b/helm-chart/images/image-cleaner/image-cleaner.py
@@ -20,6 +20,9 @@ logging.basicConfig(format='%(asctime)s %(message)s',
                     level=logging.INFO)
 
 
+annotation_key = "hub.jupyter.org/image-cleaner-cordoned"
+
+
 def get_absolute_size(path):
     """
     Directory size in bytes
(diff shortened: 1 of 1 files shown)


betatim:
Would uncordoning be a legitimate use of a [`atexit`](https://docs.python.org/3/library/atexit.html) handler in Python? It almost always feels like you shouldn't use a exit handler because there are subtleties about when they run, when they don't run, what you can do, etc but this seems like a case where it would be useful. Or maybe a `try ... finally` block?

minrk:
I think try/finally is exactly the right pattern. I just didn't want to make the diff that big for some reason :/. I can take a stab at that another day, unless someone else wants to.

PR #1127, opened 2020-07-15
Add new help message to cycle through on loading page

sgibson91:
This PR adds a new help message to cycle through that points to our discourse post explaining how users can speed up their Binder launches. It also rephrases another message to remove the term "whitelisting" in reference to ad blocking.


betatim:
Merged!

PR #1137, opened 2020-08-03
Allow head requests for the health endpoint

rkevin-arch:
We're considering running our own binderhub, and one of the uptime checkers we're using only sends HEAD requests. Tornado doesn't like the HEAD requests and return a 405, so this patch makes it respond to HEAD requests as well as GET. I've tested it in minikube and it should work.

### DIFF
--- a/binderhub/health.py
+++ b/binderhub/health.py
@@ -151,7 +151,12 @@ class HealthHandler(BaseHandler):
         }
         return usage
 
-    async def get(self):
+    async def check_all(self):
+        """Runs all health checks and returns a tuple (overall, checks).
+
+        `overall` is a bool representing the overall status of the service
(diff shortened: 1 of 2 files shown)


rkevin-arch:
Not sure what the failing test case is (test_dataverse?), but it doesn't look like this code change caused it. Let me know if it's something I need to fix.

consideRatio:
@rkevin-arch I think the dataverse test is unstable for other reasons - thank you for adding the test!

rkevin-arch:
What's the status of this PR? Let me know if there's anything I can do to get this merged.

betatim:
Thanks a lot. Sorry that this slipped through the cracks.

Merged (will get deployed to mybinder.org in a few hours).

PR #1139, opened 2020-08-10
fix default value of hub_url in config

bitnik:
In https://github.com/gesiscss/persistent_binderhub/issues/5 we found out that if someone installs BinderHub with auth enabled, 
deployment fails with 
`Error: render error in "persistent_binderhub/charts/binderhub/templates/deployment.yaml": template: persistent_binderhub/charts/binderhub/templates/deployment.yaml:98:74: executing "persistent_binderhub/charts/binderhub/templates/deployment.yaml" at <"/">: invalid value; expected string
` 

The error indicates https://github.com/jupyterhub/binderhub/blob/23ae2061decdf8e9f7b94829cabe602b6ed36519/helm-chart/binderhub/templates/deployment.yaml#L98 and it happens only if the deployment is not an upgrade from a standard BinderHub deployment (auth disabled), but it is an installation of BinderHub directly with authentication included. The reason is that on cloud-based environments `BinderHub.hub_url` is missing during the first round of the installation (https://binderhub.readthedocs.io/en/latest/zero-to-binderhub/setup-binderhub.html#install-binderhub), because it is provided after the first installation when JupyerterHub is ready (https://binderhub.readthedocs.io/en/latest/zero-to-binderhub/setup-binderhub.html#connect-binderhub-and-jupyterhub).

This PR fixes the default value of  `BinderHub.hub_url` in helm config, so helm won't complain about it during the first installation and then user can set `BinderHub.hub_url` when it is available. It also fixes it `binderhub_config.py`, so it uses default value of hub_url "" instead of `LazyConfigValue`.

### DIFF
--- a/helm-chart/images/binderhub/binderhub_config.py
+++ b/helm-chart/images/binderhub/binderhub_config.py
@@ -84,7 +84,8 @@ if os.getenv('BUILD_NAMESPACE'):
     c.BinderHub.build_namespace = os.environ['BUILD_NAMESPACE']
 
 if c.BinderHub.auth_enabled:
-    hub_url = urlparse(c.BinderHub.hub_url)
+    hub_url = get_value('config.BinderHub.hub_url', '')
+    hub_url = urlparse(hub_url)
     c.HubOAuth.hub_host = '{}://{}'.format(hub_url.scheme, hub_url.netloc)
     if 'base_url' in c.BinderHub:
         c.HubOAuth.base_url = c.BinderHub.base_url
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -95,7 +95,7 @@ spec:
               key: "binder.hub-token"
         {{- if .Values.config.BinderHub.auth_enabled }}
         - name: JUPYTERHUB_API_URL
-          value: {{ (print (.Values.config.BinderHub.hub_url | trimSuffix "/") "/hub/api/") }}
+          value: {{ print (.Values.config.BinderHub.hub_url | default "" | trimSuffix "/") "/hub/api/" }}
         - name: JUPYTERHUB_BASE_URL
           value: {{ .Values.jupyterhub.hub.baseUrl | quote }}
         - name: JUPYTERHUB_CLIENT_ID


bitnik:
Actually now I am thinking that instead of my first changes, it would be better to set the default value in `values.yaml` as

```
config:
  BinderHub:
    hub_url: ''
```

what do you think?

helm-chart/images/binderhub/binderhub_config.py
```diff
@@ -84,7 +84,8 @@ def get_value(key, default=None):
     c.BinderHub.build_namespace = os.environ['BUILD_NAMESPACE']
 
 if c.BinderHub.auth_enabled:
-    hub_url = urlparse(c.BinderHub.hub_url)
+    hub_url = get_value('config.BinderHub.hub_url', '')
+    hub_url = urlparse(hub_url)
     c.HubOAuth.hub_host = '{}://{}'.format(hub_url.scheme, hub_url.netloc)
```
minrk, line 89: we can be more consistent about getting information off of the config object instead of going all the way back to the yaml. We have similar conditions for when other optional values are missing:

```suggestion
    if "hub_url" in config.BinderHub:
        hub_url = urlparse(config.BinderHub. hub_url)
        c.HubOAuth.hub_host = '{}://{}'.format(hub_url.scheme, hub_url.netloc)
```

but does this work if hub_url is not specified? What happens if hub_url is empty, does the binder auth still work? If it won't work, we should probably raise here instead of producing a known-invalid value

bitnik: thanks for the review Min! 

> but does this work if hub_url is not specified? What happens if hub_url is empty, does the binder auth still work? If it won't work, we should probably raise here instead of producing a known-invalid value

if hub_url is empty, binder auth doesnt work (but I guess it would still work if hub and binder had the same domain). Btw if hub_url is empty, binder launching wouldnt work neither. That is why this step (https://binderhub.readthedocs.io/en/latest/zero-to-binderhub/setup-binderhub.html#connect-binderhub-and-jupyterhub) is necessary, isnt it?

bitnik: I wanted to accept your suggestion already, because I wont be available the coming week. If there will be more changes required, I will do it when I am back

bitnik:
@minrk I am not sure how to continue here. Could you do another review?

consideRatio:
> Actually now I am thinking that instead of my first changes, it would be better to set the default value in `values.yaml` as
> 
> ```
> config:
>   BinderHub:
>     hub_url: ''
> ```
> 
> what do you think?

I agree! That is what we do in z2jh as well I think, and it makes it clearer to the reader of the config than if the default value is different from the configured empty string etc.

PR #1144, opened 2020-08-25
add build_memory_request config

minrk:
allows separate memory_request and memory_limit on the build pod

required to implement plan in https://github.com/jupyterhub/mybinder.org-deploy/issues/1529

the build pod doesn't itself consume much resources, but this allows separate request and limit *for* the build, applied via the build pods.

limit is actually applied on the build via repo2docker flag and request allows kubernetes to assign builds more appropriately across nodes

Took the opportunity to make Build's many constructor arguments explicitly keyword-only via `*`.

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -36,9 +36,28 @@ class Build:
         API instead of having to invent our own locking code.
 
     """
-    def __init__(self, q, api, name, namespace, repo_url, ref, git_credentials, build_image,
-                 image_name, push_secret, memory_limit, docker_host, node_selector,
-                 appendix='', log_tail_lines=100, sticky_builds=False):
+
+    def __init__(
+        self,
+        q,
+        api,
+        name,
+        *,
+        namespace,
+        repo_url,
+        ref,
+        build_image,
+        docker_host,
+        image_name,
+        git_credentials=None,
+        push_secret=None,
+        memory_limit=0,
+        memory_request=0,
+        node_selector=None,
+        appendix="",
+        log_tail_lines=100,
+        sticky_builds=False,
+    ):
         self.q = q
         self.api = api
         self.repo_url = repo_url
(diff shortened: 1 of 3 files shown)


binderhub/build.py
```diff
@@ -36,9 +36,28 @@ class Build:
...
+        q,
+        api,
+        name,
+        *,
+        namespace,
+        repo_url,
+        ref,
+        build_image,
+        docker_host,
+        image_name,
```
betatim, line 51: Keyword only with no default value is a crazy new world in Python 👍

binderhub/build.py
```diff
@@ -255,7 +274,7 @@ def submit(self):
                         volume_mounts=volume_mounts,
                         resources=client.V1ResourceRequirements(
                             limits={'memory': self.memory_limit},
-                            requests={'memory': self.memory_limit}
+                            requests={'memory': self.memory_request},
```
betatim, line 277: The travis error is about `self.memory_request` not existing: `AttributeError: 'Build' object has no attribute 'memory_request'` I think we need to assign it in the constructor as it is new and `Build` doesn't use traitlets

binderhub/build.py
minrk: I know, I love it!

binderhub/build.py
minrk: Fixed, thanks! I write so much in traitlets-land now that I've forgotten how to write normal Python constructors...

minrk: Should we make Build a Configurable? Lots of BinderHub's configuration is ultimately only for Build, and we need to do quite a bit of passing around through settings, constructors, etc. because where it's exposed as config (BinderHub) is not where it's used (Build).

betatim: Moving to a configurable sounds like a good idea. Only caveat is that I don't know why it isn't already a configurable and maybe it was a on purpose decision?

betatim:
Merged! Next up the mybinder.org-deploy PR :D

`Build` as configurable sounds like a new PR..

PR #1149, opened 2020-09-08
fix: launch retrying

bitnik:
When a launch fails, binder doesnt retry but raises "Internal Server Error" because variable `duration` is not defined:

https://github.com/jupyterhub/binderhub/blob/de1d0cf8a19f7cbbebc4a7d45f68b7b3ec13948e/binderhub/builder.py#L522-L568

The problem is with line 534 where `duration` is defined. When line 529 (`launcher.launch`) fails, line 562 (inside except block) logs a retry message including duration. But it fails because variable `duration` is not defined.

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -531,14 +531,8 @@ class BuildHandler(BaseHandler):
                                                     server_name=server_name,
                                                     repo_url=self.repo_url,
                                                     extra_args=extra_args)
-                duration = time.perf_counter() - launch_starttime
-                LAUNCH_TIME.labels(status="success", retries=i).observe(duration)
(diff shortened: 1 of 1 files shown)


betatim:
LGTM. If we can find a mybinder.org link that demonstrates the breakage we can quickly confirm that it is broken now and then fixed when we deploy this

bitnik:
Thanks to both of you :)

In standard binderhub there are 2 specific errors can cause this, one is "Failed to create temporary user" which i think we cant produce easily and one is that "too long to launch". In authenticated BinderHub it would be easier to produce because there we could cause "user already has a running server" or "user already has the max named servers"

PR #1153, opened 2020-09-28
demote logs of health/liveness checks to debug-level

minrk:
copy logging logic from jupyterhub, where we do the same thing

this prevents logs of Binder pods from looking like this:

```
binder-96f6c97bb-4958n binder [I 200928 10:46:04 web:2250] 200 GET /health (10.12.32.7) 4.50ms
binder-96f6c97bb-4958n binder [I 200928 10:46:04 web:2250] 200 GET /health (10.12.32.7) 4.53ms
binder-96f6c97bb-4958n binder [I 200928 10:46:04 web:2250] 200 GET /health (10.12.32.7) 4.70ms
binder-96f6c97bb-4958n binder [I 200928 10:46:04 web:2250] 200 GET /about (10.128.0.36) 1.04ms
binder-96f6c97bb-4958n binder [I 200928 10:46:04 web:2250] 200 GET /versions (10.12.34.130) 0.55ms
binder-96f6c97bb-4958n binder [I 200928 10:46:05 web:2250] 200 GET /versions (10.12.34.130) 0.48ms
binder-96f6c97bb-4958n binder [I 200928 10:46:05 web:2250] 200 GET /versions (10.12.34.130) 0.47ms
binder-96f6c97bb-4958n binder [I 200928 10:46:05 web:2250] 200 GET /versions (10.12.34.130) 0.52ms
binder-96f6c97bb-4958n binder [I 200928 10:46:05 web:2250] 200 GET /versions (10.12.34.130) 0.48ms
binder-96f6c97bb-4958n binder [I 200928 10:46:09 web:2250] 200 GET /about (10.128.0.36) 1.24ms
```

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -117,7 +117,7 @@ spec:
             name: binder
         livenessProbe:
           httpGet:
-            path: {{ default "/" .Values.config.BinderHub.base_url }}about
+            path: {{ default "/" .Values.config.BinderHub.base_url }}versions
             port: binder
           initialDelaySeconds: 10
           periodSeconds: 5
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -30,6 +30,7 @@ from .build import Build
 from .builder import BuildHandler
 from .health import HealthHandler
 from .launcher import Launcher
+from .log import log_request
 from .registry import DockerRegistry
 from .main import MainHandler, ParameterizedMainHandler, LegacyRedirectHandler
 from .repoproviders import (GitHubRepoProvider, GitRepoProvider,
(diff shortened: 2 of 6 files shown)


consideRatio:
:heart: excellent work! I love it when the logs isn't cluttered ;D

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -117,7 +117,7 @@ spec:
             name: binder
         livenessProbe:
           httpGet:
-            path: {{ default "/" .Values.config.BinderHub.base_url }}about
+            path: {{ default "/" .Values.config.BinderHub.base_url }}versions
```
bitnik, line 120: Hi @minrk! I know this PR is already merged. But I was on vacation and now have time to go over updates in BinderHub. And I am now just curious if there was a specific reason to choose `/versions` path instead of `/health`?

minrk: If a pod fails a `liveness` check it should be killed and restarted, so the liveness check should only fail if the pod itself is unhealthy or unready. Our health endpoint describes the health of the whole application including pod quotas, etc. If we used `/health` for a liveness probe, whenever we hit a pod quota or if the Hub is restarting, the binder pod itself would be restarted. So I went with the simplest handler we have to measure only "is this pod responding to basic requests promptly". That way, it doesn't get added to rotation while it's not ready to accept http requests, but that's the only condition checked.

We could have a different health check (or modify the one we have) that truly measures only pod health and use that, but `/versions` was the closest to `/about` without serving a human-facing page that I didn't want to demote in the logs.

bitnik: Thank you!

PR #1157, opened 2020-10-01
Enable CodeQL code scanning as a GitHub Action

betatim:
This GitHub Action uses CodeQL to scan our code for security problems.

I'd be interested to see how well it works (or not). Can we enable it for a while to see if we like it or find it annoying/noisy?

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '29 13 * * 5'
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '29 13 * * 5'
```
consideRatio, line 15:
```suggestion
    # At 13:29 on Friday - https://crontab.guru/
    - cron: '29 13 * * 5'
```

consideRatio:
LGTM, will merge following CI tests completed. I added a comment to help quickly understand the cron specification.

consideRatio:
@betatim :'( it seems like we have an unrelated CI failure, where a pip install -r /tmp/requirements.txt fail, because PyYaml dependency yarl fails to install, because gcc fails to build something...

Unrelated I assume right? I dare not merge myself, but I +1 to merge if you also think it is unrelated and safe to merge.

betatim:
I'll investigate the build failure. Unrelated to the PR but if we can't build the hub image then that is a problem that needs fixing :D

PR #1158, opened 2020-10-01
[MRG] Use slim-buster as base image for binderhub and image-cleaner

betatim:
Using Alpine as base distro means we have to build wheels for a lot of
packages which often requires a compiler and other native dependencies.
By using a debian base image we get wheels from pypi for free. The
BinderHub image uses buster already so while the individual image might
be larger the layers will be shared with an image that is already
present in the cluster.

Found during #1157 where the CI failed.

### DIFF
--- a/helm-chart/images/binderhub/Dockerfile
+++ b/helm-chart/images/binderhub/Dockerfile
@@ -6,7 +6,7 @@ ARG DIST=buster
 # The build stage
 # ---------------
 FROM python:3.8-$DIST as build-stage
-# ARG DIST has to be defined again to made available inside this build stage.
+# ARG DIST is defined again to be made available in this build stage's scope.
 # ref: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
 ARG DIST
 
--- a/helm-chart/images/binderhub/Dockerfile
+++ b/helm-chart/images/binderhub/Dockerfile
@@ -1,8 +1,8 @@
 # Using multi-stage builds
-ARG DIST=slim-buster
+ARG DIST=buster
 FROM buildpack-deps:$DIST as build-stage
 # ARG has to occur twice to be used in both contents and FROM
-ARG DIST=slim-buster
+ARG DIST=buster
 
 RUN echo "deb http://deb.nodesource.com/node_14.x $DIST main" > /etc/apt/sources.list.d/nodesource.list \
  && curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
@@ -24,7 +24,7 @@ RUN python3 setup.py bdist_wheel
 
 # The final stage
 # ---------------
-FROM python:3.8-$DIST
+FROM python:3.8-slim-$DIST
 WORKDIR /
 
 # Copy the built binderhub python wheel from the build-stage
--- a/helm-chart/images/binderhub/Dockerfile
+++ b/helm-chart/images/binderhub/Dockerfile
@@ -1,8 +1,8 @@
 # Using multi-stage builds
-ARG DIST=buster
+ARG DIST=slim-buster
 FROM buildpack-deps:$DIST as build-stage
 # ARG has to occur twice to be used in both contents and FROM
-ARG DIST=buster
+ARG DIST=slim-buster
 
 RUN echo "deb http://deb.nodesource.com/node_14.x $DIST main" > /etc/apt/sources.list.d/nodesource.list \
  && curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
--- a/helm-chart/images/binderhub/Dockerfile
+++ b/helm-chart/images/binderhub/Dockerfile
@@ -22,8 +22,7 @@ COPY . /tmp/binderhub
 WORKDIR /tmp/binderhub
 
 # Build the binderhub python library into a wheel and save it to the ./dist
-# folder. Also build pycurl into a wheel as doing it requires curl-config and
-# gcc that aren't available in the slim version of the final stage.
+# folder. There are no pycurl wheels so we build our own in the build stage.
 RUN python setup.py bdist_wheel
 RUN pip wheel pycurl --wheel-dir ./dist
 
--- a/helm-chart/images/binderhub/Dockerfile
+++ b/helm-chart/images/binderhub/Dockerfile
@@ -13,9 +13,10 @@ ARG DIST
 # Install node as required to package binderhub to a wheel
 RUN echo "deb http://deb.nodesource.com/node_14.x $DIST main" > /etc/apt/sources.list.d/nodesource.list \
  && curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
-RUN apt-get update && \
-    apt-get install --yes \
-        nodejs
+RUN apt-get update \
+ && apt-get install --yes \
+        nodejs \
+ && rm -rf /var/lib/apt/lists/*
 
 # Copy the whole git repository to /tmp/binderhub
 COPY . /tmp/binderhub
@@ -33,9 +34,11 @@ FROM python:3.8-slim-$DIST
 WORKDIR /
 
 # The slim version doesn't include git as required by binderhub
-RUN apt-get update && \
-    apt-get install --yes \
-        git
+RUN apt-get update \
+ && apt-get install --yes \
+        git \
+ && rm -rf /var/lib/apt/lists/*
+
 
 # Copy the built wheels from the build-stage. Also copy the image
 # requirements.txt built from the binderhub package requirements.txt and the
--- a/helm-chart/images/binderhub/Dockerfile
+++ b/helm-chart/images/binderhub/Dockerfile
@@ -33,6 +33,11 @@ RUN pip wheel pycurl --wheel-dir ./dist
 FROM python:3.8-slim-$DIST
 WORKDIR /
 
+# The slim version doesn't include git as required by binderhub
+RUN apt-get update && \
+    apt-get install --yes \
+        git
+
 # Copy the built wheels from the build-stage. Also copy the image
 # requirements.txt built from the binderhub package requirements.txt and the
 # requirements.in file using the ./dependency script.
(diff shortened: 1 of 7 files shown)


helm-chart/images/image-cleaner/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM python:3.8-alpine3.11
+ARG DIST=buster
```
minrk, line 1: `slim-buster`?

helm-chart/images/image-cleaner/Dockerfile
```diff
@@ -1,4 +1,5 @@
...
+FROM python:3.7-$DIST
```
minrk, line 2: Is there a reason for the variant and not the Python version to be an ARG? Maybe change the whole FROM:

```suggestion
FROM python:3.8-slim-buster
```

betatim: Changed

betatim:
Let's see what the tests say. This now switches the cleaner from alpine to slim-buster (shouldn't be controversial) and upgrade the BinderHub image from Python 3.7 to 3.8 (shouldn't break things but who knows).

helm-chart/images/binderhub/Dockerfile
```diff
@@ -1,8 +1,8 @@
 # Using multi-stage builds
-ARG DIST=buster
+ARG DIST=slim-buster
```
minrk, line 2: I'm not sure buildpack-deps has `slim` variants (there are kind un-slim by definition). So maybe need DIST to be `buster` and add `slim` below in `python:3.8-slim-$DIST`?

betatim:
How can it be that #1214 built successfully but the tests in this PR fail?

manics:
https://pypi.org/project/pycurl/7.43.0.5/#files doesn't have any Linux wheels, perhaps because each distribution has it's own curl and openssl libraries? I assume this means 7.43.0.5 would still have been built?

Using the slim image here may mean curl-config is missing:
```
$ docker run -it --rm python:3.8-buster curl-config
Usage: curl-config [OPTION]

Available values for OPTION include:

  --built-shared says 'yes' if libcurl was built shared
  --ca        ca bundle install path
  --cc        compiler
  --cflags    pre-processor and compiler flags
  --checkfor [version] check for (lib)curl of the specified version
  --configure the arguments given to configure when building curl
  --features  newline separated list of enabled features
  --help      display this help and exit
  --libs      library linking information
  --prefix    curl install prefix
  --protocols newline separated list of enabled protocols
  --ssl-backends output the SSL backends libcurl was built to support
  --static-libs static libcurl library linking information
  --version   output version information
  --vernum    output the version information as a number (hexadecimal)
```
```
$ docker run -it --rm python:3.8-slim-buster curl-config
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"curl-config\": executable file not found in $PATH": unknown.
```

consideRatio:
- [x] The size difference between alpine, slim-buster, and buster is ~15 MB, ~45 MB, ~300 MB. I agree with aiming for use of slim buster as this PR already does.
- [x] The image-cleaner image was an alpine image, but is now transitioning to be a slim-buster image. Everything works!
- [ ] The binderhub image was a buster image, but is now transitioning to be a slim-buster image. We have the curl-config error.

I'll try figure out how to make it work.

helm-chart/images/image-cleaner/Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM python:3.8-alpine3.11
+FROM python:3.8-slim-buster
 
-ADD requirements.txt /tmp/requirements.txt
-RUN pip3 install --no-cache-dir -r /tmp/requirements.txt
+COPY requirements.txt /tmp/requirements.txt
+RUN pip install --no-cache-dir -r /tmp/requirements.txt
 
-ADD image-cleaner.py /usr/local/bin/image-cleaner.py
+COPY image-cleaner.py /usr/local/bin/image-cleaner.py
```
consideRatio, line 6: COPY is best practice for some reason I forgot.

consideRatio (approved):
This looks good to me at this point and I consider it ready for merge.

helm-chart/images/binderhub/Dockerfile
```diff
@@ -1,48 +1,57 @@
...
+FROM python:3.8-slim-$DIST
 WORKDIR /
 
-# Copy the built binderhub python wheel from the build-stage
-# to the current directory within the container.
-COPY --from=build-stage /tmp/binderhub/dist/*.whl .
+# The slim version doesn't include git as required by binderhub
+RUN apt-get update && \
+    apt-get install --yes \
+        git
```
betatim, line 39: You will also have to delete the caches and other temporary files again. Otherwise the image will contain extra files.

helm-chart/images/binderhub/Dockerfile
```diff
@@ -1,48 +1,57 @@
...
-# ARG has to occur twice to be used in both contents and FROM
+# We use a build stage to package binderhub and pycurl into a wheel which we
+# then install by itself in the final image which is relatively slimmed.
 ARG DIST=buster
 
+
+# The build stage
+# ---------------
+FROM python:3.8-$DIST as build-stage
+# ARG DIST has to be defined again to made available inside this build stage.
```
betatim, line 9: Please use the original comment or fix the english of the new comment

requirements.txt
```diff
@@ -1,12 +1,19 @@
-# Keep this file in sync with `doc/doc-requirements.txt`!
+# About pycurl:
+# - pycurl is strongly recommended for performance but not strictly required by
+#   BinderHub.
+# - pycurl require both `curl-config` and `gcc` to be available when installing.
```
betatim, line 4:
```suggestion
# - pycurl requires both `curl-config` and `gcc` to be available when installing
#   it from source.
```

requirements.txt
```diff
@@ -1,12 +1,19 @@
-# Keep this file in sync with `doc/doc-requirements.txt`!
+# About pycurl:
+# - pycurl is strongly recommended for performance but not strictly required by
+#   BinderHub.
```
betatim, line 3: This is hard to understand. The binderhub package authors (us) decided to make pycurl a required dependency by including it in the list of packages required in `setup.py`. So pycurl isn't optional for the user of binderhub.

The package authors could have decided to not make it a dependency. This is when it was optional. So I would rephrase it like this
```suggestion
# - pycurl is a dependency but not listed here, only in `setup.py`
```

requirements.txt
```diff
@@ -1,12 +1,19 @@
...
+# - pycurl require both `curl-config` and `gcc` to be available when installing.
+# - The binderhub package require pycurl, but it is a dependency augmented to
+#   this files listed dependencies from within setup.py so that it can be
+#   either omitted as in our ReadTheDocs build that does code inspection of the
+#   binderhub package, or be installed in a separate build step as in the
+#   BinderHub docker image.
```
betatim, line 9:
```suggestion
# - the reason we do not list pycurl here is that it is not needed when our
#   code is imported to generate the documentation and it is tricky to install
#   in the ReadTheDocs build environment.
# - instead we manually add pycurl to the list of dependencies in setup.py
#   which is the list used when installing the binderhub package.
```

setup.py
```diff
@@ -13,6 +13,9 @@
         l.strip() for l in f.readlines()
         if not l.strip().startswith('#')
     ]
+    # pycurl is augmented to requirements.txt as a dependency for the binderhub
+    # package for reasons explained within requirements.txt.
```
betatim, line 17:
```suggestion
    # manually add pycurl here, see comment in requirements.txt
```

helm-chart/images/binderhub/Dockerfile
```diff
@@ -1,48 +1,57 @@
...
 
 # Copy the whole git repository to /tmp/binderhub
 COPY . /tmp/binderhub
 WORKDIR /tmp/binderhub
 
-# Build binderhub the python library
-RUN python3 setup.py bdist_wheel
+# Build the binderhub python library into a wheel and save it to the ./dist
+# folder. Also build pycurl into a wheel as doing it requires curl-config and
+# gcc that aren't available in the slim version of the final stage.
```
betatim, line 26:
```suggestion
# folder. There are no pycurl wheels so we build our own in the build stage.
```

betatim (commented):
I added some suggestions and a comment.

betatim:
Thanks for taking the time to investigate this. In the future I'd prefer it if you could leave a comment instead of pushing changes to my PRs. For this PR the thing I was missing was that this PR uses the slim image variant and in #1214 the "full" image was being used. This together with manics observation that even pycurl 7.43.0.5 didn't provide linux wheels.

While it is nice to swap things like `ADD` for `COPY` in the `Dockerfile`, mixing these in with other changes adds noise to this PR. This is because the changes are not required as well as being unrelated to what is being worked on in this PR. It would be nicer to make a separate PR to make these kinds of style improvements.

helm-chart/images/binderhub/Dockerfile
consideRatio: Thanks! I mistook the comment below from Dockerfile's best practices that we may not need to do the `rm -rf /var/lib/apt/lists/*` but it only meant we don't need to do `apt-get clean` in debian/docker base images.

> Official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.

consideRatio:
> In the future I'd prefer it if you could leave a comment instead of pushing changes to my PRs.

I hear you @betatim, I will! I make an exception for the commits I push right now because I think you meant for me to push the fixes requested in your comments.

> While it is nice to swap things like ADD for COPY in the Dockerfile, mixing these in with other changes adds noise to this PR. This is because the changes are not required as well as being unrelated to what is being worked on in this PR. It would be nicer to make a separate PR to make these kinds of style improvements.

I agree it does, since it would likely be a merge conflict to change that in a separate PR, it would be something that would need to be added to a todo list though, and at this point I feel it is a work/benefit compromise I'm no longer happy to make. I think it could have been made into a dedicated commit though, together with the Dockerfile best practices note about the change.

> For other items (files, directories) that do not require ADD’s tar auto-extraction capability, you should always use COPY.
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

PR #1160, opened 2020-10-04
Remove federation related docs to split using and operating a bhub docs

choldgraf:
This is a general docs improvement that does a bit of restructuring and tidying. It also complements https://github.com/jupyterhub/binder/pull/213 as I moved over some content to these docs from there.

cc @consideRatio in case he wants to take a look at these too ;-)


doc/conf.py
```diff
@@ -114,7 +109,10 @@
 # further.  For a list of options available for each theme, see the
 # documentation.
 #
-html_theme_options = {}
+html_theme_options = {
+  "github_url": "https://github.com/jupyterhub/binderhub",
+  "twitter_url": "https://twitter.com/mybinderteam",
+}
```
consideRatio, line 115: Just like in the other PR, we can allow the edit option right?

```suggestion
html_theme_options = {
  "use_edit_page_button": True,
  "github_url": "https://github.com/jupyterhub/binderhub",
  "twitter_url": "https://twitter.com/mybinderteam",
}
```

doc/doc-requirements.txt
```diff
@@ -1,6 +1,6 @@
 # documentation specific extra packages
 sphinx>=1.7
```
consideRatio, line 2: If i recall correctly, having a dedicated entry for sphinx as well as the order of entriers matters greatly in this file to make everything work smoothly on RTD. This relates to RTD coming with a pre-installed sphinx version and given how `pip` will try to please the various requirements.

This is from z2jh's `requirements.txt, which followed https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1658 and https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1723.

```shell
# Order matters:
# - myst-parser depends on "sphinx>=2,<3"
# - pydata-sphinx-theme depends on "sphinx"
# - sphinx-copybutton depends on "sphinx>=1.8"
#
# Listing either pydata-sphinx-theme or sphinx-copybutton first will make the
# myst-parser constraints on sphinx be ignored, so myst-parser should go
# first. This is only relevant if sphinx==1.* is already installed in the
# environment, which it is on ReadTheDocs.
#
myst-parser
pydata-sphinx-theme
pyyaml
sphinx-autobuild
sphinx-copybutton
```

```suggestion
```

choldgraf:
thanks - both applied! I also turned on readthedocs preview builds and if that works, will remove the circleci build (since it's just for docs preview)

consideRatio (approved):
Other than the requirements.txt details that I wish no-one need to experience debugging more than me and I think @sgibson91 have already ended up doing, this LGTM! Thank you for your excellent work to make the documentation so great @choldgraf !

doc/doc-requirements.txt
```diff
@@ -1,6 +1,5 @@
 # documentation specific extra packages
-sphinx>=1.7
```
betatim, line 2: can we keep sphinx as a dependency here? I use `doc/doc-requirements.txt` to install things in my dev environment so having this is nice because it brings in sphinx and remembers for me that it has to be >=1.7

betatim:
WDYT of the new title? I couldn't resist trying to find an alternative to the very generic one

choldgraf:
I just added back in Sphinx - I *think* at a location that does not trigger the bug @consideRatio mentioned? Is that right @consideRatio ?

Anything else to take care of here?

consideRatio:
@betatim this LGTM, merge agreement?

sgibson91 (approved):
LGTM! 🎉

Issue #1156, opened 2020-09-30
Persist repository build logs for later access


betatim:
### Proposed change
When a repository is being built we stream the output of repo2docker to the website the user sees. It would be great if this log would also be available after the build has failed. For example builds can take a very long time (hours or more) in which users close the tab or otherwise leave. They have no way to recover the build log once the build has failed. The only option is to restart the build and this time wait. It is also hard to copy&paste text from the build log to share for debugging purposes. In both cases it would help to have a stable URL that for some time shows the build log.

<img width="1169" alt="Screenshot 2020-09-30 at 19 30 28" src="https://user-images.githubusercontent.com/1448859/94719440-71b83a80-0353-11eb-9d48-5dec0eaf6ac9.png">

This is an attempt to make a actionable issue out of #155.

### Alternative options
For successful builds we could store the build log in the container image as a special file. This way users could access it from "within" their binder. This would not help users for which the build fails, as they'd never get a launching binder.

### Who would use this feature?
People who have long running builds that need debugging or otherwise want to share/look at the output of a build.

### (Optional): Suggest a solution
A new endpoint in BinderHub that outputs the build log of the last build for a given "repo spec". The log would be overwritten the next time the same spec is built. It would be a reasonably stable URL that is easy to discover. An alternative would be to assign a "build number" to each build, this would be super stable but creates the challenge of how a user would discover what the build number of their build was.

Build logs can be large. This means it is probably not feasible to store them in the process memory of BinderHub. In memory or on disk in the pod storage also means that on a cluster that runs several instances of the binderhub pod you'd need a mechanism for sending requests for the log to the right pod. This points towards the need for an additional service in which to store the logs.

The BinderHub process already sees a copy of the build log as it streams it to users. It could stream it to a log sink at the same time. Or the log could be archived independent of the BinderHub process directly from the repo2docker pod.

manics:
Using an s3 compatible object store might be an option? It's easier to setup on public cloud than managing a K8S persistent volume, and it also means we don't need an endpoint to download the logs, if the bucket is public you can give people a direct link.

You can optionally specify a TTL for auto-deletion, and it should work out cheaper than a disk volume.

minrk:
@manics That sounds like a great solution. We would need to:

1. collect / process logs in binderhub prior to deletion of the pod (logs are actually JSON, what folks want is terminal-processed ANSI output of log messages)
2. upload those to public blob store with TTL, etc.
[...]

PR #1161, opened 2020-10-04
WIP: Upload build logs to S3

manics:
Playing around with S3 logs again (https://github.com/jupyterhub/binderhub/issues/1156). This time uploading repo2docker logs from the BinderHub side instead of the [repo2docker side](https://github.com/jupyterhub/repo2docker/pull/967).

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -457,6 +457,12 @@ class BinderHub(Application):
         """
     )
 
+    s3_logs_endpoint = Unicode("", help="S3 endpoint", config=True)
+    s3_logs_access_key = Unicode("", help="S3 access key ", config=True)
+    s3_logs_secret_key = Unicode("", help="S3 secret key", config=True)
+    s3_logs_bucket = Unicode("", help="S3 bucket", config=True)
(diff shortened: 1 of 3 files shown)


minrk:
Following comments in https://github.com/jupyterhub/repo2docker/pull/967 , if we do it on the binderhub side, I think a sidecar container in the build pod might be a better way to go because we don't have to think about the consequences of binder pod replicas and restarts. A build can have 0, 1, or 2 (or more) RequestHandlers and binder pods watching it, whereas doing it in repo2docker (or in the build pod at least) should mean the lifecycle and everything is more naturally aligned.

Cases we would need to handle if we did at the level in this PR:

- 0 watchers if no requests are currently connected
- 2 (or more) watchers if multiple requests to watch a build of the same image are routed to different binder replicas. There's still one build pod, but multiple pods watching the build logs. This may even happen with multiple requests to one pod, I'd need to refresh my memory, though.

manics:
Thanks... after reading the feedback on https://github.com/jupyterhub/repo2docker/pull/967 I think r2d is the best (and easiest) place for it, let's continue on https://github.com/jupyterhub/repo2docker/pull/967

PR #1165, opened 2020-10-19
Mounting binderhub_config.py instead of letting it be part of the image

iwilltry42:
Revival of #1076 after screwing up branches :roll_eyes: 
FYI @consideRatio


consideRatio (commented):
Thank you for this work! I realize now, i think we should only set the args in k8s and let ENTRYPOINT in dockerfile, because another change we want is to start this pod with a tool called tini i think, and such change should not need to be coupled with the helm chart.

Hmmmm, or should entrypoint for dockerfile be set to just tini, or also include the call to binderhub? Not so important perhaps. Id say lets mirror z2jh for now with regards to that for now.

/ Cheers frlm my mobile during my commute

helm-chart/binderhub/templates/configmap.yaml
```diff
@@ -19,3 +19,6 @@ data:
   values.yaml: |
     {{- $values | toYaml | nindent 4 }}
 
+  {{- /* Glob files to allow them to be mounted by the binderhub pod */ -}}
+  {{- /* key=filename: value=content */ -}}
+  {{- (.Files.Glob "files/*").AsConfig | nindent 2 }}
```
consideRatio, line 24: Nitpick: add back a blank line.

consideRatio:
Aaargh sorry, perhaps tini isnt needed, anyhow, that is another PR.

Two easy requests:
- nitpick about new line in end of file
- inconsistency between splitting dockerfiles cmd into entrypoint/cmd and k8s config of command/args (these sets are the same, but differently named, in k8s/dockerfile). I think id suggest to set Entrypoint in dockerfile and not setting command in k8s is, letting k8s only influence the args to binderhub, while the dockerfile can be free to install and use tini later by changing the entrypoint to wrap the call to binderhub or similar. Hmmm

consideRatio:
@iwilltry42 hey btw thank you for work on k3d :)

I think this PR is in a good enough state for merge already pretty much, i probably overthink this a bit.

helm-chart/binderhub/templates/configmap.yaml
consideRatio, line 24:
```suggestion
  {{- (.Files.Glob "files/*").AsConfig | nindent 2 }}

```

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -64,6 +64,13 @@ spec:
       containers:
       - name: binder
         image: {{ .Values.image.name }}:{{ .Values.image.tag }}
+        command:
+          - python3
+          - -m
+          - binderhub
+        args:
```
consideRatio, line 71:
```suggestion
        args:
```

helm-chart/images/binderhub/Dockerfile
```diff
@@ -38,11 +38,6 @@ RUN pip install --no-cache-dir \
         *.whl \
         -r requirements.txt
 
-# Copy the binderhub configuration to use.
-# FUTURE: Follow the approach taken in Z2JH:
-#         https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1478
-COPY helm-chart/images/binderhub/binderhub_config.py .
-
-CMD ["python3", "-m", "binderhub"]
+CMD ["python3", "-m", "binderhub", "--config", "/etc/binderhub/config/binderhub_config.py"]
```
consideRatio, line 41:
```suggestion
ENTRYPOINT ["python3", "-m", "binderhub"]
CMD ["--config", "/etc/binderhub/config/binderhub_config.py"]
```

consideRatio:
Thank you @iwilltry42! :heart:

iwilltry42:
Thanks for your feedback and the merge @consideRatio :)
However, I agree with your arguments and I would generally add `tini` to the Dockerfile and then set `ENTRYPOINT ["tinit", "--"]` and `CMD ["python3", "-m", "binderhub", "--config", "/etc/binderhub/config/binderhub_config.py"]`.
Then in the Kubernetes manifest we could "duplicate"/"override" the `CMD` with the `args` for the sake of completeness (i.e. people see what it does just from the manifest :thinking:

Issue #1154, opened 2020-09-30
Allow builds of private repositories, using the logged-in user's credentials


rprimet:
This issue is related to #1117, and possibly also to https://github.com/alan-turing-institute/hub23-deploy/issues/272

### Proposed change
Provide a way of dynamically fill repo provider / git credentials for a forge (e.g. gitlab instance) using the currently logged-in user information. This would help set up binderhub instances for use with private repositories for organizations that run a forge (e.g. gitlab instance).

### Alternative options
The alternative option today is to have a 'technical' user (e.g. 'binderhub') on the forge (e.g. gitlab instance) that has (at least) read access to repositories. Then, a personal access token is created for that technical user and passed to the binderhub via the configuration system in `GitLabRepoProvider.private_token`. This token will be used to pull all repositories for all users.

This is not ideal:

- from a usability perspective, it requires the user to grant access to their repos and revoke it as needed, and involves several steps that need to be documented on the private binderhub instance
- from a security perspective, I am unsure of the implications. The project access level that has to be granted to the technical user are sometimes higher than a simple read access (e.g. on gitlab, I believe that the 'reporter' access level is required at least) -- but even when running as the logged-in user, binder may require capabilities that are not strictly necessary because of the somewhat coarse grain of the gitlab permissions model.

### Who would use this feature?
Organizations who would like to run a private binderhub for their teams along with a forge (e.g. gitlab instance) and who want to build and launch private repositories. 

### (Optional): Suggest a solution
Since the repo provider will have to resolve private refs using the currently logged-in user identity, it might be useful to pass user information to the repo providers, at construction time. I'm unsure how clean it is design-wise (and looking for feedback/comments!) but an option might be to pass the handler to RepoProviders at init time.

Then, assuming that a repo provider can access the `current_user` for the handler, a sketch implementation of a private repo provider for a gitlab instance might look like this:

```
class AuthGitLabProvider(GitLabRepoProvider):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.handler = kwargs.get('handler')
        ud = self.get_user_data(self.handler.get_current_user()['name'])
        self.access_token = ud['auth_state']['access_token']

    def get_user_data(self, username):
        r = requests.get(c.HubOAuth.api_url + f'/users/{username}',
            headers={
             'Authorization': 'token %s' % c.HubOAuth.api_token,
            }
        )
        r.raise_for_status()
        return r.json()

    @default('git_credentials')
[...]

Issue #1117, opened 2020-07-01
User hook for the build endpoint


jtpio:
### Proposed change

This issue is related to the idea mentioned in this Discourse topic: https://discourse.jupyter.org/t/binderhub-with-private-gitlab-and-user-scopes/3502

Looking at the code, it seems like there is (at the moment) no hook or option that could be set to tweak the behavior of the `/build` endpoint, or more generally of the `buidler`.

The idea is to be able to implement fine-grained access control to BinderHub based on the JupyterHub authenticator used to authenticate users. 

The use case is summarized as follows:

- Users authenticate to their BinderHub using their private GitLab instance as the authenticator
- This means that each BinderHub user now corresponds to a GitLab user
- When they enter the repository in the input field, they can only build repositories they have access to
- If they don't have access, the `Could not resolve ref for my-project/repo. Double check your URL.` would ideally be shown
- This would happen before triggering a new build
- [Optional] The UI only shows `GitLab` in the dropdown menu (after configuring `repo_providers`). This looks like it should be solved by https://github.com/jupyterhub/binderhub/pull/1038 :tada: 

### Alternative options

An alternative option might be to add an extra build handler to the main app, and change the frontend to use that endpoint instead.

However this adds a lot of complexity to the BinderHub admin as it would require maintaining custom Docker images and helm charts with these changes.

### Who would use this feature?

Those who want to have a custom BinderHub setup implementing user access based on the user access pattern from the JupyterHub authenticator (GitLab, GitHub). 

### (Optional): Suggest a solution

Provided that an access token was generated according to: https://binderhub.readthedocs.io/en/latest/zero-to-binderhub/setup-binderhub.html#accessing-private-repositories

For a `binderhub` user that has read-only access to all repositories.

And the token set as:

```yaml
config:
  GitLabRepoProvider:
    private_token: "<access token>"
``` 

At the moment it's possible to have control on the *launch* behavior, by providing the following snippet to the helm chart config:

https://github.com/jupyterhub/binderhub/blob/b6446b12b30f741d9e82b7aec1498ede4776cd79/helm-chart/binderhub/values.yaml#L66-L119

However users can still trigger a build to a repository they do not have access to.

It looks like this could be implemented by providing a custom `RepoProvider` (in the helm config value, that could derive from an existing one).

But it would require some user specific information to be passed to the `RepoProvider` to be able to decide whether or not it is possible to resolve the ref for that user, probably somewhere around this line:

[...]

PR #1169, opened 2020-10-22
Allow dynamic repository credentials for authenticated Binderhub instances.

rprimet:
This PR is related to issue #1154 

## Proposed change
Make it possible to fill repo provider / git credentials for a forge (e.g. gitlab instance) dynamically using the currently logged-in user information. This would help set up binderhub instances for use with private repositories for organizations that run a forge (e.g. gitlab instance).

## Scope
This change is meant for authenticated Binderhubs (e.g. an instance of BinderHub that is authenticated against a Gitlab instance and whose main purpose is to build repositories hosted in that instance). It does not address building private repositories from an anonymous Binderhub instance (which would probably entail UI changes and/or a complex authentication workflow?).

In that use case, an auth-aware RepoProvider could assume that auth information is available.

## Issues

- The example RepoProvider (`AuthGitLabProvider`) in issue #1154 makes a blocking http call in the initializer to retrieve auth info -- this should be made asynchronous, but I'm not sure how.

- It should be possible to fill default traitlets config attributes (e.g. `default_git_credentials`) using dynamic credentials, after retrieving those credentials asynchronously. How can that be implemented?

- The RepoProviders should probably made aware of the auth info in order to be able to fetch private repositories, but are RepoProviders the right place to implement authorization / access controls ? If not, would a pre_build_hook as suggested in #1117 help?

## Testing this PR locally

- Install this branch of Binderhub locally using minikube as per the CONTRIBUTING instructions

- Start minikube (e.g. run `minikube start --memory 8192`) and note the minikube IP (run `minikube ip`)

- Register for an account at gitlab.com (user will be called `user1`)

- Create a private project 

- Create an application on gitlab.com ("settings" on your user profile) for authentication purposes (e.g. `bhubtest`). 

    * In the `redirect URL` field, use `http://<minikube ip>:30123/hub/oauth_callback` (note that the minikube IP may change when minikube restarts, usually alternating between two values. If so, put both in the `redirect_uri` field to ease testing). 
    * Give the application the `api` and `read_repository` scopes. 
    * Note the values of Application ID and Application Secret. 
    * **Note**: it would be nice to restrict the scope to `read_api` now that gitlab offers this...

- Install the `requests` package (used by the example repo provider) e.g. `pip install requests`

- change the file `testing/minikube/binderhub_auth_config.py` so that it looks like this

```
import json
import os
import requests
from traitlets import default

from binderhub.repoproviders import GitLabRepoProvider

here = os.path.abspath(os.path.dirname(__file__))
load_subconfig(os.path.join(here, 'binderhub_config.py'))

c.BinderHub.base_url = '/'
c.BinderHub.auth_enabled = True
# configuration for authentication
hub_url = urlparse(c.BinderHub.hub_url)
c.HubOAuth.hub_host = '{}://{}'.format(hub_url.scheme, hub_url.netloc)
c.HubOAuth.api_token = c.BinderHub.hub_api_token
c.HubOAuth.api_url = c.BinderHub.hub_url + '/hub/api/'
c.HubOAuth.base_url = c.BinderHub.base_url
c.HubOAuth.hub_prefix = c.BinderHub.base_url + 'hub/'
c.HubOAuth.oauth_redirect_uri = 'http://127.0.0.1:8585/oauth_callback'
c.HubOAuth.oauth_client_id = '<client ID from the newly-created gitlab application>'

class AuthGitLabProvider(GitLabRepoProvider):
    def __init__(self, *args, handler, **kwargs):
        self.handler = handler
        ud = self.get_user_data(self.handler.get_current_user()['name'])
        self.access_token = ud['auth_state']['access_token']

    def get_user_data(self, username):
        r = requests.get(c.HubOAuth.api_url + f'/users/{username}',
            headers={
             'Authorization': 'token %s' % c.HubOAuth.api_token,
            }
        )
        r.raise_for_status()
        return r.json()

    @default('git_credentials')
    def _default_git_credentials(self):
        if self.access_token:
            return r'username=oauth2\npassword={token}'.format(token=self.access_token)
        return ""

c.BinderHub.repo_providers = {'gl': AuthGitLabProvider}
```

- change the file `testing/minikube/jupyterhub-helm-auth-config.yaml` so that it looks like this

```
cull:
  users: false
hub:
  services:
    binder:
      oauth_no_confirm: true
      oauth_redirect_uri: "http://127.0.0.1:8585/oauth_callback"
      # gitlab application id
      oauth_client_id: <newly-created Gitlab app ID>

custom:
  binderauth_enabled: true

singleuser:
  # to make notebook servers aware of hub
  cmd: jupyterhub-singleuser

auth:
  type: gitlab
  gitlab:
    callbackUrl: http://<minikube IP>:30123/hub/oauth_callback
    # retrieved from the GitLab UI (see above)
    clientId: "<newly-created Gitlab app ID>"
    clientSecret: "<newly-created Gitlab app secret>"
  state:
    enabled: true
    cryptoKey: "<output of `openssl rand -hex 32`>"
```

- Install jupyterhub by running `./testing/minikube/install-hub --auth`

- Run binderhub: `python3 -m binderhub -f testing/minikube/binderhub_auth_config.py`

- Access http://127.0.0.1:8585 using your browser. You will be prompted to login using gitlab.com

- Authorize the application, you should end up at the familiar BinderHub UI (phew!)

- Try building your private repository: in the provider dropdown, select gitlab.com and enter the "namespace" (user1/project) for your private project. Check that it builds and launches

- Register a new gitlab user (`user2`), create a new private project for user2. 

- Check that user2 can build its own private project, but not user1's

(Note that whenever the minikube IP changes due to a restart, the file `testing/minikube/jupyterhub-helm-auth-config.yaml` should be updated accordingly and `./testing/minikube/install-hub --auth` should be run again)

### DIFF
--- a/binderhub/base.py
+++ b/binderhub/base.py
@@ -2,10 +2,13 @@
 
 import json
 from ipaddress import ip_address
+import os
 
 from http.client import responses
 from tornado import web
 from tornado.log import app_log
+from tornado.httpclient import AsyncHTTPClient, HTTPRequest, HTTPError
+
 from jupyterhub.services.auth import HubOAuthenticated, HubOAuth
 
 from . import __version__ as binder_version
@@ -19,6 +22,7 @@ class BaseHandler(HubOAuthenticated, web.RequestHandler):
         super().initialize()
         if self.settings['auth_enabled']:
             self.hub_auth = HubOAuth.instance(config=self.settings['traitlets_config'])
+            self.current_user_model = None
 
     def prepare(self):
         super().prepare()
@@ -71,14 +75,46 @@ class BaseHandler(HubOAuthenticated, web.RequestHandler):
         spec = self.request.path[idx + len(prefix) + 1:]
         return spec
 
-    def get_provider(self, provider_prefix, spec):
+    async def get_provider(self, provider_prefix, spec):
         """Construct a provider object"""
         providers = self.settings['repo_providers']
         if provider_prefix not in providers:
             raise web.HTTPError(404, "No provider found for prefix %s" % provider_prefix)
 
+        async def api_request(url, *args, **kwargs):
+            headers = kwargs.setdefault('headers', {})
+            headers.update({'Authorization': 'token %s' % self.hub_auth.api_token})
+            hub_api_url = os.getenv('JUPYTERHUB_API_URL', '') or self.hub_auth.api_url
+            request_url = hub_api_url + url
+            req = HTTPRequest(request_url, *args, **kwargs)
+
+            try:
+                return await AsyncHTTPClient().fetch(req)
+            except HTTPError as e:
+                app_log.error("Error accessing Hub API (using %s): %s", request_url, e)
+
+        async def get_current_user_model():
+            """Get the current user model.
+            The user auth_state is only accessible to admin users.
+            """
+            if not self.settings['auth_enabled']:
+                return None
+            
+            if self.current_user_model is None:
+                username = self.get_current_user()['name']
+                resp = await api_request(
+                    f'/users/{username}',
+                    method='GET',
+                )
+                self.current_user_model = json.loads(resp.body.decode('utf-8'))
+
+            return self.current_user_model
+
         return providers[provider_prefix](
-            config=self.settings['traitlets_config'], spec=spec)
+                    config=self.settings['traitlets_config'], 
+                    spec=spec,
+                    user_model=await get_current_user_model()
+                )
 
     def get_badge_base_url(self):
         badge_base_url = self.settings['badge_base_url']
--- a/binderhub/base.py
+++ b/binderhub/base.py
@@ -149,6 +149,9 @@ class VersionHandler(BaseHandler):
 
     # demote logging of 200 responses to debug-level
     log_success_debug = True
+    # allow version-check requests from banned hosts
+    # (e.g. mybinder.org federation when blocking cloud datacenters)
+    skip_check_request_ip = True
 
     async def get(self):
         self.set_header("Content-type", "application/json")
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -229,7 +229,7 @@ class BuildHandler(BaseHandler):
 
         # get a provider object that encapsulates the provider and the spec
         try:
-            provider = self.get_provider(provider_prefix, spec=spec)
+            provider = await self.get_provider(provider_prefix, spec=spec)
         except Exception as e:
             app_log.exception("Failed to get provider for %s", key)
             await self.fail(str(e))
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -2,6 +2,7 @@
 Handlers for working with version control services (i.e. GitHub) for builds.
 """
 
+import asyncio
 import hashlib
 from http.client import responses
 import json
@@ -10,7 +11,6 @@ import time
 import escapism
 
 import docker
-from tornado.concurrent import chain_future, Future
 from tornado import gen
 from tornado.httpclient import HTTPClientError
 from tornado.web import Finish, authenticated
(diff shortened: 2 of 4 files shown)


minrk:
Thanks for opening this! I'm not sure passing the handler to the repoprovider is the right implementation, but having a concrete use case is very helpful. Some thoughts:

- when deploying binderhub in an authenticated context, the Handler object should have the authenticated user model available as `self.current_user`. This should already have the info fetched by `get_user_data(self, username)`. If it doesn't, that's something I think we can fix in HubAuth on the jupyterhub side without needing to re-identify the user a second time
- if we are talking about standardizing on the auth info provided by HubAuth, we can pass `user=self.current_user` to RepoProviders instead of the Handler itself.
- I believe passing the user model to the repo provider solves the need to make the request async in this particular case, but if we need a new async API to load info, I think we can add it.

rprimet:
Hi @minrk, 

Thanks for your reply!

Regarding the Handler's `current`, at least in the configuration described above, it does not seem to contain all info fetched by `get_user_data` (crucially, the `auth_state` attribute seems to be missing).

Regarding the other points, yes, at least for this use case limiting ourselves to passing the user info to the RepoProvider would be enough.

meeseeksmachine:

https://discourse.jupyter.org/t/a-question-about-binderhub-authentication-and-privacy/7758/2

adriendelsalle:
Hi!

That would be definitely a great feature !

> when deploying binderhub in an authenticated context, the Handler object should have the authenticated user model available as self.current_user. This should already have the info fetched by get_user_data(self, username). If it doesn't, that's something I think we can fix in HubAuth on the jupyterhub side without needing to re-identify the user a second time

`self.current_user` is only exposing basic information (not incl. `auth_state`) about the user because it makes use of `get_current_user_token` or `get_current_user_cookie`, right ?
Since it requires the requester to be a hub admin to get the user `auth_state`, we can't get this info from token or cookie.

I would propose to fetch the user model once and only when `auth_enabled` is `True`, and pass a `user_model` kwarg to repo providers (defaulting to `None` when `auth_enabled` is `False`). This proposition is [here](https://github.com/adriendelsalle/binderhub/commit/44ef87f1e87b5aaa78732a7b981fa1868769f8b8)!
Maybe we could/should define a trait in the base `binderhub.RepoProvider` class.

Then, you could just write your configuration as follow:
``` python
from binderhub.repoproviders import GitLabRepoProvider
from traitlets import default

class AuthGitLabProvider(GitLabRepoProvider):
    def __init__(self, *args, user_model, **kwargs):
        self.access_token = user_model['auth_state']['access_token']

    @default('git_credentials')
    def _default_git_credentials(self):
        if self.access_token:
            return r'username=oauth2\npassword={token}'.format(token=self.access_token)
        return ""

c.BinderHub.repo_providers = {'gl': AuthGitLabProvider}
```
@rprimet @minrk Do you think it's a good way to do ?

rprimet:
@adriendelsalle that looks good to me!

adriendelsalle:
I would be happy to open a PR on your fork if you add me as a collaborator on it!

rprimet:
OK I just did, but I'm not sure if it would be valuable (what's the workflow?), maybe opening a PR on the main repo referencing this one would be better? Sorry for the naive question :-)

meeseeksmachine:

https://discourse.jupyter.org/t/private-gitlab-access-for-binderhub/504/9

g-braeunlich:
Any chances, that we can get this merged?

g-braeunlich:
I tested it. Worked perfectly with the update of @adriendelsalle !

rprimet:
Seems that I messed something up while trying to sync with the current master. Yet https://github.com/jupyterhub/binderhub/compare/master...rprimet:dynamic_credentials_for_repoproviders still seems to show the correct diff so I'm puzzled as to why...

adriendelsalle:
> Seems that I messed something up while trying to sync with the current master. Yet [master...rprimet:dynamic_credentials_for_repoproviders](https://github.com/jupyterhub/binderhub/compare/master...rprimet:dynamic_credentials_for_repoproviders) still seems to show the correct diff so I'm puzzled as to why...

Did you rebased or merged ? only rebase :).
This should be fixed now, I pushed on your fork.

rprimet:
@adriendelsalle great thanks!

rprimet:
Is there anything else (e.g. tests) to do on this PR?

nreith:
When will this be merged and available? I have a use case where users would want to share only with other people authorized on their repo.

larsbonczek:
Hey @minrk, could you please take a look at this again? :)

PR #1170, opened 2020-10-23
Update BinderHub error message for GitHub main/master branches

sgibson91:
This PR updates the error message return when a commit ref is not resolved. When the repo provider is GitHub and the resolved branch is either `main` or `master`, we ask the user if that's the branch the meant indicating a change to the GitHub platform.

~~I tried to test this locally using the minikube instructions [here](https://github.com/jupyterhub/binderhub/blob/master/CONTRIBUTING.md#one-time-installation) but ran into a "connection refused" error at server launch time.~~

~~`[E 201023 10:54:22 builder:556] Retrying launch of https://github.com/binder-examples/requirements after error (duration=0s, attempt=3): ConnectionRefusedError(61, 'Connection refused')`~~

Ok, I got the minikube deployment to work. (I think my problem was that I edited code while waiting for something to build. `pip install -e .` means the package gets auto-updated and minikube doesn't like that.)

Screenshot 2020-10-23 at 13 23 46

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -254,16 +254,12 @@ class BuildHandler(BaseHandler):
         if ref is None:
             error_message = ["Could not resolve ref for %s. Double check your URL." % key]
 
-            split_key = key.split("/")
-            repo_provider = split_key[0].split(":")[0]
-            resolved_branch = split_key[-1]
-
-            if repo_provider == "gh":
+            if provider.name == "GitHub":
                 error_message.append('GitHub recently changed default branches from "master" to "main".')
 
-                if resolved_branch == "master":
+                if provider.unresolved_ref == "master":
                     error_message.append('Did you mean the "main" branch?')
-                elif resolved_branch == "main":
+                elif provider.unresolved_ref == "main":
                     error_message.append('Did you mean the "master" branch?')
 
             else:
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -250,8 +250,26 @@ class BuildHandler(BaseHandler):
         except Exception as e:
             await self.fail("Error resolving ref for %s: %s" % (key, e))
             return
+
         if ref is None:
-            await self.fail("Could not resolve ref for %s. Double check your URL." % key)
+            error_message = ["Could not resolve ref for %s. Double check your URL." % key]
+
+            split_key = key.split("/")
+            repo_provider = split_key[0].split(":")[0]
+            resolved_branch = split_key[-1]
+
+            if repo_provider == "gh":
+                error_message.append('GitHub recently changed default branches from "master" to "main".')
+
+                if resolved_branch == "master":
+                    error_message.append('Did you mean the "main" branch?')
+                elif resolved_branch == "main":
+                    error_message.append('Did you mean the "master" branch?')
+
+            else:
+                error_message.append("Is your repo public?")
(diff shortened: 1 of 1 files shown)


binderhub/builder.py
```diff
@@ -250,8 +250,26 @@ async def get(self, provider_prefix, _unescaped_spec):
...
             await self.fail("Error resolving ref for %s: %s" % (key, e))
             return
+
         if ref is None:
-            await self.fail("Could not resolve ref for %s. Double check your URL." % key)
+            error_message = ["Could not resolve ref for %s. Double check your URL." % key]
+
+            split_key = key.split("/")
+            repo_provider = split_key[0].split(":")[0]
+            resolved_branch = split_key[-1]
```
minrk, line 259: unresolved_ref = key.rsplit("/", 1)[-1]

binderhub/builder.py
```diff
@@ -250,8 +250,26 @@ async def get(self, provider_prefix, _unescaped_spec):
...
+
+            if repo_provider == "gh":
```
minrk, line 261: "gh" is already available as `provider_prefix` from the function. Could also use `isinstance(repo_provider, GitHubRepoProvider)` or `provider.name == "GitHub"`

binderhub/builder.py
```diff
@@ -250,8 +250,26 @@ async def get(self, provider_prefix, _unescaped_spec):
...
+                error_message.append('GitHub recently changed default branches from "master" to "main".')
+
+                if resolved_branch == "master":
```
minrk, line 264: resolved_branch should be available as `provider.unresolved_ref` so we shouldn't need to parse out the request string

minrk (commented):
Great! I think we don't have to re-parse the string and can use provider attribute access instead, but otherwise 👍

sgibson91:
Thanks @minrk! I pushed your suggestions :)

minrk:
Great, thanks!

Now the question is when to change the default branch name in the form

sgibson91:
That was going to be the next thing I worked on today after sorting the CI for staging. I think it's a common enough problem now that switching will be fine. Can you point me to the right section of code please?

minrk:
Yes, it's mostly in `updateRepoText()` where we set the placeholder and label values, and then in `getBuildFormValues()` where we pick a default value if ref is empty. I think it makes sense to store the default value in the `placeholder` attr instead of what we do now, which is duplicate the label text in the placeholder. Then to get the default we can do `$("#ref").val().trim() || $("#ref").attr("placeholder")` instead of `|| 'master'`.

These are in `static/index.js` Make sure to run `npm run webpack:watch` while working on this code

choldgraf:
I'm +1 on switching the default to `main` for github, now that it is the default repository. For any repositories that *already* have a binder link, the `master` branch is already coded in their link. So I think the question is "of the population of repositories creating their *first* Binder link, is it more disruptive to use `master` vs. `main` right now?" IMO we could just switch now since the balance will continue to tip more and more towards `main` over time anyway.

manics:
Can we answer that question from the mybinder stats, e.g. using https://binderlytics.herokuapp.com/binder-launches/binder ?

sgibson91:
#1186 will actually make our default ref `HEAD` which will resolve to the correct default branch for us 🎉

PR #1172, opened 2020-10-28
Update UI and launch main branch by default for GitHub repos

### LINKED ISSUE #1168: New repos on Github no longer work by default with MyBinder
New repos created on Github seem to be given the default branchname `main` rather than `master` ([New GitHub Repositories Default to Main Branch](https://www.infoq.com/news/2020/10/github-main-branch/)).

Pasting the repo URL into MyBinder, the build shows a failed build with message: ` "Could not resolve ref for gh:ORG/REPO/master. Double check your URL."`

Relates to: https://github.com/jupyterhub/binderhub/issues/1145

sgibson91:
This PR updates the placeholder for the "Git branch, commit or tag" field to be `main` and also defaults to launching the `main` branch for GitHub repos.

fixes #1168 

Screenshot of how the UI now looks:

Screenshot 2020-10-28 at 15 38 37

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -101,7 +101,7 @@ function updateRepoText() {
   }
   $("#repository").attr('placeholder', text);
   $("label[for=repository]").text(text);
-  $("#ref").attr('placeholder', tag_text);
+  $("#ref").attr('placeholder', 'main');
   $("label[for=ref]").text(tag_text);
 }
 
@@ -121,7 +121,7 @@ function getBuildFormValues() {
     repo = encodeURIComponent(repo);
   }
 
-  var ref = $('#ref').val().trim() || 'main';
+  var ref = $('#ref').val().trim() || $("#ref").attr("placeholder");
   if (providerPrefix === 'zenodo' || providerPrefix === 'figshare' || providerPrefix === 'dataverse' ||
       providerPrefix === 'hydroshare') {
     ref = "";
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -63,6 +63,10 @@ function updateRepoText() {
   var text;
   var provider = $("#provider_prefix").val();
   var tag_text = "Git branch, tag, or commit";
+  var placeholder = "master";
+  if (provider === "gh") {
+    placeholder = "main";
+  }
   // first enable branch/ref field, some providers later disable it
   $("#ref").prop("disabled", false);
   $("label[for=ref]").prop("disabled", false);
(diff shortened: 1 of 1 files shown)


binderhub/static/js/index.js
```diff
@@ -101,7 +101,7 @@ function updateRepoText() {
   }
   $("#repository").attr('placeholder', text);
   $("label[for=repository]").text(text);
-  $("#ref").attr('placeholder', tag_text);
+  $("#ref").attr('placeholder', 'main');
```
minrk, line 104: I think we may want this to be conditional on the provider. `main` is the default on github, but `master` is still the default (and not changeable, I believe) on gists. GitLab also still defaults to `master`, as does git itself.

So I think at the very least, we want `main` for github and `master` for gist. Then the question is: do we 'push' folks elsewhere to use 'main' and use that for gitlab and/or git? I'm okay either way, and would lean slightly toward the push choice given the reason for the change, and only use 'master' by default for gist where it is hardcoded and cannot be chosen by users.

binderhub/static/js/index.js
```diff
@@ -121,7 +121,7 @@ function getBuildFormValues() {
     repo = encodeURIComponent(repo);
   }
 
-  var ref = $('#ref').val().trim() || 'master';
+  var ref = $('#ref').val().trim() || $("#ref").attr("placeholder");
```
minrk, line 124: 👍

binderhub/static/js/index.js
sgibson91: Something like bed2e0a?

minrk: Not quite, since `tag_text` isn't "master", it's the description (a duplicate of the label). I think we need a separate variable for `placeholder` that's `main` for github, `master` for gist, and either main or master the rest (`master` would preserve existing behavior, `main` would apply the new github convention to all providers by default).

For example:

```javascript
var placeholder = "main";
if (provider == "gist") {
    placeholder = "master";
}
$("#ref").attr('placeholder', placeholder);
```

sgibson91: I went for the preserving existing behaviour model as I'm unsure where other communities are on this c1b7dc6

sgibson91: I think it'll be "easy" now to switch to the push choice you suggested, but maybe we prepare a blog post to release when we do it? I don't like the idea of breaking behaviour without some kind of "announcement" informing the users.

minrk (approved):

binderhub/static/js/index.js
sgibson91: I opened this issue to discuss this further https://github.com/jupyterhub/binderhub/issues/1175

minrk:
Great!

PR #1174, opened 2020-10-29
Make helm template formatting follow best practices

consideRatio:
This PR is free from any changes to the logic or assumptions on the configuration, it only only adjusts formatting.

### DIFF
--- a/helm-chart/binderhub/templates/configmap.yaml
+++ b/helm-chart/binderhub/templates/configmap.yaml
@@ -3,9 +3,11 @@ apiVersion: v1
 metadata:
   name: binder-config
 data:
-  # FIXME: Stop splitting up parts into configmap/secret as it adds complexity,
-  #        instead only use a secret. For the equivalent PR in z2jh, see:
-  #        https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1682
+  {{- /*
+  FIXME: Stop splitting up parts into configmap/secret as it adds complexity,
+         instead only use a secret. For the equivalent PR in z2jh, see:
+         https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1682
+  */}}
   {{- /* Important: only pick relevant subset of values. Including too much here can trigger unnecessary binderhub pod restarts */}}
   {{- $values := pick .Values "config" "cors" "dind" "extraConfig" }}
   {{- /* make a copy to avoid modifying the original */ -}}
--- a/helm-chart/binderhub/templates/configmap.yaml
+++ b/helm-chart/binderhub/templates/configmap.yaml
@@ -3,19 +3,22 @@ apiVersion: v1
 metadata:
   name: binder-config
 data:
-{{- /* Important: only pick relevant subset of values. Including too much here can trigger unnecessary binderhub pod restarts */}}
-{{- $values := pick .Values "config" "cors" "dind" "extraConfig" }}
-{{- /* make a copy to avoid modifying the original */ -}}
-{{- $_ := set $values "config" (merge dict .Values.config) }}
-{{- $_ := set $values.config "BinderHub" (merge dict .Values.config.BinderHub) }}
-{{- /* trim secret values. Update here if new secrets are added! */ -}}
-{{- /* every 'omit' here should be matched with a corresponding 'pick' in secret.yaml */ -}}
-{{- if $values.config.GitHubRepoProvider }}
-{{- $_ := set $values.config "GitHubRepoProvider" (omit .Values.config.GitHubRepoProvider "client_id" "client_secret" "access_token") }}
-{{- end }}
-{{- if $values.config.GitLabRepoProvider }}
-{{- $_ := set $values.config "GitLabRepoProvider" (omit .Values.config.GitLabRepoProvider "private_token" "access_token") }}
-{{- end }}
+  # FIXME: Stop splitting up parts into configmap/secret as it adds complexity,
+  #        instead only use a secret. For the equivalent PR in z2jh, see:
+  #        https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1682
+  {{- /* Important: only pick relevant subset of values. Including too much here can trigger unnecessary binderhub pod restarts */}}
+  {{- $values := pick .Values "config" "cors" "dind" "extraConfig" }}
+  {{- /* make a copy to avoid modifying the original */ -}}
+  {{- $_ := set $values "config" (merge dict .Values.config) }}
+  {{- $_ := set $values.config "BinderHub" (merge dict .Values.config.BinderHub) }}
+  {{- /* trim secret values. Update here if new secrets are added! */ -}}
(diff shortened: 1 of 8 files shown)


minrk:
I ran `helm template ./binderhub -f minikube-binder.yaml` and compared before/after this PR. to double-check that the results are as-intended:

```diff
--- before.yaml	2020-10-29 13:45:19.000000000 +0100
+++ after.yaml	2020-10-29 13:45:32.000000000 +0100
@@ -1104,6 +1104,9 @@
 metadata:
   name: binder-config
 data:
+  # FIXME: Stop splitting up parts into configmap/secret as it adds complexity,
+  #        instead only use a secret. For the equivalent PR in z2jh, see:
+  #        https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1682
   values.yaml: |
     config:
       BinderHub:
@@ -1483,7 +1486,6 @@
   labels: {}
 spec:
   type: NodePort
-  
   selector:
     app: binder
     name: binder
@@ -1548,7 +1550,7 @@
         - name: IMAGE_GC_DELAY
           value: "5"
         - name: IMAGE_GC_THRESHOLD_TYPE
-          value: relative
+          value: "relative"
         - name: IMAGE_GC_THRESHOLD_HIGH
           value: "80"
         - name: IMAGE_GC_THRESHOLD_LOW
@@ -1852,13 +1854,11 @@
         component: binder
         release: RELEASE-NAME
         heritage: Helm
-        
       annotations:
         # This lets us autorestart when the configmap changes!
-        checksum/config-map: ae0a43fd9e9d31da75751fbe6db89e48ba413df226cce43f853e55258ff9533d
+        checksum/config-map: 1c586d1a2901830eaccb36194aa1451e3eaba06538202254afb0e2d00e0b1141
         checksum/secret: 958722d545be64139fadb4d84570157e6d4d430874e91e4a663c8dc880fbf84e
     spec:
-      
       nodeSelector: {}
       serviceAccountName: binderhub
       volumes:
@@ -1871,7 +1871,6 @@
       - name: docker-socket
         hostPath:
           path: /var/run/docker.sock
-      
       containers:
       - name: binder
         image: jupyterhub/k8s-binderhub:0.2.0-n246.h8c5b0c9
@@ -1885,11 +1884,10 @@
             name: secret-config
           - mountPath: /var/run/docker.sock
             name: docker-socket
-          
         resources:
-            requests:
-              cpu: 0.2
-              memory: 100Mi
+          requests:
+            cpu: 0.2
+            memory: 100Mi
         imagePullPolicy: IfNotPresent
         env:
         - name: BUILD_NAMESPACE
@@ -1901,7 +1899,6 @@
             secretKeyRef:
               name: binder-secret
               key: "binder.hub-token"
-        
         ports:
           - containerPort: 8585
             name: binder
```

so everything looks right!

helm-chart/binderhub/templates/configmap.yaml
```diff
@@ -3,19 +3,22 @@ apiVersion: v1
...
-{{- /* every 'omit' here should be matched with a corresponding 'pick' in secret.yaml */ -}}
-{{- if $values.config.GitHubRepoProvider }}
-{{- $_ := set $values.config "GitHubRepoProvider" (omit .Values.config.GitHubRepoProvider "client_id" "client_secret" "access_token") }}
-{{- end }}
-{{- if $values.config.GitLabRepoProvider }}
-{{- $_ := set $values.config "GitLabRepoProvider" (omit .Values.config.GitLabRepoProvider "private_token" "access_token") }}
-{{- end }}
+  # FIXME: Stop splitting up parts into configmap/secret as it adds complexity,
+  #        instead only use a secret. For the equivalent PR in z2jh, see:
+  #        https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1682
```
minrk, line 8: Did you mean for this comment to be included in the actual configmap, or as a template comment?

minrk (approved):

helm-chart/binderhub/templates/configmap.yaml
consideRatio: I didn't mind if it was included in the configmap, but as the checksum is changed by doing this, I figure should make it a helm template to avoid triggering a restart of the binder-pod for no reason.

consideRatio:
Thank you for the review @minrk! Good review idea to provide a diff like that!

betatim (approved):

helm-chart/binderhub/templates/configmap.yaml
```diff
@@ -3,19 +3,24 @@ apiVersion: v1
...
-{{- $_ := set $values.config "GitHubRepoProvider" (omit .Values.config.GitHubRepoProvider "client_id" "client_secret" "access_token") }}
-{{- end }}
-{{- if $values.config.GitLabRepoProvider }}
-{{- $_ := set $values.config "GitLabRepoProvider" (omit .Values.config.GitLabRepoProvider "private_token" "access_token") }}
-{{- end }}
+  {{- /*
+  FIXME: Stop splitting up parts into configmap/secret as it adds complexity,
+         instead only use a secret. For the equivalent PR in z2jh, see:
+         https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1682
+  */}}
```
betatim, line 10: Why add comments like this into the code instead of making an issue? I think we should try and keep issues as "the one place to go" to find out about things that could be done. Even if only because it is much easier to clean up and close stale issues than clean up and remove stale comments like this.

consideRatio: While browsing through the files I realized this, and knew that I could create an issue, but I also felt like advertising this within this context so that everyone working in this file get a notice and encouragement to fix this.

consideRatio: Created https://github.com/jupyterhub/binderhub/issues/1176 to represent this as well.

betatim:
The mybinder.org `kubeval` check now fails https://github.com/jupyterhub/mybinder.org-deploy/pull/1679/checks?check_run_id=1327653768 any idea why/which change here lead to that?

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -102,25 +102,25 @@ spec:
...
         {{- if .Values.jupyterhub.hub.allowNamedServers }}
         - name: JUPYTERHUB_ALLOW_NAMED_SERVERS
           value: "true"
         - name: JUPYTERHUB_NAMED_SERVER_LIMIT_PER_USER
-          value: {{ .Values.jupyterhub.hub.namedServerLimitPerUser | quote  }}
+          value: {{ .Values.jupyterhub.hub.namedServerLimitPerUser | quote }}
+        {{- end }}
         {{- end }}
+        {{- with .Values.extraEnv }}
+        {{- . | toYaml | indent 8 }}
```
consideRatio, line 116: This is the error. It must be nindent

consideRatio: Resolved by https://github.com/jupyterhub/binderhub/pull/1178

PR #1177, opened 2020-10-29
Attempt fix of RTD build failures + modernize RTD config

### LINKED ISSUE #1173: Read the docs HTML build failure
The HTML build is failing. I want to focus right now on bumping README badges for travis.org -> .com migration but the thing to do I think is:

1. Start using a readthedocs configuration like in z2jh, of v2 kind, name it .readthedocs.yml for aesthetics while we are at it.
2. Make sphinx be >=2 in dependencies I think, depends on stuff installed. See doc-requirements.txt of z2jh for notes relevant for understanding some parts about doc-requirements that are used on RTD.

consideRatio:
This PR organized dependencies in a way that makes them less prone to issues, and stops making RTD build and publish PDF/htmlzip alongside its standard HTML output. It is too hard to maintain all of these, so lets stick with only HTML builds.

Closes #1173


choldgraf:
this LGTM

as an aside, I find it really annoying that RTD uses Sphinx 1.x by default

doc/doc-requirements.txt
```diff
@@ -1,15 +1,25 @@
...
 pandas
 ruamel.yaml
-pydata_sphinx_theme
-sphinx>=1.7
+
 # install BinderHub dependencies. We manually list them here because some
 # dependencies (like pycurl) can't be installed on ReadTheDocs and aren't
 # needed to build the docs.
-kubernetes==9.0.*
+kubernetes
```
consideRatio, line 22: I unpinned this, which is currently at 0.12.0 I think, and the latest KubeSpawner part of z2jh 0.10.0 works fine with 0.12.0.

.readthedocs.yml
```diff
@@ -0,0 +1,24 @@
...
+
+# Required (RTD configuration version)
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: doc/conf.py
+
+# Optionally build your docs in additional formats such as PDF and ePub
+formats: []
```
consideRatio, line 12: This is a key part of the PR, to exclude HTMLZIP by default for example which causes failures.

PR #1179, opened 2020-10-29
Reduce complexity by merging configmap/secret into secret

### LINKED ISSUE #1176: Put all config in a k8s Secret, don't split it up into ConfigMap/Secret
We currently have a configmap.yaml and secret.yaml, and split up information from .Values into them based on if the content is sensitive or not. This is a cumbersome practice that adds complexity which is prone to break - this is based on experience of it breaking :D In z2jh we decided to merge them.

I added a note in the configmap saying....

```
  FIXME: Stop splitting up parts into configmap/secret as it adds complexity,
         instead only use a secret. For the equivalent PR in z2jh, see:
         https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1682
```

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1682 can provide an example on how we went about it in z2jh, and I suggest that we mimic this here.

consideRatio:
This PR is meant to close #1176.

The value of doing this is simply to reduce the complexity of things. As one can see from the deleted code in the Helm templates, it is quite messy code that is hard to parse and generally hard to maintain. I really hope to make this Helm chart easier to maintain by reducing complexity wherever we can.

## Review notes
This is ready for merge in my mind, but it would be good to eyeball it on mybinder.org's staging environment as well.


helm-chart/binderhub/templates/secret.yaml
```diff
@@ -3,18 +3,17 @@ apiVersion: v1
...
+  {{- /* Make it easy to mount certain parts without traversing YAML */}}
+  binder.hub-token: {{ .Values.jupyterhub.hub.services.binder.apiToken | quote }}
+
+  {{- /* Stash away all Helm template values */}}
+  values.yaml: |
+    {{- .Values | toYaml | nindent 4 }}
+
+  {{- /* Glob files to allow them to be mounted by the binderhub pod */}}
+  {{- /* key=filename: value=content */}}
+  {{- (.Files.Glob "files/*").AsConfig | nindent 2 }}
```
consideRatio, line 16: data -> stringData to not need to base64 encode everything, which makes everything easier to read etc.

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -70,8 +65,7 @@ spec:
         volumeMounts:
           - mountPath: /etc/binderhub/config/
             name: config
-          - mountPath: /etc/binderhub/secret/
-            name: secret-config
+            readOnly: true
```
consideRatio, line 68: Just mount one file now into /etc/binderhub/config. I made it readOnly as well because why not, it improves performance a bit for k8s.

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -44,9 +42,6 @@ spec:
       {{- end }}
       volumes:
       - name: config
-        configMap:
-          name: binder-config
-      - name: secret-config
```
consideRatio, line 49: I removed one volume definition related to the configmap, but let the volume be named config.

helm-chart/binderhub/files/binderhub_config.py
```diff
@@ -1,46 +1,24 @@
...
-            with open(path) as f:
-                values = yaml.safe_load(f)
-            cfg = _merge_dictionaries(cfg, values)
-        else:
-            print(f"No config at {path}")
-    return cfg
+    path = "/etc/binderhub/config/values.yaml"
+    print(f"Loading {path}")
+    with open(path) as f:
+        return yaml.safe_load(f)
```
consideRatio, line 21: I removed logic about merging the different YAML files, now we just have one.

helm-chart/binderhub/files/binderhub_config.py
```diff
@@ -1,46 +1,24 @@
-from collections.abc import Mapping
 import os
+
+from collections.abc import Mapping
 from functools import lru_cache
 from urllib.parse import urlparse
+
 import yaml
```
consideRatio, line 7: I just ordered imports like I recall having seen some linter suggest. Hmm... Dunno... What would `black` do here?

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -30,9 +30,7 @@ spec:
         {{- end }}
         {{- end }}
       annotations:
-        # This lets us autorestart when the configmap changes!
```
betatim, line 33: Can we keep this comment?

betatim:
Are there any steps people need to take to transition from the CM+Secret setup to the Secret only setup? I think it is all automated.

helm-chart/binderhub/templates/secret.yaml
betatim: Do you know why it used to be b64 encoded?

consideRatio: [ConfigMap resources](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#configmap-v1-core) have `data` and `binaryData`, while [Secret resources](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#secret-v1-core) has `data` and `stringData`.

- stringData expects plain strings and will write out plain strings when mounted.
- binaryData expects base64 encoded strings and will base64 decode those strings before writing out the data to a file on mount

The `data` field on ConfigMap defaults to be a stringData kind of field, while the `data` field on a Secret defaults to be a binaryData kind of field. I'm not sure why they have different defaults though. Base64 doesn't add security in any way though.

consideRatio:
> Are there any steps people need to take to transition from the CM+Secret setup to the Secret only setup? I think it is all automated.

It is all automated since Helm is always creating/updating these resources. It could make sense to have a k8s upgrade step like we have in z2jh to verify upgrades between changes work btw. I created https://github.com/jupyterhub/binderhub/issues/1195 to represent that wish.

consideRatio:
The test failure can be ignored, a rebase didn't make it run with updated tests apparently, but the error is fixed in #1201.

helm-chart/binderhub/templates/secret.yaml
```diff
@@ -3,18 +3,17 @@ apiVersion: v1
...
-  {{- $_ := set $values.config "GitLabRepoProvider" (pick $cfg.GitLabRepoProvider "access_token" "private_token") }}
-  {{- end }}
-  values.yaml: {{ $values | toYaml | b64enc | quote }}
+stringData:
+  {{- /* Make it easy to mount certain parts without traversing YAML */}}
+  binder.hub-token: {{ .Values.jupyterhub.hub.services.binder.apiToken | quote }}
+
+  {{- /* Stash away all Helm template values */}}
+  values.yaml: |
+    {{- .Values | toYaml | nindent 4 }}
```
minrk, line 12: We've lost the logic [here](https://github.com/jupyterhub/binderhub/pull/1179/files#diff-1b900e3e30b9fbc60551b87d5f21bfbc14277cae86ba9441f391b2898a220958L11) which I think is important: not including quite all of values.yaml, since restarting the binder pods on every possible change to values is unnecessarily disruptive.

It can still be simpler since we don't have to do secret splitting, and it's okay to cast a slightly wider-than-strictly-necessary net in the name of simplicity, but I think restarting binder pods on every helm upgrade is too far, which this would do.

If we get that behavior back, then 👍 this is ready to go.

consideRatio: Nice catch, thanks for the review @minrk! I updated this in dc3df7f

consideRatio:
Hi @minrk @betatim! Just a friendly reminder of this maintenance PR.

consideRatio:
@minrk does this LGTY?

sgibson91 (approved):
I think this is good to merge once the conflicts have been resolved?

manics:
Ping @consideRatio! @sgibson91 has approved this subject to conflicts.

consideRatio:
Rebased, I think this is ready for review again!

helm-chart/binderhub/templates/deployment.yaml
manics: Looks like this comment is still missing.

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -27,9 +27,7 @@ spec:
         {{- . | toYaml | nindent 8 }}
         {{- end }}
       annotations:
-        # This lets us autorestart when the configmap changes!
-        checksum/config-map: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
-        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+        checksum/config: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
```
consideRatio, line 30:
```suggestion
        # This lets us autorestart when the secret's pass-through config changes
        checksum/config: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
```

PR #1180, opened 2020-10-29
Dependency freezer script, bumped image deps, and doc-requirements.txt cleanup

consideRatio:
## Script to (re)freeze the image environment

@minrk built a script that I often use like `./dependency freeze --upgrade` to freeze a requirements.in file into a requirements.txt file. It has worked great in Z2JH, with new contributors understanding how to use it right away. I copied that script here.

The Python dependencies relevant to the binderhub Docker image seem to be the requirements of the BinderHub package, and the `google-cloud-logging` package. Due to this I created `requirements.in` with the `google-cloud-logging` package and use the BinderHub's requirements.txt (referenced as binderhub.in) in the root folder of this project to generate /helm-chart/images/binderhub/requirements.txt using an adjusted copy of @minrk's excellent script.

## BinderHub image

- Python 3.7 --> 3.8 in the binderhub image
- Node 12 --> 14 to build the binderhub package HTML/JS resources in the binderhub image build step
- `kubernetes` (Python library) was bumped 0.9.0 --> 0.12.0

## doc-requirements.txt documented and cleaned up

I noted that there was some issues to run `make html`, and that there were some parts that were no longer relevant in doc-requirements.txt, so I trimmed them.

PR #1184, opened 2020-10-30
Bump Z2JH to 0.10.6, and let binderhub's image match JupyterHub and Kubernetes client version (12.0.1)

consideRatio:
## z2jh breaking changes - [CHANGELOG](http://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/master/CHANGELOG.md)

- __Network policies are now created by default__:
  If a k8s cluster has a network policy enforcer, such as Calico, and have custom communication to z2jh pods, this may be an issue. Currently we run tests in z2jh and this repo with those enabled though and things seem to work.
- __Other breaking changes fail fast__:
  All other breaking changes will happen during Helm template rendering and fail with a nice error message before any change to k8s is made.

## mybinder.org-deploy's upgrade after this PR is merged

1. __NetworkPolicies__:
   Network policies on mybinder.org's deployment have not been used before but will be that now if the k8s cluster has a network policy controller (Calico or various others) to enforce them. Any pod communicating with the Hub or user servers without going through the autohttps or proxy pod will be blocked unless they have the relevant access label set on them.

   __Suggestion__: For mybinder.org-deploy, disable network policies explicitly, upgrade to latest jupyterhub, and then enable them as a separate step.

2. __OVH and `singleuser.imagePullSecret` configuration__:
   The OVH cluster use this configuration which has now been replaced with a chart wide configuration that can be set in `imagePullSecret` (root level z2jh configuration) instead, and if that is done, will be referenced on all JupyterHub helm chart pods.

   __Suggestion__: For mybinder.org-deploy and the OVH configuration, make `singleuser.imagePullSecret` go to `imagePullSecret` instead.

### DIFF
--- a/helm-chart/binderhub/requirements.yaml
+++ b/helm-chart/binderhub/requirements.yaml
@@ -1,4 +1,5 @@
 dependencies:
-- name: jupyterhub
-  version: "0.10.2"
-  repository: "https://jupyterhub.github.io/helm-chart"
+  # https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tags
+  - name: jupyterhub
+    version: "0.10.6"
+    repository: "https://jupyterhub.github.io/helm-chart"
--- a/helm-chart/binderhub/requirements.yaml
+++ b/helm-chart/binderhub/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: jupyterhub
-  version: "0.9.0"
+  version: "0.10.2"
   repository: "https://jupyterhub.github.io/helm-chart"
--- a/helm-chart/images/binderhub/requirements.in
+++ b/helm-chart/images/binderhub/requirements.in
@@ -3,11 +3,12 @@
 #
 #    ./dependencies freeze --upgrade
 #
-google-cloud-logging
+# FIXME: google-cloud-logging was pinned to avoid introducing a potentially
+#        breaking change as major version 2 was recently released. See:
+#        https://github.com/googleapis/python-logging/blob/master/UPGRADING.md
+google-cloud-logging==1.*
 
-# pins for our image
-# kubernetes is pinned to 9 because we have observed a performance issue in 12
-# where listing pods took ~6s instead of ~1s
-kubernetes==9.*
-# jupyterhub is pinned to match the JupyterHub Helm chart's version of jupyterhub
-jupyterhub==1.1.*
+# jupyterhub and kubernetes is pinned to match the JupyterHub Helm chart's
+# version of jupyterhub
+jupyterhub==1.2.*
+kubernetes==12.*
--- a/helm-chart/images/binderhub/requirements.txt
+++ b/helm-chart/images/binderhub/requirements.txt
@@ -7,38 +7,38 @@
 alembic==1.4.3            # via jupyterhub
 async-generator==1.10     # via jupyterhub
 attrs==20.3.0             # via jsonschema
-cachetools==4.1.1         # via google-auth
-certifi==2020.11.8        # via kubernetes, requests
+cachetools==4.2.0         # via google-auth
+certifi==2020.12.5        # via kubernetes, requests
 certipy==0.1.3            # via jupyterhub
-cffi==1.14.3              # via cryptography
-chardet==3.0.4            # via requests
-cryptography==3.2.1       # via pyopenssl
-docker==4.3.1             # via -r binderhub.in
+cffi==1.14.4              # via cryptography
+chardet==4.0.0            # via requests
+cryptography==3.3.1       # via pyopenssl
+docker==4.4.0             # via -r binderhub.in
 entrypoints==0.3          # via jupyterhub
 escapism==1.0.1           # via -r binderhub.in
-google-api-core[grpc]==1.23.0  # via google-cloud-core, google-cloud-logging
-google-auth==1.23.0       # via google-api-core, kubernetes
-google-cloud-core==1.4.3  # via google-cloud-logging
+google-api-core[grpc]==1.24.1  # via google-cloud-core, google-cloud-logging
+google-auth==1.24.0       # via google-api-core, kubernetes
+google-cloud-core==1.5.0  # via google-cloud-logging
 google-cloud-logging==1.15.1  # via -r requirements.in
 googleapis-common-protos==1.52.0  # via google-api-core
-grpcio==1.33.2            # via google-api-core
+grpcio==1.34.0            # via google-api-core
 idna==2.10                # via requests
 ipython-genutils==0.2.0   # via traitlets
 jinja2==2.11.2            # via -r binderhub.in, jupyterhub
 jsonschema==3.2.0         # via -r binderhub.in, jupyter-telemetry
 jupyter-telemetry==0.1.0  # via jupyterhub
-jupyterhub==1.1.0         # via -r binderhub.in, -r requirements.in
-kubernetes==9.0.1         # via -r binderhub.in, -r requirements.in
+jupyterhub==1.2.2         # via -r binderhub.in, -r requirements.in
+kubernetes==12.0.1        # via -r binderhub.in, -r requirements.in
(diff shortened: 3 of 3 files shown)


consideRatio:
The bump to JupyterHub 1.2.0 have led to a failing assumption in the test named test_auth. Generally, I don't understand these parts so well, perhaps I can recruit some help from @bitnik or @minrk who have developed this test and such?

Here is the test failure: https://travis-ci.com/github/jupyterhub/binderhub/jobs/420930577#L1144

```
    @pytest.mark.auth_test
    async def test_auth(app, path, authenticated, use_session):
        url = f'{app.url}{path}'
        r = await async_requests.get(url)
        assert r.status_code == 200, f"{r.status_code} {url}"
        if not authenticated:
            # not authenticated, we should get the page and be done
            assert r.url == url
            return
    
        # submit login form
        assert "/hub/login" in urlparse(r.url).path
        r2 = await async_requests.post(r.url, data={'username': 'dummy', 'password': 'dummy'})
        assert r2.status_code == 200, f"{r2.status_code} {r.url}"
        # verify that we landed at the destination after auth
>       assert r2.url == url
E       AssertionError: assert 'http://10.30...XJsIjogIi8ifQ' == 'http://127.0.0.1:8585/'
E         - http://127.0.0.1:8585/
E         + http://10.30.0.181:30123/hub/api/oauth2/authorize?client_id=binder-oauth-client-test&redirect_uri=http%3A%2F%2F127.0.0.1%3A8585%2Foauth_callback&response_type=code&state=eyJ1dWlkIjogIjU2YTJhOTJhYmYyNjQ3ZDFiYjY5NzVlMDkwNmNiYjY1IiwgIm5leHRfdXJsIjogIi8ifQ
```

bitnik:
@consideRatio thanks for doing this!

At first look, it looks like `oauth_no_confirm` config is ignored

bitnik:
I think https://github.com/jupyterhub/jupyterhub/pull/3234 will fix the issue

bitnik:
We should also upgrade JupyterHub version here

https://github.com/jupyterhub/binderhub/blob/1619d74a24d20bf101f1ac893071b57ed1b3e940/helm-chart/images/binderhub/requirements.txt#L4

consideRatio:
## Action points
- [ ] Patch JupyterHub with jupyterhub/jupyterhub#3234
- [ ] Release JupyterHub 1.2.1 preferably
- [ ] Update Z2JH to use the patched JupyterHub and release 0.10.1 preferably
- [ ] Update this PR to use Z2JH with a patched JupyterHub

betatim:
#1180 is merged

consideRatio:
Nice! I think like #1176 this merge may benefit from active monitoring on mybinder.org's staging environment. I'll be around to help it work all the way.

consideRatio:
@minrk @betatim I let the bump of kubernetes/python-client remain to 12.0.0 in this PR, which means both the hub image and binder image will use kubernetes 12.0.0 if this PR is merged, as compared to the hub image using 10.0.1 and the binder image using 9.0.1.

I'd like to better understand the question I raise here https://github.com/jupyterhub/mybinder.org-deploy/issues/1687#issuecomment-720526045 to feel safe about this merge.

betatim:
I agree with Erik that we should not merge this PR until we understand if there is a performance regression in v12 of the Python kubernetes client or not (xref https://github.com/jupyterhub/mybinder.org-deploy/issues/1687#issuecomment-720505798).

I don't really know how to test this other than setting up a BinderHub with this PR and then generating traffic to it that is roughly similar to gke.mybinder.org and then timing how long requests take with v9 and v12. Maybe we can see the slowdown with less load already, I guess we'd be just looking for a smaller effect.

betatim:
On suggestions for migrating mybinder.org. I don't quite understand (1) NetworkPolicies. We currently use network policies to limit traffic so we can't disable them. Or at the very least we'd have to be very sure that after disabling them we can turn them on quickly again as they are used to limit traffic from the user launched Pods to the internet.

From reading https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CHANGELOG.md#0100---2020-10-29 I am thinking that what changed is that the z2jh chart now adds some network policies of its own. These limit traffic to the chp and jupyterhub pod. This means we will need to add a network policy that matches requests from the BinderHub pod to the jupyterhub pod to this helm chart. Even for no mybinder.org deploys right? One thing I am not clear on: The BinderHub pod talks to `hub_url` (a public hostname) in order to contact the JupyterHub pod. Does this mean it is automatically covered by the network policies that the z2jh helm chart installs? What if the user configures `hub_url_local`?

consideRatio:
@betatim it is correct that z2jh will in 0.10.0+ create k8s NetworkPolicy resources by default which it didn't before. If you are using mybinder.org-deploy specific NetworkPolicy resources already, it means you have a NetworkPolicy controller that enforce them in your k8s cluster.

You have the option in z2jh to disable the creation of the networkpolicy resources, there are four.

- hub.networkPolicy.enabled=false
- proxy.chp.networkPolicy.enabled=false
- proxy.traefik.networkPolicy.enabled=false
- singleuser.networkPolicy.enabled=false

They describe what connections are allowed to be established for specific pods they target. proxy.traefik is the autohttps pod, proxy.chp is the normal proxy pod.

By default, all http/https traffic to the proxy/autohttps pods are allowed, so if you communicate with the hub through those, you are good, but http traffic directly to the hub without going through the proxy isn't allowed by default, so any pod that tries to do that will need to have a certain label (`hub.jupyter.org/network-access-hub: "true"`).

But, since the creation of network policy resources in the z2jh Helm chart has been disabled in previous releases and they haven't been explicitly enabled in mybinder.org-deploy, then I suggest that they are disabled first, and then we upgrade z2jh to verify in isolation that it works separately from also verifying that network policies doesn't cause trouble.

consideRatio:
@betatim I have pushed a pin of kubernetes-client/python to v9. At this point, I think this PR is safe to merge, and if mybinder.org-deploy also upgrades this in two steps which involves first explicitly disabling the creation of network policy resources, and then enabling them, then it will be easier to pin potential issues.

consideRatio:
## Status update
- [ ] this PR pins binder pod's image version of kubernetes to v9, which matches z2jh v0.9.1 which is currently used.
- [ ] hub pod's image version of kubernetes: currently v12.0.1 (we think we got performance regression with v12.0.0)
- [ ] kubespawner 0.12.0+ requires kubernetes>=0.10.1, at least because the KubeIngressProxy class separate from KubeSpawner that we need requires it, but perhaps also because KubeSpawner has started to require it since then.

I cannot ping z2jh to kubernetes==9 without pinning kubespawner==0.11 instead of kubespawner==0.14.1 which contain very relevant AKS performance and stability fixes.

If wanted for testing purposes, I can provide z2jh dev releases with:
- hub image pins kubernetes==9 and kubespawner==0.11
- hub image pins kubernetes==10 but kubespawner is free to become 0.14.1
- hub image pins kubernetes==11 but kubespawner is free to become 0.14.1
- hub image pins kubernetes==12 but kubespawner is free to become 0.14.1

OH! But z2jh 0.9.1 has kubernetes==10.0.1 in the hub image already and you have been fine with that!

## Action plan
1. I provide a dedicated dev release of z2jh for use of testing, which pins kubernetes==10 in the hub image
2. I update the binderhub pods kubernetes client to use kubernetes==10
3. We merge this
4. We try it out in mybinder.org-deploy, which I think should not cause any performance degradation.
5. We test kubernetes==12 in hub and binder images and see if we get a regression in performance, this test can either be in production or in a ci environment if someone develops a test for that.

Sounds okay?

consideRatio:
## Status update
I have pinned z2jh to a version, that in turn has pinned `kubernetes==10.*`, which has been used for a very long time without issues in the hub pod, also by mybinder.org. I have pinned the same version in the binderhub pod as well.

consideRatio:
I'll go ahead and merge this to try it on mybinder.org now during a sunday night close to xmas.

PR #1186, opened 2020-10-30
use HEAD for default git ref

### LINKED ISSUE #1175: Migrate ALL providers to expecting main branch, not just GitHub, and announce the change
The work done in https://github.com/jupyterhub/binderhub/pull/1172 means that binderhub now looks for the `main` branch if a ref isn't provided, rather than `master`. This is probably a change we should roll out across all providers but I, at least, am unfamiliar with where these other communities are in this process and this change will be breaking previously established behaviour, possibly disrupting a lot of users. Therefore, I think we should have some form of "announcement" (blog post, twitter thread...) ready to go when we do make this change, just so users aren't blind-sided and they can find information on how to fix their links.

_Note: This change does not affect gists over which the user has no control of the ref naming._

### TODOs

- [ ] Draft a blog post/twitter thread to announce the change
- [ ] Change the behaviour of the `placeholder` variable added in PR #1172 

```
placeholder = "main";
if (provider === "gist") {
    placeholder = "master";
}
```

minrk:
for all git-based providers. The result is a stable URL that resolves to the 'default ref' of any git repo anywhere, that works no matter what the default branch is named, and even across branch renames over time.

HEAD always resolves to the default branch, so we don't need to do anything special like `ls-remote` to figure out the name, since HEAD itself is a valid ref.

Changes:

- verify with more paramtrized tests that all git-based providers (git, github, gitlab, gist) support HEAD as a ref
- gist was the only one needing a fix, since "master" was handled specially by the provider, not git or github
- Make git handling of an unresolved ref more consistent with the rest (return None instead of raising an error)
- update `placeholder` in form used for the default value to HEAD

closes #1175


binderhub/repoproviders.py
```diff
@@ -882,7 +881,7 @@ def get_resolved_ref(self):
                              "'GistRepoProvider.allow_secret_gist = True'")
 
         all_versions = [e['version'] for e in ref_info['history']]
-        if (len(self.unresolved_ref) == 0) or (self.unresolved_ref == 'master'):
+        if self.unresolved_ref in {"", "HEAD", "master"}:
```
minrk, line 884: gist was the only provider that didn't support HEAD because it's the only one where we resolve ref names in the provider instead of calling out to git/github/etc. I believe if we used `ls-remote` as in the git provider, this would not have needed any changes.

sgibson91 (approved):
This looks great to me!

sgibson91:
Here's the error(s) I found in the CI

```
    def test_github_ref(repo, unresolved_ref, resolved_ref):
        spec = f"{repo}/{unresolved_ref}"
        provider = GitHubRepoProvider(spec=spec)
        slug = provider.get_build_slug()
        assert slug == repo.replace("/", "-")
        full_url = provider.get_repo_url()
        assert full_url == f"https://github.com/{repo}"
        ref = IOLoop().run_sync(provider.get_resolved_ref)
        if resolved_ref == True:
            # True means it should resolve, but don't check value
            assert ref is not None
>           assert sha1_validate(ref)
E           NameError: name 'sha1_validate' is not defined
```

minrk:
@sgibson91 sorry! Fixed.

sgibson91:
Fab, thank you!

betatim:
Such a smart idea!!!

PR #1189, opened 2020-11-02
Improve debugging of health-check failures

minrk:
- log exceptions instead of hiding them
- log unhealthy services
- for debug-by-default handlers, only do so if requests complete in less than 1 second.
  It's always worth seeing a request if the duration is slow.

This should help debugging issues with latest deployments where health checks have started timing out

### DIFF
--- a/binderhub/log.py
+++ b/binderhub/log.py
@@ -78,6 +78,7 @@ def log_request(handler):
     """
     status = handler.get_status()
     request = handler.request
+    request_time = 1000.0 * request.request_time()  # seconds to milliseconds
     if status == 304 or (
         status < 300
         and (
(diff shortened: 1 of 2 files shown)


binderhub/log.py
```diff
@@ -78,6 +78,8 @@ def log_request(handler):
     """
     status = handler.get_status()
     request = handler.request
+    request_time = 1000.0 * handler.request.request_time()  # milliseconds
```
consideRatio, line 81:
```suggestion
    request_time = 1000.0 * handler.request.request_time()  # micro- to milli-seconds
```

Is this correct?

minrk: No, seconds to milliseconds. Added clarifying comment

consideRatio: Woops ah

consideRatio (approved):
LGTM!

PR #1201, opened 2020-11-03
CI: stop asserting local binderhub version match remote version

### LINKED ISSUE #1200: Test failures of the Helm chart diagnosed along with an easy fix
Following #1188 we are not longer always rebuilding our docker images for the Helm chart. Due to this, we run into this error.

```
=================================== FAILURES ===================================
______________________________ test_about_handler ______________________________
app = <binderhub.tests.conftest.RemoteBinderHub object at 0x7f7e74491730>
    @pytest.mark.remote
    async def test_about_handler(app):
        # Check that the about page loads
        r = await async_requests.get(app.url + "/about")
        assert r.status_code == 200
        assert "This website is powered by" in r.text
>       assert binder_version in r.text
E       assert '0.2.0+354.g83ad58e.dirty' in '<!DOCTYPE html>\n<html>\n<head>\n  <title>Binder</title>\n  \n<meta property="og:description" content="Reproducible, ...        \n        </div>\n      </div>\n      \n    </div>\n  </div>\n</div>\n\n\n  \n\n\n  \n  \n  \n</body>\n</html>'
E        +  where '<!DOCTYPE html>\n<html>\n<head>\n  <title>Binder</title>\n  \n<meta property="og:description" content="Reproducible, ...        \n        </div>\n      </div>\n      \n    </div>\n  </div>\n</div>\n\n\n  \n\n\n  \n  \n  \n</body>\n</html>' = <Response [200]>.text
```

This is caused by the test_environment being out of sync with the built image. I that we remove the version match assertion as a resolution as the other part of the test is still relevant.

consideRatio:
Closes #1200 by no longer asserting the remote test environment will have the binderhub version according to versioneer which use git commits to decide its version, compared to the built image.

The image would rebuild if binderhub the Python package had changed at all, but it wouldn't if something unrelated has been committed. The \_version.py script doesn't care about this though, and decides the binderhub package is even newer than the image and observes a version mismatch.

Due to this, I suggest we stop asserting this. I don't see this assertion to be helpful to make unless we tighten down the test to only be relevant in a few cases, and that doesn't seem to be helpful enough to motivate it. Due to this, I'll go ahead and remove this assertion from the /about test.

### DIFF
--- a/binderhub/tests/test_main.py
+++ b/binderhub/tests/test_main.py
@@ -80,7 +80,6 @@ async def test_about_handler(app):
     r = await async_requests.get(app.url + "/about")
     assert r.status_code == 200
     assert "This website is powered by" in r.text
-    assert binder_version in r.text
 
 
 @pytest.mark.remote


betatim:
It would be good to add a modified version of this assertion back. The test didn't mean to check that the versions were the same but that the version appears in the page. I think using the local version was an easy way to check that the expected text appears. A bit more tricky to do if we don't know what version will appear but it would be good to keep testing that a vresion is displayed.

consideRatio:
I agree @betatim, i felt i overstepped, and added it back in https://github.com/jupyterhub/binderhub/pull/1203/files

PR #1202, opened 2020-11-03
[MRG] Speed up repository builds in the test suite

### LINKED ISSUE #1199: Reduce the image build tests to reduce CI times from ~13 minutes
https://github.com/jupyterhub/binderhub/blob/71896a68432dd97500f36b03399c8922953c7a7a/binderhub/tests/test_build.py#L18-L28

This test is run four times as it pass four different versions of the parameters. I'd like to make it faster, because we wait several minutes here and I bet we can test the same details with at least one less build.

- The first build takes a long time, its the first.
- The second is done very quickly because it is the same as the first in practice and has been completed already.
- The third is another build which takes significant time, and if the point is to test building a branch, we can perhaps instead create a branch matching that commit, or create a tag matching this commit instead?
- The fourth is a new build again, and takes some time.

## I suggest the following
- [x] Optimize third test, which is about using a not already resolved (https://github.com/jupyterhub/binderhub/commit/db9029f4ccdfe0987e3fc41b2436c5fe2d6f01d9), to reference to point to a tag to the same commit we have already built.
  - It seems that two new commits have been added, so master is now different from what it was before. Since those commits are about performance, we want to start pointing to master again with the first two builds, which is commit hash 20c4fe55a9b2c5011d228545e821b1c7b1723652.
- [x] We stop doing a second build from scratch, and only do one. The second new repo build is to test against GitLab as well as GitHub, but I think we can settle for just GitHub as its repo2docker's job to fix the functionality of all different providers mainly.
- [x] We trivialize the build. Ah, this is already done in https://github.com/jupyterhub/binderhub/commit/be975033516cf3c71cfa5e4256cc6309a4069c9f
- [ ] We could perhaps even do `docker pull` on something that makes us only do a very silly final step in the build rather than a full rebuild that does all the repo2docker stuff from scratch?

consideRatio:
Closes #1199 by https://github.com/jupyterhub/binderhub/pull/1202#issuecomment-729297913

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -143,16 +143,16 @@ jobs:
       - name: Run main tests
         if: matrix.test == 'main'
         # running the "main" tests means "all tests that aren't auth"
-        run: pytest -m "not auth" -vx --cov binderhub
+        run: pytest -m "not auth" -vx --cov binderhub --durations=0
       - name: Run auth tests
         if: matrix.test == 'auth'
         # running the "auth" tests means "all tests that are marked as auth"
-        run: pytest -m "auth" -vx --cov binderhub
+        run: pytest -m "auth" -vx --cov binderhub --durations=0
       - name: Run helm tests
         if: matrix.test == 'helm'
         run: |
             export BINDER_URL=http://localhost:30901
-            pytest -m "remote" -vx --cov binderhub
+            pytest -m "remote" -vx --cov binderhub --durations=0
       - name: Kubernetes namespace report
         if: ${{ failure() }}
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -143,16 +143,16 @@ jobs:
       - name: Run main tests
         if: matrix.test == 'main'
         # running the "main" tests means "all tests that aren't auth"
-        run: pytest -m "not auth" -vx --cov binderhub --durations=0
+        run: pytest -m "not auth" -vx --cov binderhub --durations=10
       - name: Run auth tests
         if: matrix.test == 'auth'
         # running the "auth" tests means "all tests that are marked as auth"
-        run: pytest -m "auth" -vx --cov binderhub --durations=0
+        run: pytest -m "auth" -vx --cov binderhub --durations=10
       - name: Run helm tests
         if: matrix.test == 'helm'
         run: |
             export BINDER_URL=http://localhost:30901
-            pytest -m "remote" -vx --cov binderhub --durations=0
+            pytest -m "remote" -vx --cov binderhub --durations=10
       - name: Kubernetes namespace report
         if: ${{ failure() }}
         run: |
--- a/binderhub/tests/test_build.py
+++ b/binderhub/tests/test_build.py
@@ -24,11 +24,11 @@ from .utils import async_requests
         quote("https://github.com/consideratio/cached-minimal-dockerfile", safe='')
     ),
     "git/{}/596b52f10efb0c9befc0c4ae850cc5175297d71c".format(
-        quote("https://github.com/consideratio/cached-minimal-dockerfile", safe='')
+        quote("https://github.com/binderhub-ci-repos/cached-minimal-dockerfile", safe='')
     ),
     # gh/ GitHub repo provider
-    "gh/consideratio/cached-minimal-dockerfile/HEAD",
-    "gh/consideratio/cached-minimal-dockerfile/596b52f10efb0c9befc0c4ae850cc5175297d71c",
+    "gh/binderhub-ci-repos/cached-minimal-dockerfile/HEAD",
+    "gh/binderhub-ci-repos/cached-minimal-dockerfile/596b52f10efb0c9befc0c4ae850cc5175297d71c",
     # gl/ GitLab repo provider
     "gl/binderhub-ci-repos%2Fcached-minimal-dockerfile/HEAD",
     "gl/binderhub-ci-repos%2Fcached-minimal-dockerfile/596b52f10efb0c9befc0c4ae850cc5175297d71c",
--- a/binderhub/tests/test_build.py
+++ b/binderhub/tests/test_build.py
@@ -15,19 +15,29 @@ from binderhub.build import Build
 from .utils import async_requests
 
 
+# We have optimized this slow test, for more information, see the README of
+# https://github.com/binderhub-ci-repos/minimal-dockerfile.
 @pytest.mark.asyncio(timeout=900)
 @pytest.mark.parametrize("slug", [
-    "gh/binderhub-ci-repos/requirements/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd",
-    "git/{}/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd".format(
-        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
+    # git/ Git repo provider
+    "git/{}/HEAD".format(
+        quote("https://github.com/consideratio/cached-minimal-dockerfile", safe='')
     ),
-    "git/{}/master".format(
-        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
+    "git/{}/596b52f10efb0c9befc0c4ae850cc5175297d71c".format(
+        quote("https://github.com/consideratio/cached-minimal-dockerfile", safe='')
     ),
-    "gl/minrk%2Fbinderhub-ci/0d4a217d40660efaa58761d8c6084e7cf5453cca",
+    # gh/ GitHub repo provider
+    "gh/consideratio/cached-minimal-dockerfile/HEAD",
+    "gh/consideratio/cached-minimal-dockerfile/596b52f10efb0c9befc0c4ae850cc5175297d71c",
+    # gl/ GitLab repo provider
+    "gl/binderhub-ci-repos%2Fcached-minimal-dockerfile/HEAD",
+    "gl/binderhub-ci-repos%2Fcached-minimal-dockerfile/596b52f10efb0c9befc0c4ae850cc5175297d71c",
 ])
 @pytest.mark.remote
 async def test_build(app, needs_build, needs_launch, always_build, slug, pytestconfig):
+    """
+    Test build a repo that is very quick and easy to build.
+    """
     # can't use mark.github_api since only some tests here use GitHub
     if slug.startswith('gh/') and "not github_api" in pytestconfig.getoption('markexpr'):
         pytest.skip("Skipping GitHub API test")


consideRatio:
The thing to consider about this PR is about the removal of the gitlab build. I think it is far more valuable to save any local developer or someone waiting for the tests to complete to save time than to do what repo2docker already does in its tests.

betatim:
As I wrote in the issue: I disagree about removing the GitLab test. For local development you can simply not run tests you don't care about/only run tests you care about. For CI runs saving a few minutes is not worth it because CI minutes are something we have a lot of compared to human time required to react to downtime on mybinder.org. The BinderHub repo is special in the sense that we have to have a very high probability that `master` will "just work" on mybinder.org so that we can constantly deploy it without people wondering "will deploying this break mybinder.org or not? I am not sure and I don't have 1hr to watch and react to a a breakage so I'd rather not deploy now". This is a problem because this leads to the "diff" between deployments getting larger, which in turn increases the chances of something breaking and the time to then debug it because the diff is so big. Which in turn reduces the likelihood that people choose to deploy changes. So it starts a vicious spiral. We've been in that position before where changes were not deployed for weeks or months. Let's not go back there.

We can speed up the GitLab test by modifying the contents of the repository to for example something like https://github.com/binder-examples/minimal-dockerfile which is about as fast to build as it gets.

---

Why is there an issue and this PR? It feels like living in a parallel universe because there are two places to discuss the same thing. I think there is no need to create an issue if a PR for that issue is created straight away. Let's either only make a PR or  make an issue which is only followed up with a PR after discussion.

binderhub/tests/test_build.py
```diff
@@ -17,17 +17,26 @@
...
-    "gh/binderhub-ci-repos/requirements/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd",
-    "git/{}/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd".format(
+    "gh/binderhub-ci-repos/requirements/20c4fe55a9b2c5011d228545e821b1c7b1723652",
+    "git/{}/20c4fe55a9b2c5011d228545e821b1c7b1723652".format(
         quote("https://github.com/binderhub-ci-repos/requirements", safe='')
     ),
     "git/{}/master".format(
         quote("https://github.com/binderhub-ci-repos/requirements", safe='')
     ),
-    "gl/minrk%2Fbinderhub-ci/0d4a217d40660efaa58761d8c6084e7cf5453cca",
```
betatim, line 27: Please don't remove this test. Instead lets try and simplify the repository so it builds more quickly

binderhub/tests/test_build.py
```diff
@@ -17,17 +17,26 @@
...
 ])
 @pytest.mark.remote
 async def test_build(app, needs_build, needs_launch, always_build, slug, pytestconfig):
+    """
+    Test build an image a set of times against a repo which is very simple to
+    build.
+    
+    This test should only be run once when the build is actually taking place
+    without cache. If any commits are made to master of
```
betatim, line 35: I don't really understand the first sentence (line 34 and 35). This test checks that several different ways of specifying a build slug work and get handled correctly.

betatim:
Updated the title of the PR to something more descriptive. WDYT?

binderhub/tests/test_build.py
minrk: I agree that we should keep this test.

binderhub/tests/test_build.py
```diff
@@ -17,17 +17,26 @@
 
 @pytest.mark.asyncio(timeout=900)
 @pytest.mark.parametrize("slug", [
-    "gh/binderhub-ci-repos/requirements/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd",
-    "git/{}/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd".format(
+    "gh/binderhub-ci-repos/requirements/20c4fe55a9b2c5011d228545e821b1c7b1723652",
+    "git/{}/20c4fe55a9b2c5011d228545e821b1c7b1723652".format(
```
minrk, line 21: Should we use HEAD here to avoid manually changing the ref for the tests? That also tests the default behavior.

binderhub/tests/test_build.py
consideRatio: Yeah line 34 and 35's communication didn't turn out well. My intention was to communicate that while the test_build test case run several times with different arguments as input, it is meant to avoid making time consuming builds by being able to use caches of previous builds.

binderhub/tests/test_build.py
consideRatio: When the /master test was added, it was added with the motivation of testing a not already full git hash.

consideRatio:
> Why is there an issue and this PR? It feels like living in a parallel universe because there are two places to discuss the same thing. I think there is no need to create an issue if a PR for that issue is created straight away. Let's either only make a PR or make an issue which is only followed up with a PR after discussion.

I understand the friction you have experienced @betatim, I mean to avoid that. I first created the issue, and an hour later or so I decided to invest my time to try address it in a PR.

> We can speed up the GitLab test by modifying the contents of the repository to for example something like https://github.com/binder-examples/minimal-dockerfile which is about as fast to build as it gets.

Wieee yes! Let's do https://github.com/binder-examples/minimal-dockerfile builds! To not need to rebuild the entire repo2docker boilerplate is the biggest improvement we can get I think.

binderhub/tests/test_build.py
betatim: I think adding a test case for HEAD is a good idea, especially because that is the default now.

I'd also keep the explicit reference case. My reasoning is that it tests a case that isn't the default and infrequently used/developed for. As a developer you might test master/main/HEAD explicitly but rarely think "ah yes, let's make sure explicit commit hashes still work". This makes it a good task for robots because they don't get bored of testing the same stuff over and over again.

Maybe we can remove the manual work of having to update the commit hash by making the test find out the latest hash and then updating the build slug. Like a templated build slug. However we commit so rarely to the CI repos used here that I'd not spend a lot of time on building this automation and instead would add a comment to let performance improvers from the future know that there is a low hanging fruit here.

betatim:
> an hour later I decided to work on it

nods. It is tricky because you want to create an issue when you think of it and don't know if you'll have time to work on it or not. So when you then shortly after do have time, what to do? I don't know. Maybe we should then post in the issue saying "Please discuss in the PR"? No good ideas from my side :(

> Wieee yes! Let's do binder-examples/minimal-dockerfile builds!

Lets keep using the GitLab repository but change its contents, not switch to the binder-examples repo.

consideRatio:
>  Maybe we should then post in the issue saying "Please discuss in the PR"? No good ideas from my side :(

I think that was a good idea. Updating the issue to reflect there is now an open PR, and declaring where the discussion should go, and in general I figure it should go in the PR where some of the discussion always will need to go as it is where the changes are declared practically.

## Action points
- [ ] Define a minimal dockerfile build. I suggest the [truly-minimal branch of binder-examples/minimal-dockerfile](https://github.com/binder-examples/minimal-dockerfile/tree/truly-minimal) initially.
- [ ] Use HEAD in test_build, but make it concrete by discussing exactly how many times we run test_build with what arguments.
- [ ] Reference a GitLab repo with the minimal dockerfile build chosen.
- [ ] Reference a GitHub repo with the minimal dockerfile build chosen.

consideRatio:
### How many times do we run test_build, and with what arguments?

I think this is mutually agreed, so I'll make that explicit.
- To build a minimalistic image with repo2docker, using a custom Dockerfile.
- To not hardcoding a full git commit hash in this specific test

## GitLab or not?
I accept building a GitLab repo alongside the GitHub repo if needed, but would like to have a motivation that it is still needed even though we have a [GitLab resolver test in `test_repoproviders.py`](https://github.com/jupyterhub/binderhub/blob/e769c8f5ab6e8e97093b09cca98b0add4aa5863f/binderhub/tests/test_repoproviders.py#L419-L440) that has been updated several times since `test_build` added the GitLab test.

## Dedicated hash, HEAD, master?
Since we have [this test](https://github.com/jupyterhub/binderhub/blob/e769c8f5ab6e8e97093b09cca98b0add4aa5863f/binderhub/tests/test_repoproviders.py#L459-L480) trying to resolve various URLs etc, I think the only thing we need to test in test_build, is if we manage to handle weird URLs in practice, but not the resolution of references.

Due to this, I suggest the test_build function is run twice.

### Currently in master

```
@pytest.mark.parametrize("slug", [
    "gh/binderhub-ci-repos/requirements/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd",
    "git/{}/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd".format(
        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
    ),
    "git/{}/master".format(
        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
    ),
    "gl/minrk%2Fbinderhub-ci/0d4a217d40660efaa58761d8c6084e7cf5453cca",
])
```

### Suggested
I suggest we combine the dedicated commit hash, the URL quoting of the commit hash, and the master branch reference into one.

```
@pytest.mark.parametrize("slug", [
    "gh%2Fbinderhub-ci-repos/minimal/HEAD",
    "gl%2Fbinderhub-ci-repos/minimal/HEAD",
])
```

consideRatio:
## Changes
We keep testing builds with gh/gl/git, but we now use a minimalistic Dockerfile instead.

## Outcome
[Before](https://github.com/jupyterhub/binderhub/runs/1396912568): 15m14 + 3m25s + 18m04s
[After](https://github.com/jupyterhub/binderhub/runs/1396912568?check_suite_focus=true): 7m30s + 3m11s + 10m23s

Almost twice as fast runs now.

binderhub/tests/test_build.py
```diff
@@ -17,17 +17,17 @@
 
 @pytest.mark.asyncio(timeout=900)
 @pytest.mark.parametrize("slug", [
-    "gh/binderhub-ci-repos/requirements/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd",
-    "git/{}/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd".format(
-        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
+    "gh/binder-examples/minimal-dockerfile/truly-minimal",
```
betatim, line 20: We should continue to use repos from the binderhub-ci-repos organisation. The example repositories get changed by humans a lot without knowing that changing them might break the CI of binderhub.

binderhub/tests/test_build.py
```diff
@@ -17,17 +17,17 @@
...
+    "git/{}/truly-minimal".format(
+        quote("https://github.com/binder-examples/minimal-dockerfile", safe='')
     ),
-    "git/{}/master".format(
-        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
-    ),
-    "gl/minrk%2Fbinderhub-ci/0d4a217d40660efaa58761d8c6084e7cf5453cca",
+    "gl/consideRatio%2Fbinderhub-ci-repo/HEAD",
```
betatim, line 24: Please don't change the repository and instead get your changes merged into `minrk/binderhub-ci` (or more ambitiously create a binderhub-ci org on Gitlab and lets use that)

binderhub/tests/test_build.py
betatim: In particular my request is:
1. to create a `binderhub-ci-repos/minimal-dockerfile` repository
1. use an explicit SHA1 for the tests instead of referring to a branch

Using a fixed hash means that we always get a known input when the tests are executed. A frequent reasons for tests that used to work suddenly not working any more (the famous "tests fail but not because of my changes so please merge with red CI" comment PR authors make) is because an input to our CI has changed.

consideRatio:
## Performance testing
/gh was started first, and took ~80 seconds while the other took ~60 seconds, it is because it was started first, or because github is slower for some reason?

image

I moved /gh tests to be last instead and put /gl tests first, I confirm that the first repo-provider to run is slower by 14-19 seconds. I think this is because the image-builder container spins up faster the second/third time.

image.

For each followup run on a repo-provider, everything goes very quick because it will never start a image-builder pod at all I think.

consideRatio:
@betatim @minrk I have documented the new minimal-dockerfile and cached-minimal-dockerfile repos in https://github.com/binderhub-ci-repos/minimal-dockerfile. I want to highlight some outcomes of this setup.

1. We have a daily automatic build/push of a minimal docker image in https://github.com/binderhub-ci-repos/minimal-dockerfile.
2. We have two repos named cached-minimal-dockerfile on github/gitlab repo that are will be able to remain frozen in time forever as their Dockerfile only reference the pre-build minimal-dockerfile in a FROM statement and their readme delegate info to the https://github.com/binderhub-ci-repos/minimal-dockerfile repo's readme file.

Together, this makes us able to update the minimal-dockerfile with changes without thinking about commit changes, as we only reference cached-minimal-dockerfile, and by automatically updating the minimal-dockerfile image by daily rebuilds, we ensure that don't start to fail because the built image has become outdated. It also makes our build duration go from ~20s to ~5s.

I have now optimized as much as I can, and it is not a question on how much tests variations should be run. Each additional repo provider we add adds ~65+5 seconds if we do a HEAD+SHA test against a cached-minimal-dockerfile. In the PRs current state we test 3 repo providers with HEAD+SHA.

binderhub/tests/test_build.py
consideRatio: I created a new organization and invited a few people to it!

binderhub/tests/test_build.py
consideRatio: > to create a binderhub-ci-repos/minimal-dockerfile repository

Done in both GitHub and GitLab, using the exact same git content in both repos which makes both have the same SHA etc.

> use an explicit SHA1 for the tests instead of referring to a branch

Done.

> have a test for gh/binderhub-ci-repos/minimal-dockerfile/SHA1, git/quote("binderhub-ci-repos/minimal-dockerfile")/SHA1 and gh/gh/binderhub-ci-repos/minimal-dockerfile/HEAD. Having three builds of the same thing does not use more time than building it once or twice because we have docker layer caching

Done, but note that after the build has been optimized as much as I think is possible, there is a ~65 second penalty per git repo provider. Notes about this are in https://github.com/jupyterhub/binderhub/pull/1202#issuecomment-729046409.

binderhub/tests/test_build.py
consideRatio: Comment no longer around to cause confusion!

consideRatio:
@betatim okay to merge?

betatim:
I've not had time to review this yet. Is there a reason to be in a hurry merging this optimisation?

.github/workflows/test.yml
```diff
@@ -143,16 +143,16 @@ jobs:
       - name: Run main tests
         if: matrix.test == 'main'
         # running the "main" tests means "all tests that aren't auth"
-        run: pytest -m "not auth" -vx --cov binderhub
+        run: pytest -m "not auth" -vx --cov binderhub --durations=0
```
betatim, line 146: I think showing the durations for all tests creates very large logs. How about the top 10? This is also the value we use in repo2docker. (same for all three instances of `--durations`

binderhub/tests/test_build.py
```diff
@@ -15,19 +15,29 @@
...
     ),
-    "git/{}/master".format(
-        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
+    "git/{}/596b52f10efb0c9befc0c4ae850cc5175297d71c".format(
+        quote("https://github.com/consideratio/cached-minimal-dockerfile", safe='')
     ),
-    "gl/minrk%2Fbinderhub-ci/0d4a217d40660efaa58761d8c6084e7cf5453cca",
+    # gh/ GitHub repo provider
+    "gh/consideratio/cached-minimal-dockerfile/HEAD",
+    "gh/consideratio/cached-minimal-dockerfile/596b52f10efb0c9befc0c4ae850cc5175297d71c",
```
betatim, line 31:
```suggestion
    "gh/binderhub-ci-repos/cached-minimal-dockerfile/HEAD",
    "gh/binderhub-ci-repos/cached-minimal-dockerfile/596b52f10efb0c9befc0c4ae850cc5175297d71c",
```

binderhub/tests/test_build.py
```diff
@@ -15,19 +15,29 @@
...
-    "git/{}/d687a7f9e6946ab01ef2baa7bd6d5b73c6e904fd".format(
-        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
+    # git/ Git repo provider
+    "git/{}/HEAD".format(
+        quote("https://github.com/consideratio/cached-minimal-dockerfile", safe='')
     ),
-    "git/{}/master".format(
-        quote("https://github.com/binderhub-ci-repos/requirements", safe='')
+    "git/{}/596b52f10efb0c9befc0c4ae850cc5175297d71c".format(
+        quote("https://github.com/consideratio/cached-minimal-dockerfile", safe='')
```
betatim, line 27:
```suggestion
        quote("https://github.com/binderhub-ci-repos/cached-minimal-dockerfile", safe='')
```

.github/workflows/test.yml
betatim, line 146:
```suggestion
        run: pytest -m "not auth" -vx --cov binderhub --durations=10
```

.github/workflows/test.yml
```diff
@@ -143,16 +143,16 @@ jobs:
...
       - name: Run auth tests
         if: matrix.test == 'auth'
         # running the "auth" tests means "all tests that are marked as auth"
-        run: pytest -m "auth" -vx --cov binderhub
+        run: pytest -m "auth" -vx --cov binderhub --durations=0
```
betatim, line 150:
```suggestion
        run: pytest -m "auth" -vx --cov binderhub --durations=10
```

.github/workflows/test.yml
```diff
@@ -143,16 +143,16 @@ jobs:
...
       - name: Run helm tests
         if: matrix.test == 'helm'
         run: |
             export BINDER_URL=http://localhost:30901
-            pytest -m "remote" -vx --cov binderhub
+            pytest -m "remote" -vx --cov binderhub --durations=0
```
betatim, line 155:
```suggestion
            pytest -m "remote" -vx --cov binderhub --durations=10
```

betatim:
Thanks for working on this optimisation

consideRatio:
Thank you for getting this merged @betatim! :heart:

PR #1205, opened 2020-11-04
[MRG] Replace gen.coroutine/yeild with modern async/await

consideRatio:
This should be fine considering py35 has reached end of life and async functions that were generators is supported in py36, which was the last piece gen.coroutine could be needed for still.

### DIFF
--- a/binderhub/registry.py
+++ b/binderhub/registry.py
@@ -6,7 +6,7 @@ import json
 import os
 from urllib.parse import urlparse
 
-from tornado import gen, httpclient
+from tornado import httpclient
 from tornado.httputil import url_concat
 from traitlets.config import LoggingConfigurable
 from traitlets import Dict, Unicode, default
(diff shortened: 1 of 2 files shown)


binderhub/tests/conftest.py
```diff
@@ -86,13 +86,6 @@ def load_mock_responses(host):
     MockAsyncHTTPClient.mocks.update(records)
 
 
-def pytest_collection_modifyitems(items):
-    """add asyncio marker to all async tests"""
-    for item in items:
-        if inspect.iscoroutinefunction(item.obj):
-            item.add_marker('asyncio')
```
minrk, line 93: I think we should keep this mark based on https://github.com/pytest-dev/pytest-asyncio/issues/124 where folks are asking for pytest-asyncio to stop implicit marking itself, since it causes compatibility problems. There's no harm in having it, as far as I can tell.

consideRatio: Ah okay!

consideRatio:
This PR is the BinderHub equivalent of https://github.com/jupyterhub/binderhub/pull/1205.

consideRatio:
Thank you for review/merge @betatim! :tada: :heart:

PR #1206, opened 2020-11-05
[MRG] Add commit title to chartpress message

betatim:
Having the title as well as the PR number helps people recognise what
the chart is about.

Follow up to #1197.


ci/publish
```diff
@@ -45,7 +45,7 @@ if [ "${TRAVIS_TAG:-}" == "" ]; then
     #       jupyterhub/binderhub@123abcd which will be understood as a reference
     #       by GitHub.
     PR_OR_HASH=$(git log -1 --pretty=%B | head -n1 | sed 's/^.*\(#[0-9]*\).*/\1/' | sed 's/^\([0-9a-f]*\)-.*/@\1/')
-    EXTRA_MESSAGE="${TRAVIS_REPO_SLUG}${PR_OR_HASH}"
+    EXTRA_MESSAGE="${TRAVIS_REPO_SLUG}${PR_OR_HASH} ${LATEST_COMMIT_TITLE}"
```
consideRatio, line 48: This environment variable isn't part of the CI system itself, but was previously named this even though it was mostly rendering to `#123` which was extracted from the latest commit title with `sed`.

The actual latest commit title can be extracted like this:

```
LATEST_COMMIT_TITLE=$(git log -1 --pretty=%B | head -n1)
```

I have made a mistake in #1197 I note also, which is that for the extraction of the commit hash, there also needs to be an inclusion of that (`%h`) for `sed` to work with.

```diff
-    PR_OR_HASH=$(git log -1 --pretty=%B | head -n1 | sed 's/^.*\(#[0-9]*\).*/\1/' | sed 's/^\([0-9a-f]*\)-.*/@\1/')
+    PR_OR_HASH=$(git log -1 --pretty=%h-%B | head -n1 | sed 's/^.*\(#[0-9]*\).*/\1/' | sed 's/^\([0-9a-f]*\)-.*/@\1/')
```

betatim: Thanks! I added a line to set the commit title.

ci/publish
```diff
@@ -45,7 +45,8 @@ if [ "${TRAVIS_TAG:-}" == "" ]; then
     #       jupyterhub/binderhub@123abcd which will be understood as a reference
     #       by GitHub.
     PR_OR_HASH=$(git log -1 --pretty=%B | head -n1 | sed 's/^.*\(#[0-9]*\).*/\1/' | sed 's/^\([0-9a-f]*\)-.*/@\1/')
```
betatim, line 47:
```suggestion
    PR_OR_HASH=$(git log -1 --pretty=%h-%B | head -n1 | sed 's/^.*\(#[0-9]*\).*/\1/' | sed 's/^\([0-9a-f]*\)-.*/@\1/')
```
is this what you meant as a fix to the `PR_OR_HASH` @consideRatio ?

minrk (approved):

ci/publish
consideRatio: Yepp @betatim thank you!! This makes the `@123abcd` hash be created successfully when the commit title doesn't container `#123` as only the GitHub merge commits can be assumed to do.

consideRatio (approved):
This LGTM!

I was considering adding the title like this before but was afraid to break henchbot logic I didn't know about how it worked. I think though, that I suggest we merge this and find out if it does and improve the reliability of henchbot if needed instead.

consideRatio:
This will cause cause conflicts when merged, I'd like to make it easier by tackling them in this PR rather than in the two PRs #1209 and https://github.com/betatim/binderhub/pull/2 which changes the previous mentioned PR

PR #1188, opened 2020-10-31
CI / Local dev refresh and Helm2->3 transition

consideRatio:
With a easy to understand CI system without tangled configuration that influence too many things at once, I hope that it will help myself and others to make contributions to this project in the long term. I wanted to update this step by step towards being easier to understand, maintain, and update it with tricks learned over time.

## Work to be done
- [x] Inline comment files on how they are used
- [x] Collect testing config into testing folder
- [x] Test with helm3 instead of helm2
- [x] Create a dedicated publishing step
- [x] Minimize rebuilds by configuring chartpress as well as possible
  - [x] Develop the rebuildOnContextPathChanges configuration option in chartpress (https://github.com/jupyterhub/chartpress/pull/98 - __Review help wanted!__)
  - [x] Release chartpress 0.7.0 (https://github.com/jupyterhub/chartpress/pull/99)
  - [x] Bump to use chartpress 0.7.0 in this repo

## Review notes
- __No tooling changes__. We still use `minikube` and I have not introduced new tooling (except the helm v2 to v3 migration).
- __No changes to BinderHub or Helm chart itself__. This PR only impacts the CI system local development instructions, and the guide with regards to now having instructions on how to use Helm 3 rather than Helm 2.
- __CONTRIBUTING.md would benefit from your review.__ I have improved them especially with regards to anyone wanting to develop the Helm chart in full rather than a BinderHub running locally interacting with a JupyterHub running in k8s.

## Positive outcomes
1. __CI refactoring reduced complexity__. This will make it easier to maintain for current maintainers, but for me, the most important part is that it can attract more maintenance contributions which makes the project more sustainable in the long term.
1. __Helm2 to Helm3 migration complete.__
1. __Helm3 based template and rendered template validation.__ `helm3 template --validate` both renders the templates but also validates them as valid k8s resources against a k8s api-server. It is a more heavy weight option to using `kubeval` and such tools as it requires a k8s api-server to speak with.
1. __Faster CI jobs__. With the chartpress feature developed, we can avoid triggering rebuilds of the BinderHub docker image for the Helm chart all the time, which can save us multiple minutes in build time.

PR #1209, opened 2020-11-08
[MRG] Add basic GitHub Actions workflow for testing

betatim:
This PR will reimplement our Travis testing setup on GitHub Actions. The goal is to retire Travis as quickly as possible and use only GitHub Actions.

Heavily based on https://github.com/manics/zero-to-jupyterhub-k8s/blob/gh-workflow/.github/workflows/test.yml


.github/workflows/test.yml
```diff
@@ -0,0 +1,90 @@
...
+            echo "List of cluster nodes:"
+            kubectl get nodes
+            echo
+            echo "Helm list output:"
+            helm list
+      - name: Setup helm chart repositories
+        shell: bash
+        run: |
+            helm repo add jupyterhub https://jupyterhub.github.io/helm-chart/
+            helm repo update
```
consideRatio, line 51: This step is included in the k3s-helm action already actually

betatim: Well, how can it know which repo to add and update it?

consideRatio: Woops my mistake, I got thinking it installed the pebble-helm-chart as part of the github action, but it doesn't. Sorry for the noise.

consideRatio:
Here is an extract from the recent pebble-helm-chart migration I attempted. I wanted to highlight the on failure/on success steps. I suggest we add a step running the [full_namespace_report](https://github.com/jupyterhub/binderhub/blob/master/ci/common#L60) function on failure.

```yaml
      - name: Run tests
        run: |
          helm lint pebble --values ci/ci-values.yaml
          helm lint pebble --strict --values ci/ci-values.yaml
          helm template pebble ./pebble --values ci/ci-values.yaml --validate
          helm install pebble ./pebble --values ci/ci-values.yaml --wait --timeout 60s
          helm test pebble
      - name: Emit report on success
        if: success()
        run: |
          kubectl get all --all-namespaces
      - name: Emit report on failure
        if: failure()
        run: |
          kubectl get all --all-namespaces
          kubectl logs pebble-test --all-containers --prefix
          kubectl logs pebble-coredns-test --all-containers --prefix
          kubectl logs deploy/pebble-coredns --all-containers --prefix
```

testing/k8s-binder-k8s-hub/binderhub-chart-config.yaml
```diff
@@ -7,7 +9,8 @@ service:
 
 config:
   BinderHub:
-    hub_url: http://192.168.99.100:30902
+    # don't set `hub_url` here, it needs to be set from Python code as we need
+    # to determine the IP at runtime.
```
consideRatio, line 13: I'm not 100%, but I think we may want to set hub_url_local: http://binder:8585.

consideRatio:
@betatim I added two commits to make the chart publishing work with GitHub actions, but there is one part missing still about DOCKER_USERNAME and DOCKER_PASSWORD. I hope that Min can help us with this and I have already pinged him to do this for z2jh and binderhub. See the concrete request in gitter if you happen to have the password for the dockerhub user jupyterhubbot

betatim:
Ah, sorry I didn't see your commits before I force pushed. I'll integrate those things again later.

consideRatio:
@betatim no worries, I created a PR for it as i still had them locally: https://github.com/betatim/binderhub/pull/2

betatim:
A bit puzzled why the tests fail. Or to put it differently: I am confused why this test ever passed. This is because when I run the code that the test exercises locally (with the same input) it raises an exception much earlier than where things go wrong in the test. Reading the code I expect what I see when I run the code in `ipython`. The output from the test:
```
_____________________________ test_hydroshare_doi ______________________________

    async def test_hydroshare_doi():
        spec = '10.4211/hs.b8f6eae9d89241cf8b5904033460af61'
    
        provider = HydroshareProvider(spec=spec)
    
>       ref = await provider.get_resolved_ref()

binderhub/tests/test_repoproviders.py:134: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/tornado/gen.py:742: in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
binderhub/repoproviders.py:374: in get_resolved_ref
    r = yield client.fetch(req)
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/tornado/gen.py:735: in run
    value = future.result()
binderhub/tests/utils.py:79: in fetch
    self._record_response(url_key, response)
```
The code in question is: https://github.com/jupyterhub/binderhub/blob/master/binderhub/repoproviders.py#L371-L374 which I would expect to fail when calling `self._parse_resource_id(self.spec)` when `spec = "10.4211/hs.b8f6eae9d89241cf8b5904033460af61"`.

So somewhere that DOI is being resolved that I can't spot :-/

betatim:
https://github.com/jupyterhub/binderhub/pull/1209#issuecomment-723645809 was a flakey test maybe or a short downtime of Hydroshare? Locally I got the same error because I misspelt the argument `specc=` not `spec=`, I think.

betatim:
The [`test_hydroshare_doi` test failed again](https://github.com/jupyterhub/binderhub/pull/1209/checks?check_run_id=1372602283) :-/

.github/workflows/test.yml
```diff
@@ -0,0 +1,321 @@
...
+            echo "curl http://localhost:30902/hub/api/" should print the JupyterHub version
+            curl http://localhost:30902/hub/api/
+            echo
+            echo "curl http://localhost:30901/health" to check BinderHub\'s health
+            curl http://localhost:30901/health
+      - name: Run tests
+        shell: bash
+        run: |
+            export BINDER_URL=http://localhost:30901
+            pytest -m "remote" -vx --cov binderhub
```
consideRatio, line 321: I'd love to see these three jobs (main, auth, helm) be combined into one job template that is parameterized into three different jobs.

I learned some tricks from @manics such as using `${{ matrix.my_setting }}` to conditionally do things in steps or conditionally run entire steps. This PR contains some examples of this: https://github.com/jupyterhub/jupyterhub/pull/3246.

betatim: Yes it would be nice to have one job and use a matrix but I am not sure that having lots of `if` statements to conditionally enable/disable steps based on the value of the matrix flags is worth it. Having three separate jobs makes it very explicit what steps should be executed when. So for now I'd like to keep it like this instead of adding ga layer of abstraction

consideRatio: > Yes it would be nice to have one job and use a matrix but I am not sure that having lots of if statements to conditionally enable/disable steps based on the value of the matrix flags is worth it. Having three separate jobs makes it very explicit what steps should be executed when. So for now I'd like to keep it like this instead of adding ga layer of abstraction

I agree things become more explicit, but like assembler which is also more explicit, it is also harder to overview and maintain. It is a quite extreme comparison and this case isn't at all that extreme, but I feel quite strongly about it, it relates to my time investment in #ci I would like to see them combined again, I put in a lot of thorough work in https://github.com/jupyterhub/binderhub/pull/1188 which included gathering all the differences into a few steps which were separated from the common.

Here is the current `.travis.yml` file with its different steps for the three test scenarios, the common scenarios are defined in `before_install` and `install`. I suggest that we do the same thing here, we have a set of common steps and then we have three test scenarios in three steps that we activate conditionally.


```yaml
    - stage: test
      name: main
      script:
        - ./testing/local-binder-k8s-hub/install-jupyterhub-chart
        - await_jupyterhub
        - pytest -m "not auth" -vx --cov binderhub

    - stage: test
      name: auth
      script:
        - ./testing/local-binder-k8s-hub/install-jupyterhub-chart --auth
        - await_jupyterhub
        - pytest -m "auth" -vx --cov binderhub

    - stage: test
      name: helm
      script:
        # Prepare:  builds images and updates default values.yaml to reference
        #           the images
        - echo "Preparing local BinderHub Helm chart"
        - helm dependency update ./helm-chart/binderhub
        - (cd helm-chart && chartpress)
        - git --no-pager diff

        # Validate: renders Helm chart templates and validates the rendered k8s
        #           manifests against the k8s api-server.
        - |
          helm template --validate binderhub-test helm-chart/binderhub \
              --values testing/k8s-binder-k8s-hub/binderhub-chart-config.yaml \
              --set config.BinderHub.hub_url=http://$(minikube ip):30902 \
              --set config.BinderHub.access_token=$GITHUB_ACCESS_TOKEN

        # Install:  installs the local BinderHub Helm chart and await it to
        #           become ready.
        - echo "Installing local BinderHub Helm chart with JupyterHub as a dependency into k8s cluster"
        - |
          helm upgrade --install binderhub-test helm-chart/binderhub \
              --values testing/k8s-binder-k8s-hub/binderhub-chart-config.yaml \
              --set config.BinderHub.hub_url=http://$(minikube ip):30902 \
              --set config.BinderHub.access_token=$GITHUB_ACCESS_TOKEN
        - await_binderhub

        # Test
        - export BINDER_URL=http://$(minikube ip):30901
        - echo "Running tests against $BINDER_URL"
        - pytest -m "remote" -vx --cov binderhub
```

betatim: In travis it was nice because we could share all the setup steps and then have steps that were different. The only way I know to share steps between jobs in GH Actions is to make steps conditional.

This is now implemented but I don't think it is easy to read because you have to check the `if:` for every step. However I don't have a strong enough opinion on this so will just do what ever others want.

manics: A compromise could be to move the common functionality into new bash functions in `ci/common`, and keep the three separate jobs?

betatim:
Is there a way to attach the k3s ports to the "main" network interface of a node? This could be useful to avoid having to route traffic from BinderHub to JupyterHub differently from traffic from a client to JupyterHub.

.github/workflows/test.yml
consideRatio: Thank you @betatim :heart:

@manics I think we can extract some bash functions, but I think we should let the github workflow be easy to parse in form of "await_jupyterhub" or similar, rather than "do_everything_script". I believe we already have extracted most relevant parts though.

.github/workflows/test.yml
```diff
@@ -0,0 +1,169 @@
+name: Tests
+
+on:
+  pull_request:
+  push:
+    branches: [master, main]
+
```
consideRatio, line 7:
```suggestion

defaults:
  run:
    shell: bash

```

betatim:
An alternative to using the "two URLs" setup (which is unusual compared to production deploys) would be to setup a hostname to use for `hub_url`. On the host machine it would point to 127.0.0.1 and in the BinderHub pod it would point to `proxy-public`. On the host we can append to `/etc/hosts` and in the pod we can use [something like this](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/f529f88bad4e7e4dc72eb5219b0d4ddf80b6d90e/dev-config-pebble.yaml) to setup the hostname.

consideRatio:
> An alternative to using the "two URLs" setup (which is unusual compared to production deploys) would be to setup a hostname to use for hub_url. On the host machine it would point to 127.0.0.1 and in the BinderHub pod it would point to proxy-public. On the host we can append to /etc/hosts and in the pod we can use something like this to setup the hostname.

I think the two URL setup is quite easy to understand and reason with, it is something I believe is unavoidable when testing cert acquisition with a local ACME server.

betatim:
Why would it be unavoidable?

I think setting up two DNS records is easy to do and avoids our tests taking a different path compared to production deploys. Therefore it is simpler and easier to understand.

consideRatio:
> Why would it be unavoidable?

I don't have all the pieces to be 100% sure about this, but from my memory of the long struggle I had to get things working was that I wanted to meet multiple wishes.

To use 1) a pre-existing domain name that doesn't require a modification to /etc/hosts and 2) can be overridden in a k8s cluster's DNS server to fool the local ACME server Pebble without unwanted consequences like we would have by overriding localhost meaning.

I have explored the ways to get local ACME server's to have a domain name point to a k8s local ip and have the same domain name point to the k8s cluster's IP several times, and have not come up with another solution.

betatim:
I've incorporated Erik's PR with changes for publishing the chart. I've not tested them/know too little about that to know if they will work or not though. Would be good for someone to take a look at them.

betatim:
I think this is ready for reviewing. Maybe @minrk can take a look over.

I think there are a few more things we could do (like fix the code scanning) but I'd address those in a new PR because there are PRs with actual content that are waiting to have working CI again. The most important thing to check once we have this merged is if the publishing of the chart works (so that henchbot and mybinder.org deploys can resume).

.github/workflows/test.yml
```diff
@@ -0,0 +1,170 @@
...
+        run: |
+            export BINDER_URL=http://localhost:30901
+            pytest -m "remote" -vx --cov binderhub
+      - name: Kubernetes namespace report
+        if: ${{ failure() }}
+        run: |
+          # Display debugging information
+          . ci/common
+          full_namespace_report
+      - name: "Upload code coverage stats"
```
manics, line 162: There's a [codecov action](https://github.com/codecov/codecov-action) that I've successfully used in
https://github.com/manics/jupyterhub-ansiblespawner/blob/4efa38c2016501675dabd23a269562d4af7ca1a1/.github/workflows/workflow.yml#L64-L67
Could be an option? But in the interests of fixing CI it's also fine to leave this for a later.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,91 @@
...
+          fetch-depth: 0
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+
+      - name: Install chart publishing dependencies (chartpress, helm)
+        run: |
+          . ./ci/common
+          setup_helm
+          pip install --no-cache-dir chartpress
```
consideRatio, line 47: We need a oneliner to install helm here since setup_helm is removed

.github/workflows/test.yml
betatim: I used this explicit snippet because I had it handy from trying to figure out why codecov (the action? the tool?) wasn't working for https://github.com/jupyterhub/repo2docker/pull/982

Switched to the action here

.github/workflows/publish.yml
betatim: I added the function back. The only reason I removed it was because I thought "the k3s action sets it all up".

.github/workflows/publish.yml
```diff
@@ -0,0 +1,87 @@
...
+          # chartpress requires the full history
+          fetch-depth: 0
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+
+      - name: Install chart publishing dependencies (chartpress, helm)
+        run: |
+          . ./ci/common
+          setup_helm v3.4.0
```
betatim, line 42: Does someone know a way to share/get this version from the action defined in `test.yml`? Would be nice but not a big deal I think

consideRatio: I googled a bit and didn't find a solution. I think workflows run independently of each other, so either you need to explicitly import one from the other somehow or declare something outside both that is read by both, like a github secret variable. I think those approaches are introducing more complexity that makes it harder to both maintain and understand to motivate not breaking the DRY principle.

We can use the stable version in both as an alternative, by not providing an explicit version. The installer script to get helm gives you the latest version with a delay of a day or so i think. At least I got 3.4.0 when 3.4.1 had been out for some hours already.

consideRatio:
Wieee, @betatim do you also get that nice feeling? ;D

image

.github/workflows/publish.yml
betatim: For me it is ok to repeat it here as it shouldn't matter much which version is used here. Especially as there is no easy way to share. And probably easy to find if it does lead to breakage.

betatim:
I get the feeling that we had to invest a lot of hours to get back to where we were before :D

consideRatio:
Wieee! With my own LGTM and the [MRG] title update and all comments resolved etc its time for a merge I think!

Let's see if things work as they are intended!

consideRatio:
Publish workflow run failed: https://github.com/jupyterhub/binderhub/actions/runs/361986200

image

Test workflow run seems to succeed: https://github.com/jupyterhub/binderhub/actions/runs/361986201

PR #1210, opened 2020-11-08
[MRG] Make the build pod clean up task run in the background

betatim:
We need to create a task, not just convert it into a future for the
build pod clean up function to run in the background all the time.

Noticed this because some tests printed a warning that this coroutine wasn't being awaited. I don't remember why there was an `ensure_future()` here. IMHO that would be useful for converting things that aren't a coroutine but won't get them executed.

Instead we now create a task which runs in the background. I think this is the right thing to do and equivalent to `IOLoop.current().add_callback()` from Tornado (which we could also use if we wanted to?).

Two questions for reviewing:
1. do `create_task` and `add_callback` achieve the same thing?
1. if yes, should we prefer one over the other? For example for cleanup during shutdown or something like that?

Because Travis is paused right now "all checks passed" as a PR status doesn't mean all our tests passed. They are "green" because the unit tests are not being executed at all :D

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -731,7 +731,7 @@ class BinderHub(Application):
         )
         self.http_server.listen(self.port)
         if self.builder_required:
-            asyncio.ensure_future(self.watch_build_pods())
+            asyncio.create_task(self.watch_build_pods())
         if run_loop:
             tornado.ioloop.IOLoop.current().start()
 


consideRatio:
This change will introduce a dependency on Py37 rather than just Py36 as we currently have according to [the python documentation](https://docs.python.org/3/library/asyncio-task.html#creating-tasks).

```python
async def coro():
    ...

# In Python 3.7+
task = asyncio.create_task(coro())
...

# This works in all Python versions but is less readable
task = asyncio.ensure_future(coro())
...
```

The ensure_future part will schedule the coroutine to be executed, but exactly where is not something I'm sure of, or if it will keep executing if not awaited etc.

betatim:
Ok, then this isn't needed. Super weird naming though :)

PR #1212, opened 2020-11-13
Syntax error fix to publishing workflow

consideRatio:
#1209 needed a followup to fix the following error. It was caused by `var_name = $1` where varName was interpreted as a command to execute rather than a variable to assign to, so this PR fixes it by doing `var_name=$1`. I also bumped the Helm version.

[image](https://github.com/jupyterhub/binderhub/actions/runs/361986200)


betatim:
Oh wow, shell be so picky about syntax :D

PR #1213, opened 2020-11-13
Fix chart publishing workflow

consideRatio:
Followup of #1209 and #1212 with, hopefully, final fix.

### DIFF
--- a/ci/publish
+++ b/ci/publish
@@ -6,6 +6,7 @@
 # Exit on errors, assert env vars, log commands
 set -eux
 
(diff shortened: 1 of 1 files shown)


ci/publish
```diff
@@ -6,6 +6,7 @@
 # Exit on errors, assert env vars, log commands
 set -eux
 
+cd helm-chart
```
betatim, line 9: This is the change that fixes the problem. The other two changes are clean up/prettiness only.

PR #1216, opened 2020-11-17
[MRG] Manual trigger of tests and fix tag pushes

consideRatio:
While it is possible to push on tags by just defining the `tags` keyword as well as branches by specififying the `branches` keyword, whenever one of these have a value set, the other is ignored unless it has a value as well.

> If you define only tags or only branches, the workflow won't run for events affecting the undefined Git refs

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

Due to this, we won't trigger on tags as it currently is, but this PR fixes this. This PR also adds the ability to manually trigger the test workflow which can be useful as a sanity check sometimes, for example if a PR has a failing build but you wonder if its related to the PR or if a regression in the main branch.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,7 +4,7 @@ name: Publish helm chart and docker images
 on:
   push:
     branches: [main, master]
-    tags:
+    tags: ["**"]
 
 defaults:
   run:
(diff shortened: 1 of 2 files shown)


consideRatio:
Ok to merge? I'll self merge this soon, it is a bugfix of the CI system causing pushed tags to not lead to a published chart and the ability to trigger manual runs of the tests.

consideRatio:
Thanks Tim!

PR #1222, opened 2020-12-03
[MRG] ci: bugfix of chartpress fixed

consideRatio:
Chartpress 1.0.0 contained a regression, and our publishing pipeline was using chartpress 1.0.0 instead of 0.7.0, so it was affected. This made the image version used by binderhub since chartpress 1.0.0 was released in 21 Nov 2020 be the same version as the image-cleaner image.

The only PR that was influenced by this mishap was a PR merged ~2 hours. With this commit, it should not happen again.

@sgibson91 with this merged, we should get a Henchbot PR for mybinder.org-deploy that will be correctly containing #1220 within the image as set by the binderhub helm chart.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -40,7 +40,7 @@ jobs:
         run: |
           . ./ci/common
           setup_helm v3.4.1
-          pip install --no-cache-dir chartpress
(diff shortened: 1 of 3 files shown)


sgibson91 (approved):
Great work, thanks @consideRatio! ✨

PR #1225, opened 2020-12-04
set _request_timeout on most kubernetes API requests

minrk:
avoids indefinite hangs when there are connection issues

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -12,7 +12,7 @@ from kubernetes import client, watch
 from tornado.ioloop import IOLoop
 from tornado.log import app_log
 
-from .utils import rendezvous_rank
+from .utils import rendezvous_rank, KUBE_REQUEST_TIMEOUT
 
 
 class Build:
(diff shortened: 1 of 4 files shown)


sgibson91 (approved):
LGTM! 🎉

binderhub/build.py
```diff
@@ -320,10 +325,11 @@ def submit(self):
...
                 for f in w.stream(
-                        self.api.list_namespaced_pod,
-                        self.namespace,
-                        label_selector="name={}".format(self.name),
-                        timeout_seconds=30,
+                    self.api.list_namespaced_pod,
+                    self.namespace,
+                    label_selector="name={}".format(self.name),
+                    timeout_seconds=30,
+                    _request_timeout=(3, 30),
```
consideRatio, line 332: There is a hardcoded reference here to the same timeout tuple config, and below to (3, None), should perhaps at least this be the same as the constant in utils?

minrk: yup, I missed this one after I added the constant. Fixed!

PR #1229, opened 2020-12-13
Do not mount host docker.sock into image cleaner pod if it is not enabled

rkevin-arch:
Currently if you run binderhub with dind and no host Docker daemon, the image cleaner pod still declares `dockerlib-host` and `dockersocket-host` as volumes even if `imageCleaner.host.enabled` is false, even if the host docker socket is nonexistent. This PR adds a check and only adds the host docker socket as a volume if `imageCleaner.host.enabled` is true.

The image cleaner pod actually would still work, because no containers in that pod actually use the volumes. Still, it would be nice to remove them from the pod declaration altogether.


consideRatio:
LGTM! Thanks @rkevin-arch!

PR #1233, opened 2020-12-21
binder image: jh 1.1 -> 1.2, k8s-client 9 -> 12

consideRatio:
This is a followup to the z2jh chart bump. We hesitated a bit upgrading this because we observed issues so I'll monitor the binderhub pod specifically along with this bump.


consideRatio:
Oh. I got confused, this was bumped as part of #1184 after all which i initially thought and then confused myself of it not being done.

PR #1235, opened 2021-01-06
Revert to kubernetes API client v9

betatim:
Newer versions use a lot more CPU than v9. Going back to this version to
confirm if the CPU load on mybinder.org drops again as soon as we deploy
this change.


betatim:
The CI failures are related to figshare and fixed in #1236

consideRatio:
Thanks for working to verify if this is the cause of issues or not @betatim!

PR #1236, opened 2021-01-06
Change regex to match new(?) figshare URL scheme

betatim:
It looks like figshare.com URLs now have a new layout: https://figshare.com/articles/code/Binder-ready_openSenseMap_Analysis/9782777 is what I get when I follow https://doi.org/10.6084/m9.figshare.9782777.v1.

Based on the regex we use I think the old URLs looked like https://figshare.com/articles/Binder-ready_openSenseMap_Analysis/9782777 (or so).

With this new regex we can parse the new URLs. Maybe there is a way to make this more robust to future changes but my regex-fu isn't strong enough.

Is it a good idea to switch to using `/(\d+)(/)?(\d+)?$` as regex (to get the numbers we need), a separate check that `/articles/` appears in the URL and that the hostname contains `figshare`?

### DIFF
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -249,7 +249,7 @@ class FigshareProvider(RepoProvider):
     Users must provide a spec consisting of the Figshare DOI.
     """
     name = Unicode("Figshare")
-    url_regex = re.compile(r"(.*)/articles/([^/]+)/(\d+)(/)?(\d+)?")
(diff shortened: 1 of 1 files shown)


consideRatio:
Thanks for solving this!

If you have time it would be nice to have a docstrings or inline comments that:
- examplifies the URLs the regex will work against
- warns about that this pattern has evolved over time

I have not learned what figshare is at this point so I didn't dare try to solve it when I saw this error first.

Issue #1237, opened 2021-01-06
ci system following two merged PRs get stuck


consideRatio:
I merged #1236 and #1235 which both seemed fine to me, but now our tests fail in a new way. Did the combination of these PRs introduce a new bug? I'm clueless.

Test failure: https://github.com/jupyterhub/binderhub/actions/runs/466459518

![image](https://user-images.githubusercontent.com/3837114/103787603-11195780-503e-11eb-9c38-f14bb1807a14.png)

---

From the namespace report, it seems like the binder pod became ready and running though...

```
NAME                             READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/proxy            1/1     1            1           10m
deployment.apps/user-scheduler   2/2     2            2           10m
deployment.apps/hub              1/1     1            1           10m
deployment.apps/binder           1/1     1            1           10m
```

The actual test seem to be...

```
  . ci/common
  await_binderhub
  echo "curl http://localhost:30901/health" to check BinderHub\'s health
  curl http://localhost:30901/health
```

And the initial output of this test is...
```
deployment "proxy" successfully rolled out
deployment "hub" successfully rolled out
Waiting for deployment "binder" rollout to finish: 0 of 1 updated replicas are available...
deployment "binder" successfully rolled out
curl http://localhost:30901/health to check BinderHub's health
<THIS IS WHERE IT ALL GETS STUCK: in the "curl http://localhost:30901/health" statement>
```

It seems to me that the /health endpoint now fail to respond or perhaps streams a response or similar that is never completed?

consideRatio:
Here is a test run from #1238 that provides more details about the state of things associated with new failure. Nothing seems wrong to me, but I'm not sure what causes the curl request to get stuck.

https://github.com/jupyterhub/binderhub/pull/1238/checks?check_run_id=1657413955

betatim:
I think it has happened before that the `curl` call gets stuck. Restarting the CI run seems to fix it. So I think this is a flakey test/setup kind of issue :-/

consideRatio:
No because the echoed statement is executes

manics:
Is there a potential race condition? Could the check in
https://github.com/jupyterhub/binderhub/blob/f76e37e334e49749bc883eade8ac63da70e4abdb/ci/common#L22
succeed before `/health` and/or the network routing is ready?

consideRatio:
[...]

PR #1238, opened 2021-01-06
ci: sync full_namespace_report from z2jh

consideRatio:
Updates to the full_namespace_report to provide more details about the k8s cluster, which will help us debug test failures etc.

This is an example of how it can look following the update. In short, it is smart to describe and print logs of pods with issues, and it also always prints the logs of certain pods we declare important, such as deploy/binder, deploy/hub, deploy/proxy.

full-namespace-report


consideRatio:
ci only change, self-merging to use it in to debug #1237.

PR #1242, opened 2021-01-06
[MRG] helm chart: add readinessProbe and let probes be configurable

### LINKED ISSUE #1237: ci system following two merged PRs get stuck

consideRatio:
My diagnosis of #1237 is that for whatever reason, the ci test fails because it is `curl`ing the /version endpoint before it is ready. Further, from testing in various dummy PRs, that this is an outcome influenced quite reliably based on if the binder image has been recently rebuilt or not.

When the binder image has been rebuilt recently, it means that it will be available without a download step that risks queing behind other image downloads etc, which makes pods startup in sequence. When pods startup in sequence, its plausible that the binder pod starts late and won't get a chance to startup properly.

With this PR, we introduce a readinessProbe that will never make the pod unready before the livenessProbe triggers a restart. Due to this, the purpose becomes solely to ensure that the pod is started properly before accepting network traffic.

Benefits:
- Provides a clear criteria to know if it has properly started up useful in CI system, closes #1237.
- Helps when multiple replicas are used to ensure traffic isn't redirected to a recently started byt not yet responsive binder pod.

Drawbacks:
- Can clutter logs, _but only_ if we use debug logging because of an exception to /health and /version is made to only log successful requests if debug mode is enabled.
- Can delay readiness with up to 5 seconds because `periodSeconds: 5` of actual readiness. This value is a compromise of not cluttering logs and not waiting too long for readiness on startup.

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -145,6 +145,18 @@ jupyterhub:
       enabled: false
 
 deployment:
+  readinessProbe:
+    enabled: true
+    initialDelaySeconds: 0
+    periodSeconds: 5
+    failureThreshold: 1000  # we rely on the liveness probe to resolve issues if needed
+    timeoutSeconds: 3
+  livenessProbe:
+    enabled: true
+    initialDelaySeconds: 10
+    periodSeconds: 5
+    failureThreshold: 3
+    timeoutSeconds: 10
   labels: {}
 
 dind:
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -121,10 +121,23 @@ spec:
         ports:
           - containerPort: 8585
             name: binder
+        {{- if .Values.deployment.readinessProbe.enabled }}
+        readinessProbe:
+          httpGet:
+            path: {{ .Values.config.BinderHub.base_url | default "/" }}versions
+            port: binder
+          initialDelaySeconds: {{ .Values.deployment.readinessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.deployment.readinessProbe.periodSeconds }}
+          timeoutSeconds: {{ .Values.deployment.readinessProbe.timeoutSeconds }}
+          failureThreshold: {{ .Values.deployment.readinessProbe.failureThreshold }}
+        {{- end }}
+        {{- if .Values.deployment.livenessProbe.enabled }}
         livenessProbe:
           httpGet:
             path: {{ .Values.config.BinderHub.base_url | default "/" }}versions
             port: binder
-          initialDelaySeconds: 10
-          periodSeconds: 5
-          timeoutSeconds: 10
+          initialDelaySeconds: {{ .Values.deployment.livenessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.deployment.livenessProbe.periodSeconds }}
+          timeoutSeconds: {{ .Values.deployment.livenessProbe.timeoutSeconds }}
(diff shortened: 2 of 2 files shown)


helm-chart/binderhub/values.yaml
```diff
@@ -145,6 +145,18 @@ jupyterhub:
       enabled: false
 
 deployment:
+  readinessProbe:
+    enabled: true
+    initialDelaySeconds: 0
+    periodSeconds: 5
+    failureThreshold: 1000  # we rely on the liveness probe to resolve issues if needed
```
betatim, line 152: For my education: what does the comment mean/try to say? Why not use a startup probe which seems to be made for exactly our case of "pod can take time to start up"?

I think setting the failure threshold to 1000 means that the readiness probe will never(*) become "not ready" again after it becomes ready for the first time.

Why not use an initial delay similar to the liveness probe to cover the expected time the process needs to start up (I think the initial delay only starts counting after the container image has been pulled but couldn't confirm that).

(*) in reality it would become "not ready" if the probe fails for ~5000s, so about 1.3hours

consideRatio: __readinessProbe__
- Required to make a pod Ready.
- A failure of this -> the container is becoming Unready
- Ready is required for the pod to receive traffic from k8s Services
- Ready is required for `kubectl rollout status --wait` checks of a deployment resource.
- A readinessProbe with a very long failureThreshold makes it act as a delay to become Ready during startup, but will then always remain Ready.
- To let a readinessProbe fail following having become ready, the purpose would be to temporarily remove it from receiving traffic from k8s Services, perhaps to recover or similar, but this is only useful in specific situations I don't think is applicable to our Binder pod.

__livenessProbe__
- A failure of this -> the container is getting restarted

__startupProbe__
- k8s 1.16+ required
- doesn't influence the pod's Ready status
- serves the purpose only to complement the livenessProbe for slow startups, see [this docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes)

__Meaning of comment?__
I meant to convey that we rely on the livenessProbe's functionality to recover from bad states, not by simply making the pod Unready which can be unhelpful.

__Reference__
startupProbe and such is discussed also in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1732#issuecomment-704288546

consideRatio:
Merging to close #1237 and test if the updated henchbot (https://github.com/henchbot/mybinder.org-upgrades/pull/5) works as it should.

Issue #793, opened 2019-02-18
font files are not served under base url


bitnik:
In index.css there are 3 font files loaded with relative path:
https://github.com/jupyterhub/binderhub/blob/74b3d69c1c37a2ca05cff816b81b68ba515ac0ae/binderhub/static/index.css#L3-L16
But for both mybinder (base url is '/') and GESIS binder (base url is '/binder/') they are loaded from `/static/dist/...` in styles.css:
- https://mybinder.org/static/dist/styles.css
- https://notebooks.gesis.org/binder/static/dist/styles.css

I don't know if this is because how they are loaded in index.css or because of [webpack config](https://github.com/jupyterhub/binderhub/blob/master/webpack.config.js).

Actually this is not a problem for GESIS binder. I just wanted to report the issue.

betatim:
I think webpack rewrites the URLs when it creates `styles.css` from the input CSS files.

I am not sure if there is a problem with how things currently work or if this issue is about improving the documentation (I'd welcome that :) ) or something else?

bitnik:
But isn't it a problem that they are always served under `/static/dist/...` without considering the base url of the application? If base url is different than '/', those file won't be found. Maybe I am totally wrong :) I really don't know much about css files and webpack.

betatim:
Ah! I think I agree with you that who ever re-writes that path to be `/static/dist/...` needs to take care to also include the base URL. I read your comment about `notebooks.gesis.org/binder/static/dist/styles.css` as "the files are being loaded from that URL, the fonts are somehow included in the CSS (by magic of webpack) and it includes the base URL and for mybinder.org it doesn't so all is as expected".

Now that I actually looked at the two CSS files you linked I see the problem. (Maybe I should have had a second coffee for breakfast.) -> We need to communicate the base URL to webpack. Maybe @captainsafia or @yuvipanda can help out with advice on how to do that best.

PR #1247, opened 2021-01-22
fix URLs of files in styles.css for binderhubs that are not served at "/"

bitnik:
fixes https://github.com/jupyterhub/binderhub/issues/793

Overrides the publicPath for the loader used in `ExtractTextPlugin`. By default it uses the publicPath of the application ("/static/dist/"), when it rewrites the URLs in styles.css. But this makes these files available only if BinderHub is served under "/" (base_url is "/"), and not available if BinderHub has a different base_url, e.g. "/binder/". So if we set publicPath relative to root ("./"), these files will be always available.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -15,7 +15,11 @@ module.exports = {
                 test: /\.css$/,
                 use: ExtractTextPlugin.extract({
                     fallback: "style-loader",
-                    use: "css-loader"
+                    use: "css-loader",
+                    // Set publicPath as relative path ("./").
+                    // By default it uses the `output.publicPath` ("/static/dist/"), when it rewrites the URLs in styles.css.
+                    // And it causes these files unavailabe if BinderHub has a different base_url than "/".
+                    publicPath: "./"
                 })
             },
             {


consideRatio:
This LGTM as far as I can tell! :heart: for the inline comment!

bitnik:
Yes, I tested it locally.

> Would a test capture if this caused a UI error? 

I dont think so. Before this change, it caused only that font files were not found and it used the fonts from bootstrap.

betatim:
@consideRatio there hasn't been a henchbot PR to deploy this. Could you investigate if this is related to the change in https://github.com/jupyterhub/mybinder.org-deploy/pull/1753? We needed one change in henchbot already but maybe there is more?

consideRatio:
Oh no :(

I have no logs or similar to debug it as far as I know so I think we require help from @henchc. I opened https://github.com/henchbot/mybinder.org-upgrades/issues/6 about running the henchbot in a GitHub workflow also to help with providing information and such about its execution.

betatim:
While we wait for moving somewhere else I think you should be able to run henchbot locally somehow to debug this

PR #1248, opened 2021-01-22
upgrade webpack and all other JS packages

bitnik:
This PR upgrades all JS packages to the latest version.

I had similar problem what @betatim had at https://github.com/jupyterhub/binderhub/pull/787; while initializing the terminal, it complained that parent elements are hidden and same happened while calling fit addon. I will point my solution, which I am not happy about, inline the code. ~~I also upgraded `event-source-polyfill` package, but to be honest I don't know what it exactly does.~~

If we mange to merge this, my next attempt will be to separate bundles for index and loading pages.

Pinging @yuvipanda :)


consideRatio:
@bitnik wiee! I tried upgrading things before but failed as bumping broke things that I failed to restore.

binderhub/static/js/index.js
```diff
@@ -231,19 +228,35 @@ function build(providerSpec, log, path, pathType) {
...
+    if (!$panelBody.hasClass('hidden')) {
+      fitAddon.fit();
+    }
+  };
 
-  log.open(document.getElementById('log'), false);
+  // https://xtermjs.org/docs/api/terminal/classes/terminal/#open says that
+  // `parent` must be visible (have dimensions) when open is called 
+  // as several DOM- based measurements need to be performed when this function is called.
+  $panelBody.removeClass('hidden');
```
bitnik, line 249: Because `open` method needs the parent elements to be visible, here I made them visible, then called the open method and then made them hidden again. How can we improve this?

binderhub/static/js/index.js
```diff
@@ -231,19 +228,35 @@ function build(providerSpec, log, path, pathType) {
...
     convertEol: true,
     disableStdin: true
   });
+  const fitAddon = new FitAddon();
+  log.loadAddon(fitAddon);
+
+  var $logContainer = $("div#log-container");
+  var $panelBody = $("div.panel-body");
+  log.fit = function () {
+    if (!$panelBody.hasClass('hidden')) {
```
bitnik, line 241: To prevent errors, fit is called only if parent is visible. Because fit is called every time window is resized (see [line 256](https://github.com/jupyterhub/binderhub/pull/1248/files#diff-ecd046a09fce6b476b2a7e29adaae5537510474c0431c720b4a973d38e51b595R256-R258) and this may happen when panel is hidden.

package.json
```diff
@@ -4,15 +4,15 @@
   "description": "`BinderHub`",
   "devDependencies": {
     "bootstrap": "^3.4.0",
-    "clipboard": "^1.7.1",
-    "css-loader": "^0.28.7",
-    "event-source-polyfill": "^0.0.12",
-    "extract-text-webpack-plugin": "^3.0.2",
```
bitnik, line 10: "extract-text-webpack-plugin" is deprecated (https://github.com/webpack-contrib/extract-text-webpack-plugin)

package.json
```diff
@@ -4,15 +4,15 @@
...
-    "file-loader": "^1.1.6",
```
bitnik, line 11: "file-loader"  is also deprecated (see https://webpack.js.org/loaders/file-loader/). [asset/resource](https://webpack.js.org/guides/asset-modules/) is used instead.

betatim:
What do you think of always having the logs open as a way to get around the "fit" problem? The discussion that followed https://github.com/jupyterhub/binderhub/pull/787#issuecomment-469671808 sounded like there was support for doing that (both as a feature and as a way to avoid difficulties with fit)

bitnik:
hmm I like the compact view but I think I am slightly more for having it by default as well.

Btw maybe then we should also rename "Build Logs" to "Logs" as it also shows launch logs or errors.

bitnik:
I thought there will be inputs from others, so decided to wait for doing the changes, then didn't have time to come back to this. Anyways, my latest commit (https://github.com/jupyterhub/binderhub/pull/1248/commits/31372aeb4e949e70f6b143bde96d4cd18a947f21) makes the log panel open by default but it can be still toggled by clicking on the bar anytime. Do you prefer to remove toggling and have it always open?

manics:
I think it's fine to have the logs open by default.

My first thought is not to bother with the added complexity of making it hidable, though are there advantages to doing so that I haven't thought of?

bitnik:
I can't think of any neither. I just made a commit to have it always open. I have done the changes gradually, so if we want to change back to any previous state, we should be able to do it easily.

yuvipanda:
Wanna try dust this off and get it merged, @bitnik? I can provide review this time :) 

Other than the merge conflicts, I'd really like us to keep the log window closed by default. If not, it's just a big black box that takes up a lot of space and attention the moment we come to the page...

yuvipanda:
I've separated this out into multiple smaller PRs:

- https://github.com/jupyterhub/binderhub/pull/1362
- https://github.com/jupyterhub/binderhub/pull/1365

bitnik:
sorry, I couldn't react on time, unfortunately I dont have much time these days. And I am happy that you take this over :)

yuvipanda:
@bitnik thank you so much for doing the actual hard work :) I hope you're enjoying whatever it is that you are doing now! <3

yuvipanda:
https://github.com/jupyterhub/binderhub/pull/1371 is the last PR split from this.

PR #1249, opened 2021-01-23
Pin chartpress to 1.0.3

betatim:
A new version of chartpress was released and now our chart publishing job thinks that it does not need to publish anything. More details in https://github.com/henchbot/mybinder.org-upgrades/issues/7#issuecomment-766126450

Let's merge this and see if this brings the chart publishing back to life. If not we know it wasn't the chartpress release, if yes we can use the old version of chartpress until there is a fix.

It might make sense to keep the chartpress version pinned for a "long time" and rely on explicit upgrades of it so that we don't have to investigate why our machinery has started failing. In particular as there were no problems with the old chartpress version from a BinderHub point of view.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,7 +30,7 @@ jobs:
         run: |
           . ./ci/common
           setup_helm v3.4.1
-          pip install --no-cache-dir chartpress==1.0.*
+          pip install --no-cache-dir chartpress==1.0.3
 
       - name: Setup push rights to jupyterhub/helm-chart
         # This was setup by...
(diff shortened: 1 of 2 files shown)


consideRatio:
@betatim the issue is the chartpress.yaml config of binderhub isn't correct I think, it did not consider a change of the webpack.config file to be relevant in order to trigger a version bump.

I'd like to have this unpinned still because if there is a time when we want a new feature, we suddenly must consider why this was pinned etc. And at this point, it was a configuration error unrelated to the version bump afaik.

betatim:
If #1250 fixes it then that is good and we don't need to pin chartpress.

betatim:
#1250 fixed it

PR #1251, opened 2021-01-23
[MRG] Improve contribution document

betatim:
This adds back some sections that got lost over time and fixes typos, broken links, etc.

I followed all the individual guides to check that they work, except for the "helm chart development" one.

PR #1252, opened 2021-01-23
[MRG] In unit tests close client connection once 'ready' message arrives

betatim:
In our tests when we pretend to be a web browser we were waiting for
BinderHub to close the network connection while streaming the response.
Because BinderHub waits "a long time" after sending the "ready" message
this was slowing down the tests, as they sat around twiddling their
thumbs.

This is a follow up/fix for behaviour observed in https://github.com/jupyterhub/binderhub/pull/1251#issuecomment-766155383

I think this means we don't explicitly test that the last message the hub sends is the "ready" message as we close the stream when we get it. So if other messages are sent (by accident) we wouldn't know. I think that is fine or could be tested explicitly in a different test. I can't think of a case were a "ready" message is sent in a successful build, which would lead us astray here.

-> nice optimisation to make the tests a bit quicker.

### DIFF
--- a/binderhub/tests/test_build.py
+++ b/binderhub/tests/test_build.py
@@ -52,6 +52,13 @@ async def test_build(app, needs_build, needs_launch, always_build, slug, pytestc
             events.append(event)
             assert 'message' in event
             sys.stdout.write(event['message'])
+            # this is the signal that everything is ready, pod is launched
+            # and server is up inside the pod. Break out of the loop now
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
LGTM, nice catch!

consideRatio:
OMG! YES! The time things took dropped EXTENSIVELY! I got very frustrated about the time it took to run these tests before.

I'm confused that this triggered a image rebuild though. Hm...

image

PR #1254, opened 2021-01-26
Update binder API example to include file/urlpath

MridulS:
Now it's possible to use the example API to directly launch binder repos with urlpath/filepath.

An example tested
```
python binder-api.py voila-dashboards/voila --ref stable --urlpath voila/render/notebooks/bqplot.ipynb
```

### DIFF
--- a/examples/binder-api.py
+++ b/examples/binder-api.py
@@ -67,7 +67,7 @@ if __name__ == '__main__':
             elif opts.urlpath:
                 url = "{url}{urlpath}?token={token}".format(**evt, urlpath=opts.urlpath)
             else:
-                url = "{url}notebooks/?token={token}".format(**evt)
+                url = "{url}?token={token}".format(**evt)
             print("Opening %s" % url)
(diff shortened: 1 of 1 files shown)


examples/binder-api.py
```diff
@@ -53,15 +55,19 @@ def build_binder(repo,
...
                 message=evt['message'].rstrip(),
             ))
         if evt.get('phase') == 'ready':
-            url = "{url}?token={token}".format(**evt)
+            if opts.filepath:
+                url = "{url}notebooks/{filepath}?token={token}".format(**evt, filepath=opts.filepath)
+            elif opts.urlpath:
+                url = "{url}{urlpath}?token={token}".format(**evt, urlpath=opts.urlpath)
+            else:
+                url = "{url}notebooks/?token={token}".format(**evt)
```
minrk, line 70:
```suggestion
                url = "{url}?token={token}".format(**evt)
```

minrk:
Great!

PR #1255, opened 2021-01-27
build_docker_config added, enables augmentation of the build pod's docker config

g-braeunlich:
## PR Summary by Erik

The BinderHub configuration `build_docker_config` is added to allow an augmentation of the build pods docker config (`config.json`) that previously have only contained container registry credentials when `use_registry` was set to to True.

A motivation for adding this configuration option is to enable the ability to configure network proxy settings for docker, so for example traffic going to download an apt package during the build process is done through a proxy. Such configuration is documented [here](https://docs.docker.com/network/proxy/) for example.

This PR does not introduce any breaking changes, but it does rename a k8s Secret (managed directly by the Helm chart) from `binder-push-secret` to `binder-build-docker-config`. By doing so, a build pod starting up in the wrong moment could get stuck in pending by referencing the old name but not being able to mount it. If a pod is already running everything will be fine though, the content mounted won't be deleted just because the mounted k8s Secret was deleted.

I think this PR makes sense, is well scoped, and while could cause a disruption during a very brief moment in time for new builds does the right thing to rename the k8s Secret.

---

## Original PR summary

This is the build pod part of the PR  #990
It will allow to build images with repo2docker behind a proxy.
Prerequisite:
https://github.com/jupyterhub/repo2docker/pull/1003#pullrequestreview-574870630

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -297,7 +297,7 @@ class BinderHub(Application):
     )
 
     push_secret = Unicode(
-        'binder-push-secret',
+        'binder-build-docker-config',
         allow_none=True,
         help="""
         A kubernetes secret object that provides credentials for pushing built images.
@@ -374,6 +374,22 @@ class BinderHub(Application):
         if parts.scheme != 'unix' or parts.netloc != '':
             raise TraitError("Only unix domain sockets on same node are supported for build_docker_host")
         return proposal.value
+
+    build_docker_config = Dict(
+        None,
+        allow_none=True,
+        help="""
+        A dict which will be merged into the .docker/config.json of the build container (repo2docker)
+        Here, you could for example pass proxy settings as described here:
+        https://docs.docker.com/network/proxy/#configure-the-docker-client
+
+        Note: if you provide your own push_secret, this values wont
+        have an effect, as the push_secrets will overwrite
+        .docker/config.json
+        In this case, make sure that you include your config in your push_secret
+        """,
+        config=True
+    )
 
     hub_api_token = Unicode(
         help="""API token for talking to the JupyterHub API""",
@@ -693,6 +709,7 @@ class BinderHub(Application):
                 "build_memory_limit": self.build_memory_limit,
                 "build_memory_request": self.build_memory_request,
                 "build_docker_host": self.build_docker_host,
+                "build_docker_config": self.build_docker_config,
                 "base_url": self.base_url,
                 "badge_base_url": self.badge_base_url,
                 "static_path": os.path.join(HERE, "static"),
--- a/helm-chart/binderhub/templates/_helpers.tpl
+++ b/helm-chart/binderhub/templates/_helpers.tpl
@@ -16,9 +16,9 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 
 {{/*
-Render docker config.json for the registry-publishing secret.
+Render docker config.json for the registry-publishing secret and other docker configuration.
 */}}
-{{- define "registryDockerConfig" -}}
+{{- define "buildDockerConfig" -}}
 
 {{- /* default auth url */ -}}
 {{- $url := (default "https://index.docker.io/v1" .Values.registry.url) }}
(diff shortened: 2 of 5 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/binder-behind-outbound-proxy/7428/5

g-braeunlich:
Just found out:
if `push_secret` is set, then this will mount a read only secret over `.docker/config.json`
This will prevent the proxy functionality

manics:
Looks liek these are the relevant lines:
https://github.com/jupyterhub/binderhub/blob/e2f23e90f8e9e805262f1d201d0dcded730e80e4/binderhub/build.py#L248-L253
https://github.com/jupyterhub/binderhub/blob/7cedaa7e3089525fd1eb94faa08ad978e68831f8/helm-chart/binderhub/templates/secret.yaml#L19-L27

g-braeunlich:
Exactly. What do you think? Should I try to include the proxy vars into the secret in the `registryDockerConfig` helper?

manics:
I'm not sure. Adding the proxy config to `registryDockerConfig` is an easy solution. On the other hand the name of the parameter `registryDockerConfig` is no longer correct, and it'd be nice to separate the configuration (proxy vars, registry credentials) from the implementation in repo2docker (read those vars, create the required config file). Anyone else have thoughts?

g-braeunlich:
@manics I just pushed a proposal

helm-chart/binderhub/templates/_helpers.tpl
```diff
@@ -36,11 +36,11 @@ Render docker config.json for the registry-publishing secret.
...
-    }
-  }
-}
+{{- /* prepare a dict containing the registry credentials */ -}}
+{{- $dockerConfig := dict "auths" (dict $url (dict "auth" (printf "%s:%s" $username .Values.registry.password | b64enc))) }}
+{{- if .Values.config.BinderHub.buildDockerConfig }}
+{{- $dockerConfig := merge $dockerConfig $dockerConfig .Values.config.BinderHub.buildDockerConfig }}
+{{- end }}
+
+{{ $dockerConfig | toPrettyJson }}
```
consideRatio, line 45: I wanted to verify that this template renders what we already did, and it does except for some blank lines. This suggestion makes everything render like before.

```suggestion
{{- /* initialize a dict to represent a docker config with registry credentials */}}
{{- $dockerConfig := dict "auths" (dict $url (dict "auth" (printf "%s:%s" $username .Values.registry.password | b64enc))) }}

{{- if .Values.config.BinderHub.buildDockerConfig }}
{{- $dockerConfig := merge $dockerConfig $dockerConfig .Values.config.BinderHub.buildDockerConfig }}
{{- end }}

{{- $dockerConfig | toPrettyJson }}
```

consideRatio (approved):
I made a code suggestion refactoring some helm template logic that I'd like to see accepted, but otherwise this LGTM and feel good about merging this!

consideRatio:
@betatim @sgibson91 @manics @MridulS this was the PR brought  for discussion during our team meet.

I suggest we go for a merge and updated the description to summarize the situation as i see it.

binderhub/app.py
```diff
@@ -297,7 +297,7 @@ def _valid_badge_base_url(self, proposal):
     )
 
     push_secret = Unicode(
-        'binder-push-secret',
+        'binder-build-docker-config',
```
manics, line 300: Is this change of default a breaking change?

consideRatio: I discussed it in the PR summary i updated above. Yes it can in theory cause a new build pod get stuck pending if started in a timeframe of a second for 1 replica binder deployments or a 10 second window for 2 replica deployments. I consider this acceptable.

If chart is upgraded, the old secret us gone, and an old binderhub replica awaiting termination until another binder pod starts up during a rolling upgrade, would starts a new build pod - then it would get stuck pending not being able to mount old k8s secret.

binderhub/app.py
```diff
@@ -375,6 +375,22 @@ def docker_build_host_validate(self, proposal):
             raise TraitError("Only unix domain sockets on same node are supported for build_docker_host")
         return proposal.value
 
+    build_docker_config = Dict(
```
manics, line 378: Hello from the distant future! I'm reviewing the BinderHub config properties as part of https://github.com/jupyterhub/binderhub/pull/1521 and as far as I can tell this property isn't used. It's passed to `builder.py`, and referenced in a conditional, but its value is never used. Instead the helm chart references `.Values.config.BinderHub.buildDockerConfig` in `helm-chart/binderhub/templates/_helpers.tpl`.

Can you remember what's happening here?

PR #1258, opened 2021-02-05
bump jupyterhub chart to 0.11.1

minrk:
gets kubespawner 0.15, jupyterhub 1.3

adds hub.config passthrough

changes auth config, shouldn't be much else backward-incompatible

@consideRatio any other expected changes I missed that you think might affect binderhub?


consideRatio:
No I think we are good to go, for reference [here is the changelog](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/CHANGELOG.md).

consideRatio (commented):
LGTM!

PR #1259, opened 2021-02-08
fix: location of hub config should be hub.config

consideRatio:
Followup to #1258 that had typo with regards to where the correct configuration was nested within values.yaml.

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -63,6 +63,9 @@ jupyterhub:
   rbac:
     enabled: true
   hub:
+    config:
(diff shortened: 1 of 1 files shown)


helm-chart/binderhub/values.yaml
```diff
@@ -54,6 +54,9 @@ cors: &cors
   allowOrigin:
 
 jupyterhub:
+  config:
+    JupyterHub:
+      authenticator_class: nullauthenticator.NullAuthenticator
```
consideRatio, line 59: WRONG AGAIN! :D

minrk (approved):

PR #1262, opened 2021-02-10
add BinderHub.ban_networks

minrk:
to ban requests from certain sources, e.g. AWS

health endpoint is not included, for reasons covered in https://github.com/jupyterhub/mybinder.org-deploy/issues/1828

We could also block only builds/launches, rather everything but health check. That would save on cost.

see https://github.com/jupyterhub/mybinder.org-deploy/issues/1828 for reasoning about doing this at the application level instead of a NetworkPolicy

I'm still exploring the performance costs of this and possible optimizations using netaddr when we have thousands of networks to check.

### DIFF
--- a/binderhub/base.py
+++ b/binderhub/base.py
@@ -1,9 +1,11 @@
 """Base classes for request handlers"""
 
 import json
+from ipaddress import ip_address
 
 from http.client import responses
 from tornado import web
+from tornado.log import app_log
 from jupyterhub.services.auth import HubOAuthenticated, HubOAuth
 
 from . import __version__ as binder_version
@@ -16,6 +18,23 @@ class BaseHandler(HubOAuthenticated, web.RequestHandler):
         super().initialize()
         if self.settings['auth_enabled']:
             self.hub_auth = HubOAuth.instance(config=self.settings['traitlets_config'])
+
+    def prepare(self):
+        super().prepare()
+        # check request ips early on all handlers
+        self.check_request_ip()
+
+    skip_check_request_ip = False
+
+    def check_request_ip(self):
+        """Check network block list, if any"""
+        if self.skip_check_request_ip or not self.settings.get("ban_networks"):
+            return
+        request_ip = ip_address(self.request.remote_ip)
+        for network in self.settings["ban_networks"]:
+            if request_ip in network:
+                app_log.warning(f"Blocking request from {request_ip} in {network}")
+                raise web.HTTPError(403, f"Requests from {request_ip} are not allowed")
 
     def get_current_user(self):
         if not self.settings['auth_enabled']:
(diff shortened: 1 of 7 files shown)


binderhub/base.py
```diff
@@ -17,6 +19,23 @@ def initialize(self):
...
+        self.check_request_ip()
+
+    skip_check_request_ip = False
+
+    def check_request_ip(self):
+        """Check network block list, if any"""
+        if self.skip_check_request_ip or not self.settings.get("ban_networks"):
+            return
+        request_ip = ip_address(self.request.remote_ip)
+        for network in self.settings["ban_networks"]:
```
manics, line 34: If performance is too bad I found a nice trick in https://codereview.stackexchange.com/a/183221
Order the list of `ban_networks` by the integer representation of the first IP, then you can easily do a search to find the relevant one for `request_ip`.

minrk: netaddr.IPSet [has an O(1) implementation](https://netaddr.readthedocs.io/en/latest/_modules/netaddr/ip/sets.html#IPSet.__contains__), assuming you can preconstruct a set (we can). I'm looking at comparing a version using the stdlib vs pulling in netaddr as a dependency.

minrk:
Added the O(1) test based on netaddr, but without pulling in the dependency. It's a fixed cost at ~100-150µs.

Unnecessarily detailed analysis [here](https://gist.github.com/c1a81f415e5b4dea17d08a27b6ca9a95)

minrk:
The only thing I see as left to discuss is that the configuration is a Set of CIDRs. Having a single flat config value is not the best for https://github.com/jupyterhub/mybinder.org-deploy/issues/1828 where we would ideally merge multiple sources of config (i.e. the short manual ban list and the generated datacenter collections). These are easy to merge in the container, or even from multiple *Python* config files (traitlets.config supports `c.BinderHub.ban_networks.update({'a', ...})`), but can't be done in simple yaml/json.

I see the main choice being:

- accept dict of `cidr: message` where message can be a truthy string. Dicts are merged in things like helm values files, so more naturally fit in the multi-config-source. This also gives us a handle to use for custom messages about why each ip block is banned (do we want this?)
- handle the multi-source merge in mybinder.org-deploy customization, and leave it as a simple set here. This can either be a Python snippet to load additional data mounted in an extra volume, or a flat list in config with something like an `# insert after here` anchor

manics:
LGTM! Took me a while to figure out `ip_in_network_set` but it makes perfect sense, given an IP there's only one canonical CIDR for each prefix from 1-32.

For configuration I like the sound of option 1, dict of `cidr: message`. As well as allowing multiple configs to be merged it also allows future extensibility. For example if `message` is the datacentre or rule name this could be logged and grepped to get stats to see how often IPs are blocked from each source. It might even be worth making message a dict to support a detailed log message as well as a user-friendly message, but that can be left for later.

minrk:
@manics sounds good! ban_networks is now a dict of `cidr: message`

manics (approved):
Looks great!

PR #1266, opened 2021-02-18
builder: set default headers in set_default_headers

minrk:
avoids wrong content-type in errors raised during prepare/authenticate

which would lead to "failed to connect to event stream"
due to:

> EventSource's response has a MIME type ("text/html") that is not "text/event-stream".
> Aborting the connection.

in the browser console

### DIFF
--- a/binderhub/tests/test_auth.py
+++ b/binderhub/tests/test_auth.py
@@ -1,6 +1,7 @@
 """Test authentication"""
 
 import ipaddress
+import json
 from unittest import mock
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
LGTM!

minrk:
In particular, blocked requests due to #1262 would produce these errors. @consideRatio added a test since you looked at it.

binderhub/tests/test_auth.py
```diff
@@ -94,6 +101,22 @@ def reset():
...
         assert r.status_code in status
 
+    ban_message = "Requests from local are not allowed"
     if status == 403:
         # check error message on 403
-        assert "Requests from local are not allowed" in r.text
+        assert ban_message in r.text
+
+    if banned and path.startswith("/build"):
+        # /build/ is event-stream, so allow connecting with200
```
betatim, line 110:
```suggestion
        # /build/ is event-stream, so allow connecting with status 200
```
i think that is the missing word here

minrk:
Thanks for the review! With this one, I think we can deploy https://github.com/jupyterhub/mybinder.org-deploy/pull/1829 to start blocking datacenters on mybinder.org.

PR #1269, opened 2021-02-19
docker registry: handle subdirectories in custom registries

### LINKED ISSUE #1267: Binderhub always tries to build ignoring any prebuilds
Using my private gitlab repository binderhub will always clone and build an image even if one already exists in the docker registry. It'll even finish the build and then after pushing say the image is already built. Is this known fixable behaviour?

Thanks

ctr26:
Fixed image name splitting in manifest for this case:
"dockerhub.ebi.ac.uk/bioimagearchive/k8s-jupyterhub/binder-ebibiostudies-2dbiostudies-2dnotebooks-c728a8:7f8fe051826f19933283b7a4dfe975b62a1e17ba"

Tested on:

"registry.hub.docker.com/nvidia/cuda:11.2.1-devel-ubuntu20.04"
----
Background:

Binderhub can pull my images and push them to our gitlab docker registry, but it can't find manifest files, I think this patch fixes this and doesn't break every other container repo

closes #1267

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -306,7 +306,7 @@ class BuildHandler(BaseHandler):
         if self.settings['use_registry']:
             for _ in range(3):
                 try:
-                    image_manifest = await self.registry.get_image_manifest(*'/'.join(image_name.split('/')[-2:]).split(':', 1))
+                    image_manifest = await self.registry.get_image_manifest(*image_name.split('/',1)[-1].split(':', 1))
                     image_found = bool(image_manifest)
                     break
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

ctr26:
I should note that this PR solves this problem better and more completely than mine:

https://github.com/jupyterhub/binderhub/pull/986

minrk:
@ctr26 thanks. Note that we've reverted this because it broke several mybinder.org federation members. Let's try to revive #986

PR #1272, opened 2021-03-01
skip ip check in VersionHandler

minrk:
matches health handler

this is also used in the mybinder.org federation redirector


sgibson91 (approved):
LGTM 🚀

PR #1273, opened 2021-03-10
avoid preload_content when checking quotas

minrk:
this request lists all user pods and can take a long time when there are a lot of users

this should dramatically reduce cpu load with lots of user pods, especially with more recent kubernetes client versions

xref: https://github.com/jupyterhub/mybinder.org-deploy/pull/1859#issuecomment-795023498

I didn't add it to *all* kube api requests, though doing so would probably save some cpu. This is the main one that I expect to cost a lot.

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -2,6 +2,7 @@
 Handlers for working with version control services (i.e. GitHub) for builds.
 """
 
+import asyncio
 import hashlib
 from http.client import responses
 import json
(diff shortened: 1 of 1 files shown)


binderhub/builder.py
```diff
@@ -485,19 +485,16 @@ async def launch(self, kube, provider):
...
+            _preload_content=False,
         )
-        # concurrent.futures.Future isn't awaitable
-        # wrap in tornado Future
-        # tornado 5 will have `.run_in_executor`
-        tf = Future()
-        chain_future(f, tf)
-        pods = await tf
-        for pod in pods.items:
+        resp = await asyncio.wrap_future(f)
```
minrk, line 490: Now that we're native asyncio, this can be simpler

PR #1274, opened 2021-03-10
[MRG] Stop preloading k8s API response when listing all DIND pods

betatim:
This gets called every time a new build is started, if (as mybinder.org has) "sticky builds" are enabled. If this isn't enough we could also think about caching the response as the names of the nodes with DIND pods in the cluster probably doesn't change all that often (cache the response for 5min or so)

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -183,14 +183,16 @@ class Build:
         repository prefer to schedule on the same node in order to reuse the
         docker layer cache of previous builds.
         """
-        dind_pods = self.api.list_namespaced_pod(
+        resp = self.api.list_namespaced_pod(
             self.namespace,
             label_selector="component=dind,app=binder",
             _request_timeout=KUBE_REQUEST_TIMEOUT,
+            _preload_content=False,
         )
+        dind_pods = json.loads(resp.read())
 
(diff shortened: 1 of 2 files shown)


minrk:
`test_sticky_builds_affinity` needs to update its mocks to produce the no-preload response. Something like:

```python
    mock_response = mock.MagicMock()
    mock_response.read.return_value = json.dumps(
        {
            "items": [
                {
                    "spec": {"nodeName": name},
                }
                for name in ["node-a", "node-b"]
            ]
        }
    )

    mock_k8s_api = mock.MagicMock()
    mock_k8s_api.list_namespaced_pod.return_value = mock_response
```

betatim:
Tests are now fixed

minrk:
FWIW, from [these measurements](https://nbviewer.jupyter.org/gist/minrk/45ce7c14f71227809b2178bf05fd1562/preload-content.ipynb) the benefit of `_preload_content=False` is ~2-3x with 75 pods, 10-20x with 600 pods, and negligible with 1 pod. I think we can safely stick with the slightly nicer default results, as long as the number of items for a given response are reasonably bounded. We definitely don't *have* to use `_preload_content=False` on every API request.

PR #1275, opened 2021-03-11
Add prose between code snippets

betatim:
When reading this page it was tricky to parse this section because input and output looked the same. I added some text in between the snippets to break up the flow a bit.

### DIFF
--- a/doc/zero-to-binderhub/setup-prerequisites.rst
+++ b/doc/zero-to-binderhub/setup-prerequisites.rst
@@ -48,13 +48,19 @@ with your Kubernetes cluster.
 
    helm version
 
+Which will output something similar to:
+
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


manics (approved):
LGTM!

Another pattern I've used is to include the output in the same block, but prefix the command with the shell prompt `$ `

betatim:
> Another pattern I've used is to include the output in the same block, but prefix the command with the shell prompt `$ `

Is there a clever bit of CSS or rST we can use so that when people highlight the line to copy&paste it the `$` doesn't get copied? That would be the luxury version I think

manics:
I'm not aware of anything. Since we're using [MyST Parser](https://github.com/executablebooks/MyST-Parser) maybe they'd be willing to add an extension for this use-case?

PR #1283, opened 2021-03-29
Gitlab registry support

rappertomate:
This PR provides support for GitLab image registry and possibly also further non-Docker Hub image registries, the latter not tested.
New code was tested against following image registries using minikube:

* Docker Hub
* Image registry of Gitab.com
* Image registry of a private GitLab instance

Besides parsing image names, we also append the query param `service=container_registry` when sending requests to read manifests. I have noticed that without, both Gitlab.com and my private instance, can't be accessed properly. See [Gitlab docs](https://docs.gitlab.com/ee/administration/packages/container_registry.html#enable-the-container-registry) for more context. I can tell that in case of Docker Hub, the extra param is no issue and i assume same applies for other Image registries.

Note that this is mostly not my own work, most of the code is based on [#986 ](https://github.com/jupyterhub/binderhub/pull/986).

For reference, the relevant parts of an example config using Gitlab.com image registry:
```
config:
  BinderHub:
    use_registry: true
    image_prefix: "registry.gitlab.com/<gitlab-user>/<repo>/"
  DockerRegistry:
    token_url: "https://gitlab.com/jwt/auth"
    url: "https://registry.gitlab.com"
    username: <deploy-token-username>
    password: <deploy-token-pw>
registry:
  url: https://registry.gitlab.com
  username: <deploy-token-username>
  password: <deploy-token-pw>
```

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -71,13 +71,13 @@ def _get_image_basename_and_tag(full_name):
         image_name = full_name
         tag = 'latest'
 
-    if re.fullmatch('[a-z0-9]{4,40}/[a-z0-9\._-]{2,255}', image_name):
+    if re.fullmatch('[a-z0-9]{4,40}/[a-z0-9._-]{2,255}', image_name):
         # if it looks like a Docker Hub image name, we're done
         return image_name, tag
-    else:
-        # if the image isn't implied to origin at Docker Hub, the first part has to be a registry
-        image_basename = '/'.join(image_name.split('/')[1:])
-        return image_basename, tag
+    # if the image isn't implied to origin at Docker Hub,
+    # the first part has to be a registry
+    image_basename = '/'.join(image_name.split('/')[1:])
+    return image_basename, tag
 
 
 def _generate_build_name(build_slug, ref, prefix='', limit=63, ref_length=6):
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -7,6 +7,7 @@ import hashlib
 from http.client import responses
 import json
 import string
+import re
 import time
 import escapism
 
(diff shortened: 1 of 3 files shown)


ctr26:
Would be very hyped to see this PR go through

binderhub/builder.py
```diff
@@ -55,6 +56,30 @@
...
+    """
+    # the tag is either after the last (and only) colon, or not given at all,
+    # in which case "latest" is implied
+    tag_splits = full_name.rsplit(':', 1)
+    if len(tag_splits) == 2:
+        image_name = tag_splits[0]
+        tag = tag_splits[1]
+    else:
+        image_name = full_name
+        tag = 'latest'
```
yuvipanda, line 72: In the current implementation, I think the tag is left unspecified if it isn't present. Let's preserve that?

binderhub/builder.py
```diff
@@ -55,6 +56,30 @@
...
+
+    if re.fullmatch('[a-z0-9]{4,40}/[a-z0-9._-]{2,255}', image_name):
```
yuvipanda, line 74: Could we not instead just check for the number of `/` present? The full validator for image names is at https://github.com/distribution/distribution/blob/main/reference/regexp.go, and I'm not sure if this is exactly right. Can we get away with not using regexes?

yuvipanda (changes requested):
Thank you for opening this! I've left some comments, will be happy to continue reviewing and getting this in. \o/

binderhub/builder.py
rappertomate: I wouldn't mind. If i see correctly, the tag is passed to `registry.get_image_manifest()` where the following happens:
`"{}/v2/{}/manifests/{}".format(self.url, image, tag)`
So if tag is an empty string, could this be a problem for the manifests API? Idk personally, as i am pretty much clueless about this.

binderhub/builder.py
rappertomate: Not so sure. What if we have self-host registries without any paths to specify user or project? The unit test shows an example:
`("weirdregistry.com/image:tag", "image", "tag")`  
vs.
`("jupyterhub/k8s-binderhub:0.2.0-a2079a5", "jupyterhub/k8s-binderhub", "0.2.0-a2079a5")`

This distinction cannot be covered by just counting `/`. And the purpose of this PR should be to cover these edge cases as well imo.

binderhub/builder.py
yuvipanda: @rappertomate I think tag is just always required to be specified when pushing, and is never left empty. I wouldn't mind raising an exception if tag is empty or None, but we can tackle that in another PR. I mostly just wanted to make sure the behavior doesn't change.

I think once that's done I'll happily merge this.

yuvipanda:
I just merged this, let's deploy this and see how it goes :) I think the behavior change with `:latest` is acceptable, and leaving the tag out was a bug we never hit earlier. Thanks for pushing this through, @rappertomate! Please test it and let me know if it works?

betatim:
Have you deployed this yet @yuvipanda ?

rappertomate:
I gave it a quick shot on minikube and it works as expected. I have tested it with a private GitLab instance using Helm values as described in the initial post of this PR. I'm not involved with operating any BinderHub instance anymore these days which is why I'm not available to test this more thoroughly.  
As far as i can tell everything looks goof though.

ctr26:
I'll give it a go too

yuvipanda:
@betatim https://github.com/jupyterhub/mybinder.org-deploy/pull/1991 just deployed it

Issue #1223, opened 2020-12-03
Authenticated Binderhub fails when username has spaces


TomAugspurger:
### Bug description

After deploying a binderhub with authorization, I noticed that I was unable to start a jupyterhub session after building an image.
I determined that my Jupyterhub username having spaces was the root cause. My username below is `tom augspurger`.

#### Expected behaviour

Launch a singleuser pod

#### Actual behaviour

Here's the log from the binderhub pod:

```
2020-12-03T18:51:03.029409273Z     HTTPServerRequest(protocol='http', host='20.73.224.239', method='GET', uri='/build/gh/pangeo-gallery/default-binder/HEAD', version='HTTP/1.1', remote_ip='10.244.0.1')
2020-12-03T18:51:03.029437373Z     Traceback (most recent call last):
2020-12-03T18:51:03.029444373Z       File "/usr/local/lib/python3.8/site-packages/tornado/web.py", line 1704, in _execute
2020-12-03T18:51:03.029449573Z         result = await result
2020-12-03T18:51:03.029454073Z       File "/usr/local/lib/python3.8/site-packages/binderhub/builder.py", line 336, in get
2020-12-03T18:51:03.029459673Z         await self.launch(kube, provider)
2020-12-03T18:51:03.029464872Z       File "/usr/local/lib/python3.8/site-packages/binderhub/builder.py", line 543, in launch
2020-12-03T18:51:03.029472972Z         server_info = await launcher.launch(image=self.image_name,
2020-12-03T18:51:03.029478772Z       File "/usr/local/lib/python3.8/site-packages/binderhub/launcher.py", line 181, in launch
2020-12-03T18:51:03.029484172Z         user_data = await self.get_user_data(username)
2020-12-03T18:51:03.029488572Z       File "/usr/local/lib/python3.8/site-packages/binderhub/launcher.py", line 116, in get_user_data
2020-12-03T18:51:03.029492972Z         resp = await self.api_request(
2020-12-03T18:51:03.029497272Z       File "/usr/local/lib/python3.8/site-packages/binderhub/launcher.py", line 94, in api_request
2020-12-03T18:51:03.029502072Z         return await AsyncHTTPClient().fetch(req)
2020-12-03T18:51:03.029506572Z     tornado.httpclient.HTTPClientError: HTTP 400: Bad Request
[...]

Issue #1134, opened 2020-07-28
Users with punctuation in their usernames get 404 instead of notebook


trapnine:
### Bug description
Binder fails to redirect to the new Jupyter server if the logged-in user has punctuation in their username. The punctuation appears unescaped in the browser's address bar after the redirect and the user sees a 404 page. So far we've seen this with apostrophe and parentheses, as in "John O'Malley" or "Jon Jones (cool guy)".

If the user clicks the 'Control Panel' button on the 404 page and navigates to the JupyterHub server manually, everything works. The special characters are not a problem once successfully in JupyterHub. 

The results are the same whether in 'multiple named server' or 'single-server' mode.

#### Expected behaviour
Users with apostrophes or parentheses in their username are redirected to their new notebook after the binder launch.

#### Actual behaviour
Users with apostrophes or parentheses in their username are redirected to a 404 page after the binder launch.

The URL they end up at shows raw punctuation in the browser address bar, e.g.:
https://hub.binder.example.com/user/jon%20jones%20(cool%20guy)/tree/index.ipynb

Any spaces in the name are properly escaped to %20.
The Jupyter server is started and properly named, and the aks pod name is OK, e.g.:
jupyter-jon-20jones-20-28cool-20guy-29-index-2es3b6hx3y

### How to reproduce
1) log in to binder with a username containing apostrophes or parentheses
2) attempt to launch a known good notebook
3) see a 404 page after redirect

### Your personal set up
 - OS:
Linux VMs on Azure Kubernetes
 - Version:
helm chart versions tried:
  binderhub-0.2.0-n200.h305d0e8
  binderhub-0.2.0-n175.he98f51d
 - Configuration:
Followed Zero-to-binderhub setup instructions configuring Azure ACR, Azure AD, Azure Kubernetes service, and https ingress with letsencrypt.

PR #1285, opened 2021-04-06
Escape usernames while launching servers in auth mode

MridulS:
This will let authenticated binderhubs deal with
usernames which aren't escaped.
Example: 'User 1', 'something+spam@example.com'

Fixes https://github.com/jupyterhub/binderhub/issues/1223 and fixes https://github.com/jupyterhub/binderhub/issues/1134

### DIFF
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -6,7 +6,7 @@ import json
 import random
 import re
 import string
-from urllib.parse import urlparse
+from urllib.parse import urlparse, quote
 import uuid
 import os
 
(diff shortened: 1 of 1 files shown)


MridulS:
Just wanted to bump this up :)

binderhub/launcher.py
```diff
@@ -161,11 +161,12 @@ async def launch(self, image, username, server_name='', repo_url='', extra_args=
         """
         # TODO: validate the image argument?
 
+        escaped_username = quote(username, safe='@~')
```
yuvipanda, line 164:
```suggestion
        # Matches the escaping that JupyterHub does https://github.com/jupyterhub/jupyterhub/blob/c00c3fa28703669b932eb84549654238ff8995dc/jupyterhub/user.py#L427
        escaped_username = quote(username, safe='@~')
```

yuvipanda:
Thanks, @MridulS! \o/

PR #1288, opened 2021-04-08
Send filepath urls to JupyterLab instead of classic notebook

SylvainCorlay:
Now that JupyterLab 3 has been out for some time, and most lab extensions have been updated to work well with that version, it might be a good time to make the switch from classic notebook to lab.

In this PR, I am changing the default redirect URL from `tree` to `doc/tree` to enable the JupyterLab "simple mode" that is the closest to the classic notebook user interface (doc).

- Obviously, it is still possible to opt out using the `urlpath` instead of a `filepath`.
- Also referencing the [poll about defaulting to JupyterLab in JupyterHub](https://discourse.jupyter.org/t/poll-should-zero-to-jupyterhub-switch-to-jupyterlab-as-the-default-user-interface/8001) on discourse.

### DIFF
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -64,7 +64,7 @@ class ParameterizedMainHandler(BaseHandler):
         social_desc = f"{SPEC_NAMES[provider_prefix]}: {spec}"
         nbviewer_url = None
         if provider_prefix == "gh":
-            # we can only produce an nbviewer URL for github right now
+            # We can only produce an nbviewer URL for github right now
             nbviewer_url = 'https://nbviewer.jupyter.org/github'
             org, repo_name, ref = spec.split('/', 2)
             # NOTE: tornado unquotes query arguments too -> notebooks%2Findex.ipynb becomes notebooks/index.ipynb
--- a/binderhub/static/js/src/image.js
+++ b/binderhub/static/js/src/image.js
@@ -44,9 +44,8 @@ BinderImage.prototype.launch = function(url, token, path, pathType) {
     if (pathType === "file") {
       // trim trailing / on file paths
       path = path.replace(/(\/$)/g, "");
-      // /tree is safe because it allows redirect to files
-      // need more logic here if we support things other than notebooks
-      url = url + "/tree/" + encodeURI(path);
+      // /doc/tree is safe because it allows redirect to files
+      url = url + "/doc/tree/" + encodeURI(path);
     } else {
       // pathType === 'url'
       url = url + "/" + path;


SylvainCorlay:
We could eventually decide to start the jupyterlab server instead of using the server extension, but I am not doing this now so that users can opt out.

SylvainCorlay:
> I created an issue in repo2docker a few weeks ago to discuss the transition, I think we should keep repo2docker and binderhub in sync if possible:
jupyterhub/repo2docker#1026

Thanks! I missed that one.

> I think we should keep repo2docker and binderhub in sync if possible.

My understanding is that the main thing to change in repo2docker would be in the `start_container` function to start `jupyter lab` instead of `jupyter notebook`. (Would there be anything else?) However some users might prefer to keep the `notebook` command for a while to be able to opt out completely until any issue that arises is resolved.

betatim:
Is it a ridiculous idea to modify the default URL generated by BinderHub (https://mybinder.org/v2/gh/binder-examples/requirements/HEAD?urlpath=doc/tree instead of https://mybinder.org/v2/gh/binder-examples/requirements/HEAD) to include the new URL path instead of changing the behaviour when no URL path is specified?

manics (approved):
As explained in https://github.com/jupyterhub/repo2docker/issues/1026#issuecomment-826579752 this PR _only_ changes the URLs generated for the filepath option, and therefore only affects new URLs, so I think it's fine to merge.

However perhaps we should hold off until our binder-examples are updated to JupyterLab 3.0? I just tried to check how `/doc/tree` appears with https://github.com/binder-examples/conda, then discovered it's still using a cached image with JupyterLab 2.

minrk:
How long have we been building with jupyterlab 3? Maybe we should do an image cache clean of images older than that?

manics:
https://github.com/jupyterhub/mybinder.org-deploy/pull/1839 was merged on 18 Feb.

SylvainCorlay:
Let me know if anything else is needed from me in this PR!

SylvainCorlay:
Just following up. Is there a blocker for this to get in?

SylvainCorlay:
It could be quite interesting to get this in, especially with lab 3.1 being out!

agitter:
I support migrating to JupyterLab, but I wanted to note that this change breaks some existing Binder links.  I was teaching a workshop today and participants got a 404 error with the URL https://mybinder.org/v2/gh/carpentries-incubator/ml4bio-workshop/gh-pages?filepath=illustration.ipynb

It took some digging to find this pull request and change the URL to https://mybinder.org/v2/gh/carpentries-incubator/ml4bio-workshop/gh-pages?urlpath=tree/illustration.ipynb to revert to the previous behavior.

PR #1290, opened 2021-04-22
A description is a bit paraphrased

WhiteBlackGoose:
Wasted a couple of minutes not understanding what "text below" refers to. So... I think it might make sense to make it a bit clearer


consideRatio (approved):
Thanks, LGTM!

betatim (approved):

PR #1293, opened 2021-05-03
Revert "build_docker_config added, enables augmentation of the build pod's docker config"

betatim:
This proposes to revert jupyterhub/binderhub#1255. When we deployed this to mybinder.org today we ended up with all(?) build pods getting stuck in a pending state for about two hours. At which point we reverted the deploy.

Screenshot 2021-05-03 at 17 05 26

The marker just before 15:30h is the deploy. The blue shaded graph shows the number of "running" build requests ("running" includes "container creating"). At 17:00h I deleted all build pods and reverted the deploy.

The summary of the original PR says that there is no breaking change and that maybe for a brief moment new build pods would be using the wrong secret/a secret that doesn't exist. Does someone involved in the PR have an idea what happened/went wrong when deploying this?

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -297,7 +297,7 @@ class BinderHub(Application):
     )
 
     push_secret = Unicode(
-        'binder-build-docker-config',
+        'binder-push-secret',
         allow_none=True,
         help="""
         A kubernetes secret object that provides credentials for pushing built images.
@@ -374,22 +374,6 @@ class BinderHub(Application):
         if parts.scheme != 'unix' or parts.netloc != '':
             raise TraitError("Only unix domain sockets on same node are supported for build_docker_host")
         return proposal.value
-
-    build_docker_config = Dict(
-        None,
-        allow_none=True,
-        help="""
-        A dict which will be merged into the .docker/config.json of the build container (repo2docker)
-        Here, you could for example pass proxy settings as described here:
-        https://docs.docker.com/network/proxy/#configure-the-docker-client
-
-        Note: if you provide your own push_secret, this values wont
-        have an effect, as the push_secrets will overwrite
-        .docker/config.json
-        In this case, make sure that you include your config in your push_secret
-        """,
-        config=True
-    )
 
     hub_api_token = Unicode(
         help="""API token for talking to the JupyterHub API""",
@@ -709,7 +693,6 @@ class BinderHub(Application):
                 "build_memory_limit": self.build_memory_limit,
                 "build_memory_request": self.build_memory_request,
                 "build_docker_host": self.build_docker_host,
-                "build_docker_config": self.build_docker_config,
                 "base_url": self.base_url,
                 "badge_base_url": self.badge_base_url,
                 "static_path": os.path.join(HERE, "static"),
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -248,9 +248,9 @@ class Build:
         )]
 
         if self.push_secret:
-            volume_mounts.append(client.V1VolumeMount(mount_path="/root/.docker", name='docker-config'))
+            volume_mounts.append(client.V1VolumeMount(mount_path="/root/.docker", name='docker-push-secret'))
             volumes.append(client.V1Volume(
-                name='docker-config',
+                name='docker-push-secret',
                 secret=client.V1SecretVolumeSource(secret_name=self.push_secret)
             ))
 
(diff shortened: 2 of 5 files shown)


betatim:
cc @g-braeunlich @consideRatio  @manics

betatim:
If we need some time to figure out why this didn't work I'd propose to merge this and make a second attempt to land this feature. Maybe we can even find a test that would have caught this to stop it from coming back?

g-braeunlich:
oh. :grimacing:

manics:
The change in secret name is breaking
https://github.com/jupyterhub/binderhub/pull/1255#pullrequestreview-638083801

consideRatio:
Hmmm this would have happened if the binderhub Pods didn't restart as part of the upgrade, and started using the new k8s secret name for the new build pods being created.

Since the build pods keep increasing and fail with pending due to failure to see the old k8s secret around, it indicates to me that the software that creates the build pods didn't reconfigure itself with the new k8s secret name at least. Is it the binderhub pod that creates the build pods?

manics:
The build pods should be created by the BinderHub application https://github.com/jupyterhub/binderhub/blob/0b4462c0b1fb15e355b32b38b99c60f440666e94/binderhub/build.py#L239-L240

betatim:
It sounds like restarting the binderhub pods should be the way to fix this/not have this happen. The weird thing is that a change like this (new binderhub code -> new binderhub container image) means the binderhub pods will restart. I unfortunately didn't look at the age of the binderhub pods when things were broken so I don't know for sure if they restarted or not. However now the bhub pods are ~15h old which I think means they restarted when we reverted the deploy. So probably they also restarted when we originally deployed it.

Does anyone understand why even with restarting the bhub pods we still ended up with build pods not being able to start over a period of a few hours?

consideRatio:
This is known:

1. Tests succeeded in the binderhub CI/CD pipeline, including the creation of a build pod. So, new pods successfully referenced the new k8s Secret name. The tests we have are for fresh installations though.
2. [mybinder.org make use of two replicas](https://github.com/jupyterhub/mybinder.org-deploy/blob/7ec4218f64909c35d8e2cef8405b33c43fb2fe41/mybinder/values.yaml#L28-L32) of the binder pod (running binderhub) and will perform a rolling upgrade of the binder pod running the binderhub software.
3. `helm upgrade --wait` is used by the `mybinder.org-deploy` CI/CD script should ensure that deployments startup successfully and become ready.
4. There is no reference to the old k8s Secret name, so it indicates that an old replica of the binder pod is still running.
5. This suggests that the upgrade must have failed. Inspecting [mybinder.org-deploy pipelines, it seems like it did](https://github.com/jupyterhub/mybinder.org-deploy/runs/2492953759?check_suite_focus=true)! It failed waiting for the binder Deployment!

## Conclusion
Mybinder.org deployed a new version. This worked out well for all federation members except for GKE. A rolling upgrade was made but the new binder replicas that started failed to become Ready. The `helm upgrade` was made, but the new replica never became ready, while the old k8s secret name was removed and the new k8s secret name was added.

Why did the GKE deployment fail, why didn't the binderhub pod replica become ready? I'm not sure. I note that the GKE prod use 2 replicas, while other deployments use 1 and the staging use 1-3 based on a HPA (horizontal pod autoscaler).

manics:
Is there some magic we can add to the mybinder.org-deploy pipeline to scream/tweet/post a message to gitter when a production deploy fails?

betatim:
Attempted another deploy, same result.

Digging around the helm chart a bit I found https://github.com/jupyterhub/binderhub/blob/0b4462c0b1fb15e355b32b38b99c60f440666e94/helm-chart/binderhub/templates/deployment.yaml#L55-L58 which makes me think the name of the secret was changed in some but not all places.

consideRatio:
@betatim what are the logs from the binder pods? Are both restarted and have successfully become ready - in other words, both fully updated?

betatim:
No, they remain in the "ContainerCreating" status

betatim:
https://github.com/jupyterhub/mybinder.org-deploy/pull/1900#issuecomment-834434668 has some notes

consideRatio:
@betatim I think I found the issue. I thought i reviewed this by doing a full search across the git repo but i probably searched within some folder only...

I think #1294 will fix this.

g-braeunlich:
Sorry guys, again! This was my sloppiness!

betatim:
@consideRatio is your change in #1294 what I found in https://github.com/jupyterhub/binderhub/pull/1293#issuecomment-834452496?

betatim:
No worries.

It remains a unsolved mystery (for me) why this didn't fail on mybinder.org's staging deployment, but yeah. Let's see what happens this time.

PR #1296, opened 2021-05-13
Fix documentation issues

dolfinus:
Hello.

I've found an issue with duplicated items in the references documentation:
Screenshot_20210513_191900

It looks like the same issue has been fixed in Jupyterhub repo:
https://github.com/jupyterhub/jupyterhub/pull/3021

So I've used the same fix here:
Screenshot_20210513_192033


consideRatio:
Thanks @dolfinus !!

PR #1297, opened 2021-05-13
Run tests on push to any branch

dolfinus:
Hello.

I've made few improvements of testing workflow, each of them is described in the comments below.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,10 +3,7 @@ name: Tests
 on:
   pull_request:
   push:
-    branches: [master, main]
-    tags: ["**"]
   workflow_dispatch:
-
 
 jobs:
   # Most of the "main", "auth" and "helm" jobs are the same and only differ
@@ -18,16 +15,14 @@ jobs:
       # keep running so we can see if tests with other k3s/k8s/helm versions pass
       fail-fast: false
       matrix:
-        include:
-            - k3s-channel: v1.19
-              helm-version: v3.5.0
-              test: main
-            - k3s-channel: v1.19
-              helm-version: v3.5.0
-              test: auth
-            - k3s-channel: v1.19
-              helm-version: v3.5.0
-              test: helm
+        k3s-channel:
+        - v1.19
+        helm-version:
+        - v3.5.0
+        test:
+        - main
+        - auth
+        - helm
     steps:
       - uses: actions/checkout@v2
         with:
@@ -55,6 +50,14 @@ jobs:
         with:
           node-version: '14'
 
+      - name: Cache npm
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
       - name: Run webpack to build static assets
         run: |
             npm install
@@ -64,13 +67,22 @@ jobs:
         with:
           python-version: '3.8'
 
-      - name: Setup Python package dependencies
+      - name: Cache pip
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-python-${{ hashFiles('**/*requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-python-${{ hashFiles('**/*requirements.txt') }}
+            ${{ runner.os }}-python-
+
+      - name: Update pip
         run: |
-            # Install development requirements, package requirements, and frozen image
-            # requirements, with an increasing priority.
-            pip install -r dev-requirements.txt
-            pip install .
-            pip install -r helm-chart/images/binderhub/requirements.txt
+            pip install --upgrade pip
+            pip install --upgrade setuptools wheel
+
+      - name: Setup Python package dependencies
+        run: pip install -r dev-requirements.txt -r helm-chart/images/binderhub/requirements.txt .
 
       - name: Install JupyterHub chart for main tests
         if: matrix.test == 'main'


.github/workflows/test.yml
```diff
@@ -2,9 +2,8 @@ name: Tests
 
 on:
   pull_request:
-  push:
```
dolfinus, line 5: If someone pushes commits to a new branch not mentioned in the list below, tests will not be executed. That does not look great.
Instead it's better to limit branch names which are triggering the workflow in case of creating new pull request, because most of them are created with target_branch=master/main.

.github/workflows/test.yml
```diff
@@ -18,16 +17,14 @@ jobs:
...
-            - k3s-channel: v1.19
-              helm-version: v3.5.0
-              test: main
-            - k3s-channel: v1.19
-              helm-version: v3.5.0
-              test: auth
-            - k3s-channel: v1.19
-              helm-version: v3.5.0
-              test: helm
+        k3s-channel:
```
dolfinus, line 20: Instead of describing each case separately it's possible to just list all the options and then Github will run a build for every possible combination.

.github/workflows/test.yml
```diff
@@ -55,6 +52,14 @@ jobs:
         with:
           node-version: '14'
 
+      - name: Cache npm
```
dolfinus, line 55: Using package cache allows to decrease the installation time.

.github/workflows/test.yml
```diff
@@ -64,13 +69,22 @@ jobs:
...
+            ${{ runner.os }}-python-
+
+      - name: Update pip
         run: |
-            # Install development requirements, package requirements, and frozen image
-            # requirements, with an increasing priority.
-            pip install -r dev-requirements.txt
-            pip install .
-            pip install -r helm-chart/images/binderhub/requirements.txt
+            pip install --upgrade pip
```
dolfinus, line 83: Package manager as well as libraries used for installing dependencies should always be upgraded to the most recent version.

.github/workflows/test.yml
```diff
@@ -64,13 +69,22 @@ jobs:
...
+            pip install --upgrade setuptools wheel
+
+      - name: Setup Python package dependencies
+        run: pip install -r dev-requirements.txt -r helm-chart/images/binderhub/requirements.txt .
```
dolfinus, line 87: Well, you're installing some set of dependencies, then other one with more strict checks and then another one there all the versions are fixed. This leads to downloading, installing and then uninstalling lots of dependencies' versions which is completely useless:
```
...
 Installing collected packages: six, rsa, protobuf, cffi, cachetools, pytz, markupsafe, googleapis-common-protos, google-auth, cryptography, attrs, sqlalchemy, ruamel.yaml, grpcio, google-api-core, websocket-client, prometheus-client, jinja2, google-cloud-core, alembic, kubernetes, jupyterhub, google-cloud-logging
  Attempting uninstall: six
    Found existing installation: six 1.16.0
    Uninstalling six-1.16.0:
      Successfully uninstalled six-1.16.0
  Attempting uninstall: rsa
    Found existing installation: rsa 4.7.2
    Uninstalling rsa-4.7.2:
      Successfully uninstalled rsa-4.7.2
  Attempting uninstall: cffi
    Found existing installation: cffi 1.14.5
    Uninstalling cffi-1.14.5:
      Successfully uninstalled cffi-1.14.5
  Attempting uninstall: cachetools
    Found existing installation: cachetools 4.2.2
    Uninstalling cachetools-4.2.2:
      Successfully uninstalled cachetools-4.2.2
  Attempting uninstall: markupsafe
    Found existing installation: MarkupSafe 2.0.0
    Uninstalling MarkupSafe-2.0.0:
      Successfully uninstalled MarkupSafe-2.0.0
  Attempting uninstall: google-auth
    Found existing installation: google-auth 1.30.0
    Uninstalling google-auth-1.30.0:
      Successfully uninstalled google-auth-1.30.0
  Attempting uninstall: cryptography
    Found existing installation: cryptography 3.4.7
    Uninstalling cryptography-3.4.7:
      Successfully uninstalled cryptography-3.4.7
  Attempting uninstall: attrs
    Found existing installation: attrs 21.2.0
    Uninstalling attrs-21.2.0:
      Successfully uninstalled attrs-21.2.0
  Attempting uninstall: sqlalchemy
    Found existing installation: SQLAlchemy 1.4.15
    Uninstalling SQLAlchemy-1.4.15:
      Successfully uninstalled SQLAlchemy-1.4.15
  Attempting uninstall: ruamel.yaml
    Found existing installation: ruamel.yaml 0.17.4
    Uninstalling ruamel.yaml-0.17.4:
      Successfully uninstalled ruamel.yaml-0.17.4
  Attempting uninstall: websocket-client
    Found existing installation: websocket-client 0.59.0
    Uninstalling websocket-client-0.59.0:
      Successfully uninstalled websocket-client-0.59.0
  Attempting uninstall: prometheus-client
    Found existing installation: prometheus-client 0.10.1
    Uninstalling prometheus-client-0.10.1:
      Successfully uninstalled prometheus-client-0.10.1
  Attempting uninstall: jinja2
    Found existing installation: Jinja2 3.0.0
    Uninstalling Jinja2-3.0.0:
      Successfully uninstalled Jinja2-3.0.0
  Attempting uninstall: alembic
    Found existing installation: alembic 1.6.2
    Uninstalling alembic-1.6.2:
      Successfully uninstalled alembic-1.6.2
  Attempting uninstall: kubernetes
    Found existing installation: kubernetes 17.17.0
    Uninstalling kubernetes-17.17.0:
      Successfully uninstalled kubernetes-17.17.0
  Attempting uninstall: jupyterhub
    Found existing installation: jupyterhub 1.4.1
    Uninstalling jupyterhub-1.4.1:
      Successfully uninstalled jupyterhub-1.4.1
Successfully installed alembic-1.5.4 attrs-20.3.0 cachetools-4.2.1 cffi-1.14.4 cryptography-3.4.4 google-api-core-1.26.0 google-auth-1.25.0 google-cloud-core-1.6.0 google-cloud-logging-1.15.1 googleapis-common-protos-1.52.0 grpcio-1.35.0 jinja2-2.11.3 jupyterhub-1.2.2 kubernetes-9.0.1 markupsafe-1.1.1 prometheus-client-0.9.0 protobuf-3.14.0 pytz-2021.1 rsa-4.7 ruamel.yaml-0.16.12 six-1.15.0 sqlalchemy-1.3.23 websocket-client-0.57.0
```

Instead you can pass all the requirements in the one line. Pip will check the entire set of constraints and then fetch the most suitable ones. That should decrease installation time.

Also pip can detect incompatibility of dependencies' versions. But this can cause CI/CD failure only is you're installing all the dependencies in the same pip run.
If you installed some dependency in one pip call and then installed another dependency which is not compatible with a previous one in another pip call, you'll see a warning instead of an error.
IMHO it's better to see an error in just your CI/CD workflow before the release than receiving a lot of issues after publishing a broken release.

.github/workflows/test.yml
manics: Could we remove the branch restriction instead and run tests on all pushes? Is anyone aware of a reason to keep it?

dolfinus: The only reason of  using such a restriction I know is a bit unexpected Github Actions behavior. In case then you've pushed a commit into some branch and then created a pull request using this branch as a source, 2 copies of workflow will be started - one for `push` event and another one for `pull_request` event.

If someone is using Git Flow (way of naming and working with branches in large projects), it can be quite common situation. For example, you've created branch `feature/abc` and then created a pull request into `dev` or `release-X.Y` branch of the same repo. All the tests here will be run twice. But limiting a list of target branches of pull request could prevent this situation.

If pull requests are created mostly from forked repos instead of branches of the same repo, you'll never face with such an issue, so this restriction could be removed.

betatim: A note on the suggested development flow (in case that helps decide things) here: fork&PR is the recommended way to make changes, even for maintainers. We encourage people to open a work-in-progress PR as soon as possible, even if they are just doing exploratory work. This means there should almost never be additional branches in the upstream repository and most contributors get the benefits of CI "for free".

We do not have `release-X.Y` branches as all changes target HEAD and each merge into the main branch is a "release". We never backport changes to earlier "releases".

However some of the maintainers sometimes create a branch (a la `betatim-patch-1` etc) when to fix a small typo or such. And then make a PR. In these cases we'd have the tests run twice.

Overall it seems there are reasons to keep what is in the config now and reasons to change. Going for the principle of "make the smallest number of changes possible": what workflow/positive feature that is currently not possible would be enabled by making this change?

dolfinus: Well, for example I've found few bugs, created a fork and made several branches for each bug to then create separated pull requests.
But now pushing commits leads to running tests only if in case of master branch. So I'm not sure that my changes are not causing some issues until I made a pull request to the original repo.

manics: I'm strongly in favour of running CI tests on pushes to all branches in all forks unless there's a reason not to, because it means contributors can experiment in their own forks without opening PRs. This is particularly the case now that people opening their first PR need to be approved before CI tests are run (an unfortunate consequence of cryptominers).

dolfinus: Any updates here?

manics: My preference is for
```yaml
on:
  pull_request:
  push:
  workflow_dispatch:
```
and not restrict this to particular branches. Same for `diff.yml`

dolfinus: OK, fixed this.

minrk:
Thanks!

PR #1301, opened 2021-05-28
Add container lifecycle hooks support to chart

sgaist:
This merge request implements support for adding lifecycle hooks to the binder container.

One of the main use case here is to be able to add one (or more) custom CA to the system without having to rebuild the base image.

The combination of extraVolumes, extraVolumesMounts and lifecyle allows to do that in a simple manner.

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -70,6 +70,10 @@ spec:
         args:
           - --config
           - /etc/binderhub/config/binderhub_config.py
+        {{- with .Values.lifecycle }}
+        lifecycle:
+          {{- . | toYaml | nindent 10 }}
+        {{- end }}
         volumeMounts:
           - mountPath: /etc/binderhub/config/
             name: config
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -211,6 +211,7 @@ ingress:
     #     - chart-example.local
 
 initContainers: []
+lifecycle: {}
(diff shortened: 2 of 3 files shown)


helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -70,6 +70,10 @@ spec:
         args:
           - --config
           - /etc/binderhub/config/binderhub_config.py
+        {{- with .Values.lifecycleHooks }}
```
consideRatio, line 73:
```suggestion
        {{- with .Values.lifecycle }}
```

helm-chart/binderhub/values.yaml
```diff
@@ -211,6 +211,7 @@ ingress:
     #     - chart-example.local
 
 initContainers: []
+lifecycleHooks: {}
```
consideRatio, line 214:
```suggestion
lifecycle: {}
```

helm-chart/binderhub/templates/deployment.yaml
sgaist: Fixed

helm-chart/binderhub/values.yaml
sgaist: Fixed

sgaist:
@consideRatio One additional question, would a similar change be accepted for the dind daemonset ?

I juste realized that I had to pass "--insecure-registry" as extraArgs for it to be able to push the generated image. That makes for an inconsistent experience when configuring the whole setup.

consideRatio:
I would think so yepp!

PR #1309, opened 2021-06-16
add build token, apply rate limit by request ip

minrk:
build token is a signed JWT, used to validate requests originating from our pages.

A build token is only valid for a short time, and only valid for a single build spec and request origin. The token is issued by the request for the html page served at `/v2/:spec` and added in the javascript via url param to `/build/v2/:spec?build_token=...`.

Similar to an xsrf token, but actually allows cross-site requests within the federation as long as federation members share a signing key (can be improved to only distributing public keys with RS256 for signing instead of HS256).

rate limit is calculated by remote ip address and only applied on requests made *without* a build token, i.e. via raw requests to the event stream API.

Default rate limit is 10 per hour.

Most abuse occurs via floods of requests to the api from single ips, but we don't want to limit traffic from e.g. a single-room workshop as aggressively, which might also originate from a single IP. Hence the need to distinguish 'regular human' page requests from API requests or cross-origin requests.

This is *not* a robust security measure, as dedicated abuser can still request valid build tokens (I won't say how, but it's not hard). Indeed, it's *less* robust than a standard XSRF token approach, but those prohibit the way our federation works, hence the jwt. xsrf is also trivially circumventable by scripts, it only protects against malicious websites visited by valid users (not a problem we have, I believe). But at least it's more tedious, and they'll need to keep requesting tokens as they are only valid for one spec at a time.

If we find that these requests are still getting abused, we can use a similar rate limit mechanism to put a captcha in front of the request for the build token itself if we start seeing lots of traffic from an ip. Properly implemented, that should be a genuinely robust marker for requests by humans, which is what we are really after.

I added a boolean `build_token` field to the event schema, so we'll be able to see how many builds are issues with a token vs without, as a proxy measure for whether it's via the API or a regular browser request.

### DIFF
--- a/binderhub/ratelimit.py
+++ b/binderhub/ratelimit.py
@@ -1,0 +1,95 @@
+"""Rate limiting utilities"""
+
+import time
+from traitlets import Integer, Dict, Float, default
+from traitlets.config import LoggingConfigurable
+
+
+class RateLimitExceeded(Exception):
+    """Exception raised when rate limit is exceeded"""
+
+
+class RateLimiter(LoggingConfigurable):
+    """Class representing a collection of rate limits
+
+    Has one method: `.increment(key)` which should be called when
+    a given actor is attempting to consume a resource.
+    `key` can be any hashable (e.g. request ip address).
+    Each `key` has its own rate limit counter.
+
+    If the rate limit is exhausted, a RateLimitExceeded exception is raised,
+    otherwise a summary of the current rate limit remaining is returned.
+
+    Rate limits are reset to zero at the end of `period_seconds`,
+    not a sliding window,
+    so the entire rate limit can be consumed instantly
+    """
+
+    period_seconds = Integer(
+        3600,
+        config=True,
+        help="""The rate limit window""",
+    )
+
+    limit = Integer(
+        10,
+        config=True,
+        help="""The number of requests to allow within period_seconds""",
+    )
+
+    clean_seconds = Integer(
+        600,
+        config=True,
+        help="""Interval on which to clean out old limits.
+
+        Avoids memory growth of unused limits
+        """,
+    )
+
+    _limits = Dict()
+
+    _last_cleaned = Float()
+
+    @default("_last_cleaned")
+    def _default_last_cleaned(self):
+        return self.time()
+
+    def _clean_limits(self):
+        now = self.time()
+        self._last_cleaned = now
+        self._limits = {
+            key: limit for key, limit in self._limits.items() if limit["reset"] > now
+        }
+
+    @staticmethod
+    def time():
+        """Mostly here to enable override in tests"""
+        return time.time()
+
+    def increment(self, key):
+        """Check rate limit for a key
+
+        key: key for recording rate limit. Each key tracks a different rate limit.
+        Returns: {"remaining": int_remaining, "reset": int_timestamp}
+        Raises: RateLimitExceeded if the request would exceed the rate limit.
+        """
+        now = int(self.time())
+        if now - self._last_cleaned > self.clean_seconds:
+            self._clean_limits()
+
+        if key not in self._limits or self._limits[key]["reset"] < now:
+            # no limit recorded, or reset expired
+            self._limits[key] = {
+                "remaining": self.limit,
+                "reset": now + self.period_seconds,
+            }
+        limit = self._limits[key]
+        # keep decrementing, so we have a track of excess requests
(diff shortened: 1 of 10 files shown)


manics:
It looks like the timeout is a fixed window beginning from the first request, as opposed to a sliding window. I think it's worth documenting that clearly to avoid others being surprised, since it means you could end up with `2 * limit - 1` requests in `period_seconds`.

minrk:
@manics yes! I believe this is the most common way to implement API rate limiting. GitHub does the same (I copied their rate limit headers). I added some docstrings to that effect.

Notes from the meeting today:

- land this as a proof of concept to make sure it works as advertised. That means distributing a shared build_token_secret across federation members

Things to do to improve this, but not before deploying this as a proof of concept:

- using `/.well-known/jwks.json` URLs to distribute JWK public keys, so we don't have to distribute shared keys
- Since most abuse is from direct API request that would lack build tokens, we can shift our datacenter bans to only non-token build requests to solve https://github.com/jupyterhub/mybinder.org-deploy/issues/1828#issuecomment-814037292

binderhub/ratelimit.py
```diff
@@ -0,0 +1,95 @@
...
+            # no limit recorded, or reset expired
+            self._limits[key] = {
+                "remaining": self.limit,
+                "reset": now + self.period_seconds,
+            }
+        limit = self._limits[key]
+        # keep decrementing, so we have a track of excess requests
+        # which indicate abuse
+        limit["remaining"] -= 1
+        if limit["remaining"] <= 0:
```
manics, line 90:
```suggestion
        if limit["remaining"] < 0:
```
Otherwise this could raise with `Rate limit exceeded (by 0) ...`

binderhub/tests/test_ratelimit.py
```diff
@@ -0,0 +1,81 @@
...
+    with mock.patch.object(r, "time", lambda: now):
+        limit = r.increment("1.2.3.4")
+    assert limit == {
+        "remaining": 9,
+        "reset": reset,
+    }
+    assert r._limits == {
+        "1.2.3.4": limit,
+    }
+    for i in range(1, 9):
```
manics, line 23: `(0, 9)` or `(1, 10)`?: `r.increment` is called once prior to this loop, so you should be able to call it a further 9 times without an exception.

binderhub/ratelimit.py
betatim: Good point. I think we'd not only end up with a slightly weird exception text but also we'd have a rate limit that is one less than what is configured. That is because we subtract one from the remaining limit and then check. So on the last valid request the remaining limit will be zero, because we just decreased it to that.

binderhub/static/js/src/image.js
```diff
@@ -8,6 +9,10 @@ export default function BinderImage(providerSpec) {
 
 BinderImage.prototype.fetch = function() {
   var apiUrl = BASE_URL + "build/" + this.providerSpec;
+  if (BUILD_TOKEN) {
+      apiUrl = apiUrl + `?build_token=${BUILD_TOKEN}`;
```
betatim, line 13: TIL: we can't set headers for `EventSource` requests and that there is a thing called "readable streams" based on `fetch()` https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams

Sounds like this is the replacement/"do everything and more that EventStream can do" for event streams (based on GH comments by Chrome devs ;) )

binderhub/tests/test_ratelimit.py
betatim:
```suggestion
    for i in range(1, 10):
```
so we can keep the `assert` about the remaining requests the same below

binderhub/tests/test_ratelimit.py
```diff
@@ -0,0 +1,81 @@
...
+        assert limit == {
+            "remaining": 9 - i,
+            "reset": reset,
+        }
+
+    for i in range(5):
+        with pytest.raises(RateLimitExceeded):
+            r.increment("1.2.3.4")
+
+    assert r._limits["1.2.3.4"]["remaining"] == -4
```
betatim, line 34:
```suggestion
    assert r._limits["1.2.3.4"]["remaining"] == -5
```

betatim:
Merging this now because I think it is good enough. Next thing is to roll this out across the federation 🎉

minrk:
Thanks @betatim!

PR #1314, opened 2021-06-25
Normalise the value of Host and Origin headers

betatim:
A follow up to #1309 

The origin header can contain the scheme (eg `https://mybinder.org`) and not just `hostname:port` like the host header.

Noticed this when looking at the requests sent to launch `binder-examples/requirements`. I directly visited `https://mybinder.org/v2/gh/binder-examples/requirements/HEAD` which sends only a `Host` header with value `mybinder.org`. I think with #1309 the response from this request would contain the build token (with a origin of `mybinder.org`).

Next `https://mybinder.org/build/gh/binder-examples/requirements/HEAD` is fetched which also only has a `host: mybinder.org` header. It returns a status 307 pointing to `https://ovh.mybinder.org/build/gh/binder-examples/requirements/HEAD`. The request to this URL contains a `Host: ovh.mybinder.org` and `Origin: https://mybinder.org` header.

We give preference to the origin header, but because that contains the scheme we end up with a build token that has `mybinder.org` in it and a origin of `https://mybinder.org`. In this PR I add a bit of code that parses the origin and returns the `netloc` (hostname + port) which should be in the same format as the `Host` header.

### DIFF
--- a/binderhub/base.py
+++ b/binderhub/base.py
@@ -1,6 +1,7 @@
 """Base classes for request handlers"""
 
 import json
+import urllib.parse
 
 import jwt
 from http.client import responses
@@ -46,6 +47,22 @@ class BaseHandler(HubOAuthenticated, web.RequestHandler):
             )
             raise web.HTTPError(403, f"Requests from {message} are not allowed")
 
+    def token_origin(self):
+        """Compute the origin used by build tokens
+
+        For build tokens we check the Origin and then the Host header to
+        compute the "origin" of a build token.
+        """
+        origin_or_host = self.request.headers.get("origin", None)
+        if origin_or_host is not None:
+            # the origin header includes the scheme, which the host header
+            # doesn't so we normalize Origin to the format of Host
+            origin_or_host = urllib.parse.urlparse(origin_or_host).netloc
+        else:
+            origin_or_host = self.request.headers.get("host", "")
+
+        return origin_or_host
+
     def check_build_token(self, build_token, provider_spec):
         """Validate that a build token is valid for the current request
 
@@ -69,9 +86,7 @@ class BaseHandler(HubOAuthenticated, web.RequestHandler):
             app_log.error(f"Failure to validate build token for {provider_spec}: {e}")
             raise web.HTTPError(403, "Invalid build token")
 
-        origin = self.request.headers.get(
-            "origin", self.request.headers.get("host", "")
-        )
+        origin = self.token_origin()
         if decoded["origin"] != origin:
             app_log.error(f"Rejecting build token from mismatched origin {decoded}")
             raise web.HTTPError(403, "Invalid build token")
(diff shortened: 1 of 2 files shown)


betatim:
Thinking about this I wonder what `Host` and `Origin` headers our "prime" BinderHub (gke.mybinder.org) sees  when we proxy access to it in the federation redirector. Does the request from the federation proxy send on the original header values or does it replace them? My guess is that `Host` has to be set to the BinderHub's actual value (gke.mybinder.org) because otherwise the ingress won't route traffic to it

manics:
Since you've already spent time debugging this is it worth writing up your findings in the form of a simple test, especially as the `Host` and `Origin` (not to mention `X-Forwarded-Host` and `Forwarded`) headers can get complicated.

minrk:
Origin and Host are both normally present for cross-origin requests. Host gets updated in a potentially complicated way as it passes through proxies and may or may not reflect what the Browser's request host was (hopefully it is! Things get bad when it doesn't, which generally indicates a misconfigured proxy). Origin should always arrive unmodified from the original request.

The standard way to detect cross-origin requests is to compare Origin and Host if Origin is set.

Origin is *not* usually set for same-origin requests, in which case Host is your only source of info. However, browsers differ in when they set Origin (some always set it for XMLHTTPRequest, some always exclude it if Origin matches Host according to the browser)

minrk (approved):

manics:
Shall we merge and add a test later? Or wait?

minrk:
Merged this one. Test in #1317

betatim:
Thanks a lot!!

The thing I still don't have straight in my head (and not sure where to discuss it) is the following: the request to the page `https://mybinder.org/v2/gh/binder-examples/requirements/HEAD` is proxied by the federation-redirector. It is always sent to the "prime" hub in the federation (gke.mybinder.org). This means the token embedded in the website will be for `gke.mybinder.org`. The browser then (directly, not through the proxy) makes a request to https://ovh.mybinder.org/build/gh/binder-examples/requirements/HEAD (or some other random member of the federation). This request will contain the token that was embedded in the HTML served from gke.mybinder.org. 

Will this work or will the token get rejected because it was made for a different member of the federation? And if this is what happens, how do we fix it?

The reason the HTML page is proxied (always to the prime) is because it seemed too difficult to organise for it to be served by the same host as the subsequent request to the build endpoint. Federation members might go away, be full, etc, etc between the two requests.

minrk:
The use of `Origin` should mean that the token is associated with the webpage the human is looking at (Origin will be mybinder.org for the federation redirector). However, there was a complication that the federation redirector does not (and cannot for kubernetes reasons) preserve the Host header, since that's how the ingress controller picks the ingress. Since our proxy is a weird one, I made it set Origin (if it wasn't already) to the originating Host. So Origin is always set for requests through the federation-redirector, and both proxied and redirected requests should look the same with Origin set on both (I truly wish browsers always did this)

PR #1317, opened 2021-06-28
exercise the host/origin field in build tokens

minrk:
Origin is preferred when present, but not always present.

Test for #1314


manics (approved):
👍

Issue #1318, opened 2021-06-30
Support running without kubernetes, just docker


yuvipanda:
### Proposed change

There's nothing inherent about binderhub that requires kubernetes. What it needs really is:

1. A way to spawn repo2docker
2. A way to tell a JupyterHub to run users with a specific image

This is doable without kubernetes! If you look at where we use kubernetes, it is mostly only in [build.py](https://github.com/jupyterhub/binderhub/blob/master/binderhub/build.py). We already support another build backend - [FakeBuild](https://github.com/jupyterhub/binderhub/blob/00bd8304e99e9a4edbdfcca6fd98293fb1fb2e6a/binderhub/build.py#L409).

I'd like to support an additional backend that lets you run binderhub on a single host, with docker + a jupyterhub with dockerspawner. This will make it much easier for people to set up 'small' binderhubs internally, and also exercise our abstraction for future users (who might want it on HPCs, for example).

### Alternative options

Decide that we have a *hard* dependency on kubernetes, and other setups will not be supported. I don't think we should do this though.

### Who would use this feature?

1. People who want to run binderhub on a single node, since they don't expect a lot of traffic (or it is authenticated)
2. People who want to code for different backends eventually. This would also involve finishing up @manics's work in https://github.com/jupyterhub/repo2docker/pull/848 in repo2docker, but that can happen independent of work here.

### (Optional): Suggest a solution

1. Turn `Build` in https://github.com/jupyterhub/binderhub/blob/00bd8304e99e9a4edbdfcca6fd98293fb1fb2e6a/binderhub/build.py#L18 into an abstract, well defined interface. I think this involves definint `__init__`, `submit` and `stream_logs`
2. Implement a `DockerBuild` that implements `Build` on top of the docker API
3. Investigate features we need that are deployed as part of the helm chart, rather than in the python package. Refactor as needed (keeping in mind needs of simplicity for mybinder.org)
3. Move our `kubernetes` imports to a file, and make it conditional. This helps us make `kubernetes` an optional dependency!
4. Add tests to make sure things work with just docker!

betatim:
Collecting other issues that propose the same idea.

https://github.com/jupyterhub/binderhub/issues/1061

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/building-a-the-littlest-binderhub/9824/1

minrk:
[...]

PR #1319, opened 2021-06-30
Make build class configurable

g-braeunlich:
This will enable to replace binderhub.build.Build by a custom class.
One usecase would be to configure custom options for the repo2docker
command.

See: https://discourse.jupyter.org/t/make-target-repo-dir-configurable-in-binder/9617

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -31,6 +31,7 @@ from traitlets import (
     TraitError,
     Unicode,
     Union,
+    Type,
     default,
     observe,
     validate,
@@ -245,6 +246,16 @@ class BinderHub(Application):
 
         Set to false to use only local docker images. Useful when running
         in a single node.
+        """,
+        config=True
+    )
+
+    build_class = Type(
+        Build,
+        help="""
+        The class used to build repo2docker images.
+        
+        Must inherit from binderhub.build.Build
         """,
         config=True
     )
@@ -740,6 +751,7 @@ class BinderHub(Application):
                 "repo_providers": self.repo_providers,
                 "rate_limiter": RateLimiter(parent=self),
                 "use_registry": self.use_registry,
+                "build_class": self.build_class,
                 "registry": registry,
                 "traitlets_config": self.config,
                 "google_analytics_code": self.google_analytics_code,
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -21,7 +21,6 @@ from tornado.log import app_log
 from prometheus_client import Counter, Histogram, Gauge
 
 from .base import BaseHandler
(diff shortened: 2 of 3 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/make-target-repo-dir-configurable-in-binder/9617/6

binderhub/app.py
```diff
@@ -249,6 +250,13 @@ def _valid_badge_base_url(self, proposal):
         config=True
     )
 
+    build_class = Type(
+        Build,
+        help="""The class used to build repo2docker images
```
yuvipanda, line 255:
```suggestion
        help="""
        The class used to build repo2docker images.
        
        Must inherit from binderhub.build.Build
```

yuvipanda:
I was just about to start doing similar things for https://github.com/jupyterhub/binderhub/issues/1318, this looks great.

I suggested a bit of documentation change :)

binderhub/builder.py
```diff
@@ -368,7 +368,7 @@ async def get(self, provider_prefix, _unescaped_spec):
         else:
             push_secret = None
 
-        BuildClass = FakeBuild if self.settings.get('fake_build') else Build
+        BuildClass = FakeBuild if self.settings.get('fake_build') else self.settings.get('build_class')
```
manics, line 371: Do we still need the `fake_build` setting or can we set `c.BinderHub.build_class` instead in:
https://github.com/jupyterhub/binderhub/blob/1957bfb0df4bffd2cd31d6915a6d132426c614ae/testing/local-binder-mocked-hub/binderhub_config.py#L15
This would also serve to validate this change.

yuvipanda: That would be great.

g-braeunlich: Guess, I could successfully remove it (plz have a look!).
Do you want me to move the `FakeBuild` class to `testing/local-binder-mocked-hub/binderhub_config.py`?
This would be possible now.

manics (approved):
Thanks, this LGTM!

binderhub/builder.py
manics: @yuvipanda any preference on where `FakeBuild` should be defined? Or can we merge this now?

consideRatio (approved):
LGTM from i can tell!

binderhub/builder.py
g-braeunlich: I guess I better should not have asked :stuck_out_tongue_closed_eyes:

manics: Two approvals and three weeks later... I'm merging 😁. We can move the class to a different location in another PR.

PR #1323, opened 2021-06-30
Allow running dev setup without kubernetes

yuvipanda:
- Talk to kubernetes to get list of running user pods only
  if quotas are set. This saves a network call, and allows
  running locally without a kubernetes setup (for frontend dev).
  This does meean we lose the log value of number of other
  pods currently running a given repo, though.
- Optionally make `kube` parameter to `Build` `None` - this works
  fine for FakeClient. We should abstract this out better so
  `Build` is abstract and `KubeBuild` is kubernetes specific, but
  that can come later

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -333,9 +333,6 @@ class BuildHandler(BaseHandler):
             else:
                 image_found = True
 
-        # Launch a notebook server if the image already is built
-        kube = self.settings['kubernetes_client']
-
         if image_found:
             await self.emit({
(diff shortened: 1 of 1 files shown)


yuvipanda:
Just fixing things along the way as I try to follow the contributing guide (https://github.com/jupyterhub/binderhub/blob/master/CONTRIBUTING.md#develop-user-interface).

manics:
Not for this PR, but would be nice if we could also develop the front-end without Docker :smiley:

PR #1330, opened 2021-07-06
Bump Chart.yaml apiVersion (embeds requirements.yaml content into Chart.yaml)

consideRatio:
In this PR I migrate Chart.yaml's apiVersion: 1 -> 2. The purpose is just general maintenance to make life easier in the end by keeping things somewhat updated.

I'm not sure if a henchbot or similar creates PRs to mybinder.org will be influenced by this, but if it is, I think it is a concern somewhat out of scope for consideration when reviewing this PR.


yuvipanda:
Thanks, @consideRatio!

PR #1331, opened 2021-07-07
Add schema.yaml for helm3 compliant chart validation

### LINKED ISSUE #1329: values.schema.json - like in the z2jh Helm chart
### Proposed change
Z2JH has a schema.yaml file where we write documentation about options. Helm 3 introduced the ability to bundle a Helm chart with a schema.values.json file, and that is exactly what we have already in schema.yaml, so in z2jh we just make that become schema.values.json using a script that runs before we package and publish the chart.

I suggest we create the same setup for the BinderHub Helm chart. It has been a blessing to have such schema validation - it captures many errors even for experienced users, and it does so in a very early stage with a nice error message directly by using the `helm` CLI without any custom plugins etc.

### Alternative options

To do nothing and let users run into errors that are hard to debug, but end up being YAML syntax errors :)

Everyone

0. Read [about schema.values.json in helm docs](https://helm.sh/docs/topics/charts/#schema-files) trial a schema validation failure in z2jh so you see how it works in practice
   ```shell
   # helm 3.5+ required for z2jh specifically btw, and 3+ for schema validation
   helm template --repo https://jupyterhub.github.io/helm-chart/ jupyterhub --set hub.asdf=test-validation
   ```
1. Define a [schema.yaml file like in z2jh](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/schema.yaml) for binderhub [in the helm-chart/binderhub](https://github.com/jupyterhub/binderhub/tree/master/helm-chart/binderhub) folder
   It has been a big win using YAML rather than JSON in z2jh. We can make use of anchors etc there that we couldn't make use of in JSON. So I suggest we keep using YAML that is rendered into a JSON file when we package the Helm chart.
2. Copy the script to transform the schema.yaml into schema.values.json and verify it works.
   Example [scripts from z2jh here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/main/tools). The generate-json-schema.py script is relevant and the validate- against-schema.py is relevant as well.
   Note that we want a dedicated lint-and-validate-values.yaml file because it is not plausible to test all kinds of templates as part of the real helm install test we do.
3. Add GitHub workflow / job to validate the Helm chart based on the schema.
   Example [from z2jh here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/adda42a2766e823473554a105baf3c37d55c9208/.github/workflows/test-chart.yaml#L27-L104).
4. Make sure to package generated values.schema.json as part of the publication process.
   [Example from z2jh](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/adda42a2766e823473554a105baf3c37d55c9208/.github/workflows/publish.yml#L128-L135) where the generation script is run before packaging the chart.
5. Make sure to add a .helmignore to exclude schema.yaml
   Example [in z2jh here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/adda42a2766e823473554a105baf3c37d55c9208/jupyterhub/.helmignore#L1-L9).

consideRatio:
Closes #1329 by mimicing https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2033 that converts a schema.yaml file into values.schema.json before publishing and package it with the Helm chart so `helm` automatically can perform validation logic whenever `helm template`, `helm install` or `helm upgrade` is run.

My process to create the schema.yaml file was to look in values.yaml and add schema entries for everything there. I copied various schema.yaml entries from z2jh's schema.yaml file to save me some work, but a lot of descriptions is still just having a TODO note. I consider it out of scope for this PR to also write descriptions for everything in the schema.yaml. The descriptions are currently not used anyhow, but they can be used to generate a configuration reference in the future.

I've tried this schema against mybinder.org's Helm chart, adjusted the schema by adding `hpa` and `networkPolicy` to not break mybinder.org-deploy and because these configuration options may very well be upstreamed to BinderHub itself.

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -117,7 +117,7 @@ spec:
         {{- end }}
         {{- end }}
         {{- with .Values.extraEnv }}
-        {{- include "jupyterhub.extraEnv" . | nindent 8 }}
+        {{- . | toYaml | nindent 8 }}
         {{- end }}
         ports:
           - containerPort: 8585
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -24,10 +24,7 @@ spec:
         release: {{ .Release.Name }}
         heritage: {{ .Release.Service }}
         {{- with .Values.deployment.labels }}
-        # Because toYaml + indent is super flaky
-        {{- range $key, $value := .Values.deployment.labels }}
-        {{ $key }}: {{ $value | quote }}
-        {{- end }}
+        {{- . | toYaml | trimSuffix "\n" | nindent 8 }}
         {{- end }}
       annotations:
         # This lets us autorestart when the configmap changes!
@@ -120,7 +117,7 @@ spec:
         {{- end }}
         {{- end }}
         {{- with .Values.extraEnv }}
-        {{- . | toYaml | nindent 8 }}
+        {{- include "jupyterhub.extraEnv" . | nindent 8 }}
         {{- end }}
         ports:
           - containerPort: 8585
--- a/helm-chart/binderhub/templates/secret.yaml
+++ b/helm-chart/binderhub/templates/secret.yaml
@@ -6,7 +6,7 @@ type: Opaque
 data:
   {{- $values :=  dict "config" dict }}
   {{- $cfg := .Values.config }}
-  binder.hub-token: {{ .Values.jupyterhub.hub.services.binder.apiToken | b64enc | quote }}
+  binder.hub-token: {{ .Values.jupyterhub.hub.services.binder.apiToken | required "jupyterhub.hub.services.binder.apiToken must be explicitly set!" | b64enc | quote }}
   {{- /* every 'pick' here should be matched with a corresponding 'omit' in secret.yaml */ -}}
   {{- if $cfg.GitHubRepoProvider }}
   {{- $_ := set $values.config "GitHubRepoProvider" (pick $cfg.GitHubRepoProvider "client_id" "client_secret" "access_token") }}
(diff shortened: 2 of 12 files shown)


helm-chart/binderhub/templates/secret.yaml
```diff
@@ -6,7 +6,7 @@ type: Opaque
 data:
   {{- $values :=  dict "config" dict }}
   {{- $cfg := .Values.config }}
-  binder.hub-token: {{ .Values.jupyterhub.hub.services.binder.apiToken | b64enc | quote }}
+  binder.hub-token: {{ .Values.jupyterhub.hub.services.binder.apiToken | required "jupyterhub.hub.services.binder.apiToken must be explicitly set!" | b64enc | quote }}
```
consideRatio, line 9: Note that this would just error when b64enc receives a null value with a less obvious error message, so adding this `required` pipe isn't a breaking change.

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -120,7 +117,7 @@ spec:
         {{- end }}
         {{- end }}
         {{- with .Values.extraEnv }}
-        {{- . | toYaml | nindent 8 }}
+        {{- include "jupyterhub.extraEnv" . | nindent 8 }}
```
yuvipanda, line 120: This couples a helper function from the `jupyterhub` chart here. We aren't doing that elsewhere in our chart. Can we put this in a different PR so we can discuss that separately?

yuvipanda (changes requested):
One comment, otherwise good to merge!

helm-chart/binderhub/templates/deployment.yaml
consideRatio: Reverted in a4c368a!

consideRatio: The tests of relevance succeeded, but there was an intermittent failure with polling zenodo - restarted jobs.

yuvipanda:
Thanks, @consideRatio!

betatim:
I think this PR broke the helm chart publishing action. It looks like the fix would be to install a Python yaml library. Do you want to follow up this PR with a fix?

PR #1332, opened 2021-07-07
helm-chart: enhance extraEnv like in z2jh with jupyterhub.extraEnv helper

consideRatio:
### Feature summary

extraEnv can now not only be a list, but also an object. The implementation reuse a z2jh named template that transforms a extraEnv entry into a list to be augmented in a containers `env` specification.

Key outcomes are to:
- allow extraEnv config to be a dictionary so it can merge instead and be overridden as a list would be
- allow extraEnv in dictionary format to contain advanced env entries as well (not just `name: value`, see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1757)
- reuse jupyterhub Helm chart logic, both in schema.yaml by copy paste and in the logic to render by reference to the named template.

### Notes about implementation

BinderHub the Helm chart depends on Z2JH. This PR reuse logic in the Z2JH Helm chart by referencing a globally accessible named Helm template defined in Z2JH. It is a named Helm template that takes either a list or mapping and emits a YAML formatted list.

To reference certain named template in Z2JH is currently an explicitly recommended practice in Z2JH ([see fullnameOverride docs](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#fullnameoverride)). If we want to refresh the BinderHub helm chart I think a relevant PR is to make BinderHub to stop referencing Z2JH resource names as hardcoded, but instead references the names in z2jh via the globally accessible named templates that evaluates to the actual name.

image

### DIFF
--- a/helm-chart/binderhub/schema.yaml
+++ b/helm-chart/binderhub/schema.yaml
@@ -492,9 +492,38 @@ properties:
   extraVolumes: *extraVolumes-spec
   extraVolumeMounts: *extraVolumeMounts-spec
   extraEnv:
-    type: array
+    type: [object, array]
+    additionalProperties: true
     description: |
-      Env to add to the binder Pods.
+      Environment variables to add to the binder pods.
+
+      String literals with `$(ENV_VAR_NAME)` will be expanded by Kubelet which
+      is a part of Kubernetes.
+
+      ```yaml
+      extraEnv:
+        # basic notation (for literal values only)
+        MY_ENV_VARS_NAME1: "my env var value 1"
+
+        # explicit notation (the "name" field takes precedence)
+        CHP_NAMESPACE:
+          name: CHP_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+
+        # implicit notation (the "name" field is implied)
+        PREFIXED_CHP_NAMESPACE:
+          value: "my-prefix-$(CHP_NAMESPACE)"
+        SECRET_VALUE:
+          valueFrom:
+            secretKeyRef:
+              name: my-k8s-secret
+              key: password
+      ```
+
+      For more information, see the [Kubernetes EnvVar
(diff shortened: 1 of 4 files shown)


helm-chart/binderhub/schema.yaml
```diff
@@ -492,9 +492,38 @@ properties:
...
+              fieldPath: metadata.namespace
+
+        # implicit notation (the "name" field is implied)
+        PREFIXED_CHP_NAMESPACE:
+          value: "my-prefix-$(CHP_NAMESPACE)"
+        SECRET_VALUE:
+          valueFrom:
+            secretKeyRef:
+              name: my-k8s-secret
+              key: password
```
consideRatio, line 522:
```suggestion
        # explicit notation (the "name" field takes precedence)
        BINDER_NAMESPACE:
          name: BINDER_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace

        # implicit notation (the "name" field is implied)
        PREFIXED_BINDER_NAMESPACE:
          value: "my-prefix-$(BINDER_NAMESPACE)"
        SECRET_VALUE:
          valueFrom:
            secretKeyRef:
              name: my-k8s-secret
              key: password
```

yuvipanda:
Can you add a note to Chart.yaml mentioning that when upgrading, check to make sure no inherited used templates are broken? Happy to merge after that.

consideRatio:
> check to make sure no inherited used templates are broken? Happy to merge after that.

Do you mean to verify that something like `helm template --validate` still works? I think we can add that to a CI system instead than adding notes about doing it manually.

yuvipanda:
> Do you mean to verify that something like `helm template --validate` still works - which mean to validate rendered templates against a k8s-apiserver for validity? It is already part of the ci system.

Fair enough! I just don't want to accidentally end up in a situation where refactoring in z2jh breaks binderhub. But the CI system means there will be an error thrown if refactoring there actually breaks binderhub.

PR #1334, opened 2021-07-09
Update DOCKER_REGISTRY_URL to the recommended one

MridulS:
The recommended hostname ( https://github.com/docker/hub-feedback/issues/2113#issuecomment-866298153 ) for docker registry is `https://registry-1.docker.io`.

This was fixed on mybinder deployment https://github.com/jupyterhub/mybinder.org-deploy/pull/1937, updating the URL here so people depending on dockerhub in their own binderhub deployments don't hit this issue.

### DIFF
--- a/binderhub/registry.py
+++ b/binderhub/registry.py
@@ -128,7 +128,7 @@ class DockerRegistry(LoggingConfigurable):
         url = urlparse(self.url)
         if ("." + url.hostname).endswith(".gcr.io"):
             return "https://{0}/v2/token?service={0}".format(url.hostname)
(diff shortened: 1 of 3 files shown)


yuvipanda:
Thanks, @MridulS. Once merged, can we remove the override from mybinder.org-deploy?

MridulS:
Yeah we could revert https://github.com/jupyterhub/mybinder.org-deploy/pull/1937, let me know if I should do it :)

yuvipanda:
@MridulS that would be great!

PR #1335, opened 2021-07-10
Add button to view raw logs during a build

manics:
Adds a `view raw` button to view the current build logs as plain text. Suggested by @bollwyvl in https://discourse.jupyter.org/t/binder-fails-to-compile-pygraphviz/9912/5 (at the end of the post).

This works by storing all log messages in an array, and creating a URL Blob object containing the logs when `view raw` is clicked. Note this won't update with new logs. You have to click `view logs` again.

Using `FakeBuild`:
image

Clicking on `view raw` opens a new tab:
image

I've also tested this in Chrome.

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -165,7 +165,7 @@ function build(providerSpec, log, path, pathType) {
 
   image.onStateChange('*', function(oldState, newState, data) {
     if (data.message !== undefined) {
-      log.write(data.message);
+      log.writeAndStore(data.message);
       log.fit();
     } else {
       console.log(data);
@@ -225,6 +225,7 @@ function setUpLog() {
     convertEol: true,
     disableStdin: true
   });
+  var logMessages = [];
 
   log.open(document.getElementById('log'), false);
   log.fit();
@@ -235,12 +236,12 @@ function setUpLog() {
 
   var $panelBody = $("div.panel-body");
   log.show = function () {
-    $('#toggle-logs button').text('hide');
+    $('#toggle-logs button.toggle').text('hide');
     $panelBody.removeClass('hidden');
   };
 
   log.hide = function () {
-    $('#toggle-logs button').text('show');
+    $('#toggle-logs button.toggle').text('show');
     $panelBody.addClass('hidden');
   };
 
@@ -252,7 +253,20 @@ function setUpLog() {
     }
   };
 
+  $('#view-raw-logs').on('click', function(ev) {
+    const blob = new Blob([logMessages.join('')], { type: 'text/plain' });
+    this.href = window.URL.createObjectURL(blob);
+    // Prevent the toggle action from firing
+    ev.stopPropagation();
+  });
+
   $('#toggle-logs').click(log.toggle);
+
+  log.writeAndStore = function (msg) {
+    logMessages.push(msg);
+    log.write(msg);
+  }
+
   return log;
 }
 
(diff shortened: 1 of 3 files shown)


bollwyvl:
Best feature I didn't have to write! It might also be possible to pull it directly out of the xterm, but it looks like it only stores the last 1000 lines, so this might be better.

It would be _slightly_ more ergonomic if it showed a useful filename, but wouldn't consider that a blocker, and it's apparently annoying to make it do the thing without actually pushing it to the DOM, which seems silly.

binderhub/static/js/index.js
```diff
@@ -252,7 +253,20 @@ function setUpLog() {
     }
   };
 
+  $('#view-raw-logs').on('click', function(ev) {
+    const blob = new Blob([logMessages.join('')], { type: 'text/plain' });
+    this.href = window.URL.createObjectURL(blob);
+    // Prevent the toggle action from firing
+    ev.stopPropagation();
+  });
```
consideRatio, line 261: Is this understanding correct?
1. Line 257: The blob constant contain all log messages concatenated into a string, and of a declared type of text/plain.
2. Line 258: An data based URL destination is via createObjectURL.
3. Line 258: The link named with `id=view-raw-logs` gets its `href` field automatically set to the data based URL.

I don't understand the following though:
- Line 259-260: Hmmm... This is a click event registered on a single link element, isn't it? Would that event propagate/bubble to the toggle button somehow? What confuses me is that the toggle button isn't a parent element of the `a` link that the `view-raw-logs` element is. I'm not sure how various click events can triggered registered functions in general, but my understanding of `stopPropegation` is to mark the click event as handled somehow which I guess it is and then perhaps that's a sensible practice - but unrelated to the toggle button.

consideRatio:
Nice work @manics! What I think I understand LGTM, but I'm not sure about a not so critical part that I don't understand.

manics:
Having a random `blob:` URL isn't great. An alternative is to force the file to be downloaded:
```diff
diff --git a/binderhub/static/js/index.js b/binderhub/static/js/index.js
index 289e392..70c160f 100755
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -256,6 +256,7 @@ function setUpLog() {
   $('#view-raw-logs').on('click', function(ev) {
     const blob = new Blob([logMessages.join('')], { type: 'text/plain' });
     this.href = window.URL.createObjectURL(blob);
+    this.download = 'log.txt';
     // Prevent the toggle action from firing
     ev.stopPropagation();
   });
```
image

I can't decide what's better. What do you think?

binderhub/static/js/index.js
manics: The toggle function is attached to the header, not the button:
- https://github.com/jupyterhub/binderhub/blob/938324b95becd269da56decb4aff434af5fd9eda/binderhub/static/js/index.js#L255
- https://github.com/jupyterhub/binderhub/blob/938324b95becd269da56decb4aff434af5fd9eda/binderhub/templates/index.html#L149

manics:
Just realised there's also a `binderhub/templates/loading.html` that needs to be updated

binderhub/static/js/index.js
consideRatio: Ah, thanks!

consideRatio:
> I can't decide what's better. What do you think?

I'm used to opening an URL where i see a lot of blank text that I can select-all -> copy on. I think that because I'm used to that, like GitHub's "view raw" button. Was that one of the options via the blob url? Then I'm +1 for that!

manics:
That's how it behaves at the moment- you'll get a new tab with a random in-browser blob URL.

consideRatio (approved):
This LGTM and make sense!

PR #1337, opened 2021-07-15
Let z2jh seed jupyterhub.hub.services.binder.apiToken

consideRatio:
This feature helps user not need to specify a randomly generated token. This token is needed still, but can now be generated and stored in a k8s automatically by z2jh 1.1.0+ (bumped in #1340) that includes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2312.

The actual change to make this work is below and to simply stop passing an api token explicitly. So, when mybinder.org-deploy for example starts using this version, or at a later time, they can simply delete the apiToken from the git config.

```diff
         - name: JUPYTERHUB_API_TOKEN
           valueFrom:
             secretKeyRef:
-              name: binder-secret
-              key: "binder.hub-token"
+              name: "{{ include "jupyterhub.hub.fullname" . }}"
+              key: hub.services.binder.apiToken
```

### Why not to worry about the one test failure

The helm diff test fails because it use the same config to render templates before and after this PR, but before this PR you were required to pass an api token explicitly. In the future, the before and after comparison will succeed as both the "before" and "after" will accept omitting the api token.

### DIFF
--- a/doc/zero-to-binderhub/setup-binderhub.rst
+++ b/doc/zero-to-binderhub/setup-binderhub.rst
@@ -27,30 +27,11 @@ the container registry. For more information on getting a registry password, see
 :ref:`setup-registry`. We'll copy/paste the contents of this file in the steps
 below.
 
-Create two random tokens by running the following commands then copying the
-outputs.::
-
-    openssl rand -hex 32
-    openssl rand -hex 32
-
-.. note::
-
-   This command is run **twice** because we need two different tokens.
-
 Create ``secret.yaml`` file
 ---------------------------
 
-Create a file called ``secret.yaml`` and add the following::
-
-  jupyterhub:
-    hub:
-      services:
-        binder:
-          apiToken: "<output of FIRST `openssl rand -hex 32` command>"
-    proxy:
-      secretToken: "<output of SECOND `openssl rand -hex 32` command>"
-
-Next, we'll configure this file to connect with our registry.
+Create a file called ``secret.yaml``, we'll set sensitive values to pass to our
+Helm chart in this file.
 
 If you are using ``gcr.io``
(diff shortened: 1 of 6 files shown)


doc/zero-to-binderhub/setup-binderhub.rst
```diff
@@ -27,30 +27,11 @@ the container registry. For more information on getting a registry password, see
...
-  jupyterhub:
-    hub:
-      services:
-        binder:
-          apiToken: "<output of FIRST `openssl rand -hex 32` command>"
-    proxy:
-      secretToken: "<output of SECOND `openssl rand -hex 32` command>"
-
-Next, we'll configure this file to connect with our registry.
+Create a file called ``secret.yaml``, we'll set sensitive values to pass to our
```
yuvipanda, line 33:
```suggestion
If you have any sensitive config values to set, create a file called ``secret.yaml`` to store them in.
By default, you can leave this empty.
```

doc/zero-to-binderhub/setup-binderhub.rst
```diff
@@ -27,30 +27,11 @@ the container registry. For more information on getting a registry password, see
...
+Helm chart in this file.
```
yuvipanda, line 34:
```suggestion
```

doc/zero-to-binderhub/setup-binderhub.rst
yuvipanda: Not sure of the exact wording, but let's clarify that we don't expect anything in this file by default?

yuvipanda: Actually I just got confused by rst heading markers, we still need stuff here. Ignore me.

yuvipanda:
Thanks, @consideRatio!

PR #1342, opened 2021-07-26
custom registry spec

ltetrel:
This PR add some missing details on how to properly configure the cluster nodes when using a custom/private registry.

### DIFF
--- a/doc/zero-to-binderhub/setup-binderhub.rst
+++ b/doc/zero-to-binderhub/setup-binderhub.rst
@@ -263,7 +263,7 @@ Second, you will need to instruct BinderHub about the token URL::
 If you setup your own local registry using
 `native basic HTTP authentication <https://docs.docker.com/registry/deploying/#native-basic-auth>`__
 (htpasswd), you can set ``token_url`` to ``None``.
-You will also need to configure the kubernetes nodes to authentificate to your private registry.
+You will also need to configure the kubernetes nodes to authenticate to your private registry.
 Instructions can be followed on the `official kubernetes documentation <https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry>`_.
 
 .. note::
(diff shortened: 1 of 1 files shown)


doc/zero-to-binderhub/setup-binderhub.rst
```diff
@@ -263,6 +263,8 @@ Second, you will need to instruct BinderHub about the token URL::
 If you setup your own local registry using
 `native basic HTTP authentication <https://docs.docker.com/registry/deploying/#native-basic-auth>`__
 (htpasswd), you can set ``token_url`` to ``None``.
+You will also need to configure the kubernetes nodes to authentificate to your private registry.
```
manics, line 266:
```suggestion
You will also need to configure the kubernetes nodes to authenticate to your private registry.
```

doc/zero-to-binderhub/setup-binderhub.rst
```diff
@@ -263,6 +263,8 @@ Second, you will need to instruct BinderHub about the token URL::
 If you setup your own local registry using
 `native basic HTTP authentication <https://docs.docker.com/registry/deploying/#native-basic-auth>`__
 (htpasswd), you can set ``token_url`` to ``None``.
+You will also need to configure the kubernetes nodes to authenticate to your private registry.
+Instructions can be followed on the `official kubernetes documentation <https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry>`_.
```
consideRatio, line 267: I think the statement is a bit too specific, because what's really needed is that all pods working with the images needs to be authenticated to pull them, and builder pods needs to be authenticated to push to them.

To configure the pods to have access, we can make use of `imagePullSecrets`. I don't remember how this helm chart does it any more at this point, but if this chart provides such configuration, that is also an option.

See https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod for details about this.

ltetrel: I suggested this change because in my case it was by far the easiest solution (for an already spawned cluster). But if you think it is not necessary we can close this PR.

consideRatio:
Thank you @ltetrel for your PR!

I'll opt to close this because I think the proper recommendation is to configure the pods that require access to the registry to have a imagePullSecrets configured instead of configuring the nodes to have access. I believe that is the minimal requirement for proper function, which seems like a safer recommendation than configuring the nodes themselves which could expose the credentials to other pods in the k8s cluster unrelated to this deployment of the Helm chart.

PR #1348, opened 2021-08-24
CPU requests and limits for build pod

noroutine:
This PR makes CPU request and limit for build pod to be configurable in config, just like memory settings
Trait was implemented to be flexible in take values and accepts integer, floats, string representations of those, as well as millicore spec as per https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu


betatim:
Thanks for this PR! I was surprised to see that we don't already have this, which means this is for sure useful :D

Looking and thinking about the code now.

Some immediate thoughts:
1. will this work when BinderHub is not running on kubernetes? Recently @manics and/or @yuvipanda did some work on making that possible. Would be good to know how this interacts. If it would require extra work it doesn't need to be addressed here (can be a new PR)
2. do we need to have a new traitlet type? The reason I'm asking is because that adds a lot of code for which we'd need some tests, etc. We did the same for the memory spec, so we probably need it. But who knows maybe we got smarter since then and now no longer need it?
3. there are no tests :( can we copy from/be inspired by the memory spec test suite?

noroutine:
Hey guys

I've just pushed some tests to cover that trait, let me know if i can do any more from my side to get this through!

Kind regards

consideRatio (approved):
LGTM, nice!

consideRatio:
@manics @betatim tests has been added and the code looks good to me.

I also think that the config is reasonable in a non-k8s context, it could be relevant to have such requests/limits there as well even though maybe it doesn't - and then it can be documented they will be ignored for such environments later.

Go for merge?

binderhub/utils.py
```diff
@@ -42,6 +42,86 @@ def rendezvous_rank(buckets, key):
...
+
+    """
+
+    # Default to allowing None as a value
+    allow_none = True
+
+    def validate(self, obj, value):
+        """
+        Validate that the passed in value is a valid cpu specification
+        in the K8s CPU meaning.
```
betatim, line 60:
```suggestion
        using Kubernetes formatting.
```

betatim (approved):

betatim:
Not sure why the tests are failing :(

manics:
If you go to the logs for the failing check: https://github.com/jupyterhub/binderhub/pull/1348/checks?check_run_id=3697325231
and expand `Show hub logs` you'll see the Hub errors:
```
[E 210924 07:41:15 web:1789] Uncaught exception GET /services/binder/build/gh/binderhub-ci-repos/cached-minimal-dockerfile/HEAD (::1)
    HTTPServerRequest(protocol='http', host='localhost:8000', method='GET', uri='/services/binder/build/gh/binderhub-ci-repos/cached-minimal-dockerfile/HEAD', version='HTTP/1.1', remote_ip='::1')
      File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/binderhub/builder.py", line 402, in get
        self.build = build = BuildClass(
    TypeError: __init__() got an unexpected keyword argument 'cpu_limit'
```
Work on https://github.com/jupyterhub/binderhub/issues/1318 went a lot faster than expected! The new parameter needs to be added to https://github.com/jupyterhub/binderhub/blob/a8e3444a20d01b1937d3a92d59fb3651e46b5b58/binderhub/build_local.py#L65-L93

The alternative is to add this as a configurable Trailets on the Build class when it's refactored as discussed in https://github.com/jupyterhub/binderhub/issues/1318#issuecomment-922824039. That way you can avoid adding a parameter to BinderHub, the configurable should be passed down to the class, and since it can be a configurable on e.g. KubernetesBuilder it can use Kubernetes specific config. The downside is that'll delay the addition of this new feature.

However there's another issue. The build pod runs repo2docker, which does the build in another Docker container (by communicating directly with the Docker daemon through a socket). Since the Docker build takes place outside K8S the K8S pod limits won't apply, they somehow need to be passed to repo2docker.

Based on this comment:
https://github.com/jupyterhub/binderhub/blob/a8e3444a20d01b1937d3a92d59fb3651e46b5b58/binderhub/build.py#L132-L136
a memory request is useful because it guarantees a certain amount of memory for the build pod. If I understand correctly the build pod won't use that memory so most of it will be free on the K8S node, which means the Docker container where the build occurs can use it. `memory_limit` has an effect because it's passed down to repo2docker, which then sets the memory limit using the Docker build API:
https://github.com/jupyterhub/binderhub/blob/a8e3444a20d01b1937d3a92d59fb3651e46b5b58/binderhub/build.py#L193-L195
https://github.com/jupyterhub/repo2docker/blob/3eef69f7454ddb9086af38b2ece5f632af05920f/repo2docker/app.py#L161-L169

This means `cpu_request` will effectively reserve some CPU for the build, but `cpu_limit` will have no effect unless it's also added to repo2docker.

noroutine:
Hi gentlemen

Unfortunate it failed again.

I was a bit out on leave, and checking now above, trying to understand what best action on making this work again.
Will dive into this again this week, but have to understand what shall I do regarding repo2docker code

Thank you

manics:
@noroutine You'll need to add cpu limit support to repo2docker, then you can come back here and pass the appropriate flags when repo2docker is run.

noroutine:
Thank you @manics

PR #1350, opened 2021-09-03
Remove unnecessary env var pass-throughs in binderhub_config.py

yuvipanda:
- hub api token was already being picked up from the env var by traitlet
- build_namespace is now also picked up from the env var by traitlet

### DIFF
--- a/helm-chart/binderhub/files/binderhub_config.py
+++ b/helm-chart/binderhub/files/binderhub_config.py
@@ -78,9 +78,6 @@ if allow_origin:
         }
     })
 
-if os.getenv('BUILD_NAMESPACE'):
-    c.BinderHub.build_namespace = os.environ['BUILD_NAMESPACE']
-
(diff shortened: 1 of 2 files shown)


helm-chart/binderhub/files/binderhub_config.py
```diff
@@ -4,8 +4,6 @@
 from urllib.parse import urlparse
 import yaml
 
-c.BinderHub.hub_api_token = os.environ['JUPYTERHUB_API_TOKEN']
-
```
consideRatio, line 8:
```python
    hub_api_token = Unicode(
        help="""API token for talking to the JupyterHub API""",
        config=True,
    )
    @default('hub_api_token')
    def _default_hub_token(self):
        return os.environ.get('JUPYTERHUB_API_TOKEN', '')
```

consideRatio (approved):
LGTM nice cleanup!

PR #1351, opened 2021-09-03
Move custom BinderSpawner config to traitlets

yuvipanda:
Currently, we have two config variables that are specified in `jupyterhub.custom`
so they can be read by code in the JupyterHub custom config. However, we can
do this in a much more structured way by using actual traitlets. Better validation,
documentation and 'cleaner'. 

This also helps move config away from z2jh specific things (`z2jh.get_config`) 
towards the more generic traitlets.

### DIFF
--- a/helm-chart/binderhub/schema.yaml
+++ b/helm-chart/binderhub/schema.yaml
@@ -262,17 +262,15 @@ properties:
       BinderHub Helm chart.
     required: [hub]
     properties:
-      custom:
-        type: object
-        description: |
-          DEPRECATED
       hub:
         type: object
         additionalProperties: true
+        required: [services]
         properties:
           services:
             type: object
             additionalProperties: true
+            required: [binder]
             properties:
               binder:
                 type: object
@@ -538,6 +536,6 @@ properties:
   # Deprecated
   cors:
     type: object
-    additionalProperties: false
+    additionalProperties: true
     description: |
       DEPRECATED.
--- a/helm-chart/binderhub/schema.yaml
+++ b/helm-chart/binderhub/schema.yaml
@@ -255,17 +255,6 @@ properties:
       manifest as either the binder pod going into `Error` or `CrashLoopBackoff` states, or in
       some special cases, the binder pod running but... just doing very random things. Be careful!
 
-  cors: &cors-spec
-    type: object
-    additionalProperties: false
-    description: |
-      TODO
-    properties:
-      allowOrigin:
-        type: [string, "null"]
-        description: |
-          TODO
-
   jupyterhub:
     type: object
     additionalProperties: true
@@ -274,11 +263,6 @@ properties:
       BinderHub Helm chart.
     required: [hub]
     properties:
-      custom:
-        type: object
-        additionalProperties: true
-        properties:
-          cors: *cors-spec
       hub:
         type: object
         additionalProperties: true
--- a/helm-chart/binderhub/schema.yaml
+++ b/helm-chart/binderhub/schema.yaml
@@ -279,10 +279,6 @@ properties:
         additionalProperties: true
         properties:
           cors: *cors-spec
-          binderauth_enabled:
-            type: boolean
-            description: |
-              TODO
       hub:
         type: object
         additionalProperties: true
--- a/helm-chart/binderhub/schema.yaml
+++ b/helm-chart/binderhub/schema.yaml
@@ -25,7 +25,6 @@ required:
   - service
   - config
   - extraConfig
-  - cors
   - jupyterhub
   - deployment
   - dind
@@ -263,6 +262,10 @@ properties:
       BinderHub Helm chart.
     required: [hub]
     properties:
+      custom:
+        type: object
+        description: |
+          DEPRECATED
       hub:
         type: object
         additionalProperties: true
@@ -322,7 +325,7 @@ properties:
         type: array
         description: |
           List of additional initContainers.
-          
+
           See the [Kubernetes
           docs](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/)
           for more info.
@@ -531,3 +534,10 @@ properties:
       This is not a supported configuration yet by this Helm chart, but may be
       in the future. This schema entry was added because
       jupyterhub/mybinder.org-deploy configured a Networkpolicy resource here.
+
+  # Deprecated
+  cors:
+    type: object
+    additionalProperties: false
+    description: |
+      DEPRECATED.
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -67,7 +67,7 @@ jupyterhub:
 
         # image & token are set via spawn options
         from kubespawner import KubeSpawner
-        from traitlet import Bool, Unicode
+        from traitlets import Bool, Unicode
 
         class BinderSpawner(KubeSpawner):
             auth_enabled = Bool(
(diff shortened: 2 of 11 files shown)


yuvipanda:
@consideRatio you might like this, given it removes entries from custom config and moves them towards something with a better defined schema :)

helm-chart/binderhub/schema.yaml
```diff
@@ -255,17 +255,6 @@ properties:
...
-    type: object
-    additionalProperties: false
-    description: |
-      TODO
-    properties:
-      allowOrigin:
-        type: [string, "null"]
-        description: |
-          TODO
-
```
consideRatio, line 268: This is associated with line 28 in the schema, writing that the `cors` field is required. Remove it as well and anyone with this config will get an error, making anyone doing a migration get a hard error.

Another option, is to make it so that you let the schema allow this, and let NOTES.txt do a `print "migration error message" | fail` which can be more helpful.

consideRatio (changes requested):
I think this is an absolutely great refactoring, it embeds logic in the Helm chart to the Python code in BinderHub application running in the `binder` pod and the BinderSpawner running in the `hub` pod. Nice!

There is a question about how to handle this as a breaking change though. I wouldn't hesitate merging if you went for a NOTES.txt rendered message if someone was passing anything to the deprecated Helm chart config values still. To do that, first make sure the schema tolerates that. In practice, I figure:

1. Let values.yaml define a blank entry still.
2. Let schema still list it under properties so `additionalProperties: false` can be used, but remove it from being an required entry.
3. Let NOTES.txt check if the dictionary is truthy, then print a message and pipe it to `fail`.
   I'd like to see the [pattern used in Z2JH's NOTES.txt](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/4833026fc30909716bb2c0f059590cd56e02aa22/jupyterhub/templates/NOTES.txt#L89-L155) be reused for this as it makes it very clear.

yuvipanda:
@consideRatio I think I've done the changes you have asked for! Very nice way to make breaking changes, thanks for building out the system and pointing me to it!

helm-chart/binderhub/schema.yaml
```diff
@@ -551,3 +534,10 @@ properties:
       This is not a supported configuration yet by this Helm chart, but may be
       in the future. This schema entry was added because
       jupyterhub/mybinder.org-deploy configured a Networkpolicy resource here.
+
+  # Deprecated
+  cors:
+    type: object
+    additionalProperties: false
```
consideRatio, line 541:
```suggestion
    additionalProperties: true
```

Without this being true, a user won't be able to get informed via our NOTES.txt, because helm will fail even before that if someone passed something nested under this.

helm-chart/binderhub/schema.yaml
```diff
@@ -551,3 +534,10 @@ properties:
...
+    description: |
+      DEPRECATED.
```
consideRatio, line 543:
```suggestion
      DEPRECATED.

```

consideRatio (commented):
Minor fix, otherwise LGTM.

consideRatio:
@yuvipanda i reviewed and then i saw a force push making my review potentially irrelevant, not sure what changed - I'll assume I won't need to re-view it and there was just some small detail.

helm-chart/binderhub/values.yaml
```diff
@@ -39,48 +40,63 @@ config:
...
         from tornado import web
 
-        # get custom config from values.custom
-        import z2jh
-        cors = z2jh.get_config('custom.cors', {})
-        auth_enabled = z2jh.get_config('custom.binderauth_enabled', False)
-
         # image & token are set via spawn options
         from kubespawner import KubeSpawner
+        from traitlet import Bool, Unicode
```
consideRatio, line 70:
```suggestion
        from traitlets import Bool, Unicode
```

consideRatio (approved):

helm-chart/binderhub/values.yaml
```diff
@@ -39,48 +40,63 @@ config:
...
-    cors: *cors
-    binderauth_enabled: false
   rbac:
     enabled: true
   hub:
     config:
       JupyterHub:
         authenticator_class: nullauthenticator.NullAuthenticator
+      BinderSpawner:
+        auth_enabled: false
```
MridulS, line 63: A quick clarification regarding this: Will this break downstream packages which define custom Spawners and not use BinderSpawner?

consideRatio: Depends, what do they inherit from? If they inherit from BinderSpawner, everything should be fine I think.

consideRatio: This config is like writing `c.BinderSpawner.auth_enabled = False` and will influence a MyCustomBinderSpawner class as well as long as it inherits from BinderSpawner

MridulS: Thanks! We [inherit from `KubeSpawner`](https://github.com/gesiscss/persistent_binderhub/blob/2c2f87ac567d2c46b8e8848d76700e3c1da51a63/persistent_binderhub/files/jupyterhub/persistent_bhub_config.py#L39) so need to fix that up for Persistent BinderHub :)

consideRatio (changes requested):
Test failure is relevant!

yuvipanda:
@consideRatio test failures fixed!

helm-chart/binderhub/schema.yaml
consideRatio: @yuvipanda i dont think you tested the deprecation logic, this must allow unlisted properties as well i think!

Can you also set the additionalProperties to true explicitly on the other deprecated entry for consistency of having an explicit choice for all our objects?

consideRatio: Doing this, and you can test the deprecation logic

tools/generate-schema...
helm template bh helm-chart/binderhub --set cors.stuff=...

consideRatio:
I went ahead and tested the schema after making some tweaks. This now LGTM.

```
~/dev/jupyterhub/binderhub (cors-allow)$ helm template bh helm-chart/binderhub/ --set jupyterhub.custom.cors.allowedOrigin=asdf
Error: template: binderhub/templates/NOTES.txt:186:4: executing "binderhub/templates/NOTES.txt" at <fail (print $breaking_title $breaking)>: error calling fail: 

#################################################################################
######   BREAKING: The config values passed contained no longer accepted    #####
######             options. See the messages below for more details.        #####
######                                                                      #####
######             To verify your updated config is accepted, you can use   #####
######             the `helm template` command.                             #####
#################################################################################

RENAMED: jupyterhub.custom.cors.allowedOrigin has been renamed to jupyterhub.hub.config.BinderSpawner.cors_allow_origin

Use --debug flag to render out invalid YAML
~/dev/jupyterhub/binderhub (cors-allow)$ helm template bh helm-chart/binderhub/ --set cors.allowedOrigin=asd
Error: template: binderhub/templates/NOTES.txt:186:4: executing "binderhub/templates/NOTES.txt" at <fail (print $breaking_title $breaking)>: error calling fail: 

#################################################################################
######   BREAKING: The config values passed contained no longer accepted    #####
######             options. See the messages below for more details.        #####
######                                                                      #####
######             To verify your updated config is accepted, you can use   #####
######             the `helm template` command.                             #####
#################################################################################

RENAMED: cors.allowedOrigin has been renamed to config.BinderHub.cors_allow_origin

Use --debug flag to render out invalid YAML
```

helm-chart/binderhub/schema.yaml
consideRatio: Resolved by my small schema fixes commit.

consideRatio (approved):

yuvipanda:
Thanks a lot, @consideRatio! Feels good to be writing code for binderhub again :)

PR #1352, opened 2021-09-03
Make registry class configurable, add FakeRegistry

manics:
Follow-up to https://github.com/jupyterhub/binderhub/pull/1323#issuecomment-877692252

This lets you run the UI only without a working Docker daemon, following the instructions in https://github.com/jupyterhub/binderhub/blob/master/CONTRIBUTING.md#develop-user-interface

Without this Binderhub tries to query the local Docker Daemon for the image and therefore fails if you don't have Docker running:
https://github.com/jupyterhub/binderhub/blob/d8e7128b684140db80d533a3b794ebca786c40f7/binderhub/builder.py#L352-L354

This also includes some fixes for the FakeBuild class since the strings do not match the expected enum values (follow-up to https://github.com/jupyterhub/binderhub/pull/1325)

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -49,7 +49,7 @@ from .launcher import Launcher
 from .log import log_request
 from .ratelimit import RateLimiter
 from .repoproviders import RepoProvider
-from .registry import DockerRegistry
+from .registry import DockerRegistry, FakeRegistry
 from .main import MainHandler, ParameterizedMainHandler, LegacyRedirectHandler
 from .repoproviders import (GitHubRepoProvider, GitRepoProvider,
                             GitLabRepoProvider, GistRepoProvider,
@@ -256,6 +256,16 @@ class BinderHub(Application):
         The class used to build repo2docker images.
         
         Must inherit from binderhub.build.Build
+        """,
+        config=True
+    )
+
+    registry_class = Type(
+        DockerRegistry,
+        help="""
+        The class used to Query a Docker registry.
+
+        Must inherit from binderhub.registry.DockerRegistry
         """,
         config=True
     )
@@ -718,8 +728,8 @@ class BinderHub(Application):
             FileSystemLoader(template_paths)
         ])
         jinja_env = Environment(loader=loader, **jinja_options)
-        if self.use_registry and self.builder_required:
-            registry = DockerRegistry(parent=self)
+        if self.use_registry:
+            registry = self.registry_class(parent=self)
         else:
             registry = None
 
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -535,7 +535,7 @@ class FakeBuild(Build):
     Fake Building process to be able to work on the UI without a running Minikube.
     """
     def submit(self):
-        self.progress(ProgressEvent.Kind.BUILD_STATUS_CHANGE, 'Running')
+        self.progress(ProgressEvent.Kind.BUILD_STATUS_CHANGE, ProgressEvent.BuildStatus.RUNNING)
         return
 
     def stream_logs(self):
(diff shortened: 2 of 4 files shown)


binderhub/app.py
```diff
@@ -718,8 +728,8 @@ def initialize(self, *args, **kwargs):
             FileSystemLoader(template_paths)
         ])
         jinja_env = Environment(loader=loader, **jinja_options)
-        if self.use_registry and self.builder_required:
-            registry = DockerRegistry(parent=self)
+        if self.use_registry:
+            registry = self.registry_class(parent=self)
```
consideRatio, line 732: Is the builder_required traitlet still relevant, or should it be deprecated?

```
    # FIXME: Come up with a better name for it?
    builder_required = Bool(
        True,
        config=True,
        help="""
        If binderhub should try to continue to run without a working build infrastructure.
        Build infrastructure is kubernetes cluster + docker. This is useful for pure HTML/CSS/JS local development.
        """
    )
```

manics: It's still used to control whether the K8s API is setup:
https://github.com/jupyterhub/binderhub/blob/d8e7128b684140db80d533a3b794ebca786c40f7/binderhub/app.py#L697-L702
https://github.com/jupyterhub/binderhub/blob/d8e7128b684140db80d533a3b794ebca786c40f7/binderhub/app.py#L878-L879

However, it might be nice to move this into the `Build` class, but it's not clear to me how easy that is.

yuvipanda: The k8s API is set up for quota checks - I think we should have a `QuotaProvider` there eventually. I started working on one a while ago...


```
import json
from .utils import KUBE_REQUEST_TIMEOUT

from traitlets.config import LoggingConfigurable
from traitlets import Unicode

class QuotaProvider(LoggingConfigurable):
    def can_run(self):
        pass

class KubernetesQuotaProvider(QuotaProvider):

    build_namespace = Unicode(
        config=True
    )
    kubernetes_client = None

        from kubernetes.client import CoreV1Api
        import kubernetes.config
        # Keep only a single CoreV1Api
        if KubernetesQuotaProvider.kubernetes_client is None:
            KubernetesQuotaProvider.kubernetes_client = CoreV1Api()
            try:
                kubernetes.config.load_incluster_config()
            except kubernetes.config.ConfigException:
                kubernetes.config.load_kube_config()

    def get_running_servers(self, image_name):
        # Fetch info on currently running users *only* if quotas are set
        matching_pods = 0
        total_pods = 0

        # TODO: run a watch to keep this up to date in the background
        resp = KubernetesQuotaProvider.kubernetes_cliens.list_namespaced_pod(
            self.settings['build_namespace'],
            label_selector='app=jupyterhub,component=singleuser-server',
            _request_timeout=KUBE_REQUEST_TIMEOUT,
            _preload_content=False,
        )
        pods = json.loads(resp.read())
        for pod in pods["items"]:
            total_pods += 1
            for container in pod["spec"]["containers"]:
                # is the container running the same image as us?
                # if so, count one for the current repo.
                image = container["image"].rsplit(":", 1)[0]
                if image == image_name:
                    matching_pods += 1
                    break
        if matching_pods >= quota:
            app_log.error("%s has exceeded quota: %s/%s (%s total)",
                self.repo_url, matching_pods, quota, total_pods)
            await self.fail("Too many users running %s! Try again soon." % self.repo_url)
            return

        if matching_pods >= 0.5 * quota:
            log = app_log.warning
        else:
            log = app_log.info
        log("Launching pod for %s: %s other pods running this repo (%s total)",
            self.repo_url, matching_pods, total_pods)
```

yuvipanda:
Thanks, @manics!

PR #1353, opened 2021-09-05
WIP: local dev setup without Kubernetes

manics:
Closes https://github.com/jupyterhub/binderhub/issues/1318

- Add a build `LocalRepo2dockerBuild` class that runs `repo2docker` as a command line program (not in Docker)
- Run JupyterHub with DockerSpawner

Opening this so we can discuss whether we want to support any of this, what we should refactor to reduce duplication, and what should be in this repo or somewhere else.

I split the options handling out of`Build.get_cmd` so that I could easily add extra arguments by, e.g. repo2docker-podman needs `--engine podman` added to the R2D command line in a `LocalRepo2dockerBuild` subclass:
```python
class LocalRepo2dockerPodmanBuild(LocalRepo2dockerBuild):
    def get_r2d_cmd_options(self):
        return super().get_r2d_cmd_options() + [
            '--engine', 'podman',
        ]
```


binderhub/build_local.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    Call given command, yielding output line by line if capture=True.
+
+    Must be yielded from.
+    """
+    if capture:
+        kwargs["stdout"] = subprocess.PIPE
+        kwargs["stderr"] = subprocess.STDOUT
+
+    proc = subprocess.Popen(cmd, **kwargs)
```
yuvipanda, line 35: Should we use `asyncio.subprocess` instead?

binderhub/build_local.py
```diff
@@ -0,0 +1,227 @@
...
+class LocalRepo2dockerBuild(Build):
+    """Represents a build of a git repository into a docker image.
+
+    This runs a build using the repo2docker command line tool.
+    """
+
+    def __init__(
+        self,
+        q,
+        api,
```
yuvipanda, line 82: Each type of `Build` will probably have its own set of config params - `api` is only for k8s, for example. I'd say that we should probably manage k8s API object as a singleton on the class or something, and so we can stop having to pass that through. Traitlet config specific to k8s can hence also be moved to the k8s builder?

yuvipanda (commented):
I LOVE THIS! THANK YOU SO MUCH, YOU HAVE MADE ME VERY HAPPY TODAY!

We should figure out how we can pass backend specific config through in a clean way, but otherwise this structure looks great to me.

yuvipanda:
If we do this fully, it's not just a dev setup - it can be used to run binderhub without k8s (if we setup a dockerspawner based hub for it to spawn into). Is that correct?

binderhub/build_local.py
manics: I'm not sure. `build.submit` is run in a `ThreadPoolExecutor`
https://github.com/jupyterhub/binderhub/blob/be94c46009c970617be45e27c9a6ed95203b4af0/binderhub/app.py#L714-L715
https://github.com/jupyterhub/binderhub/blob/be94c46009c970617be45e27c9a6ed95203b4af0/binderhub/builder.py#L427
So does that mean `asyncio.subprocess` wouldn't do anything?

manics:
Yes, it's already working with DockerSpawner :smile:

See the second part of `testing/local-binder-local-hub/binderhub_config.py` in this PR and also `testing/local-binder-local-hub/README.md`. I included the JupyterHub config there since configuration for other apps is ignored, and it means the api_token is easily shared. Perhaps we could run BinderHub as a JupyterHub managed service?

binderhub/build_local.py
yuvipanda: Ah yeah ok - if it's run in a threadpool it can't be async.

yuvipanda:
> Perhaps we could run BinderHub as a JupyterHub managed service?

I really really want this.

binderhub/build_local.py
yuvipanda: I don't think this needs to block this PR though, we can do this in another PR

testing/local-binder-local-hub/binderhub_config.py
```diff
@@ -0,0 +1,111 @@
...
+## JupyterHub config
+from tornado import web
+
+cors = {}
+auth_enabled = False
+
+from dockerspawner import DockerSpawner
+
+# image & token are set via spawn options
+class LocalContainerSpawner(DockerSpawner):
```
yuvipanda, line 49: I'd like to move this into a mixin too.

manics: I tried... and failed .... https://github.com/jupyterhub/binderhub/pull/1358

yuvipanda:
What else do you think needs to be done before this can be made non WIP?

manics:
I'd like to add a CI job to test the non-k8s deployment.

If we're happy to make improvements in follow-up PRs  I can mark `LocalRepo2dockerBuild` as experimental so that if we change the interface we only have to worry about backwards compatability for the existing `Build` class, and clean up a few other things.

manics:
@yuvipanda I've got BinderHub running as a JupyterHub managed service :smile: https://github.com/jupyterhub/binderhub/pull/1353/commits/4e726548ea155879c8648480a82a56ae75191db1

I've split this out into three PRs so far:
- https://github.com/jupyterhub/binderhub/pull/1358 (the most complicated one)
- https://github.com/jupyterhub/binderhub/pull/1359
- https://github.com/jupyterhub/binderhub/pull/1360

manics:
Getting close.... the last bit of this experimental PR is now in https://github.com/jupyterhub/binderhub/pull/1364

PR #1354, opened 2021-09-09
use jupyterlab for default UI

minrk:
matches filepath behavior (#1288). https://github.com/jupyterhub/repo2docker/pull/1035 changed the default repo2docker behavior, but binderhub overrides the default command, so both the default and filepath urls need to be set here.


consideRatio (approved):

consideRatio:
With #1288 merged, this should absolutely be merged as well I think.

PR #1355, opened 2021-09-09
use jupyterlab for default UI

minrk:
matches filepath behavior

#1354 only took effect when auth was used (i.e. not on mybinder.org)


minrk:
adding jupyterlab to the minimal-dockerfile for CI, now that we test that lab is present

PR #1356, opened 2021-09-09
Call `git ls-remotes` asynchronously

yuvipanda:
subprocess.run blocks the entire process, preventing the
binderhub process from serving any other requests until
it returns. Can easily lock up if used with a very slow
git server.

### DIFF
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -520,7 +520,7 @@ class GitRepoProvider(RepoProvider):
             stdout, stderr = await proc.communicate()
             retcode = await proc.wait()
             if retcode:
-                raise RuntimeError("Unable to run git ls-remote to get the `resolved_ref`: {}".format(stderr))
+                raise RuntimeError("Unable to run git ls-remote to get the `resolved_ref`: {}".format(stderr.decode()))
             if not stdout:
                 return None
             resolved_ref = stdout.decode().split(None, 1)[0]
--- a/binderhub/repoproviders.py
+++ b/binderhub/repoproviders.py
@@ -13,7 +13,7 @@ import os
 import time
 import urllib.parse
 import re
-import subprocess
+import asyncio
 from urllib.parse import urlparse
 
 import escapism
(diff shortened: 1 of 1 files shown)


binderhub/repoproviders.py
```diff
@@ -514,12 +514,16 @@ async def get_resolved_ref(self):
...
-            if result.returncode:
-                raise RuntimeError("Unable to run git ls-remote to get the `resolved_ref`: {}".format(result.stderr))
-            if not result.stdout:
+            proc = await asyncio.create_subprocess_exec(
+                *command, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
+            )
+            stdout, stderr = await proc.communicate()
+            retcode = await proc.wait()
+            if retcode:
+                raise RuntimeError("Unable to run git ls-remote to get the `resolved_ref`: {}".format(stderr))
```
consideRatio, line 523: Does stderr need a .decode() ?

Can we add text=true or similar to the call and not need to uae .decode() at all?

yuvipanda: `asyncio` doesn't support `text=True` or `universal_newlines=True`:

```

self = <_UnixSelectorEventLoop running=False closed=False debug=False>
protocol_factory = <function create_subprocess_exec.<locals>.<lambda> at 0x107ed1280>
program = 'git', stdin = None, stdout = -1, stderr = -1, universal_newlines = False
shell = False, bufsize = 0, encoding = None, errors = None, text = True
args = ('ls-remote', '--', 'https://github.com/jupyterhub/zero-to-jupyterhub-k8s', 'nosuchref')
kwargs = {}

    async def subprocess_exec(self, protocol_factory, program, *args,
                              stdin=subprocess.PIPE, stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE, universal_newlines=False,
                              shell=False, bufsize=0,
                              encoding=None, errors=None, text=None,
                              **kwargs):
        if universal_newlines:
            raise ValueError("universal_newlines must be False")
        if shell:
            raise ValueError("shell must be False")
        if bufsize != 0:
            raise ValueError("bufsize must be 0")
        if text:
>           raise ValueError("text must be False")
E           ValueError: text must be False

/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py:1647: ValueError
```

So we have to manually decode. The `decode` was definitely required here, so thanks for catching that! I've added that.

consideRatio (approved):
This LGTM from what i can tell! Thanks @yuvipanda!

PR #1358, opened 2021-09-10
Move BinderSpawner into a mixin

manics:
Related to https://github.com/jupyterhub/binderhub/pull/1353#pullrequestreview-749113043

Moving this into a mixin will reduce duplication if we support multiple spawners. There are several problems though:
- Multiple inheritance with multiple classes using traitlets (no idea if it works properly or not... though it seemed to work with PodmanSpawner without auth)
- This doesn't work anyway with our current setup because we're depending on the Z2JH chart whose hub image only includes KubeSpawner, it doesn't include any of the BinderHub python files, so we can't import the mixin.


helm-chart/binderhub/values.yaml
```diff
@@ -63,83 +63,14 @@ jupyterhub:
         auth_enabled: false
     extraConfig:
       00-binder: |
-        from tornado import web
+        # This must be embedded since the Z2JH hub container won't contain BinderHub
+        from binderhub.binderspawner_mixin import BinderSpawnerMixin
```
MridulS, line 67: Would it make sense to do this embedding during the helm packaging step where the spawner file exist in `helm-chart/binderhub/files/` and is injected using the [`extraFiles` directive of z2jh](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html?highlight=extraFiles#hub-extrafiles) ?

manics: I think that requires passing `--set-file` on the Helm command line, as far as I know it can't be set in `values.yaml`.

Helm does support [embedding the content of files using `Files.Get`](https://helm.sh/docs/chart_template_guide/accessing_files/) which would be ideal and could be added to the Z2JH chart, but unfortunately it only works for files within the same chart- a file can't be passed from the parent to the child chart :disappointed: https://github.com/helm/helm/issues/1892

consideRatio: You can make a configmap part of the binderhub helm chart, then use extraVolumes/extraVolumeMounts though - right? Anything you put in a specific folder will be loaded as well automatically by jupyterhub as config.

Ita documented under hub.extraFiles what that folder was. I recall it was /usr/local/jupyterhub_config.d/ or similar.

manics: Doesn't a configmap still require embedding the py file as yaml?

consideRatio: No, files.get!

MridulS: BTW this is pretty much what we do with the [PersistentBinderSpawner class](https://github.com/gesiscss/persistent_binderhub/blob/master/persistent_binderhub/files/jupyterhub/persistent_bhub_config.py) (embed with  [extraVolumes/extraVolumeMounts](https://github.com/gesiscss/persistent_binderhub/blob/6b1b7995241a9f0cdb7a74713a889e83beffe57a/persistent_binderhub/values.yaml#L89-L93)) too.

MridulS: Also I was looking into this sometime back, and I was able to make the tests happy with https://github.com/jupyterhub/binderhub/compare/master...MridulS:persistent_bhub (the final aim of the branch is to send in the PR which augments the BinderSpawner to work with Persistence too 😄)

consideRatio: Im on mobile, but z2jh has a standalone .py file it loads into a configmap (as a string value but not hardcoded).

manics: I've just tried this, but it means the JupyterHub and BinderHub charts have to be deployed simultaneously since the configmap is part of the BinderHub chart, but mounted by the JupyterHub chart.

Our CI tests assume Z2JH can be deployed independently by extracting the `jupyterhub` chart config from the `binderhub/values.yaml`
- https://github.com/jupyterhub/binderhub/blob/8c8c0be2aba6542849269f1de440697a614ce9f5/.github/workflows/test.yml#L87-L90
- https://github.com/jupyterhub/binderhub/blob/8c8c0be2aba6542849269f1de440697a614ce9f5/testing/local-binder-k8s-hub/install-jupyterhub-chart#L35-L37

We could manually create the required configmap? The alternative which I'm investigating is to keep everything in `values.yaml` and have `setup.py` pull it out instead.

consideRatio: > Our CI tests assume Z2JH can be deployed independently by extracting the jupyterhub chart config from the binderhub/values.yaml

Oh, ah... Okay.

Well, you could do `helm template ... --show-only templates/configmap.yaml | kubectl apply -f -` as an ugly hack then, I don't see another great solution if we need to deploy binderhub logic related to jupyterhub in k8s but not the entire binderhub in a k8s context.

manics: The overall problem is it's JupyterHub specific but not k8s specific.

What do you think of the latest changes? I've put the config back into `values.yaml`, and I'm copying it into the binderhub python source in setup.py

It introduces a dependency on `pyyaml` to run `setup.py`, but given that `setup.py` already has external requirements like `npm` I don't think that matters too much.

helm-chart/binderhub/values.yaml
```diff
@@ -62,14 +62,35 @@ jupyterhub:
...
-      00-binder: |
-        from tornado import web
+      00-binderspawnermixin: |
+        """
+        Helpers for creating BinderSpawners
 
-        # image & token are set via spawn options
-        from kubespawner import KubeSpawner
+        This file is defined in helm-chart/binderhub/values.yaml and is copied to
+        binderhub/binderspawner_mixin.py by setup.py
```
consideRatio, line 70: With a migration plan, this is acceptable to me. Is this a reasonable understanding of the situation and a migration plan @yuvipanda @manics?

```suggestion
        FIXME: This file is defined in helm-chart/binderhub/values.yaml and is
               copied to binderhub/binderspawner_mixin.py by setup.py. To
               resolve this complexity, we have the following constraints to
               consider.
               
               1. This mixin is required both by the `binder` pod part of the
                  BinderHub Helm chart and the `hub` pod part of the JupyterHub
                  Helm chart.
               2. BinderHub's CI system can run locally against a JupyterHub
                  Helm chart installed in a k8s cluster, and then the code needs
                  to be available in the binderhub Python package and in the
                  JupyterHub Helm chart.
               
               A solution is to:

               1. Let the binderhub Python package contain this mixin
               2. Install the binderhub Python package in the hub pod's image
               3. Install the binderhub Python package in the binder pod's image
```

setup.py
```diff
@@ -27,6 +28,13 @@
     subprocess.check_call(['npm', 'install'])
     subprocess.check_call(['npm', 'run', 'webpack'])
 
+binderspawner_mixin_py = os.path.join(here, 'binderhub', 'binderspawner_mixin.py')
```
consideRatio, line 31:
```suggestion
# FIXME: See a related FIXME in helm-chart/binderhub/values.yaml about needing a
#        mixin both in this package, the JupyterHub Helm chart, and the
#        BinderHub Helm chart.
binderspawner_mixin_py = os.path.join(here, 'binderhub', 'binderspawner_mixin.py')
```

consideRatio (commented):
This looks acceptable to me given that we add notes making us able to recover from the workarounds in the future.

helm-chart/binderhub/values.yaml
consideRatio: I understand, I think this can be marked as resolved given that we add notes about this complexity. I've made code suggestions below about such notes - feel free to tweak them!

manics:
Thanks for reviewing!

I've added some `FIXME` comments.

Note the BinderHub code base doesn't care about the spawner code at all, it's only used by JupyterHub, but we're currently _distributing_ the custom spawner via the BinderHub repo.

The problem is we want to use the `BinderSpawnerMixin` code in multiple spawners, not just KubeSpawner. If we want to keep the current deployment setup then either we store the code in `binderhub/binderspawner_mixin.py` and copy it in to `values.yaml`, or vice-versa.

I think a nicer long term plan is to either publish the `BinderSpawnerMixin` code a separate Python package and include that in the Z2JH hub image, or build a custom JupyterHub image in this repository.

I can't put everything in `00-binder` because only the `BinderSpawnerMixin` code is part of the BinderHub Python module, the other bits (subclassing `KubeSpawner`, setting `c.JupyterHub.spawner_class = BinderSpawner`) musn't be included. Instead I've renamed it `0-binderspawnermixin` which according to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/417e84e81101f5757c5bb735bd7daf249696c45f/jupyterhub/files/hub/jupyterhub_config.py#L444 should mean it's evaluated before `00-binder`.

yuvipanda (changes requested):
IMO, we should just already override the hub image and install binderhub_spawner there. I want us to make sure we can disentangle the helm chart from the python code as much as possible, and I think keeping the primary code in the helm config is the wrong direction there.

If we want an intermediate step, I think we can just keep two copies of the file, and add a test that asserts they are identical - so we don't forget to update one when we update the other.

manics:
I'm in favour of creating a BinderHub specific JupyterHub image. IIRC there was some resistance in the past due to the maintainance cost, but I'm happy to implement that here if no-ones against it.

yuvipanda:
@manics I just had a conversation with @consideRatio, and I am now happy to actually merge this as is. The FIXMEs listed in https://github.com/jupyterhub/binderhub/pull/1358#issuecomment-919097521, and I think that's good enough. We agreed that long term option is to install `binderhub` in the upstream z2jh chart. I personally prefer having this be duplicated, but I have now decided that doesn't need to block this PR :)

manics:
Thanks both 😃.

I'm still keen to tidy this up, but we should probably have a more general team discussion about the BinderHub roadmap first.

yuvipanda:
@manics as an intermediate step, how do you feel about implementing

> If we want an intermediate step, I think we can just keep two copies of the file, and add a test that asserts they are identical - so we don't forget to update one when we update the other?

The end goal would probably be for us to consistently publish `binderhub` to pypi, and install that in z2jh itself.

PR #1361, opened 2021-09-13
Avoid using exceptions for flow control

yuvipanda:
Exceptions should be used for exceptional circumstances
only - in this case, if resolved ref is not a valid sha1.
The most common case is that the ref we get isn't a sha1 and
needs resolving. Using a conditional here is a clearer
expression of intent.


yuvipanda:
Thanks for the merge, @betatim!

PR #1363, opened 2021-09-14
CI check for binderspawnermixin code

manics:
This implement's @yuvipanda's suggestion from https://github.com/jupyterhub/binderhub/pull/1358#issuecomment-919126408

Removes the modifications to setup.py and other related files from https://github.com/jupyterhub/binderhub/pull/1358 and replaces it with a check `ci/check_embedded_chart_code.py` that the two code locations are in sync.


yuvipanda:
Thanks, @manics!

PR #1364, opened 2021-09-14
Local binder local hub configuration without K8s

manics:
The last bit of https://github.com/jupyterhub/binderhub/pull/1353
- Adds a config file for JupyterHub/BinderHub that allows BinderHub to run on Docker without K8s.
- Adds a GitHub job to run the `pytest -m remote` tests (and some fixes so that it's possible to run some tests without Kubernetes)
- See the individual commit messages for details of the other changes.

- [x] This requires https://github.com/binderhub-ci-repos/minimal-dockerfile/pull/2 since DockerSpawner doesn't override the default container user. For now the tests will fail as Jupyter won't run as root.
- [x] I've combined the JupyterHub and BinderHub configurations into a single file since trailtlets ignores anything unrelated to the application. Main benefit is not duplicating an API token between two configs, but do you think it's better to split it anyway?


yuvipanda:
@manics I think it would be nice to keep them separate regardless, just for clarity. They could in theory both use a common dependency, or a class with same name but different behaviors, etc...

testing/local-binder-local-hub/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Install local BinderHub from source
+
+    pip install -e ../..
+
+Run JupyterHub in one terminal
+
+    jupyterhub --config=binderhub_config.py
+
+BinderHub will be running as a managed JupyterHub service, go to http://localhost:8000 nd you should be redirected to BinderHub.
```
yuvipanda, line 18:
```suggestion
BinderHub will be running as a managed JupyterHub service, go to http://localhost:8000 nd you should be redirected to BinderHub.
```
```suggestion
BinderHub will be running as a managed JupyterHub service, go to http://localhost:8000 and you should be redirected to BinderHub.
```

yuvipanda (commented):
Comment about api_token.txt but otherwise happy to merge!

yuvipanda:
I've actually just fixed the `api_token.txt` thing I pointed out! I'll merge this after tests pass if nobody else gets to it first.

SO EXCITED, @manics!

manics:
The `api_token.txt` was so that you could choose between running as a managed service (token is autogenerated and passed as an env-var), and running them seperately (need to share a token across two configs). This is controlled by the `RUN_BINDERHUB_AS_JUPYTERHUB_SERVICE = True` at the top of `jupyterhub_config.py`. Maybe it's unnecessary complication though, and it's better to keep things simple and only have a managed service config?

manics:
I went ahead and removed the conditional, so now only the managed service config is kept.

yuvipanda:
Perfect, @manics! \o/

Thank you so much for working on this :)

manics:
Thanks for coming up with the idea in the first place!

PR #1371, opened 2021-09-20
Upgrade event-source-polyfill

yuvipanda:
- Use the correct 'dependencies' field in package.json, as
  event-source-polyfill is a runtime dependency.

Split up from https://github.com/jupyterhub/binderhub/pull/1248

Co-authored-by: Kenan Erdogan <kenanerdogan@gmail.com>

### DIFF
--- a/binderhub/static/js/src/image.js
+++ b/binderhub/static/js/src/image.js
@@ -1,3 +1,7 @@
+import { NativeEventSource, EventSourcePolyfill } from 'event-source-polyfill';
+
+const EventSource = NativeEventSource || EventSourcePolyfill;
+
 export default function BinderImage(providerSpec) {
   this.providerSpec = providerSpec;
   this.callbacks = {};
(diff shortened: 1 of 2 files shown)


yuvipanda:
I've tested this locally. However, the polyfill is really there to support older Internet Explorer and Edge, and I don't really have a way to test that.

binderhub/static/js/src/image.js
```diff
@@ -1,3 +1,7 @@
+import { NativeEventSource, EventSourcePolyfill } from 'event-source-polyfill';
+
+const EventSource = NativeEventSource || EventSourcePolyfill;
```
consideRatio, line 3: Do we also need to set it globally? It is a hard question to answer perhaps. Should we go for with or without setting it globally you think?

I'm really not sure at all, so this suggestion is just for your convenience rather than an actual "I think this should be done".

```suggestion
const EventSource = NativeEventSource || EventSourcePolyfill;
global.EventSource = EventSource;
```

yuvipanda: It doesn't need to be set globally. It's used explicitly only in that file, and default lexical scope of `const` means this works as it should.

PR #1375, opened 2021-09-20
Fix error in loading page

yuvipanda:
`fitAddon is not a function` was being otherwise shown.


yuvipanda:
This would've been caught by linters as an unused variable... We should get in on that.

consideRatio (approved):
Intend to merge on test success!

Issue #1373, opened 2021-09-20
Use ES6 classes in our JS


yuvipanda:
Our JS uses old school prototype based classes. I think it's time we switch to a more modern, cleaner ES6 class syntax. I think it's easier to understand what is happening, and helps newer contributors who might just have modern JS experience.

manics:
I'm in favour of using `prettier` and `black` throughout this repo with pre-commit for autoformatting (code, javascript, markdown, yml). Main thing is to decide when to make the change to avoid conflicts with pending PRs.

I know you've previously said pre-commit is too complicated but it doesn't have to be, just `pre-commit run -a` is all I ever do, I never bother with the automated hooks :smiley: 

I'm happy to make that change along with docs if you want.

consideRatio:
Yeah I've been very very happy with pre-commit since a while back, the experience got even better with pre-commit.ci.

`prettier`, `black`, and `pyupgrade` are amazing!

PR #1376, opened 2021-09-20
Replace all uses of var with const/let

yuvipanda:
var is a complex way to define variables, with complexities
around dynamic scope, hoisting, etc. Modern JS prefers
lexically scoped const (the default) and let (for variables
that get reassigned).

This PR just moves all uses of var to const or let as
appropriate.

Ref: https://github.com/jupyterhub/binderhub/issues/1373

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -26,12 +26,12 @@ import 'bootstrap/dist/css/bootstrap.min.css';
 import 'bootstrap/dist/css/bootstrap-theme.min.css';
 import '../index.css';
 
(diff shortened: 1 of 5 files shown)


yuvipanda:
This should be merged before https://github.com/jupyterhub/binderhub/pull/1377

yuvipanda:
image

https://janakachathuranga.medium.com/why-we-use-const-and-avoid-var-3528ee550fb4

consideRatio:
Thanks @yuvipanda !!! Nice work on this!

PR #1380, opened 2021-09-20
Use babel to transform JS

yuvipanda:
After this, we can generally use most modern JS
without having to look at caniuse.com and see if
the feature is supported.


package.json
```diff
@@ -12,8 +16,8 @@
   "dependencies": {
     "bootstrap": "^3.4.0",
     "clipboard": "^2.0.8",
-    "jquery": "^3.6.0",
     "event-source-polyfill": "^1.0.25",
```
consideRatio, line 19: Is event source polyfill something we won't need using babel? I recall babel to me a lot about transforming code to be compatible.

babel.config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["@babel/preset-env"]
+}
```
consideRatio, line 3:
```suggestion
}

```

consideRatio (commented):
Sounds good to me!

package.json
yuvipanda: We will definitely still need it, as it's not a *language* feature but a browser feature.

consideRatio:
Wieee!!!

PR #1381, opened 2021-09-20
Add pre-commit: pretter, black, pyupgrade, and misc

consideRatio:
### UPDATED PR STATUS: Awaiting resolution of #1390

I think there is no easy way of getting this done initially, but I figured I'll make a go for it. WDYT?

I opted to not apply prettier on js/html/css at this point as it became less readable code when only two spaces was enforced. We can run prettier twice with different settings for html/css/js vs markdown/yaml/json etc if we want to handle that in the future - of which I made a comment about.

### Related
- A team-compass issue about trialing pre-commit.ci: https://github.com/jupyterhub/team-compass/issues/379
- A team-compass issue about using pyupgrade: https://github.com/jupyterhub/team-compass/issues/453

### Action points
- [x] Extract manually made changes from this PR (@consideRatio)
  #1384
  #1385
  #1386
  #1387
  #1388
  #1389
- [x] Make this PR exclude everything but the config changes
- [x] Mitigate issues for existing PRs (everyone)
  Discussion on how in https://github.com/jupyterhub/binderhub/pull/1381#issuecomment-925045089
  A dedicated issue to handle the situation: #1390
- [ ] Merge this PR
- [ ] Open a new PR
  1. ```shell
     # create individual commits for each kind of code change
     pre-commit run --all pyupgrade
     git commit -am "pre-commit: run pyupgrade"
     pre-commit run --all black
     git commit -am "pre-commit: run black"
     pre-commit run --all reorder-python-imports
     git commit -am "pre-commit: run reorder-python-imports"
     pre-commit run --all prettier
     git commit -am "pre-commit: run prettier"
     pre-commit run --all end-of-file-fixer
     git commit -am "pre-commit: run end-of-file-fixer"
     pre-commit run --all requirements-txt-fixer
     git commit -am "pre-commit: run requirements-txt-fixer"
     
     # Verify everything now succeds
     pre-commit run --all
     ```
  2. Update helm-chart/binderhub/values.yaml with contents of binderspawner-mixin.py
- [ ] Merge the new PR
- [ ] Enable pre-commit.ci at https://github.com/organizations/jupyterhub/settings/installations/17782968/update

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,16 @@
+# flake8 is a Python linting tool run by pre-commit as declared by our
+# pre-commit-config.yaml file.
+#
+# This configuration is compatible with the autoformatter tool black, and
+# further relaxed to not bug us with too small details.
+#
+# flake8 configuration reference:
+# https://flake8.pycqa.org/en/latest/user/configuration.html
+#
+[flake8]
+max-line-length = 88
+extend-ignore = C, E, W
+
+# Adjustments to linting the binderhub repo
+builtins = c, load_subconfig
+exclude = versioneer.py,binderhub/_version.py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,65 @@
+# pre-commit is a tool to perform a predefined set of tasks manually and/or
+# automatically before git commits are made.
+#
+# Config reference: https://pre-commit.com/#pre-commit-configyaml---top-level
+#
+# Common tasks
+#
+# - Run on all files:   pre-commit run --all-files
+# - Register git hooks: pre-commit install --install-hooks
+#
+repos:
+  # Autoformat: Python code, syntax patterns are modernized
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.31.0
+    hooks:
+      - id: pyupgrade
+        args:
+          - --py37-plus
+
+  # Autoformat: Python code
+  - repo: https://github.com/psf/black
+    rev: 22.1.0
+    hooks:
+      - id: black
+        args: [--target-version=py37]
+
+  # Autoformat: Python code
+  - repo: https://github.com/asottile/reorder_python_imports
+    rev: v3.0.1
+    hooks:
+      - id: reorder-python-imports
+
+  # Autoformat: js, html, markdown, yaml, json
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.6.1
+    hooks:
+      - id: prettier
+        exclude_types:
+          # These are excluded initially as pre-commit was added but can
+          # absolutely be enabled later. If so, we should consider having a
+          # separate run of pre-commit where we configure a line spacing of 4
+          # for these file formats.
+          - html
+          - javascript
+          - css
+
+  # Misc autoformatting and linting
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
+      - id: end-of-file-fixer
+        exclude_types: [svg]
+      - id: check-case-conflict
+      - id: check-executables-have-shebangs
+      - id: requirements-txt-fixer
+
+  # Lint: Python code
+  - repo: https://github.com/PyCQA/flake8
+    rev: "4.0.1"
+    hooks:
+      - id: flake8
+
+# versioneer.py is excluded for being an imported dependency we opt to not
+# modify ourselves. This is also set in .flake8 for consistency.
+exclude: versioneer.py|binderhub/_version.py
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+helm-chart/binderhub/templates/
+http-record.doi.org.json
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -243,7 +243,7 @@ class Launcher(LoggingConfigurable):
                 else:
                     ready_event_future.cancel()
         try:
-            resp = await self.api_request(
+            await self.api_request(
                 'users/{}/servers/{}'.format(escaped_username, server_name),
                 method='POST',
                 body=json.dumps(data).encode('utf8'),


helm-chart/binderhub/values.yaml
```diff
@@ -101,7 +101,7 @@ jupyterhub:
                 # classes using traitlets?
                 # https://stackoverflow.com/questions/9575409/calling-parent-class-init-with-multiple-inheritance-whats-the-right-way
                 # https://github.com/ipython/traitlets/pull/175
-                super(BinderSpawnerMixin, self).__init__(*args, **kwargs)
+                super().__init__(*args, **kwargs)
```
consideRatio, line 104: I noted this change was made by pyupgrade @manics, is it problematic do you think? So far I've not catched it make anything that is breaking. It was run configured to be compatible with py36.

manics: I trust the CI tests :smile:

consideRatio:
Draft reason - I'm debugging a test failure

yuvipanda:
Also, if we merge this, we'll probably cause immense merge conflicts with most other open PRs...

manics:
Could we pull out the manual fixes here into their own PR and merge those?

Hopefully it'll then be easy to re-base this PR and re-run pre-commit with (hopefully) no manual changes when it's convenient?

consideRatio:
>Could we pull out the manual fixes here into their own PR and just merge those?
>
>Hopefully it'll then be easy to re-base this PR and re-run pre-commit with no manual changes when it's convenient?

I'll do that.

> Also, if we merge this, we'll probably cause immense merge conflicts with most other open PRs...

Yeah, I see no easy solution or set of action points to avoid this. I guess the only thing to do is to make a thorough work nudging all open PRs to resolving into merge/close.

betatim:
Can we run the pre-commit checks in GH Actions instead of adding yet another CI provider? The more different providers we use the higher the chances that one of them has an outage, makes changes we have to react to, etc. GH Actions seems to be running well and we use it a lot, so if we can get the tools/checks executed there that would be a win if you ask me.

On the topic of this PR changing a lot of code which makes it hard to review the changes (triggered by questions like "is this change really a good one?" as well as making life harder for existing PRs: what concrete problem or pain does a tool like pyupgrade address?

binderhub/_version.py
```diff
@@ -1,15 +1,11 @@
-
```
betatim, line 1: Can we exclude this file from these tools? It is copied directly from the versioneer project. This means if we make changes to it it becomes harder to diff our version to a new version we might one day want to upgrade to.

manics:
> On the topic of this PR changing a lot of code which makes it hard to review the changes (triggered by questions like "is this change really a good one?" as well as making life harder for existing PRs: what concrete problem or pain does a tool like pyupgrade address?

There was a discussion about pyupgrade on https://github.com/jupyterhub/jupyterhub/pull/3583 (that PR was split into two and merged). I think the big advantage of pyupgade, and autoformatters such as Black or Prettier, is it helps standardise our code which has developed over several years by several different people. If you're new to the code base the different coding styles can be very noticeable and confusing, e.g. you're not sure if the different conventions have a functional effect or not.

It also improves general readibility. It's definitely not perfect, but after a lot of prevaricating I now tend to agree with the developers of autoformatters that everyone has there own preference, so in the absence of a strong reason it's better to just go with the defaults.

You've indirectly brought up a good point about how we make these sort of wide-ranging changes though, and is perhaps something we should think about in future. It fits in with the disucssion about how we make impactful changes such as switching to JupyterLab, except this is developer focussed instead of user focussed.

binderhub/_version.py
consideRatio: I agree, will do.

consideRatio:
@betatim regarding pre-commit.ci, I wrote a comment about it in https://github.com/jupyterhub/team-compass/issues/379#issuecomment-924990376 opened to discuss usage of pre-commit.ci.

> You've indirectly brought up a good point about how we make these sort of wide-ranging changes though, and is perhaps something we should think about in future. It fits in with the discussion about how we make impactful changes such as switching to JupyterLab, except this is developer focussed instead of user focussed.

Should I perhaps open an issue dedicated to try highlight what it does and with my suggestion to start using it across our repos? I can champion this one because I'm at the stage where I'd like to see it adopted more widely after having seen it piloted successfully without any issues in jupyter/docker-stacks and jupyterhub/jupyterhub now.

manics:
How's this for a roll-out plan for the main autoformatting change:
- Make a list of all open BinderHub PRs that are still relevant, and are close to being ready to merge
- Once we the list decide on a deadline based on the changes required to those PRs, e.g. one or two months?
- Communicate on those PRs and help with getting them merged

binderhub/_version.py
consideRatio: Done!

consideRatio:
> How's this for a roll-out plan for the main autoformatting change:

I opened up a dedicated issue for this @manics - #1390

choldgraf:
Thanks @consideRatio for opening up this one - I think that these kinds of mega-PRs are quite tricky because it's so hard to review anything. I've found that the following is a useful order of operations:

1. Get agreement from the team about which conventions to automatically adopt (for this, I suggest we open up a team-compass issue that makes the proposal "Adopt `black` and `pyupgrade` as standards across our repositories")
2. The act of "approving" this should be merging this as team practice into our team compass docs.
3. If that is approved, then assume the following step has approval in all of our repositories, and we welcome implementation:
4. Make a PR that only does 2 things: 1. Adds the approved formatters to our precommit hooks, 2. Runs the pre-commit hooks on our codebase.

If there are other changes that need to be made, I think it's best that we do those separately from the "mega commits" so that it's easier to review.

As for implementing this in the `binderhub` repository, I think @manics has a good implementation proposal above (https://github.com/jupyterhub/binderhub/pull/1381#issuecomment-925045089)

consideRatio:
Thanks for input @choldgraf! I've updated this PR with instructions on how to continue based on your suggested approach.

.flake8
```diff
@@ -0,0 +1,12 @@
+# A minimal "try to respect --line-length, but don’t become crazy if you can’t"
+# configuration as copied from black's documentation:
+# https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length
+#
+[flake8]
+max-line-length = 88
+select = C,E,F,W,B,B950
```
minrk, line 7: When using auto formatters, I think it's best to ignore C, E, W, since that's the linter checking the formatter. I think linters should only be used for correctness, and it's actually bad software practice to auto-enforce any style that cannot be applied by an auto formatter.

manics:
It's been nearly 5 months since this PR was opened. It will cause disruption to [open PRs](https://github.com/jupyterhub/binderhub/pulls), but most of those are over a year old. Since there's never going to be a perfect time I think we should update this PR and merge it.

manics:
This was discussed in today's JupyterHub team meeting. We agreed to go ahead with this, and merge in two week time!

consideRatio:
@manics rebased and addressed @minrk's wishes about the flake8 config by ignoring all C, E, W kind of warnings!

manics (approved):
LGTM!

When you run pre-commit can you add a `.git-blame-ignore-revs` too? It's feature in Git that GitHub now supports, and should mean the repo-wide auto-formatting commits are skipped in the blame view:

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

choldgraf:
Woohoo for closing 6 months old PRs!

PR #1382, opened 2021-09-20
Slightly cleanup 2 js files

yuvipanda:
For badge.js:

- Use ES6 classes
- Don't reach into the DOM to get build token and baseUrl.
  This will help with writing tests in the future
- Use => functions, so we can stop writing `const that = this`.
  See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions.
  Particularly, it does not rebind `this`, so we can continue
  to access the parent context's `this`.

For badge.js:
- Stop using global constants for config. 
- Use a single function that can switch on syntax type, rather than unique functions for
  each syntax. Helps reduce boilerplate.

Eventually this could be part of a binder-client JS package that
can be used by other projects (like thebe)

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -17,7 +17,7 @@ import 'bootstrap';
 import 'event-source-polyfill';
 
 import BinderImage from './src/image';
-import { markdownBadge, rstBadge } from './src/badge';
+import { makeBadgeMarkup } from './src/badge';
 import { getPathType, updatePathText } from './src/path';
 import { nextHelpText } from './src/loading';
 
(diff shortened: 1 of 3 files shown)


yuvipanda:
The test failures seem completely unrelated?

yuvipanda:
image

definitely looks unrelated

consideRatio (approved):
Beautiful!

I spent quite a bit of time to look at this to compensate for me not being so fluent with ES6 / JS matters. I think this is good to go.

The test failures relates to https://github.com/k3s-io/k3s/issues/4054 btw.

yuvipanda:
@consideRatio do you know why the pre-commit.ci PR is failing?

consideRatio:
Because it doesnt see a pre-commit config yet, ignore it for now. I enabled it for this repo for the pre-commit related pr.

consideRatio:
Thanks yuvi!!!

Issue #1368, opened 2021-09-16
Things that can break with lab by default


minrk:
Switching to lab by default broke more than I thought. Taking some notes here, to see if there are things we can do to mitigate breakage, or at least point folks to mitigations.

The main information I think we need to surface is `?urlpath=/tree/` or `?urlpath=/notebooks/mynotebook.ipynb` allow opt-in to the classic notebook interface, for anyone who wants to preserve that behavior.

Cases that can stop working:

- Dockerfile that doesn't install lab (or not recent enough). This will 404 without any hint as to what went wrong
- environment pins down lab to too-old: I had thought only custom Dockerfiles could 404, but if folks have e.g. `jupyterlab==0.34.*` in requirements.txt/environment.yml, they can get the 404, too
- filepath and/or default URL where they just don't want to change (I view this one as purely a documentation issue, just like it was for lab users before it was the default, since nothing is actually broken)
- filepath or default ui, with extensions that don't work in lab (I don't think we can detect this, so it seems the same as previous in practice, despite users facing actual breakage)

I think the 404s are the harshest ones, and things we should consider trying to fix.

Mitigations:

- `filepath` breakage could have been avoided if we updated the form to use a *new* argument to unambiguously choose lab, and leave plain `filepath` as classic notebook. I'm ambivalent about this, because part of the *point* of filepath as it was, was to allow it to follow UI updates as they came.
- is there a way we could detect availability of jupyterlab before picking default url? This is not hard with a custom entrypoint, but we have very little available to us with custom Dockerfiles. Maybe if we mount an executable from a volume, we could do something without modifying the image?
- time-based choices - I've long thought about using last modified time to try to avoid breaking changes, and only use new options for builds since the date they were introduced (e.g. pypi time machine, etc.). I'm not sure that this is a good idea, but it intrigues me.
- we could 'advertise' a link back to classic ui via a lab extension, if folks think that would be more helpful than noisy. I'm not sure how to do it in a way that's sufficiently visible for those who need it and not annoying for everyone else

minrk:
I think this will work to avoid breakage in most of the cases we've seen so far:

if we change the default launch command to:

```python
["python3", "-c", 
"""
from distutils.version import LooseVersion as V
try:
  import jupyterlab
  major = int(jupyterlab.__version__.split(".", 1)[0])
except Exception:
  have_lab = False
else:
  have_lab = major >= 3

if have_lab:
  # if recent-enough lab is available, make it the default UI
[...]

PR #1383, opened 2021-09-21
lab default only if available, change `filepath` definition

minrk:
Changes `?filepath` definition from "open with current UI" to "open with classic notebook" to match apparent user expectations.

two pieces of #1368 :

- opening a file with jupyterlab uses `?labpath` instead of `?filepath`. `filepath` is permanently frozen to mean `/tree/...` with the classic UI
- selecting 'open a file' in the form behavior is unchanged, in that it produces a `?labpath` url, not a `filepath` url
- move default url choice down one level into `singleuser.cmd` so that we can set lab as the default url if jupyterlab 3 is available

If no filepath is specified, and lab 3 is available, lab is still the default UI, leaving it the only link with changed behavior.

@betatim proposed reserving lab-by-default to `/v3/` to make the UI change fully explicit and opt-in. I think that if we go all the way to defining a `v3` url, we should also change the server launch command to `jupyter lab` instead of `jupyter notebook`, because we know that change will be significantly more disruptive, and is also going to happen eventually. It's also a lot simpler to say "v3 starts jupyterlab" than the UI is changed, but the server is not.

We specifically chose this 'only switch default ui' to avoid breaking existing links, so if we are making new links, I think we should update the server command to the current standard, too, rather than having to do this all over again in a few months.

I think we can also communicate a lot better (and take some steps) to address the fact that UI customizations (extensions and such) are very often broken by repo2docker updates, because repo2docker sets the UI versions, not the user. The user may pin down UI, but this can break things, too. There is no good answer for now for stable links with extensions that don't need upkeep by the user.

### DIFF
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -75,7 +75,7 @@ class ParameterizedMainHandler(BaseHandler):
                 filepath = self.get_argument('filepath', '').lstrip('/')
 
 
-            # Check if we have a JupyterLab + file path, if so then use it for the filepath
+            # Check the urlpath parameter for a file path, if so use it for the filepath
             urlpath = self.get_argument('urlpath', '').lstrip('/')
             if urlpath and "/tree/" in urlpath:
                 filepath = urlpath.split('tree/', 1)[-1]
--- a/binderhub/main.py
+++ b/binderhub/main.py
@@ -70,11 +70,14 @@ class ParameterizedMainHandler(BaseHandler):
             nbviewer_url = 'https://nbviewer.jupyter.org/github'
             org, repo_name, ref = spec.split('/', 2)
             # NOTE: tornado unquotes query arguments too -> notebooks%2Findex.ipynb becomes notebooks/index.ipynb
-            filepath = self.get_argument('filepath', '').lstrip('/')
+            filepath = self.get_argument("labpath", "").lstrip("/")
+            if not filepath:
+                filepath = self.get_argument('filepath', '').lstrip('/')
+
 
             # Check if we have a JupyterLab + file path, if so then use it for the filepath
             urlpath = self.get_argument('urlpath', '').lstrip('/')
-            if urlpath.startswith("lab") and "/tree/" in urlpath:
+            if urlpath and "/tree/" in urlpath:
                 filepath = urlpath.split('tree/', 1)[-1]
 
             blob_or_tree = 'blob' if filepath else 'tree'
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -188,10 +188,29 @@ jupyterhub:
         admin: true
         apiToken:
   singleuser:
-    # start jupyter notebook for the server
-    cmd: jupyter-notebook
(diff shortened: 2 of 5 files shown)


minrk:
looks like pre-commit ci is running, despite no config (#1381)

consideRatio:
I've now disabled pre-commit.ci again in https://github.com/organizations/jupyterhub/settings/installations/17782968/update!

binderhub/main.py
```diff
@@ -70,11 +70,14 @@ async def get(self, provider_prefix, _unescaped_spec):
             nbviewer_url = 'https://nbviewer.jupyter.org/github'
             org, repo_name, ref = spec.split('/', 2)
             # NOTE: tornado unquotes query arguments too -> notebooks%2Findex.ipynb becomes notebooks/index.ipynb
-            filepath = self.get_argument('filepath', '').lstrip('/')
+            filepath = self.get_argument("labpath", "").lstrip("/")
+            if not filepath:
+                filepath = self.get_argument('filepath', '').lstrip('/')
+
 
             # Check if we have a JupyterLab + file path, if so then use it for the filepath
```
consideRatio, line 78:
```suggestion
            # Check the urlpath parameter for a file path, if so use it for the filepath
```

helm-chart/binderhub/values.yaml
```diff
@@ -188,10 +188,29 @@ jupyterhub:
...
+            have_lab = False
+        else:
+            have_lab = major >= 3
+
+        if have_lab and "NotebookApp.default_url" not in " ".join(sys.argv):
+            # if recent-enough lab is available, make it the default UI
+            sys.argv.insert(1, "--NotebookApp.default_url=/lab/")
+
+        # launch the notebook server
+        os.execvp("jupyter-notebook", sys.argv)
```
consideRatio, line 213: Do you think this could make signal transmission (SIGTERM for example) start to fail?

consideRatio (commented):
I support the suggested intentions, and the code changes looks good to me. I've raised a question and made a code suggestion about a comment but otherwise I consider this ready for merge.

helm-chart/binderhub/values.yaml
minrk: No, that's exactly what the use of `exec` here avoids. Exec *replaces* the current process, as opposed to `Popen` which launches a new process and can have signal-relay issues.

This is the Python equivalent of 

```bash
exec jupyter-notebook "$@"
```

that we use in other container entrypoints

consideRatio (approved):
LGTM! Not sure about the test failure though.

minrk:
test failure was a DNS failure to resolve discourse.jupyter.org, the kind of once-in-a-while CI networking issue that just happens sometimes. I don't think it's related.

consideRatio:
I'd like to merge this directly if tests succeed. I believe it is a very relevant fix and we see above that it has been referenced as something awaited as a fix.

minrk:
all green

consideRatio:
Thank you so much for your work on this @minrk!!! :heart: :tada:

PR #1387, opened 2021-09-22
Use f-strings where pyupgrade doesn't handle it automatically

consideRatio:
Extracted from #1381 as manual transformations that pyupgrade failed to make automatically.


manics (commented):
Can you revert the change to `binderhub/_version.py` (as discussed in https://github.com/jupyterhub/binderhub/pull/1381#discussion_r713620046)

Otherwise this looks fine.

consideRatio:
Ah, I were only on the lookout for versioneer.py - thanks! Done!

manics (approved):
That's one of the nice things about setuptools-scm, everything is done at build/install time without creating/modifying extra files 😃.

PR #1388, opened 2021-09-22
Formatting adjustments to make the code transform nicer with black autoformatting

consideRatio:
Extracted from #1381. After applying `black` autoformatting, it will order separate args we pass on separate lines. By using `key=value` instead of `key` and `value` as separate strings, we make it look nicer after `black` autoformatting is applied.

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -176,11 +176,13 @@ class Build:
     def get_r2d_cmd_options(self):
         """Get options/flags for repo2docker"""
         r2d_options = [
-            '--ref', self.ref,
-            '--image', self.image_name,
-            '--no-clean', '--no-run', '--json-logs',
-            '--user-name', 'jovyan',
-            '--user-id', '1000',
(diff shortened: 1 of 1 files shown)


binderhub/build.py
```diff
@@ -176,11 +176,13 @@ def __init__(
...
-            '--no-clean', '--no-run', '--json-logs',
-            '--user-name', 'jovyan',
-            '--user-id', '1000',
+            f"--ref={self.ref}",
+            f"--image={self.image_name}",
+            "--no-clean",
+            "--no-run",
+            "--json-logs",
+            "--user-name=jovyan",
+            "--user-id=1000",
```
consideRatio, line 185: Note I've tested repo2docker the to function with `--arg=value` and not only with `--arg value` as I know not all CLI may support that, but repo2docker does.

manics (approved):

consideRatio:
Thank you @manics!! :heart: :tada:

PR #1391, opened 2021-09-22
Rely on docker-image-cleaner in external repo

consideRatio:


helm-chart/binderhub/values.yaml
```diff
@@ -238,8 +238,8 @@ dind:
 imageCleaner:
   enabled: true
   image:
-    name: jupyterhub/k8s-image-cleaner
-    tag: 'local'
+    name: quay.io/jupyterhub/docker-image-cleaner
```
yuvipanda, line 241: Can we automate updating these with dependabot?

yuvipanda:
image
 
Nice :D

helm-chart/binderhub/values.yaml
consideRatio: Not with dependabot, but i can make it happen without. I figure i want that done for chp and traefil in z2jh as well.

consideRatio: To do this, I imagine a scan for the latest git tag and an automatic PR creation to bump to the latest git tag.

consideRatio: @manics depends, what if they push many tags, also for backports? Also, what kind of API's are easy to use to get such tags that makes us able to re-use them and not have custom logic for each image version updater because we work against both dockerhub and quay etc?

I opened https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2419 in z2jh about bumping things. I'll open another one to mimic what we do there in the end for this repo.

consideRatio: I created https://github.com/jupyterhub/binderhub/issues/1398

yuvipanda (approved):
I think this is great, but I want to give others who aren't already involved in this decision a chance to weigh in before I hit merge. I'll wait a week maybe?

/cc @sgibson91 @betatim @manics @minrk  :)

manics:
I haven't tested it, but assuming it works I can't think of any reason not to do this :smile:

betatim:
Fine for me. As always my main thoughts are around "what could break if we do this?"

minrk (approved):

betatim (approved):

consideRatio:
This PR is successfully deployed all the way to mybinder.org-deploy. It had to be complemented with https://github.com/jupyterhub/binderhub/pull/1417.

PR #950, opened 2019-09-12
Emit launch events

bitnik:
With this PR binder gets pending spawn (launch) events from jhub's progres api and sends them to bhub' UI.

First commit is copy of what @minrk showed to me during the workshop, thank you :)

Issue #886, opened 2019-07-01
Expose k8s events to users via the log


betatim:
Right now the feedback to users during the launch of a pod ends at "launching ...". This means there are long periods where "nothing" happens as far as a user can tell. We could pipe some of the kubernetes events related to the pod being launched to the user to give them something to look at and see progress.

In the background several things are happening like pulling the image, starting the init containers, etc.

Concretely the idea is to forward the events from https://gist.github.com/betatim/e8db23942746ec873c876fc4a2396b22#file-binder-launch-log-txt-L1-L11 to users.

cc @sgibson91

betatim:
#852 is related.

minrk:
JupyterHub 1.0 already provides an EventStream API for spawning servers, and KubeSpawner populates this with kubernetes events:

![progress](https://user-images.githubusercontent.com/151929/60509877-4832de00-9cce-11e9-92b5-9d367c4ca142.gif)

 We should be able to forward this to users pretty well.

PR #1393, opened 2021-09-24
Emit launch events

minrk:
Rebase of #950 (I don't have permission to push updates there)

squashed #950 to make conflict resolution easier, and fixed a couple small issues, but this works well in my own tests.

Much better feedback when launch fails: in my tests, launch failed because I hadn't set up image push/pull correctly:

Screen Shot 2021-09-24 at 12 04 09

And when launch succeeds, you get more incremental feedback about progress:

https://user-images.githubusercontent.com/151929/134658127-c2406be2-a108-4a9f-9502-77fd35e7ae9f.mov

The original PR stripped k8s timestamp/log-level prefixes from messages, but I reverted that to make it simpler. I think if we want to do that, it should be a KubeSpawner option, not something we do in BinderHub.

EDIT by @consideRatio: closes #950.
EDIT by @manics: closes https://github.com/jupyterhub/binderhub/issues/886

### DIFF
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -287,7 +287,9 @@ class Launcher(LoggingConfigurable):
             self.log.debug("Requesting progress for {username}: {progress_api_url}")
             resp_future = self.api_request(
                 progress_api_url,
-                streaming_callback=handle_chunk,
+                streaming_callback=lambda chunk: asyncio.ensure_future(
+                    handle_chunk(chunk)
+                ),
                 request_timeout=self.launch_timeout,
             )
             try:
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -1,6 +1,7 @@
 """
 Launch an image with a temporary user via JupyterHub
 """
+import asyncio
 import base64
 import json
 import random
@@ -233,7 +234,14 @@ class Launcher(LoggingConfigurable):
 
         # start server
         app_log.info(f"Starting server{_server_name} for user {username} with image {image}")
-        ready_event_container = []
+        ready_event_future = asyncio.Future()
+
+        def _cancel_ready_event(f=None):
+            if not ready_event_future.done():
+                if f and f.exception():
+                    ready_event_future.set_exception(f.exception())
+                else:
+                    ready_event_future.cancel()
         try:
             resp = await self.api_request(
                 'users/{}/servers/{}'.format(escaped_username, server_name),
@@ -257,9 +265,18 @@ class Launcher(LoggingConfigurable):
                         event = json.loads(line.split(":", 1)[1])
                         if event_callback:
                             await event_callback(event)
+
+                        # stream ends when server is ready or fails
                         if event.get("ready", False):
-                            # stream ends when server is ready
-                            ready_event_container.append(event)
+                            if not ready_event_future.done():
+                                ready_event_future.set_result(event)
+                        elif event.get("failed", False):
+                            if not ready_event_future.done():
+                                ready_event_future.set_exception(
+                                    web.HTTPError(
+                                        500, event.get("message", "unknown error")
+                                    )
+                                )
 
             url_parts = ["users", username]
             if server_name:
@@ -267,6 +284,7 @@ class Launcher(LoggingConfigurable):
             else:
                 url_parts.extend(["server/progress"])
             progress_api_url = url_path_join(*url_parts)
+            self.log.debug("Requesting progress for {username}: {progress_api_url}")
             resp_future = self.api_request(
                 progress_api_url,
                 streaming_callback=handle_chunk,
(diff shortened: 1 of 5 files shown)


binderhub/launcher.py
```diff
@@ -80,6 +83,13 @@ def _default_hub_url_local(self):
         Receives 5 parameters: launcher, image, username, server_name, repo_url
         """
     )
+    launch_timeout = Integer(
+        600,
+        config=True,
+        help="""
+         Wait this many seconds until server is ready, raise TimeoutError otherwise.
+         """,
```
consideRatio, line 91:
```suggestion
        help="""
        Wait this many seconds until server is ready, raise TimeoutError otherwise.
        """,
```

binderhub/launcher.py
```diff
@@ -223,20 +241,54 @@ async def launch(self, image, username, server_name='', repo_url='', extra_args=
...
-                    user_data = await self.get_user_data(escaped_username)
-                    if user_data['servers'][server_name]['ready']:
-                        break
-                    if not user_data['servers'][server_name]['pending']:
-                        raise web.HTTPError(500, "Image %s for user %s failed to launch" % (image, username))
-                    # FIXME: make this configurable
-                    # FIXME: Measure how long it takes for servers to start
-                    # and tune this appropriately
-                    await gen.sleep(min(1.4 ** i, 10))
+                # listen pending spawn (launch) events until server is ready
```
consideRatio, line 244:
```suggestion
                # listen to pending spawn (launch) events until server is ready
```

binderhub/launcher.py
```diff
@@ -223,20 +241,54 @@ async def launch(self, image, username, server_name='', repo_url='', extra_args=
...
+                )
+                try:
+                    await gen.with_timeout(
+                        timedelta(seconds=self.launch_timeout), resp_future
+                    )
+                except (gen.TimeoutError, TimeoutError):
+                    raise web.HTTPError(
+                        500,
+                        "Image %s for user %s took too long to launch"
+                        % (image, username),
```
consideRatio, line 284: Are you okay with f-strings? pyupgrade will upgrade some but not all. To make #1381 mergable in the future, I extracted out some manual changes regarding f-strings in #1387.

```suggestion
                        f"Image {image} for user {username} took too long to launch"
```

binderhub/launcher.py
```diff
@@ -223,20 +241,54 @@ async def launch(self, image, username, server_name='', repo_url='', extra_args=
...
+                    )
+
+                # verify that the server is running!
+                if not ready_event_container:
+                    raise web.HTTPError(
+                        500, "Image %s for user %s failed to launch" % (image, username)
```
consideRatio, line 290:
```suggestion
                        500, f"Image {image} for user {username} failed to launch"
```

helm-chart/binderhub/values.yaml
```diff
@@ -192,7 +192,6 @@ jupyterhub:
     cmd: jupyter-notebook
     # use jupyterlab for the default UI
     defaultUrl: /lab
-    events: false
```
consideRatio, line 195: Do you know why this default value is removed? These are the default values for the helm charts and I'd like there to be an entry for all valid configurations (unless its passthrough config).

consideRatio (commented):
Wieee great work! Very happy to see more information about the server launches following build!

helm-chart/binderhub/values.yaml
minrk: Yes, these are the events this PR is relaying to the user, so disabling them mostly defeats the purpose. I was thinking about this, and maybe the right thing to do is make this explicitly true here, rather than relying on the upstream chart default, which happens to be true.

consideRatio: Oh I didn't realize this was a z2jh config. Then it was acceptable to remove it, but I think indeed it should be set to true here so z2jh can modify it without breaking this, and so that we understand that binderhub requires it to be true for this to function.

Well hmm, I assume this won't break if events are disabled though, right? And I figure that the `events` serve the exact same purpose in z2jh as in binderhub, so I think it can be removed after all --- assuming it won't break something causing errors etc.

minrk: It won't break - both work, just you get the existing PR if this is disable,d and you get more info if it's enabled. I think it's sensible for binderhub to set this to true explicitly, though, even if that happens to be the current default on the upstream chart. Either way is fine by me.

minrk:
The failure is real, and I think it's due to the streaming callback being async, so it isn't guaranteed to have been called when the original request finishes. I need to adjust the waits a little bit to deal with that.

I've no idea why helm always fails and local always succeeds, though.

consideRatio (approved):
LGTM besides merge conflict!

yuvipanda:
@minrk the test failure seems real?

consideRatio (approved):
LGTM again, and with the recent fix of awaiting an async call, everything seem to work as well! @minrk I'll go for a merge if you consider yourself done at this point! Just didn't want to merge 1 second after the test succeeded without checking in with you.

minrk:
Yes, I think I finally figured out why it was failing. I'm happy with it now.

minrk:
I hope exposing these messages will motivate some work on https://github.com/jupyterhub/kubespawner/issues/524 :)

PR #1394, opened 2021-09-24
Include JS for *just* the bootstrap components we use

yuvipanda:
This reduces our bundle size from 488 KB to 452 KB with barely
any effort on our part


consideRatio:
Nice!!!

Issue #1367, opened 2021-09-15
Upgrade from bootstrap 3 to bootstrap 5


yuvipanda:
As part of upgrading our JS dependencies (See #1248), we should upgrade bootstrap too. 5 has been released for  while now.

We don't use a lot of bootstrap features so I expect the upgrade to not take a lot of work.

PR #1395, opened 2021-09-24
Migrate to bootstrap 5

yuvipanda:
Cleanup the markup to rely on bootstrap classes wherever
possible, while preserving the look and feel as much as possible.
Currently not 100% matched, particularly around font sizes.

Ref https://github.com/jupyterhub/binderhub/issues/1367

TODO:

- [ ] Investigate log show / hide behavior in the loading screen
- [x] Investigate progressbar bug (exists in main too https://github.com/jupyterhub/binderhub/issues/1439)

Screenshots

Main page:
image

Minor style changes to the button exporter + different 'copy' button (earlier it looked like a 'paste' button)
image

Some bug with the progress bar that needs to be fixed, but new styling of bootstrap 5 is present here:
image

Loading page - there's some extra padding in logs that needs fixing

image


betatim:
Do you want to get to 100% visual similarity? I think at some point I tried to upgrade to BS4 but gave up trying to make it look exactly the same and long before reaching something I thought was "good enough". This makes me think we shouldn't aim for "100% match" and instead try and use this opportunity for a "light refresh of the UI".

yuvipanda:
@betatim yeah I agree! I don't want it to be 100%, but similar enough in a way that doesn't 'look odd'.

yuvipanda:
I still need to work on the progress bars and the log viewer on the main page, and I haven't even looked at the loading page yet.

yuvipanda:
I don't think I'm going to be able to finish this one, and my long term interest is in moving the binderhub frontend to its own project anyway, without bootstrap.

PR #1402, opened 2021-10-01
Manually create the launch URL instead of using the response from JHub

MridulS:
https://github.com/jupyterhub/binderhub/pull/1393 broke binder deployments which were using a base_url with a base_url for JupyterHub. 

The `ready` [event from JupyterHub](https://github.com/jupyterhub/jupyterhub/blob/5f0077cb5b8d6507bbfe36d094fbabf7aaabe65b/jupyterhub/apihandlers/users.py#L678) embeds the `base_url` of JupyterHub already so it was creating URL with `http://<IP>:<PORT>/<Jhub_base_url>/<Jhub_base_url>/<user>/<server_name>`

Not sure if this is the right way to fix this or if this should be changed in the jupyterhub API handler.

### DIFF
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -284,7 +284,7 @@ class Launcher(LoggingConfigurable):
             else:
                 url_parts.extend(["server/progress"])
             progress_api_url = url_path_join(*url_parts)
-            self.log.debug("Requesting progress for {username}: {progress_api_url}")
+            self.log.debug(f"Requesting progress for {username}: {progress_api_url}")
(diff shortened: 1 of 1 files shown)


minrk:
Ah, I think I understand. It is just a bug here: we shouldn't be joining hub_url, we should be joining a hub *host* to the absolute path returned from the event.

Running the tests with a non-empty prefix is a good idea to catch these things (JupyterHub itself always runs tests with a prefix that's as weird as allowed to try to provoke issues)

PR #1406, opened 2021-10-05
use hub/api/health endpoint

minrk:
allows api-only deployments, where only `/hub/api` is routed to the Hub, instead of all `/hub/` urls

this duplicate endpoint of `/hub/health` was [added][] in JupyterHub 1.4, which we now require

[added]: https://github.com/jupyterhub/jupyterhub/pull/3373


consideRatio:
This LGTM and seems very safe to merge, I'll go for it.

PR #1360, opened 2021-09-11
Build class for running repo2docker locally

manics:
Adds a new Build class that runs repo2docker as a local process instead of a container. This allows you to run BinderHub without Kubernetes, and instead rely on JupyterHub with a local container spawner such as DockerSpawner.

This also enables the repo2docker options to be easily augmented in a subclass, for example
```py
class LocalRepo2dockerPodmanBuild(LocalRepo2dockerBuild):
    def get_r2d_cmd_options(self):
        return super().get_r2d_cmd_options() + [
            '--engine', 'podman',
        ]
```

Related to https://github.com/jupyterhub/binderhub/pull/1353

PR #1410, opened 2021-10-06
Implement LocalRepo2dockerBuild.stop()

manics:
This is a follow-up from @yuvipanda's comment on https://github.com/jupyterhub/binderhub/pull/1360#discussion_r707169339

I'm not sure if this is the best way to do it.....

### DIFF
--- a/binderhub/tests/test_build.py
+++ b/binderhub/tests/test_build.py
@@ -3,7 +3,7 @@
 import docker
 import json
 import sys
-from time import time
+from time import monotonic
 from unittest import mock
 from urllib.parse import quote
 from uuid import uuid4
@@ -270,10 +270,10 @@ def test_execute_cmd():
 def test_execute_cmd_break():
     cmd = ['python', '-c', 'from time import sleep; print(1, flush=True); sleep(10); print(2, flush=True)']
     lines = []
-    now = time()
+    now = monotonic()
 
     def break_callback():
-        return time() - now > 2
+        return monotonic() - now > 2
 
     # This should break after the first line
     with pytest.raises(ProcessTerminated) as exc:


binderhub/tests/test_build.py
```diff
@@ -211,3 +212,72 @@ async def test_local_repo2docker_build():
...
+
+def test_execute_cmd():
+    cmd = ['python', '-c', 'from time import sleep; print(1, flush=True); sleep(2); print(2, flush=True)']
+    lines = list(_execute_cmd(cmd, capture=True))
+    assert lines == ['1\n', '2\n']
+
+def test_execute_cmd_break():
+    cmd = ['python', '-c', 'from time import sleep; print(1, flush=True); sleep(10); print(2, flush=True)']
+    lines = []
+    now = time()
```
minrk, line 273: Not important for a test, but modern Python convention would use `time.monotonic` here to avoid system time issues. In the extremely unlikely event that you are running this two-second test during an NTP sync.

binderhub/build_local.py
```diff
@@ -38,28 +70,55 @@ def _execute_cmd(cmd, capture=False, **kwargs):
     # This should behave the same as .readline(), but splits on `\r` OR `\n`,
     # not just `\n`.
     buf = []
+    q = queue.Queue()
+
+    def read_to_queue(proc, capture, q):
+        try:
+            for c in iter(partial(proc.stdout.read, 1), b""):
```
minrk, line 77: Question: why not use asyncio subprocesses with a future to signal the break?

```python
break_future = input_arg
p = asyncio.create_subprocess_exec(cmd)
read_future = p.stdout.readline()

while not break_future.done():
    await asyncio.wait([read_future, break_future], return_when=asyncio.FIRST_COMPLETED)
    if read_future.done():
        # the read completed
        line = await read_future
        yield line
        # schedule next read
        if line and not cancel_future.done():
            read_future = p.stdout.readline()
```

This would make it an async generator (`async for line in _execute_cmd(...)`).

Running a threadsafe put for every _character_ is going to get pretty inefficient when repo2docker produces a lot of output. This is less important if localr2d is only used for testing, but if we want to move to prod, there are probably more efficient ways.

minrk: I see this is avoiding a change to make Build.submit itself a coroutine, which we should perhaps do in the future.

In that case, I think we can switch the `read_to_queue` to call `proc.stdout.readline()` since it doesn't matter anymore if it blocks. It will always unblock when EOF arrives after the process has been terminated.

binderhub/tests/test_build.py
manics: Thanks for the tip!

binderhub/build_local.py
manics: Thanks for reviewing! I've changed it to iterate through `proc.stdout`- which I think is equivalent to your suggestion, and it's a lot cleaner now. The original character-by-character handling was copied from repo2docker:
https://github.com/jupyterhub/repo2docker/blob/4da2e1e71565b26a4bf1d0f0d26ae7c2373a1fd7/repo2docker/utils.py#L34-L46

@yuvipanda and I have discussed defining the interface for the build/launch classes to make the K8s parameters optional, making this function async could be part of that. If you think it's better I'm happy to leave this PR until that refactoring is done.

minrk: 👍 to doing that as a separate task

PR #1412, opened 2021-10-07
helm-chart: support modern api-version of PDBs

### LINKED ISSUE #1396: Update kubernetes objects to v1 APIs
Just got an email from Google Cloud about the fact that with Kubernetes 1.22 some APIs will be deprecated and that we might  be using them for mybinder.org on GKE.

This issue is about finding and then upgrading the objects that still use `v1beta` or `v1alpha` versions and moving them to `v1`.

https://cloud.google.com/kubernetes-engine/docs/deprecations/apis-1-22 is a list of all the APIs that will be deprecated in v1.22.

consideRatio:
I think when a helm chart upgrades and renders this changed template, it will conclude that one PDB of old version is no longer rendered and should be removed, and one new PDB of a new version is now rendered and should be created: this kind of change makes it fine for `helm` and k8s to update this resource while changes to values within it can be troublesome.

Closes #1396 fully.

### DIFF
--- a/helm-chart/binderhub/templates/pdb.yaml
+++ b/helm-chart/binderhub/templates/pdb.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.pdb.enabled -}}
+{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+apiVersion: policy/v1
+{{- else }}
 apiVersion: policy/v1beta1
+{{- end }}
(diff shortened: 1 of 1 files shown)


consideRatio:
This is a quite trivial PR. I would like to merge it and verify that it deploys properly to mybinder.org-deploy as I've assumed, and since I'm verifying such things now I'd like to go for a merge and steward this all the way there.

PR #1428, opened 2021-11-03
update"path to a notebook" text for pathType=lab

minrk:
without this, the text on the form is always "URL to open"


consideRatio (approved):

minrk:
@consideRatio sorry, didn't notice that this was already approved!

PR #1441, opened 2021-12-21
enforce overall pod quota on attempts to launch

minrk:
otherwise, the quota field in the health check can only be used by outside consumers of the health endpoint (e.g. the mybinder.org federation redirector).

Without this, there's no way for a BinderHub to limit its own capacity, despite the appearance of such configuration.

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -303,8 +303,9 @@ class BinderHub(Application):
         The number of concurrent pods this hub has been designed to support.
 
         This quota is used as an indication for how much above or below the
-        design capacity a hub is running. It is not used to reject new launch
-        requests when usage is above the quota.
+        design capacity a hub is running.
+
+        Attempts to launch new pods once the quota has been reached will fail.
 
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
LGTM! Thanks @minrk !

minrk:
Risk of doing this for mybinder.org: a race condition where a launch is assigned to a BinderHub by the federation-redirector because it's just below quota, but arrives after it passes the quota and fails rather than being sent to another member.

A way to avoid this would be to use _two_ quotas - one soft for 'health' and one hard for rejecting launches. We could implement this only in the federation-redirector by having a 'headroom' of e.g. 5-10 pods where we consider the quota full before it's _all_ the way full.

betatim:
I like the idea of moving the "head room" decision (and related config complexity) to the redirector. With two quotas we'd have to explain to the naive, one instance BinderHub admin why there are two, why they are different and how to set them, etc. With your idea of moving this decision to the federation-redirector we keep all that complexity hidden from people who don't run a federation.

PR #1443, opened 2022-01-01
Fix Ingress API check

emrahkaya:
If the full API resource is given, Helm on ArgoCD produces wrong answer for the ".Capabilities.APIVersions.Has" template.
Checking for "networking.k8s.io/v1" is enough, and works well.


consideRatio:
There is a version issue for argo's helm library in my mind. They should use a moddr modern version. Z2jh require helm 3.5 and that can handle this.

Due to that i wish to reject this PR, i think there are more issues to follow if this would be patched anyhow, and this was just the first of several issues if using an old helm version.

I suggest making sure you use the latest argo version, and then nudging argo to use a more modern version of helm (3.5+ needed).

PR #1444, opened 2022-01-18
Support `...image.pull[Policy|Secrets]` consistently and hook into JupyterHub's chart wide configuration for imagePullSecrets

emrahkaya:
Fixing the unused image.pullSecrets value in the binderhub Deployment.

## Update by Erik

Thanks to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2546, we can hook into z2jh's chart wide configuration of imagePullSecrets. With this PR, we do that for all our pods where we specify an image.

I also added support for image.pullPolicy which wasn't supported in practice even though the schema supported it.

### DIFF
--- a/helm-chart/binderhub/templates/deployment.yaml
+++ b/helm-chart/binderhub/templates/deployment.yaml
@@ -33,6 +33,9 @@ spec:
         {{- . | toYaml | nindent 8 }}
         {{- end }}
     spec:
+      {{- with include "jupyterhub.imagePullSecrets" (dict "root" . "image" .Values.image) }}
+      imagePullSecrets: {{ . }}
+      {{- end }}
       {{- with .Values.initContainers }}
       initContainers:
         {{- . | toYaml | nindent 8 }}
@@ -70,12 +73,6 @@ spec:
 
       {{- with .Values.extraVolumes }}
       {{- . | toYaml | nindent 6 }}
-      {{- end }}
-      {{- with .Values.image.pullSecrets }}
-      imagePullSecrets:
-        {{- range . }}
-        - name: {{ . }}
-        {{- end }}
       {{- end }}
       containers:
       - name: binder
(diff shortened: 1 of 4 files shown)


emrahkaya:
Our tests here doesn't cover this case, but I've tested this my system, using a private registry.

In my tests, I have put this step in the test.yml:

```
      - name: Create an image pull secret
        run: |
          kubectl create secret docker-registry acr-secret \
            --namespace default \
            --docker-server="${{ secrets.DOCKER_REGISTRY }}" \
            --docker-username="${{ secrets.DOCKER_USERNAME }}" \
            --docker-password="${{ secrets.DOCKER_PASSWORD }}"
```

and modified `testing/k8s-binder-k8s-hub/binderhub-chart-config.yaml` file, adding

```
image:
  pullSecrets:
    - acr-secret
```
there.

Such a test might not apply here, but I'm putting this information for anyone interested using/testing it (or a proof for my tests :) ).

helm-chart/binderhub/templates/deployment.yaml
```diff
@@ -61,6 +61,12 @@ spec:
       {{- with .Values.extraVolumes }}
       {{- . | toYaml | nindent 6 }}
       {{- end }}
+      {{- with .Values.image.pullSecrets }}
+      imagePullSecrets:
+        {{- range . }}
+        - name: {{ . }}
+        {{- end }}
+      {{- end }}
```
consideRatio, line 69: I think this is now possible, allowing a binderhub user to rely on jupyterhub chart wide configuration options like below, so its only needed to be configured once. Alternatively, by specifying something special just under `image.pullSecrets`.

```yaml
jupyterhub:
  imagePullSecret: ...
  imagePullSecrets: ...
```

```suggestion
      {{- with include "jupyterhub.imagePullSecrets" (dict "root" . "image" .Values.image) }}
      imagePullSecrets: {{ . }}
      {{- end }}
```

consideRatio (approved):

Issue #1445, opened 2022-01-19
MyBinder.org's version of Docker is too old for latest Ubuntu images


pkgw:
Apologies if I'm not filing this in the right place ... I'm not quite sure where it should go.

I recently updated one of my MyBinder repos to Ubuntu 21.04, and discovered that it failed to build when actually deployed on MyBinder.org. This turns out to be an incompatibilty between Ubuntu >=21.04 and older versions of Docker:

https://stackoverflow.com/questions/66319610/gpg-error-in-ubuntu-21-04-after-second-apt-get-update-during-docker-build
https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1916485

Meanwhile, 20.10 doesn't update correctly and the version of Node.js included in 20.04 is too old for modern JupyterLab, so it's awkward to build Ubuntu-based images right now. I ended up using 20.04 with a custom-upgraded Node.js. It would be good to update the version of Docker/K8s/whatever used in the MyBinder.org service to fix the newer Ubuntus, I think.

PR #1446, opened 2022-01-19
Bump version of docker used in image builder

yuvipanda:
The version we were running was more than 2 years old,
and might also have other vulnerabilities that are not patched.

Fixes https://github.com/jupyterhub/binderhub/issues/1445


consideRatio (approved):
:tada:

PR #1447, opened 2022-01-20
Fix mini-css-extract-plugin version

emrahkaya:
Fixing the similar issue with `webpack` as in:
- https://github.com/facebook/create-react-app/issues/11930


emrahkaya:
Ok, looks like the issue is resolved with the latest version of `mini-css-extract-plugin`: 
https://github.com/webpack-contrib/mini-css-extract-plugin/releases/tag/v2.5.2 

Thus, closing this PR.

PR #1457, opened 2022-03-03
fix total pod count

minrk:
pod_quota was not being enforced because total_pods was not the number of pods, but the number of keys in the response dict (4).

This will result in overall pod quotas actually being enforced (they are not at the moment). Added in #1441


minrk:
Failures are unrelated and due to apparent networking issues on GHA runners.

PR #1459, opened 2022-03-08
reverse generated-python direction

minrk:
Make binderspawner_mixin.py canonical, instead of python-in-yaml

easier to edit with editors, linters, etc.

removes use of redundant pyyaml in favor of ruamel.yaml, needed for roundtrip yaml editing

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -68,12 +68,12 @@ jupyterhub:
         Helpers for creating BinderSpawners
 
         FIXME:
-        This file is defined in helm-chart/binderhub/values.yaml and is copied to
-        binderhub/binderspawner_mixin.py by setup.py so that it can be used by other JupyterHub
-        container spawners.
+        This file is defined in binderhub/binderspawner_mixin.py
+        and is copied to helm-chart/binderhub/values.yaml
+        by ci/check_embedded_chart_code.py
 
-        The BinderHub repo is just used as the distribution mechanism for this spawner, BinderHub
-        itself doesn't require this code.
+        The BinderHub repo is just used as the distribution mechanism for this spawner,
+        BinderHub itself doesn't require this code.
 
         Longer term options include:
         - Move BinderSpawnerMixin to a separate Python package and include it in the Z2JH Hub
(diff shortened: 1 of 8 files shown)


manics (commented):
This duplication is something that annoys me too :disappointed: . Good to merge once the tests are passing.

In future we could probably add this to pre-commit so the scripts gets updated automatically as part of linting.

helm-chart/binderhub/values.yaml
```diff
@@ -17,7 +17,7 @@ nodeSelector: {}
 
 image:
   name: jupyterhub/k8s-binderhub
-  tag: 'local'
+  tag: local
```
consideRatio, line 20: If you use a round trip, the quotes may not be left for a string that needs them i think.

Retainig explicit quotes for the tag has been relevant in the past i recall, perhaps not by chartpress generates image tags, but i'm still cautious preferring to see us having the yaml include quotes foe our placeholder values where they could be needed.

minrk: The quotes were removed _by_ the roundtrip serialization. I did not remove them. I left the change here because every re-write of the file will strip it.

consideRatio: Ah, this is how we have configured ruamel.yaml in chartpress to avoid this kind of issue.

```python
# use safe roundtrip yaml loader capable or preserving usage of no/single/double
# quotes for a string for example
yaml = YAML(typ="rt")
yaml.preserve_quotes = True
yaml.indent(mapping=2, offset=2, sequence=4)
```

The indent settings is relevant to be compatible with prettier formatting, at least if not having a rout-tripped format to retain.

```yaml
exampleList:
- without-indent-settings

# this is the kind of formatting we systematically
# use and enforce with prettier, and requires explicit
# configuration of ruamel.yaml at least when generating
# new yaml files, but I'm uncertain if it's required when
# round-tripping on existing files with existing list entries
anotherExampleList:
  - with-indent-settings
```

minrk: Got it! I remembered the indent setting (it's the same here), but forgot the quotes. Will fix.

minrk: Fixed.

manics (approved):

PR #1462, opened 2022-03-09
ensure pycurl is available in image

minrk:
pycurl wasn't working because libcurl4 is missing in python:3.8-slim-buster

I think this is contributing to turing frequently being reported as unhealthy


consideRatio:
Dependency script stuff is removed, is that as it should be?

minrk:
> Dependency script stuff is removed, is that as it should be?

Yes, the script was removed in #1433, but this reference was leftover

PR #1463, opened 2022-03-09
helm chart: fix typo in schema and support global values

consideRatio:
I figure all Helm charts should allow `global` to be set, this change makes the schema accept `global` values to be passed. It also makes fixes a harmless typo.


consideRatio:
Being a trivial PR I'll go for a self-merge on this.

PR #1472, opened 2022-04-25
support extraFiles for binderhub pod

minrk:
imports extraFiles helpers, spec from JupyterHub, applied to the binderhub pod

Possibly more general solution to #1471

### DIFF
--- a/binderhub/tests/conftest.py
+++ b/binderhub/tests/conftest.py
@@ -49,14 +49,22 @@ def pytest_configure(config):
         "markers", "auth: mark test to run only on auth environments"
     )
     config.addinivalue_line(
-        "markers", "remote: mark test to run only on remote environments"
-    )
-    config.addinivalue_line(
         "markers", "github_api: mark test to run only with GitHub API credentials"
     )
     config.addinivalue_line(
         "markers", "remote: mark test for when BinderHub is already running somewhere."
     )
+    config.addinivalue_line(
+        "markers",
+        "helm: mark test to only run when BinderHub is launched with our k8s-binderhub test config.",
+    )
+
+
+def pytest_runtest_setup(item):
+    is_helm_test = any(mark for mark in item.iter_markers(name="helm"))
+    if not item.config.getoption("--helm"):
+        if is_helm_test:
+            pytest.skip("Skipping test marked as 'helm'")
 
 
 def pytest_terminal_summary(terminalreporter, exitstatus):
--- a/helm-chart/binderhub/templates/secret.yaml
+++ b/helm-chart/binderhub/templates/secret.yaml
@@ -18,6 +18,12 @@ stringData:
   values.yaml: |
     {{- pick .Values "config" "cors" "dind" "extraConfig" | toYaml | nindent 4 }}
 
+  {{- /* Custom HTML templates */ -}}
+  {{- range $filename, $contents := .Values.templates }}
+  templates-{{ $filename }}: |
+    {{ $contents | nindent 4 }}
+  {{- end }}
+
   {{- /* Glob files to allow them to be mounted by the binderhub pod */}}
   {{- /* key=filename: value=content */}}
   {{- (.Files.Glob "files/*").AsConfig | nindent 2 }}
--- a/helm-chart/binderhub/templates/secret.yaml
+++ b/helm-chart/binderhub/templates/secret.yaml
@@ -18,15 +18,18 @@ stringData:
   values.yaml: |
     {{- pick .Values "config" "cors" "dind" "extraConfig" | toYaml | nindent 4 }}
 
-  {{- /* Custom HTML templates */ -}}
-  {{- range $filename, $contents := .Values.templates }}
-  templates-{{ $filename }}: |
-    {{ $contents | nindent 4 }}
-  {{- end }}
-
   {{- /* Glob files to allow them to be mounted by the binderhub pod */}}
   {{- /* key=filename: value=content */}}
   {{- (.Files.Glob "files/*").AsConfig | nindent 2 }}
+
+  {{- with include "jupyterhub.extraFiles.stringData" .Values.extraFiles }}
+  {{- . | nindent 2 }}
+  {{- end }}
+
+{{- with include "jupyterhub.extraFiles.data" .Values.extraFiles }}
+data:
+  {{- . | nindent 2 }}
+{{- end }}
 ---
 {{- if or .Values.config.BinderHub.use_registry .Values.config.BinderHub.buildDockerConfig }}
 kind: Secret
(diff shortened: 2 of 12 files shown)


minrk:
Hm, we have a problem in CI because it's now impossible to update the helm schema and pass tests, because the same config is passed both to the current and previous version of binderhub.

If the schema weren't so strict (i.e. rejecting addtionalProperties), this would be less of an issue.

choldgraf (commented):
This seems like a nice and relatively lightweight way to add this functionality without increasing our configuration surface area each time we want another page / section to be extensible...though we should probably improve our documentation around how to do this because many people may not be familiar with the template names they should use for over-riding etc...

consideRatio:
This seems like `[hub|singleuser].extraFiles` but specialized for templates. Should we go for the even more general approach of `extraFiles` instead of this?

- [Z2JH configuration reference](https://zero-to-jupyterhub.readthedocs.io/en/latest/resources/reference.html#hub-extrafiles)
- Z2JH code implementation
  - [extraFiles helpers defined](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/ce673fecfcef9719d0b18da5e998fa51d2f100a2/jupyterhub/templates/_helpers.tpl#L313-L384). These can perhaps also be reused directly from this helm chart as it depends on the jupyterhub Helm chart.
  - [extraFiles helpers referenced from a k8s Secret](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/ce673fecfcef9719d0b18da5e998fa51d2f100a2/jupyterhub/templates/hub/secret.yaml#L43-L50)

minrk:
> This seems like [hub|singleuser].extraFiles but specialized for templates

It is! I actually used your extraFiles implementation as reference. JupyterHub is much more infinitely extensible, so there are more general files that make sense to drop-in, especially files we as chart maintainers don't need to know about. I'm not sure I see a similar arbitrary files use case in BinderHub, so I thought this would be simpler, both to implement (the whole thing is smaller than just the helpers for extraFiles) and use (deployments don't need to know where we put templates). On the other hand, I can imagine expert JupyterHub deployers (and maintainers) liking the familiarity of extraFiles consistently applied everywhere.

helm-chart/binderhub/templates/secret.yaml
```diff
@@ -21,6 +21,15 @@ stringData:
...
   {{- (.Files.Glob "files/*").AsConfig | nindent 2 }}
+
+  {{- with include "binderhub.extraFiles.stringData" .Values.extraFiles }}
+  {{- . | nindent 2 }}
+  {{- end }}
+
+{{- with include "binderhub.extraFiles.data" .Values.extraFiles }}
+data:
+  {{- . | nindent 2 }}
+{{- end }}
```
consideRatio, line 32: Did you know/consider this option? The helpers seems general enough to function without being hard-copied to this helm chart, and are practically available in the helm chart already.

```suggestion
  {{- with include "jupyterhub.extraFiles.stringData" .Values.extraFiles }}
  {{- . | nindent 2 }}
  {{- end }}

{{- with include "jupyterhub.extraFiles.data" .Values.extraFiles }}
data:
  {{- . | nindent 2 }}
{{- end }}
```

minrk: I will try that! I thought of it, but then wasn't sure about versions, etc. I will try it that way tomorrow

consideRatio: Since binderhub pins the version of JupyterHub being the dependency, this feels quite safe. Since the helpers are general enough, I'm not worried about having binderhub starting to use this - at least no more worried than needing to update the code in two places if it improves.

manics: Can I just check, is this considered a supported use of Z2JH chat features (which therefore will be maintained in accordance with semver)?

minrk: I have two answers for that:

1. I would not consider any helpers in any chart public APIs, unless thoroughly documented to be otherwise
2. I would not consider binderhub an 'external' project, so I'm okay using these kinds of internal utilities across our _internal_ projects, especially with the strict dependency pin we use in binderhub. If jupyterhub ever breaks this, we can vendor the previous version with a copy/paste or update.

minrk: tested and works great with the imported helpers

minrk:
I went with a simple test under `pytest.mark.remote`. It's not 100% correct semantically (remote doesn't mean it's our helm test config), but it is correct _practically_ (those are the tests we run against this config).

minrk:
Looks like I wasn't quite right about the condition, trying a fix

minrk:
New test (new `helm` mark) now runs only when `--helm` is passed, verifying that extraFiles is working as intended.

helm-chart/binderhub/templates/secret.yaml
consideRatio: I appreciate the answers above about "not documented -> no commitment" and the distinction about what we recommend using to others and what we do internally.

@manics do you see an action point that would make you feel comfortable around this? Perhaps a declaration of intent for how helpers are used in z2jh? I think being explicit about such intent is fine, especially since I'm unsure about what helpers we should couple to be tracked by the changelog etc.

binderhub/tests/conftest.py
```diff
@@ -48,15 +48,23 @@ def pytest_configure(config):
     config.addinivalue_line(
         "markers", "auth: mark test to run only on auth environments"
     )
-    config.addinivalue_line(
-        "markers", "remote: mark test to run only on remote environments"
-    )
```
consideRatio, line 53: Should this really be removed? We still have tests marked as remote.

consideRatio (commented):
LGTM besides comment with regards to removal of pytest mark.

helm-chart/binderhub/templates/secret.yaml
manics: It's related to the idea behind dogfooding: https://en.m.wikipedia.org/wiki/Eating_your_own_dog_food

We know JupyterHub and Z2JH are used and extended by lots of users, many who we have no idea about, and BinderHub is a prime demonstration of how to build on top of Z2JH. If we're running into issues with extending Z2JH then it's probable that others will too. I think we should therefore aim to do whatever we'd tell others to do. I can't tell from this discussion what the feeling was- say it's acceptable for everything else to do this with the proviso that it's not a standard API, or try and keep it hidden from everyone?

Given the time pressure it's fine to merge this, but if the conclusion of the above discussion is that this shouldn't be copied by other applications then maybe create an issue for how to come up with a public solution?

binderhub/tests/conftest.py
minrk: Yes, remote was declared twice. This was a duplicate.

helm-chart/binderhub/templates/secret.yaml
minrk: Just chiming it that while I don't consider them public _now_ that doesn't mean they necessarily shouldn't be - if we want to support downstream charts using extraFiles, I think the first step is adding them to the z2jh docs.

PR #1479, opened 2022-05-25
ci: add dependabot to update github actions

consideRatio:


consideRatio:
I think this PR should be non-controversial, and since binderhub is a repo where we have got behind a lot on maintenance and review of PRs etc I'll go for a self-merge.

PR #1490, opened 2022-05-25
Update helm chart's binderhub image's frozen requirements.txt

consideRatio:
- docs: tweak notes on updating frozen image requirements.txt
- Update helm chart's binderhub image's frozen requirements.txt

### DIFF
--- a/helm-chart/images/binderhub/README.md
+++ b/helm-chart/images/binderhub/README.md
@@ -1,7 +1,7 @@
 # binderhub image
 
 The image for running binderhub itself.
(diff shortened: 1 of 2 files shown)


consideRatio:
Wieeeeeeeeee! This seems to have resolved the test failures!

Oh, perhaps also related to the actions/cache we have will cache no matter what if requirements.txt is the same. Hmmm, not happy about that as that can be the same for different python versions etc etc... Well, requirements.txt was changed, and now we got away from the broken cache.

PR #1493, opened 2022-05-25
breaking: start require k8s 1.20+ and test against k8s 1.20 and latest

consideRatio:
Is the mybinder.org federation beyond k8s 1.19 at this point?

Z2JH requires k8s 1.20 currently, see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2635. @manics made a great job documenting the status of various cloud providers in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/2591#issuecomment-1048790139. In short, k8s 1.19 is not supported by Google, Amazon, or Azure.

While we could wait to bump this until we run into a need raise the lower level, I'd prefer to not wait. If we realize there is a need to raise the lower lever, then suddenly the person fixing something ends up needing to side-track and do this breaking change first before etc. In other words, by doing this now, we avoid slowing down and blocking something else when we really need to do it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,15 +56,14 @@ jobs:
           - helm
         include:
           # Chart.yaml contains the chart's oldest supported k8s version, we
-          # want to test against that. We also test against the oldest known
-          # supported helm cli version which isn't documented anywhere, but is
-          # currently at 3.5.0. Currently we have listed 1.17 in Chart.yaml, but
-          # due to a k3s issue with their k8s 1.17 release, we can't test
-          # against k8s 1.17 and settle for k8s 1.18 for now.
-          - k3s-channel: v1.18
+          # test against that and the oldest known supported helm cli version
+          # which isn't documented anywhere, but is currently at 3.5.0. We also
+          # test with the latest k8s and helm versions.
+          #
+          - k3s-channel: v1.20
             helm-version: v3.5.0
             test: helm
-          - k3s-channel: stable
+          - k3s-channel: latest
             test: helm
             test-variation: upgrade
             # upgrade-from represents a release channel, see: https://jupyterhub.github.io/helm-chart/info.json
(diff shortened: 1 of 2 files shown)


sgibson91 (commented):
This looks ok to me, but I will leave a more "official" approval to someone with a little more context. I am unsure about the implications of this on the federation.

consideRatio:
I opened https://github.com/jupyterhub/binderhub/issues/1495 about the idea @manics, I see great value in it. Currently I'm positive to the idea.

consideRatio:
All the members are good atm, I'm not sure about gesis though, but if gesis is at 1.19 or older it makes sense to error if time is used to update the binderhub helm chart rather than the k8s cluster's version.

Turing: v1.21
GKE Staging: v1.21
GKE Prod: v1.21
OVH: v1.20

MridulS:
GESIS is still at v1.19 (I will update it ASAP) but interestingly the upgrade went through to the latest release (`0.2.0-n960.h47d7917`) without any error.

consideRatio:
@MridulS oh, that is very weird. What version of the `helm` CLI are you using? Maybe it is because binderhub is a dependency chart for the chart you actually install, but that seems off anyhow.

@MridulS btw, also note that there is a lot of misc changes of relevance to make to unstuck from old k8s versions and old helm chart dependencies. These are mentioned in https://github.com/jupyterhub/mybinder.org-deploy/issues/2182.

With regards to upgrading your k8s version, I think you should upgrade to k8s 1.21, but no further at this point, because of the migration steps needed to be done first in the issue linked above. But, it would be good if you are at 1.21 like most others at that point in time.

MridulS:
helm is at `v3.6.3` (very well could be a bug somewhere in helm) and yes [it's a dependency chart](https://github.com/gesiscss/orc/blob/master/gesisbinder/gesisbinder/requirements.yaml).

Thanks for the heads up!

I will update the cluster to v1.21 then :)

PR #1500, opened 2022-06-11
docker-image-cleaner 1.0.0-beta.2

manics:
Bring in https://github.com/jupyterhub/docker-image-cleaner/pull/51 to fix disk space issues on mybinder.org


consideRatio:
This PR needs to adjust to the breaking changes as well, about changed env var names

manics:
Updated!

consideRatio:
Now with this merged, there is automation that we can control in mybinder.org-deploy @manics, go for a "Run workflow" here https://github.com/jupyterhub/mybinder.org-deploy/actions/workflows/watch-dependencies.yaml. If the new dev version of this helm chart is available in the helm chart repo, it will be detected and a PR will be opened to bump.

PR #1511, opened 2022-07-21
[WIP] ProxyRepoProvider which routes requests to existing repoprovider based on external api

trevorspreadbury:
## What does this do?

This creates a RepoProvider subclass that allows for routing spec resolution through an external API. This has benefits for my use case and may or may not be useful enough to warrant inclusion in binderhub. The most direct use case is that this allows for a dynamic list of allowed repositories #280, which could be useful for organizations that wish to host a binderhub with the ability to regulate repositories that are available.

## How else can this be done?

This can be done through extraConfig. Due to `SPEC_NAMES` in `main.py` however, the repoprovider would have to be routed through an existing provider name, which seems somewhat hacky. Another potential update would be making all references to the list of repoproviders configurable. This might also be related to a solution to #322

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -62,6 +62,7 @@ from .repoproviders import (
     HydroshareProvider,
     RepoProvider,
     ZenodoProvider,
+    ProxyRepoProvider
 )
 from .utils import ByteSpecification, url_path_join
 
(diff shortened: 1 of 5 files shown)


consideRatio:
Hi @trevorspreadbury!

I'm struggling to understand what this class would do from the PRs description, can you clarify with a practical example or similar?

> This creates a RepoProvider subclass __that allows for routing spec resolution through an external API__.

PR #1512, opened 2022-07-25
Add `capabilities`/`privileged` to build container to support running on K8s without Docker 

manics:
I thought it'd be interesting to get BinderHub running on K8S without Docker. This is the first step:
- `build_docker_host` is optional
- support for adding capabilities to the build container, or running it as privileged (https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1SecurityContext.md)

I've only added one new config, `build_capabilities`, and overloaded it to set `privileged=True`. Alternatives include:
- separate config for `capabilities` and `privileged`
- allowing the full security_context to be specified- this would be the most flexible, but looking at https://github.com/kubernetes-client/python/issues/977 it doesn't sound like it's possible to construct the a Kubernetes object form JSON... unless someone knows a way to do so?

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -393,6 +393,13 @@ class Build:
                 client.V1EnvVar(name="GIT_CREDENTIAL_ENV", value=self.git_credentials)
             )
 
+        if "privileged" in self.build_capabilities:
+            security_context = client.V1SecurityContext(privileged=True)
+        else:
+            security_context = client.V1SecurityContext(
+                capabilities=client.V1Capabilities(add=self.build_capabilities)
+            )
+
         self.pod = client.V1Pod(
             metadata=client.V1ObjectMeta(
                 name=self.name,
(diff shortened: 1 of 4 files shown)


binderhub/build.py
```diff
@@ -382,6 +393,13 @@ def submit(self):
                 client.V1EnvVar(name="GIT_CREDENTIAL_ENV", value=self.git_credentials)
             )
 
+        if "privileged" in self.build_capabilities:
```
yuvipanda, line 396: Can we just ask for something like `build_security_context` or something and just pass it on directly through?

Part of the complexity here is that we want this to be not kubernetes specific. However, I think the way to do that is to refactor out the current class to be a KubeBuilder or something, and use traitlets directly. But until then, I think we should pass config through directly as much as possible.

yuvipanda (changes requested):
Thanks a lot for working on this, @manics!

binderhub/build.py
manics: Not easily, we'd need to recurse through the JSON dict by copying this KubeSpawner code:
https://github.com/jupyterhub/kubespawner/blob/63aaccc567d03110fb83c19cbdbfdc1a30eb5406/kubespawner/utils.py#L92-L188

Good point about the long-term plan to make this into a traitlets Configurable.... let me see if I can come up with something

manics: @yuvipanda I've added a new traitlets based build class in https://github.com/jupyterhub/binderhub/pull/1518

PR #1516, opened 2022-07-29
Using escaped username in progress api call affecting launch of serve…

pedroosorio:
Hi,

This PR aims to solve launch issues where binderhub would call the server progress api with an unescaped username, causing a 400 Bad Request to be thrown by the server. In turn, this would fail the server execution and launch another server, up to a maximum of 4 server. This behavior should also be addressed, in another PR.

Signed-off-by: Pedro Osório <pedro.osorio@atlar.pt>

### DIFF
--- a/binderhub/launcher.py
+++ b/binderhub/launcher.py
@@ -302,7 +302,7 @@ class Launcher(LoggingConfigurable):
                                     )
                                 )
 
-            url_parts = ["users", username]
+            url_parts = ["users", escaped_username]
             if server_name:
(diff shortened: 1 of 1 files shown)


piersharding:
Hi, this would be super helpful at the SKAO, as we have an internationalised user base.
Thanks,
Piers Harding.

manics (approved):
Sorry for the delay in reviewing

piersharding:
Thanks for processing this - Cheers.

PR #1518, opened 2022-08-01
Use Traitlets for passing parameters to `Build` classes

manics:
BinderHub is slowly gaining the ability to move beyond running on Kubernetes with Docker.

This is the first step to make it possible to configure parameters specific to the underlying infrastructure (Kubernetes, Docker, Podman, Local, etc). I've tried to minimise the changes.

See:
- https://github.com/jupyterhub/binderhub/issues/1318
- https://github.com/jupyterhub/binderhub/pull/1512#discussion_r931624071

The `build.py` changes look very large, but it's mostly rearranging the existing code:
- Common parameters and methods are moved out of `Build` and into the Traitlets based `BuildExecutor` base class (suggestions for alternative names welcome- if we want backwards compatibility we have to keep the generically named `Build` class as the Kubernetes builder)
- K8S specific methods are moved from `Build` to the Traitlets based `KubernetesBuildExecutor`
- The original `Build` class is now just a wrapper class around `KubernetesBuildExecutor`. The `Build.__init__` function signature is unchanged, so hopefully if someone's subclassed it their class should still work with binderHub.

`builder.py` now checks whether the BuildClass is the old existing `Build` or the newer `BuildExecutor` based classes. If it's the former the behaviour is unchanged. If it's the latter then parameters should be set using Traitlets in a config file or similar, the current BinderHub command-line args will be ignored.

For some reason switching to Traitlets surfaced a unset `IOLoop.current()`, so I've switched to avoiding tornado.ioloop.IOLoop.current()` other than in the BinderHub initialization code. Instead the loop is passed as a parameter to the `Build*` classes. I can try and split it into a separate PR if you prefer but it'll end up conflicting here due to the refactoring.

This PR currently includes unmerged PR:
- [x] https://github.com/jupyterhub/binderhub/pull/1517


binderhub/tests/test_build.py
```diff
@@ -122,15 +122,15 @@ def test_default_affinity():
         api=mock_k8s_api,
         name="test_build",
         namespace="build_namespace",
-        repo_url=mock.MagicMock(),
-        ref=mock.MagicMock(),
-        build_image=mock.MagicMock(),
-        image_name=mock.MagicMock(),
-        push_secret=mock.MagicMock(),
-        memory_limit=mock.MagicMock(),
+        repo_url=mock.MagicMock(spec=str),
```
manics, line 125: Do we need `mock.MagicMock` in these tests, or could we use a random string?

binderhub/build.py
```diff
@@ -630,3 +694,89 @@ def stream_logs(self):
...
+        self.memory_request = memory_request
+        self.docker_host = docker_host
+        self.node_selector = node_selector
+        self.appendix = appendix
+        self.log_tail_lines = log_tail_lines
+        self.git_credentials = git_credentials
+        self.sticky_builds = sticky_builds
+
+    @classmethod
+    def cleanup_builds(cls, kube, namespace, max_age):
```
manics, line 775: BinderHub hard-codes the cleanup command here:
https://github.com/jupyterhub/binderhub/blob/0fa35a06e7f7d054f8bb8dd2cba158ed7f45ac68/binderhub/app.py#L940-L944

Though since the `Build` class is hardcoded maybe it's fine to get rid of the `cleanup_builds()` method since even if it's overridden it'll never be used?

yuvipanda:
Omg so excited for this!!!!!

binderhub/build.py
manics: If we switch the default to the new traitlets class (https://github.com/jupyterhub/binderhub/pull/1521) we won't need to worry about this.

binderhub/tests/test_build.py
```diff
@@ -215,22 +215,19 @@ def test_git_credentials_passed_to_podspec_upon_submit():
     assert env["GIT_CREDENTIAL_ENV"] == git_credentials
 
 
-async def test_local_repo2docker_build():
+async def test_local_repo2docker_build(io_loop):
```
minrk, line 218: I don't think io_loop fixture should be needed here to pass explicitly to the Build. The build object should be able to manage its own reference to the calling loop if/when it needs it.

minrk (commented):
This is great!

I left some comments, mostly around the use of IOLoop.current and references to io loops.

binderhub/tests/test_build.py
manics: I've removed MagicMock for most parameters

binderhub/app.py
```diff
@@ -726,6 +730,7 @@ def init_pycurl(self):
     def initialize(self, *args, **kwargs):
         """Load configuration settings."""
         super().initialize(*args, **kwargs)
+        self.ioloop = tornado.ioloop.IOLoop.current()
```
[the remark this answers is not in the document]
manics: I've reverted most of the IOLoop changes

binderhub/build.py
```diff
@@ -49,133 +52,62 @@ def __init__(self, kind: Kind, payload: Union[str, BuildStatus]):
...
-        self.memory_request = memory_request
-        self.docker_host = docker_host
-        self.node_selector = node_selector
-        self.appendix = appendix
-        self.log_tail_lines = log_tail_lines
+    image_name = Unicode(help="Full name of the image to build. Includes the tag.")
 
-        self.stop_event = threading.Event()
-        self.git_credentials = git_credentials
+    git_credentials = Any(
```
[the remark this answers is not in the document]
manics: I've made it `CUnicode` since
https://github.com/jupyterhub/binderhub/blob/656204bb62d46be51fc7fb79cc5d7600b71a7614/binderhub/tests/test_build.py#L179-L182
passes a dict. Arguably it should be `Unicode` to avoid unexpected string conversions, but I think that can be done in a follow-up breaking PR if we want to.

binderhub/builder.py
```diff
@@ -425,34 +424,75 @@ async def get(self, provider_prefix, _unescaped_spec):
...
+                # build_image=self.settings["build_image"],
+                # memory_limit=self.settings["build_memory_limit"],
+                # memory_request=self.settings["build_memory_request"],
+                # docker_host=self.settings["build_docker_host"],
+                # node_selector=self.settings["build_node_selector"],
+                appendix=appendix,
+                # log_tail_lines=self.settings["log_tail_lines"],
+                git_credentials=provider.git_credentials,
+                # sticky_builds=self.settings["sticky_builds"],
+                main_loop=self.settings["ioloop"],
```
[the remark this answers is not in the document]
manics: I made it an external parameter after some trial and error because I was getting errors due to `IOLoop.current()` not being set in the CI tests. I've changed it to be an internal parameter that's set in `BuildExecutor.__init__()` and the tests are passing. I tried defining it as a traitlet and setting it using
```python
    main_loop = Any(allow_none=False)

    @default("main_loop")
    def _default_main_loop(self):
        return IOLoop.current()
```
but that didn't work.

binderhub/tests/conftest.py
```diff
@@ -131,11 +131,9 @@ def mock_asynchttpclient(request):
 def io_loop(event_loop, request):
```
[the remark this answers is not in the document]
manics: Done!

binderhub/build.py
```diff
@@ -49,133 +52,62 @@ def __init__(self, kind: Kind, payload: Union[str, BuildStatus]):
...
+        allow_none=True,
```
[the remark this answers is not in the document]
manics: I've updated most traitlets. I've kept `allow_none=True` for the cases where `None` is passed in `Build.__init__()`
https://github.com/jupyterhub/binderhub/blob/656204bb62d46be51fc7fb79cc5d7600b71a7614/binderhub/build.py#L86-L93

Removing `allow_none` is a cleaner, but since it's a breaking change I think that's best left for a follow-up breaking PR.

binderhub/builder.py
```diff
@@ -425,33 +425,72 @@ async def get(self, provider_prefix, _unescaped_spec):
...
+                repo_url=repo_url,
+                ref=ref,
+                image_name=image_name,
+                # push_secret=push_secret,
+                # build_image=self.settings["build_image"],
+                # memory_limit=self.settings["build_memory_limit"],
+                # memory_request=self.settings["build_memory_request"],
+                # docker_host=self.settings["build_docker_host"],
+                # node_selector=self.settings["build_node_selector"],
+                appendix=appendix,
```
manics, line 466: Should appendix be passed in from `BinderHub.appendix` which therefore overrides `BuildExecutor.appendix`, or should it be set by the `BuildExecutor.appendix` Traitlet only? On one hand it's only applicable to repo2docker which points to it being configurable in `BuildExecutor`, on the other hand changes to appendix can result in major changes to the built images which implies it's a higher-level BinderHub feature.

minrk: In general, I think it's best for configurables to only show up one place, and close to the place they are used - there are likely a few things on BinderHub that should be on BuildExecutor now that it's configurable. We probably shouldn't have things that are set on BinderHub, but only to be used on BuildExecutor. That organized config a bit better, I think, and eliminates the App->settings->Build pipeline. But it's a backward-compatibility issue to move config.

binderhub/builder.py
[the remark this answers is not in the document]
minrk: > due to IOLoop.current() not being set in the CI tests

Making tests async should solve this, because there will always be a current loop if the loop is running. This also makes the tests reflect reality more accuractely, because when an app is running, there is always a current loop.

> I tried defining it as a traitlet and...that didn't work.

My guess is the default main loop traitlet didn't work because the first access could (at least sometimes) occur in the background thread, and the current loop is per-thread. I think setting it in init is the right way to go. 👍

binderhub/build.py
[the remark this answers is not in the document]
minrk: I think this was a bug in the test - an env var can't have a value that's a dict, and [this](https://github.com/jupyterhub/binderhub/blob/master/binderhub/build.py#L382) is how we use it:

```python
        if self.git_credentials:
            env.append(
                client.V1EnvVar(name="GIT_CREDENTIAL_ENV", value=self.git_credentials)
            )
```

So I think it does strictly need to be a string. This is type info revealing a bug!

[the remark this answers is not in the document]
manics: I've changed it to `Unicode`

binderhub/builder.py
manics: I've changed it so `BinderHub.appendix` is ignored by `BuildExecutor`.

Note `BinderHub.appendix` isn't used directly, it's expanded as a Jinja2 template:
https://github.com/jupyterhub/binderhub/blob/b78cac34ee6a2324822d9596137462b022d7dec9/binderhub/builder.py#L420-L426

The appendix is empty by default so in the default case this has no effect. In the past `appendix` was set in mybinder.org-deploy, but that was subsequently removed in https://github.com/jupyterhub/mybinder.org-deploy/pull/1284/ , the relevant information is now passed at runtime, so this should be OK to remove in a later breaking PR.

binderhub/builder.py
```diff
@@ -425,33 +425,72 @@ async def get(self, provider_prefix, _unescaped_spec):
...
-            memory_request=self.settings["build_memory_request"],
-            docker_host=self.settings["build_docker_host"],
-            node_selector=self.settings["build_node_selector"],
-            appendix=appendix,
-            log_tail_lines=self.settings["log_tail_lines"],
-            git_credentials=provider.git_credentials,
-            sticky_builds=self.settings["sticky_builds"],
-        )
+        if issubclass(BuildClass, Build):
+            # Deprecated
```
consideRatio, line 429: Maybe this can be described a bit or a link to the PR description as a reference to an explanation on what is going on?

From the PR description:

> The `build.py` changes look very large, but it's mostly rearranging the existing code:
> 
> * Common parameters and methods are moved out of `Build` and into the Traitlets based `BuildExecutor` base class (suggestions for alternative names welcome- if we want backwards compatibility we have to keep the generically named `Build` class as the Kubernetes builder)
> * K8S specific methods are moved from `Build` to the Traitlets based `KubernetesBuildExecutor`
> * The original `Build` class is now just a wrapper class around `KubernetesBuildExecutor`. The `Build.__init__` function signature is unchanged, so hopefully if someone's subclassed it their class should still work with binderHub.
> 
> `builder.py` now checks whether the BuildClass is the old existing `Build` or the newer `BuildExecutor` based classes. If it's the former the behaviour is unchanged. If it's the latter then parameters should be set using Traitlets in a config file or similar, the current BinderHub command-line args will be ignored.

binderhub/builder.py
```diff
@@ -425,33 +425,72 @@ async def get(self, provider_prefix, _unescaped_spec):
...
+                    done = True
+                    failed = True
+                    # TODO: Propagate error to front-end
+
             build_starttime = time.perf_counter()
             pool = self.settings["build_pool"]
             # Start building
             submit_future = pool.submit(build.submit)
-            # TODO: hook up actual error handling when this fails
+            submit_future.add_done_callback(_check_result)
```
consideRatio, line 493: (These are changes part of #1517 now merged)

binderhub/build.py
```diff
@@ -213,72 +149,137 @@ def get_cmd(self):
...
+        """
+        pass
+
+    def stop(self):
+        """
+        Stop watching progress of build
+
+        TODO: What exactly is the purpose of this method?
+        """
+        self.stop_event.set()
```
consideRatio, line 185: @minrk introduced `stop` and `stop_event` in #628 with the PR description:

> ensures we aren't filling up the build pool with watchers no longer hooked up to any current requests

binderhub/builder.py
consideRatio:
```suggestion
        if issubclass(BuildClass, Build):
            # Deprecated, see docstring of the Build class for more details
```

binderhub/build.py
```diff
@@ -630,3 +709,89 @@ def stream_logs(self):
...
+        self.log_tail_lines = log_tail_lines
+        self.git_credentials = git_credentials
+        self.sticky_builds = sticky_builds
+
+    @classmethod
+    def cleanup_builds(cls, kube, namespace, max_age):
+        """Delete stopped build pods and build pods that have aged out"""
+        key = (kube, namespace, max_age)
+        if key not in Build._cleaners:
+            Build._cleaners[key] = KubernetesCleaner(
```
consideRatio, line 794: This seems technically wrong:

```python
if my_tuple not in empty_dict:
    empty_dict[my_tuple] = ...
```

consideRatio (approved):
Wieee this looks good to me!

binderhub/build.py
manics: I've added the explanation to the docstring.

binderhub/builder.py
manics: I've rebased this PR

manics:
Thanks for reviewing!

consideRatio:
@minrk wdyt, merge?

minrk:
Wonderful!

PR #1521, opened 2022-08-03
Breaking: Default to traitlets based Build class

manics:
This is a follow-up to https://github.com/jupyterhub/binderhub/pull/1518 that switches the default build class from `Build` to the new Traitlets based `KubernetesBuildExecutor` class.


manics:
This is ready for review!

binderhub/app.py
```diff
@@ -830,25 +881,22 @@ def initialize(self, *args, **kwargs):
 
         launch_quota = self.launch_quota_class(parent=self, executor=self.executor)
 
+        # Construct a Builder so that we can extract parameters such as the
+        # configuration or the version string to pass to /version and /health handlers
+        example_builder = self.build_class(parent=self)
```
consideRatio, line 886: I'm not confident about this situation, but wondering if it would make practical sense to work with class state on the build class itself instead of state object instance state for this information.

If so, we wouldn't have to create a dummy object, but instead just read the relevant information from the class.

What do you think?

consideRatio: I suspect this relates to `identifier` now being made a configurable traitlet. If we go for class state, the `identifier` field should probably not be a traitlet - but it doesn't seem like something that makes sense for users to configure themselves, but rather like something for the creator of the buildexecutor class to declare.

manics: It's based on a traitlets  value, which AFAIK can't be static. I've renamed the variable where it's used to make it clearer, see e38480aaf7fb89e4eaf896073a4d77ce294d431d

consideRatio: If it makes sense for this to be configurable, lets go with this as it is! Can you imagine a plausible situation when this could make sense to configure?

```python
# is this relevant/sensible to support?
c.KubernetesBuildExecutor.builder_info = <something>
```

manics: For `KubernetesBuildExecutor` the default value includes `build_image` which is a user-configurable traitlet, so unless traitlets properties can be static `builder_info` can't be a static class property.

For `config=True/False` ..... debatable. `/versions` is a public endpoint, including on authenticated BinderHubs, so one use could be to deliberately hide the `builder_info`. Not sure if that's useful..... what do you prefer?

consideRatio: Ah, well i think its not worth further thinking if we didnt come up with a clean alternative strategy by now.

---

PR LGTM ready for merge in my mind!

Thank you for amazing efforts in binderhub Simon!!!

consideRatio (approved):
Thanks for amazing work @manics!!!

@minrk, since this is a breaking change, do you mind having a look also?

minrk (approved):

consideRatio:
Thanks @manics!!! :heart: :tada:

PR #1525, opened 2022-08-16
If build pod is deleted check status instead of assuming Succeeded

manics:
Sometimes a build fails but BinderHub still attempts to launch the image.

I'm not sure if this is a race condition due to a pod exiting before BinderHub receives an error log from repo2docker, but it occurs everytime I try to build https://gist.github.com/manics/3c1a4babe6e9c368ee995c703fc0a2a9 on mybinder.org:

image

The log shows that `postBuild` fails with a non-zero error code as expected, but BinderHub continues to attempt to launch it... and obviously fails since the image wasn't built!

With this change the pod status/phase is checked, instead of always assuming it's succeeded.

This ensure BinderHub will not attempt to launch the image, but I don't think this correctly updates the front-end UI to an error status, instead the progress bar is stuck on `Building` (ignore the rest of the build logs, I'm playing with podman :smile:)
image

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -460,11 +460,22 @@ class KubernetesBuildExecutor(BuildExecutor):
                     _request_timeout=KUBE_REQUEST_TIMEOUT,
                 ):
                     if f["type"] == "DELETED":
-                        # Assume this is a successful completion
-                        self.progress(
-                            ProgressEvent.Kind.BUILD_STATUS_CHANGE,
-                            ProgressEvent.BuildStatus.COMPLETED,
+                        phase = f["object"].status.phase
+                        app_log.debug(
+                            "Pod %s was deleted with phase %s",
+                            f["object"].metadata.name,
+                            phase,
                         )
+                        if phase == "Succeeded":
+                            self.progress(
+                                ProgressEvent.Kind.BUILD_STATUS_CHANGE,
+                                ProgressEvent.BuildStatus.COMPLETED,
+                            )
+                        else:
+                            self.progress(
+                                ProgressEvent.Kind.BUILD_STATUS_CHANGE,
+                                ProgressEvent.BuildStatus.FAILED,
+                            )
                         return
                     self.pod = f["object"]
                     if not self.stop_event.is_set():
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -472,7 +472,7 @@ class KubernetesBuildExecutor(BuildExecutor):
                         if phase == "Succeeded":
                             self.progress(
                                 ProgressEvent.Kind.BUILD_STATUS_CHANGE,
-                                ProgressEvent.BuildStatus.COMPLETED,
+                                ProgressEvent.BuildStatus.BUILT,
                             )
                         else:
                             self.progress(
@@ -713,7 +713,7 @@ class FakeBuild(BuildExecutor):
                 ),
             )
         self.progress(
-            ProgressEvent.Kind.BUILD_STATUS_CHANGE, ProgressEvent.BuildStatus.COMPLETED
+            ProgressEvent.Kind.BUILD_STATUS_CHANGE, ProgressEvent.BuildStatus.BUILT
         )
         self.progress(
             "log",
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -40,9 +40,6 @@ class ProgressEvent:
         The state the build is now in
 
         Used when `kind` is `Kind.BUILD_STATUS_CHANGE`
-
-        These enum values are referenced in the front-end to display the build status,
-        hence the mismatch between the Enum name (backend) and value (frontend).
         """
 
         PENDING = "pending"
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -47,7 +47,7 @@ class ProgressEvent:
 
         PENDING = "pending"
         RUNNING = "running"
-        COMPLETED = "built"
+        BUILT = "built"
         FAILED = "failed"
         UNKNOWN = "unknown"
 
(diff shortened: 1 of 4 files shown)


minrk:
Great job tracking this down! It's been a mystery to me for ages.

I think the failure to update is because of the use of the string 'failure' instead of 'failed' [here](https://github.com/jupyterhub/binderhub/blob/c4af713bb20ed90606e9795834d160bd8a332578/binderhub/builder.py#L532). Everywhere else, we use 'failed' to indicate that the build has failed.

I'm not sure if there's a deliberate distinction between 'failure' and 'failed'? But only 'failed' is [handled in the js](https://github.com/jupyterhub/binderhub/blob/c4af713bb20ed90606e9795834d160bd8a332578/binderhub/static/js/index.js#L194).

We might reduce the likelihood of bugs like this if we used the ProgressEvent.BuildStatus enum as a `str` enum, and sent the string itself. Then it would be easier to keep track of what all valid values are.

binderhub/build.py
```diff
@@ -40,13 +40,16 @@ class BuildStatus(Enum):
...
         """
 
-        PENDING = 1
-        RUNNING = 2
-        COMPLETED = 3
-        FAILED = 4
-        UNKNOWN = 5
+        PENDING = "pending"
+        RUNNING = "running"
+        COMPLETED = "built"
```
manics, line 50: This mismatch in naming is due to
https://github.com/jupyterhub/binderhub/blob/c4af713bb20ed90606e9795834d160bd8a332578/binderhub/builder.py#L515-L517
Do you think we should rename the enum to `BUILT`?

binderhub/builder.py
```diff
@@ -529,9 +530,13 @@ def _check_result(future):
                         # Do nothing, is ok!
                         continue
                     elif progress.payload == ProgressEvent.BuildStatus.FAILED:
-                        event = {"phase": "failure"}
+                        event = {"phase": phase}
```
manics, line 533: This changes event to `{"phase": "failed"}`

binderhub/tests/test_build.py
```diff
@@ -94,6 +94,30 @@ async def test_build(app, needs_build, needs_launch, always_build, slug, pytestc
     assert r.url.startswith(final["url"])
 
 
+@pytest.mark.asyncio(timeout=120)
+@pytest.mark.remote
+async def test_build_fail(app, needs_build, needs_launch, always_build, pytestconfig):
+    """
+    Test build a repo that should fail immediately.
+    """
+    slug = "gist/manics/cc003c8264db20d22b5fa437770f655e"
```
manics, line 103: https://gist.github.com/manics/cc003c8264db20d22b5fa437770f655e
Shall I push this as a new branch `failed` in https://github.com/binderhub-ci-repos/minimal-dockerfile ?

manics (commented):
> I think the failure to update is because of the use of the string 'failure' instead of 'failed' [here](https://github.com/jupyterhub/binderhub/blob/c4af713bb20ed90606e9795834d160bd8a332578/binderhub/builder.py#L532). Everywhere else, we use 'failed' to indicate that the build has failed.

That was it! I've changed `ProgressEvent.BuildStatus` to have string values as you suggested.

Tests are currently failing due to a problem with installing Calico, https://projectcalico.docs.tigera.io/v3.23/manifests/calico.yaml is not responding.

The `"phase": "failure"` messages are from repo2docker:
- https://github.com/jupyterhub/repo2docker/blob/d415115b66f24d4165842f94a892987959c9cde3/repo2docker/app.py#L754
- https://github.com/jupyterhub/repo2docker/blob/d415115b66f24d4165842f94a892987959c9cde3/repo2docker/app.py#L816

Though repo2docker also uses `"failed"` in some places
- https://github.com/jupyterhub/repo2docker/blob/d415115b66f24d4165842f94a892987959c9cde3/repo2docker/app.py#L494
- https://github.com/jupyterhub/repo2docker/blob/d415115b66f24d4165842f94a892987959c9cde3/repo2docker/app.py#L549

Do you think it's worth standardising on `"failed"` in repo2docker?

binderhub/build.py
minrk: Probably less confusing that way, sure.

binderhub/tests/test_build.py
minrk: That sounds like a good idea

minrk:
> Do you think it's worth standardising on "failed" in repo2docker?

Yeah, if these are supposed to mean something to computers and not just humans, they should probably be consistent. Unless there's a _useful_ distinction, my guess is it was an oversight.

manics:
I'll need to dig into the failing/hanging tests

manics:
And now action-k3s-helm is failing, https://github.com/k3s-io/k3s/releases/download/latest/sha256sum-amd64.txt isn't working

consideRatio:
> And now action-k3s-helm is failing, https://github.com/k3s-io/k3s/releases/download/latest/sha256sum-amd64.txt isn't working

Oh =/ That will go away soon, they publish the checksums via another workflow triggered by the release.

manics:
Tests are passing after a rebase, though I don't know why there were failing before.

consideRatio:
I saw a 503 from gitlab, maybe intermittent issues with gitlab?

consideRatio (approved):
Wieee looks good to me!

Standardizing terminology across repositories makes perfect sense to me as well btw!

I come to think of this confusion, making me very thankful when its reasonable/possible to stick to one terminology:
- Dockerfile's ENTRYPOINT->CMD
- Kubernetes pod specifications `command` and `args`
- JupyterHub Spawner's `cmd` and how its behavior in kubespawner is like CMD/`args`.

PR #1526, opened 2022-08-21
Move pod quota checking into separate class

manics:
Whilst working on https://github.com/jupyterhub/binderhub/pull/1521 I realised the Kubernetes client is still used a few places.

This moves the quota checking code from [`builder.BuildHandler.launch`](https://github.com/jupyterhub/binderhub/blob/c4af713bb20ed90606e9795834d160bd8a332578/binderhub/builder.py#L589-L645) into it's own class. The logic should be unchanged.

After this the remaining use of the `kubernetes_client` is in the health check: https://github.com/jupyterhub/binderhub/blob/656204bb62d46be51fc7fb79cc5d7600b71a7614/binderhub/health.py#L108-L113

Note I haven't tested this on a real system yet.

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -50,6 +50,7 @@ from .launcher import Launcher
 from .log import log_request
 from .main import LegacyRedirectHandler, MainHandler, ParameterizedMainHandler
 from .metrics import MetricsHandler
+from .quota import KubernetesLaunchQuota, LaunchQuota
 from .ratelimit import RateLimiter
 from .registry import DockerRegistry
 from .repoproviders import (
@@ -304,6 +305,8 @@ class BinderHub(Application):
     pod_quota = Integer(
         None,
         help="""
+        DEPRECATED: Use c.LaunchQuota.total_quota
+
         The number of concurrent pods this hub has been designed to support.
 
         This quota is used as an indication for how much above or below the
@@ -319,6 +322,13 @@ class BinderHub(Application):
         config=True,
     )
 
+    @observe("pod_quota")
+    def _pod_quota_deprecated(self, change):
+        self.log.warning(
+            "BinderHub.pod_quota is deprecated, use LaunchQuota.total_quota"
+        )
+        self.config.LaunchQuota.total_quota = change.new
+
     per_repo_quota_higher = Integer(
         0,
         help="""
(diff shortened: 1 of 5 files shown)


binderhub/app.py
```diff
@@ -333,6 +343,17 @@ def _valid_badge_base_url(self, proposal):
         config=True,
     )
 
+    launch_quota_class = Type(
+        LaunchQuota,
+        default=KubernetesLaunchQuota,
+        help="""
+        The class used to check quotas for launched servers.
```
consideRatio, line 350: In the context of binderhub, it would be great to be explicit about what this means. I now understand it as "the amount of user servers scheduled-to-run-after-build/starting/running by binderhub in the jupyterhub that binderhub manages", but I'm open to it meaning something related to build pods as well, or that it could be coupled to possible jupyterhub user servers unmanaged by binderhub.

Is "launch quota" a name introduced in this PR?

consideRatio: Practically, does having a launch quota of 100 imply:
- you can at most have 100 build workloads, if you do, you must have 0 user servers started at this point in time
- you can at most have 100 user servers, if you do, you must have 0 build workloads running at this point in time

manics: Previously the quota was handled inside the `builder.BuildHandler.launch` method. By moving it into a separate class the `launch()` method becomes independent of the platform (k8s, docker, HPC, etc).

The quota check determines whether BinderHub should launch a repository which is why I've called the base class `LaunchQuota`. The decision on whether to allow a launch or not is an implementation detail. The current default (K8s only) method is to look at the total number of `label_selector="app=jupyterhub,component=singleuser-server"` pods, and the total number of those pods that are running an image from the repo being launched. In future there's scope for taking other factors into account such as build pods, total CPU/memory usage across all pods running a particular repo, etc.

It's not a perfect abstraction since `LaunchQuota.check_repo_quota(self, image_name, repo_config, repo_url)` takes a `repo_config` parameter, which is created by BinderHub using some parameters:
https://github.com/jupyterhub/binderhub/blob/e80f841ef43b10a5ffd54904daeb287357c638d7/binderhub/app.py#L292-L334
and a RepoProvider based class:
https://github.com/jupyterhub/binderhub/blob/e80f841ef43b10a5ffd54904daeb287357c638d7/binderhub/repoproviders.py#L134-L166

but I'm trying to minimise the changes to the codebase, assuming I've coded this correctly there should be zero change in the existing behaviour.

consideRatio: Thanks for the thorough clarfication @manics!!

consideRatio (approved):
Nice looks great! :heart: :tada: :sunflower:!!

PR #1530, opened 2022-09-12
Update https.rst

jim-fun:
nginx


consideRatio:
I think this may be correct, but I think this documentation is very outdated and needs to be significantly revised. For example, Helm charts under the helm chart repo referred to as stable is since long not maintained.

Issue #1513, opened 2022-07-25
Support running BinderHub on K8s without Docker


manics:
### Proposed change
Docker has been removed from several K8s distributions. In addition there have been requests to run BinderHub on more restricted K8s distributions such as OpenShift https://discourse.jupyter.org/t/unable-to-attach-or-mount-volumes-unmounted-volumes-dockersocket-host/14950

### Alternative options
Do nothing, though in future we may need to modify the deployment instructions to ensure Docker is available on the K8s hosts.

### Who would use this feature?
Someone who wants to run BinderHub on K8s without Docker.
Someone who wants to run BinderHub with reduced privileges.

### (Optional): Suggest a solution
There are several non-Docker container builders available, include:
- podman https://podman.io/
- buildah (used by Podman for building images) https://buildah.io/
- img https://github.com/genuinetools/img

repo2podman already works https://github.com/manics/repo2podman and it shouldn't be too hard to swap-in one of the other builders.

In theory it should be possible to run these without full privileges, with limited added capabilities, e.g.
- https://www.redhat.com/sysadmin/podman-inside-container
- https://blog.jessfraz.com/post/building-container-images-securely-on-kubernetes/

So far I've managed to get a proof-of-concept podman builder running using full privileges, supported by https://github.com/jupyterhub/binderhub/pull/1512 on AWS EKS:
```yaml
image:
  name: docker.io/manics/binderhub-dev
  tag: 2022-07-25-20-00

registry:
  url: docker.io
  username: <username>
  password: <password>

service:
  type: ClusterIP

config:
  BinderHub:
    base_url: /binder/
    build_capabilities:
      - privileged
    build_docker_host: ""
    build_image: "ghcr.io/manics/repo2podman:main"
    hub_url: /jupyter/
    hub_url_local: http://hub:8081/jupyter/
    image_prefix: <username>/binder-
    auth_enabled: false
    use_registry: true
  Application:
    log_level: DEBUG

extraConfig:
  0-repo2podman: |
    from binderhub.build import Build
    class Repo2PodmanBuild(Build):
        def get_r2d_cmd_options(self):
            return ["--engine=podman"] + super().get_r2d_cmd_options()
    c.BinderHub.build_class = Repo2PodmanBuild

jupyterhub:
  hub:
    baseUrl: /jupyter
    networkPolicy:
      enabled: false
  proxy:
    service:
      type: ClusterIP
    chp:
      networkPolicy:
        enabled: false
  scheduling:
[...]

PR #1531, opened 2022-09-16
Podman in k8s

sgaist:
This pull request is a refactor that allows to use alternatives to Docker as image builder as requested in jupyterhub/binderhub#1513. In this case, support for Podman is implemented.

The PR implements a rewrite of the DaemonSet providing the docker image build service.

The new `imageBuilderType` parameter introduced will let users select which image builder to use:
- local: same as what setting dind.enabled to false used to doo
- dind: same as what setting dind.enabled to true used to do
- pink: new builder using Podman in place of Docker

The daemonset name pink is the acronym of "podman in kubernetes".

Configuration example:

```
service:
  type: ClusterIP

config:
  BinderHub:
    debug: true
    hub_url: "https://my.host.com/jupyter"

    use_registry: true
    image_prefix: "my.registry.com/my-project/my-image-"

    build_docker_host: "unix:///local/fake_var/run/pink/docker.sock"
    build_image: "docker.io/sgaist/repo2podman:latest"

imageBuilderType: pink

pink:
  hostStorageDir: /local/fake_var/lib/pink/storage/
  hostSocketDir: /local/fake_var/run/pink/

imageCleaner:
  host:
    dockerSocket: /local/fake_var/run/pink/docker.sock
    dockerLibDir: /local/fake_var/lib/pink/storage

ingress:
  enabled: true
  ingressClassName: nginx
  hosts:
    - my.host.com
```

One advantage of this new implementation is that it should allow the addition of new build services without requiring core modifications.


sgaist:
Small note, as it is a preliminary work I have not written the documentation related to it yet.

sgaist:
Following this month's JupyterHub's meeting, @manics and myself continued a bit on the podman front discussing the podman engine for repo2docker as well as this patch content. One thing that I was not aware of was that beside podman being a drop in replacement for docker in terms of command line interface, it now also implements a docker compatible API from a service point of view.
Therefore I ran a BinderHub deployment test with the pink daemonset enabled, the standard Build class, and the image cleaner configured to watch the storage folder of podman.

All these together worked quite nicely.

This means that the environment variable addition is not strictly required for the goal of using podman in place of docker for the pod building the images. However it does allow people to choose whether they want to go full podman, full docker, or a mix of both as it would be the default.

manics:
The diff between the dind and pink K8s templates is fairly small

```diff
diff --color -Nur helm-chart/binderhub/templates/dind/daemonset.yaml helm-chart/binderhub/templates/pink/daemonset.yaml
--- helm-chart/binderhub/templates/dind/daemonset.yaml	2021-09-18 20:57:59.971717855 +0100
+++ helm-chart/binderhub/templates/pink/daemonset.yaml	2022-09-23 13:29:38.484322020 +0100
@@ -1,20 +1,20 @@
-{{- if .Values.dind.enabled -}}
+{{- if .Values.pink.enabled -}}
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: {{ .Release.Name }}-dind
+  name: {{ .Release.Name }}-pink
 spec:
   updateStrategy:
     type: RollingUpdate
   selector:
     matchLabels:
-      name:  {{ .Release.Name }}-dind
+      name:  {{ .Release.Name }}-pink
   template:
     metadata:
       labels:
-        name: {{ .Release.Name }}-dind
+        name: {{ .Release.Name }}-pink
         app: binder
-        component: dind
+        component: dind  # Lying to build so current affinity rules can work
         release: {{ .Release.Name }}
         heritage: {{ .Release.Service }}
     spec:
@@ -28,47 +28,59 @@
         operator: Equal
         value: user
       nodeSelector: {{ .Values.config.BinderHub.build_node_selector | default dict | toJson }}
-      {{- with .Values.dind.initContainers }}
+      {{- with .Values.pink.initContainers }}
       initContainers:
         {{- . | toYaml | nindent 8 }}
       {{- end }}
+      {{- with .Values.pink.daemonset.image.pullSecrets }}
+      imagePullSecrets:
+        {{- . | toYaml | nindent 8 }}
+      {{- end }}
+      securityContext:
+        fsGroup: 1000
+
       containers:
-      - name: dind
-        image: {{ .Values.dind.daemonset.image.name }}:{{ .Values.dind.daemonset.image.tag }}
+      - name: pink
+        image: {{ .Values.pink.daemonset.image.name }}:{{ .Values.pink.daemonset.image.tag }}
+        imagePullPolicy: {{ .Values.pink.daemonset.image.pullPolicy }}
+
         resources:
-          {{- .Values.dind.resources | toYaml | nindent 10 }}
+          {{- .Values.pink.resources | toYaml | nindent 10 }}
         args:
-          - dockerd
-          - --storage-driver={{ .Values.dind.storageDriver }}
-          - -H unix://{{ .Values.dind.hostSocketDir }}/docker.sock
-          {{- with .Values.dind.daemonset.extraArgs }}
-          {{- . | toYaml | nindent 10 }}
-          {{- end }}
+          - podman
+          - system
+          - service
+          - --time=0
+          - unix:///var/run/pink/docker.sock
         securityContext:
           privileged: true
+          runAsUser: 1000  # podman default user id
+
         volumeMounts:
-        - name: dockerlib-dind
-          mountPath: /var/lib/docker
-        - name: rundind
-          mountPath: {{ .Values.dind.hostSocketDir }}
-        {{- with .Values.dind.daemonset.extraVolumeMounts }}
+        - mountPath: /home/podman/.local/share/containers/storage/
+          name: podman-containers
+        - mountPath: /var/run/pink/
+          name: podman-socket
+
+        {{- with .Values.pink.daemonset.extraVolumeMounts }}
         {{- . | toYaml | nindent 8 }}
         {{- end }}
-        {{- with .Values.dind.daemonset.lifecycle }}
+        {{- with .Values.pink.daemonset.lifecycle }}
         lifecycle:
           {{- . | toYaml | nindent 10 }}
         {{- end }}
       terminationGracePeriodSeconds: 30
       volumes:
-      - name: dockerlib-dind
+      - name: podman-containers
         hostPath:
-          path: {{ .Values.dind.hostLibDir }}
-          type: DirectoryOrCreate
-      - name: rundind
+          path: {{ .Values.pink.hostStorageDir }}
+          type: Directory
+      - name: podman-socket
         hostPath:
-          path: {{ .Values.dind.hostSocketDir }}
-          type: DirectoryOrCreate
-      {{- with .Values.dind.daemonset.extraVolumes }}
+          path: {{ .Values.pink.hostSocketDir }}
+          type: Directory
+
+      {{- with .Values.pink.daemonset.extraVolumes }}
       {{- . | toYaml | nindent 6 }}
       {{- end }}
 {{- end }}
```

so a couple of alternative options are:

### Parameterise the dind templates:
- Make `.Values.dind.securityContext` fully configurable
- Allow the dind volumeMounts to be overridden (or perhaps have an enable/disable switch, and fully configure podman through `.Values.dind.daemonset.extraVolumeMounts` and `.Values.dind.daemonset.extraVolumes`)

### Add conditionals to the dind templates, e.g.
```yaml
        args:
       {{- if eq .Values.containerBuilderPod "dind" }}
          - dockerd
          - --storage-driver={{ .Values.dind.storageDriver }}
          - -H unix://{{ .Values.dind.hostSocketDir }}/docker.sock
       {{- end }}
       {{- if eq .Values.containerBuilderPod "pink" }}
          - podman
          - system
          - service
          - --time=0
          - unix:///var/run/pink/docker.sock
       {{- end }}
        {{- with .Values.dind.daemonset.extraArgs }}
        {{- . | toYaml | nindent 10 }}
        {{- end }}
```
If we decide on this route I think deprecating `.Values.dind.enabled` and replacing it with something like `.Values.containerBuilderPod` will be more extensible in the future.

sgaist:
I like the idea of a configurable `containerBuilderPod`

Should we also consider modifying the `dind` label used to keep build pods on the same node that the builder pod itself ?

manics:
Maybe we could do this in two PRs? This one for getting podman to work (so a configurable `containerBuilderPod` with no breaking changes). Then a separate PR for renaming the `dind` label to something more generic, since it's an internal breaking change- in fact a PR to change the label could be opened now, since it'll be a fairly minor conflict to resolve, and means others can agree/disagree!

sgaist:
As per your excellent idea I merge the two templates into a "new one" that covers both cases and should be easy to extend further.

I haven't touched yet the securityContext part as I wanted to first ensure that the new template would fit the bill.

binderhub/tests/test_build.py
```diff
@@ -241,6 +241,48 @@ def test_git_credentials_passed_to_podspec_upon_submit():
...
 
+def test_extra_environment_variables_passed_to_podspec_upon_submit():
+    extra_environments = {
+        "CONTAINER_HOST": "unix:///var/run/docker.sock",
+        "REGISTRY_AUTH_FILE": "/root/.docker/config.json",
+    }
+
+    mock_k8s_api = _list_dind_pods_mock()
+
+    class EnvBuild(Build):
```
manics, line 252: The `Build` class is deprecated and is only kept for backwards compatibility. Are you able to inherit from `BuildExecutor` instead, and configure it using traitlets?

docs/source/zero-to-binderhub/setup-binderhub.rst
```diff
@@ -474,4 +474,29 @@ for each DinD application::
...
+as with Docker. BinderHub supports an alternative that uses `Podman-in-Kubernetes
+(PinK) <https://www.redhat.com/sysadmin/podman-inside-kubernetes>`_. To turn
+`pink` on, you'll need to set the following configuration in your ``config.yaml`` file::
+
+    containerBuilderPod: pink
+    pink:
+      daemonset:
+        image:
+          name: quay.io/podman/stable
+          tag: v4.2.0
```
manics, line 491:
```suggestion

You can optionally override the default podman image:

    pink:
      daemonset:
        image:
          name: quay.io/podman/stable
          tag: v4.2.0
```

helm-chart/binderhub/templates/container-builder/daemonset.yaml
```diff
@@ -0,0 +1,123 @@
+{{- if ne .Values.containerBuilderPod "disabled" -}}
+{{- $builder := (index .Values .Values.containerBuilderPod) }}
+{{- $daemonset := $builder.daemonset }}
```
manics, line 3: How hard is it to also allow `containerBuilderPod="disabled" and dind.enabled` for backwards compatibility but add a deprecation message? I'd like to get this PR in, and if we can avoid a breaking change that'll make things a bit easier!

helm-chart/binderhub/templates/container-builder/daemonset.yaml
```diff
@@ -0,0 +1,123 @@
...
+          {{- if eq .Values.containerBuilderPod "pink" }}
+          args:
+            - podman
+            - system
+            - service
+            - --time=0
+            - unix:///var/run/pink/docker.sock
+          securityContext:
+            privileged: true
+            runAsUser: 1000  # podman default user id
```
manics, line 77:
```suggestion
            runAsUser: 0
```
I couldn't get this to work with the default `1000` uid, possibly due to permissions on the host mounted volumes.

helm-chart/binderhub/files/binderhub_config.py
```diff
@@ -46,10 +46,11 @@ def get_value(key, default=None):
 for section, sub_cfg in get_value("config", {}).items():
     c[section].update(sub_cfg)
 
-if get_value("dind.enabled", False) and get_value("dind.hostSocketDir"):
-    c.BinderHub.build_docker_host = "unix://{}/docker.sock".format(
-        get_value("dind.hostSocketDir")
-    )
+containerBuilderPod = get_value("containerBuilderPod")
```
manics, line 49: It took me a while to figure this out.... although `get_value()` reads from `values.yaml`
https://github.com/jupyterhub/binderhub/blob/bef386a4875f511368e38b6af1ca98f233eb6fa8/helm-chart/binderhub/files/binderhub_config.py#L15
it confusing doesn't read the full `values.yaml`, instead it's only a subset of keys:
https://github.com/jupyterhub/binderhub/blob/bef386a4875f511368e38b6af1ca98f233eb6fa8/helm-chart/binderhub/templates/secret.yaml#L18-L19
so `secret.yaml` needs to be
```yaml
  values.yaml: |
    {{- pick .Values "config" "containerBuilderPod" "cors" "dind" "extraConfig" "pink" | toYaml | nindent 4 }}
```

helm-chart/binderhub/files/binderhub_config.py
```diff
@@ -46,10 +46,11 @@ def get_value(key, default=None):
...
+if containerBuilderPod in ["dind", "pink"]:
+    hostSocketDir = get_value(f"{containerBuilderPod}.hostSocketDir")
+    if hostSocketDir:
+        c.BinderHub.build_docker_host = f"unix://{hostSocketDir}/docker.sock"
```
manics, line 53:
```suggestion
        c.KubernetesBuildExecutor.docker_host = f"unix://{hostSocketDir}/docker.sock"
```
Does this work for you?

helm-chart/binderhub/schema.yaml
```diff
@@ -349,14 +350,17 @@ properties:
...
+    enum: ["disabled", "dind", "pink"]
+    default: "disabled"
+    description: |
+      Selected builder pod type
+
   dind:
     type: object
     additionalProperties: false
     properties:
-      enabled:
```
manics, line 356: Is it reasonable to deprecate instead of making a breaking change?

manics (commented):
This looks great, sorry for taking so long to review.

> I haven't touched yet the securityContext part as I wanted to first ensure that the new template would fit the bill.

If it's OK with you could we leave this for a follow-up PR so that we can get this merged?

helm-chart/binderhub/files/binderhub_config.py
sgaist: I have completely missed that and it just hit me on Friday. It looks like I managed to test drive BinderHub with only images that were already present and did not require a fresh pull.

Thanks for the fix !

helm-chart/binderhub/files/binderhub_config.py
sgaist: Only if I also set the `build_class` traitlet from the `BinderHub` class to use `KubernetesBuildExecutor`.

Otherwise, in `builder.py` the value will be replaced when the `build` object is created starting on line 455.

binderhub/tests/test_build.py
sgaist: In this case, `KubernetesBuildExecutor` shall be used as the tested changed is implemented there but that said, the test is working using it and traitlets.

helm-chart/binderhub/schema.yaml
sgaist: Yes, we can do it that way. With a mention in the `NOTES.txt` as it seems that helm does not have yet the support for handling deprecation in the schema.

helm-chart/binderhub/templates/container-builder/daemonset.yaml
sgaist: Less that what I thought would be.

The changes required are not really intrusive and easy to remove once the deprecation period is over.

sgaist:
@manics I think it's a good idea to do the Pod Security Policy work in a separate PR. With the replacement of them in K8s 1.25 for the Pod Security admission controller, it might be better to concentrate on the latter.

helm-chart/binderhub/schema.yaml
consideRatio: I suggest we go for a hard deprecation, forcing users to transition as part of this upgrade. Since we don't have a nice version control, let's go for a decisive and verbose error that will fail. This failure will happen on template rendering, and doesn't risk breaking anything in the k8s cluster.

1. Keep what is being removed allowed in the schema, otherwise `helm` with error with "not in schema" before you can provide nice deprecation messages.
2. Let NOTES.txt handle the deprecation logic, copy paste from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/2.0.0/jupyterhub/templates/NOTES.txt#L116-L153 and adjust.

helm-chart/binderhub/templates/container-builder/daemonset.yaml
consideRatio: I'm +1 for reduced complexity and a hard failure during template rendering if the removed configuration is observed, where users are provided with information on how to migrate their config.

I wrote about it in https://github.com/jupyterhub/binderhub/pull/1531/files/6794005f7680576b5ac47684822f822d2d711866#r1040093732, its easy and straight forward based on those instructions.

There is no point in deprecating something if we don't actively communicate something to be deprecated etc, and that is not something we will be able to do without changelog and/or warning messages etc.

I think a clear hard failure during template rendering is brilliant, the worst thing that happens is that you are forced to handle a deprecation sooner rather than later when you wish to make an upgrade of binderhub.

helm-chart/binderhub/templates/container-builder/daemonset.yaml
```diff
@@ -0,0 +1,123 @@
...
+          {{- . | toYaml | nindent 10 }}
+          {{- end }}
+
+          {{- with $daemonset.lifecycle }}
+          lifecycle:
+            {{- . | toYaml | nindent 12 }}
+          {{- end }}
+
+      volumes:
+      {{- if eq .Values.containerBuilderPod "dind" }}
```
consideRatio: Can you indent the `if` statements to be on the same depth as the content within them?

Sorry for being picky about indentation in helm template manifests,

helm-chart/binderhub/files/binderhub_config.py
manics: Let's leave it as is for now then, and come back to it later

helm-chart/binderhub/schema.yaml
manics: The main reason I suggested a deprecation was so this PR could be merged as soon as possible with the decision over the breaking change being made in a follow-up. I'm happy with either option though.

manics (approved):
I've tested this, there's a few issues with Podman but they can be investigated in a follow-up as the key features are implemented!

I'm happy for this to be merged once @consideRatio is happy

docs/source/zero-to-binderhub/setup-binderhub.rst
```diff
@@ -452,20 +452,20 @@ clone a repo.
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-By default, BinderHub will build pods with the host Docker installation.
+By default, BinderHub will build images with the host Docker installation.
 This often means you are stuck with whatever version of Docker provided by your
 cloud provider. BinderHub supports an alternative that uses `Docker-in-Docker
 (DinD) <https://hub.docker.com/_/docker>`_. To turn `dind` on, you'll need to set
 the following configuration in your ``config.yaml`` file::
 
+    containerBuilderPod: dind
```
consideRatio, line 461: In the paragraph above, it sais "binderhub will build images", but we speak about a containerBuilderPod. Isn't this an imageBuilderPod?

I think of images as static artifacts that can be built, and when built they can be used to start a container.

I also want to highlight that maybe `pod` isn't what makes sense here either. In the schema description you write `Selected builder pod type`. It doesn't seem critical to declare that it must be a k8s Pod either, it must just be some `imageBuilderType` I'd say, and if that type leads to creating a DaemonSet with one pod on each node etc, or building the image locally where the binderhub software runs, then so be it.

---

Thinking of that, what happens when `imageBuilderType=disabled` is configured or `dind.enabled=False` as it was before? Is `disabled` really what we should say here, or is it something else like `local` etc? I'm not confident what it means to not have `dind.enabled=False` currently - Would binderhub would still work in some way, and if so, how - and how would we describe that imageBuildingType?

I think I'd suggest `imageBuilderMode` or `imageBuilderType` or similar, but the words container and pod doesn't seem right here.

helm-chart/binderhub/schema.yaml
consideRatio: I'm super happy about a breaking change in the helm chart's schema if implemented with a clear error message like above, and would prefer it over growing complexity.

This kind of breaking changes has been very smooth to document and migrate historically, lets do it!

helm-chart/binderhub/templates/container-builder/daemonset.yaml
sgaist: Sure thing !

I have no issue with pickiness. Clean code is better code.

helm-chart/binderhub/schema.yaml
sgaist: If I get things correctly, I should remove the warning I added in `NOTES.txt` file and make it error direcly if `.Values.dind.enabled` is found rather than the current check that checks both this and the `containerBuilderPod` value ?

tools/templates/lint-and-validate-values-dind.yaml
```diff
@@ -0,0 +1,153 @@
+pdb:
```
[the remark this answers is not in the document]
sgaist: Do you also want to test when `containerBuilderPod`'s value is disabled ?

It's the current default in `lint-and-validate-values.yaml`

I am not sure I follow you on the CI test suite VS CI test run, is it about `.github/workflows/test.yml` VS `validate-against-schema.py` ?
If so, I can update the latter to match the former more closely and so we can use separate files as your original suggestion.

docs/source/zero-to-binderhub/setup-binderhub.rst
manics: Strictly speaking the build pod shouldn't even need to use a socket, it should be possible to use a tcp connection to the docker/podman host. I managed to get it working a while back but had some other problems, but I think it's worth revisiting in the future as it takes us one step closer to having a K8s BinderHub that's not dependent on the host configuration.

sgaist: If memory serves well, in the context of Podman, I had it working with a tcp socket using repo2podman and a k8s service.
So definitively something to check again.

helm-chart/binderhub/templates/container-builder/daemonset.yaml
```diff
@@ -0,0 +1,125 @@
...
+          - name: rundind
+            mountPath: {{ $builder.hostSocketDir }}
+          {{- end }}
+          {{- if eq $builderName "pink" }}
+          args:
+            - podman
+            - system
+            - service
+            - --time=0
+            - unix:///var/run/pink/docker.sock
```
manics, line 76:
```suggestion
            - unix://{{ .Values.pink.hostSocketDir }}/docker.sock
```

sgaist: While it would make it on par with the `dind` implementation I have realized that it only works if `hostSocketDir` points to a folder that exists on both the host and pod filesystems. The default values do.

What made be aware of that issue is that I tested a deployment with the change just it case it would have some effect I did not consider. While implementing the Podman support, I have put the socket and storage in a dedicated folder structure to ensure I would not break the host system unintentionally. This folder is located in `/local`. I know this path is not part of the FHS but I am wondering whether we should take this case into consideration.

helm-chart/binderhub/templates/container-builder/daemonset.yaml
sgaist: I found out what I forgot to write down with regard to the permission issue: I have added an init container to `chown` the folder as well as one doing the same for the socket.

helm-chart/binderhub/templates/container-builder/daemonset.yaml
manics: `/var/lib/pink/` doesn't exist by default and it worked for me.

sgaist: I found the issue in my test !

The mountPath must also be changed to use `{{ .Values.pink.hostSocketDir }}` as it is done for the dind part. Otherwise, the parameter will not point to the correct location.

manics:
@consideRatio Is anything else needed here before merging?

helm-chart/binderhub/values.yaml
```diff
@@ -266,6 +267,22 @@ dind:
   hostSocketDir: /var/run/dind
   hostLibDir: /var/lib/dind
 
+# Podman in Kubernetes
+pink:
+  initContainers: []
+  daemonset:
+    image:
+      name: quay.io/podman/stable
+      tag: v4.2.0
```
consideRatio, line 276: A new image is added as a dependency, can you add automation to automatically bump it? We have a workflow to do that, and it may be straight forward.

See https://github.com/jupyterhub/binderhub/blob/992cac3844a078534f02aea8940b6e8d246ba6be/.github/workflows/watch-dependencies.yaml#L28-L31 as we have it for dind, give it your best shot and I can help fix eventual bugs following it.

consideRatio (commented):
I see that the PR description isn't updated, can you write some final remarks on what the PR does etc based on what we've arrived at @sgaist ?

Thank you soo much for your work on this!!!! :heart: :tada: :sunflower:

tools/templates/lint-and-validate-values-dind.yaml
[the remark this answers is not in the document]
consideRatio: I'm a bit hasty now in my review, but from what I can tell, the current status is excellent @sgaist!! I'll mark this as resolved!

helm-chart/binderhub/schema.yaml
consideRatio: Thanks @sgaist, excellent work!!

consideRatio:
Thanks @manics for the ping, @manics are your unresolved comments ready be resolved, or do they need to be followed up further?

helm-chart/binderhub/templates/image-cleaner.yaml
```diff
@@ -68,22 +68,33 @@ spec:
...
           type: Socket
       {{- end }}
+      {{- if eq $Values.imageBuilderType "pink" }}
+      - name: storage-pink
+        hostPath:
+          path: {{ .Values.pink.hostStorageDir }}
+          type: DirectoryOrCreate
+      - name: socket-pink
+        hostPath:
+          path: {{ .Values.pink.hostSocketDir }}/docker.sock
```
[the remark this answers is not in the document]
sgaist: I have kept `docker.sock` because it is also found in the python code and thought it would keep the coherency. Other than that I don't have strong feelings about the name and we could change it for `podman.sock`.

helm-chart/binderhub/values.yaml
sgaist: There's a bit of an inconsistency between the comment and the actual `jq` query used with regard to the `v` handling. That would cause the use of an older version of the `podman` image as they have one that is tagged without `v`. 

What about using `jq -r '[.Tags[] | match("^v?(?<version>\\d+\\.\\d+\\.\\d+)$").captures[].string ] | sort_by(split(".") | map(tonumber)) | last'` ?

helm-chart/binderhub/templates/image-cleaner.yaml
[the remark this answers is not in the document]
consideRatio: It could be retained, depends on whats involved to change it - but not left undocumented without a comment.

If you add a comment when where this could possibly get someone to think we use a socket for docker when we use actually socket for podman in practice

helm-chart/binderhub/values.yaml
consideRatio: Yes maybe, could need a strip of a v when sorting versions etc to not avoid erroring on mapping to number.

Dont want this to be a blocker for you, i can resolve this later.

helm-chart/binderhub/templates/image-cleaner.yaml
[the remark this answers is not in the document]
manics: `docker.sock` is hard coded in a lot of places. The podman socket is a docker-compatible interface, so it doesn't matter what it's called. In most cases it gets mapped to `/var/run/docker.sock` inside the container anyway.

For this reason you could also make the case for not having separate `.Values.dind.hostSocketDir` `.Values.pink.hostSocketDir` values on the host system, and that they should both use the same path since they'll never be used together.

I think it's fine as it is, but I'm happy to go with whatever you prefer.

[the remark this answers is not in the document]
sgaist: I think we can go with the dedicated socket name without too much trouble.

After rechecking the code, I saw that there is a central place that is responsible for handling the hard coded name of the socket in a deployment: the `binderhub_config.py` from the helm chart.

The original logic was doing it and its replacement as well. Therefore, it should be enough to change it there and update the templates to use `podman.sock` when `pink` is selected.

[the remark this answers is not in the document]
manics: If you change it to `podman.sock` you'll also need to modify the image-cleaner to use the new socket name since only the directory is mapped, not the socket file.

helm-chart/binderhub/values.yaml
sgaist: There are likely better (and cleaner) ways to do that as I am not that proficient with json query. I just saw the issue because there's an inconsistency in the tagging of the Podman stable images. There's both a `v3.4.4` and `3.4.4` that are one month apart and  pointing to different hashes. When testing that I had added the correct information in `watch-dependencies.yaml` I saw that it would select `3.4.4` rather than the latest of the 4 series.

consideRatio: Your example led to 4.3.1 instead of v4.3.1. The following change did the trick though. From the current line with `jq`, I added `v?` to the match pattern, and `ltrimstr("v") |` before calling `tonumber` during the sorting procedure.

```
            | jq -r '[.Tags[] | select(. | match("^v?\\d+\\.\\d+\\.\\d+$") | .string)] | sort_by(split(".") | map(ltrimstr("v") | tonumber)) | last'
```

manics:
When using pink with it's default settings it's also necessary to override the imageCleaner config.
```yml
imageCleaner:
  host:
    dockerSocket: /var/run/pink/podman.sock
    dockerLibDir: /var/lib/pink/storage
```
Since we're now making a breaking change in this PR could we get rid of these properties and use whatever is configured in `dind`/`pink`?

I've also just noticed dind uses `hostLibDir` for it's storage directory whereas pink uses `hostStorageDir`. @consideRatio do you think we should keep this as it represents the underlying container implementation, or use the same property name as an abstraction?

consideRatio:
> @consideRatio do you think we should keep this as it represents the underlying container implementation, or use the same property name as an abstraction?

Hmm hmm I'm feeling a bit out of my depth here and don't feel ready to opine about it. I think your judgement call will be acceptable no matter what though!

> When using pink with it's default settings it's also necessary to override the imageCleaner config.
> 
> ```yaml
> imageCleaner:
>   host:
>     dockerSocket: /var/run/pink/podman.sock
>     dockerLibDir: /var/lib/pink/storage
> ```
>
> Since we're now making a breaking change in this PR could we get rid of these properties and use whatever is configured in dind/pink?

But there was a local situation as well, or? Or is the imageCleaner stuff only activated as part of using dind/pink? I think if imageCleaner is tightly coupled to the use of dind/pink, then relying on configuration under there is the right call.

sgaist:
Funny you bring `imageCleaner` on as I was about to ask some questions for it.

To the best of my comprehension, the default deployment for BinderHub would be using the host Docker and a host configured cleaner.

Then you have the option to use dind and that brings two cases to consider:
- Enabling dind and configure `imageCleaner` host's paths to point to the dind socket and lib folder
- Enabling dind and disable `imageCleaner`'s host configuration

The former works because everything happens on the same node and thus all the paths are valid on the hosts. However there's a side effect. There are two containers created in the `imageCleaner` pod, one for the host and one for dind which, in the end, are checking the same volumes.

In the second case, only the container watching dind is created.

One thing we could do is refactor the cleaner template logic to follow a pattern similar to the builder daemonset and thus ensure that we have only one container doing the watching and cleanup.

One thing for that that we could do is to remove the loop in the containers part of the template.

Maybe rename `local` to `host` in the enumeration defined for `imageBuilderType` ?

consideRatio:
Thanks for thinking about this @sgaist!

When using `local`, is image building done by the binderhub software in the binder pod - using a local docker daemon? Ideally, the configuration value should clarify that directly from its name. If we update this name, I'd appreciate if we make the name verbose enough to help a person understand more from the name directly.

I think `local` means `binder_pod_local_docker_daemon` and that the binder pod is mounted a docker daemon from the host or similar, do you think that is that correct @sgaist?

Ah, but I get your point about using `host` over `local` due to this:

https://github.com/jupyterhub/binderhub/blob/0049297e69dfbc93a6a1079ef85c2b898ebad192/helm-chart/binderhub/templates/image-cleaner.yaml#L38-L41

I think as long as we have imageCleaner's terminology of `host`, it makes sense to speak of `host` instead of `local`. If you want to change that would be fine in my mind.

Confusion points are:
- The docker sockets in image cleaner pods and binder pods are mounted from `hostPath` volumes it seems, "local" and "dind".
- The image cleaner pods are being created by a daemonset, so they will run on every node - not just where the binder pod runs.

Oh I get some clarity now! If a node has been running in k8s for a long time, and pulls more and more images because its starting more and more pods of different kinds - it can run out of space. So, image cleaning on the host is not relevant only for where you are building pods using some docker daemon. It can be directly relevant just by being a node that starts a lot of pods.

At the same time, we have this k8s official description, saying that its not recommended to have external tools to do such things: https://kubernetes.io/docs/concepts/architecture/garbage-collection/#containers-images. Hmmm... I think removal of the imageCleaner host option is something worth considering, but its beyond the scope of this PR to consider I think.

consideRatio:
@sgaist / @manics, I know ~nothing about Podman. Should it be exposed to image cleaning, and if so, can it be exposed to image cleaning by our current system running https://github.com/jupyterhub/docker-image-cleaner?

manics:
For the purposes of building and storing images podman is the same as Docker- if you build an image it's going to stay around until it's deleted. The socket interface presented by podman is the same as docker, that's why I wasn't bothered about renaming the socket to `podman.sock`- as far as the client is concerned (Docker cli, docker-py, repo2docker) it shouldn't matter, except that if the socket isn't mounted to `/var/run/docker.sock` the client needs configuration to use the alternative path.

How about we only use `imageCleaner.host.*` for the local docker host, and if dind/pink are enabled ignore those values and use `dind.{hostSocketDir,hostLibDir}` / `pink.{hostStorageDir,hostSocketDir}` instead, since AFAICT there's no reason to run the cleaner with any other paths?

sgaist:
I think we can simplify some things here indeed.

Do we have an agreement on the following for this patch:
- Change `local` to `host` for the `imageBuilderType` enumeration
- Refactor the image-cleaner daemonset so it only creates the container matching the `imageBuilderType`
- Deprecate the use of `host.enabled` in the same manner as `dind.enabled`

?

consideRatio:
Agreement!
/ Erik on mobile

manics:
Sounds fine to me as well, but given the length of this PR how does everyone feel about merging as is, and doing changes in a follow-up?

consideRatio (approved):
Amazing work @sgaist!!!! Thank you for being so thorough with this!

@manics, this LGTM - agree to go for a merge?

consideRatio:
(Not to be blocking a merge as this can be done after merge)

- @sgaist could you review the PR description a final time so that it doesn't include outdated descriptions of what this PR ended up doing?

manics:
Merging!

manics:
This broke the BinderHub CI upgrade test on main:
https://github.com/jupyterhub/binderhub/blob/fb773a0891d0ed84ad397ec4eeabec7b79e2947b/.github/workflows/test.yml#L186

```
Run . ./ci/common

Installing already released binderhub version 1.0.0-0.dev.git.2920.hc64d689
using old config
Error: INSTALLATION FAILED: execution error at (binderhub/templates/NOTES.txt:194:4): 

#################################################################################
######   BREAKING: The config values passed contained no longer accepted    #####
######             options. See the messages below for more details.        #####
######                                                                      #####
######             To verify your updated config is accepted, you can use   #####
######             the `helm template` command.                             #####
#################################################################################

CHANGED:

dind:
  enabled: true

must as of version 0.3.0 be replace by

imageBuilderType: dind
```
On the bright side we've now tested the upgrade message :smiley:

consideRatio:
Thanks for following up @manics!!

sgaist:
Thanks for the reviews guys !

@consideRatio I have updated the description and I think it's now complete.

@manics can you double check it in case I missed something ?

consideRatio:
@sgaist thanks! Can you provide a leading paragraph, one or two sentence, to introduce what podman is in the PR description as well?

Relevant for the paragraph:
- link to read more about podman
- that it is similar to docker as mentioned by @manics in https://github.com/jupyterhub/binderhub/pull/1531#issuecomment-1351616863

sgaist:
@consideRatio Sure thing !
Done in three sentences but I can rework that if needed.

consideRatio:
Perfect @sgaist thanks!

PR #1543, opened 2022-10-10
Change component name from dind to image-builder

sgaist:
In parallel of the work from #1531, this change will allow the use of other builders than Docker without being tied to its nomenclature which might make the whole a bit confusing when deploying and debugging.

Part of #1513

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -296,14 +296,16 @@ class KubernetesBuildExecutor(BuildExecutor):
         """
         resp = self.api.list_namespaced_pod(
             self.namespace,
-            label_selector="component=dind,app=binder",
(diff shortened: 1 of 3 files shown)


consideRatio:
I'm not sure, about tooling relying on identifying the pods by the `component: dind` label. Maybe grafana dashboards etc or some maintenance scripts in mybinder.org-deploy?

I [looked](https://github.com/jupyterhub/mybinder.org-deploy/search?p=4&q=component) but found nothing. Seems reasonable to me!

minrk:
Yeah, if anything I would guess it's only something in grafana or possibly a network policy.

manics:
Thanks all, merging!

PR #1544, opened 2022-10-10
Update to z2jh 2 and JupyterHub 3

consideRatio:
This better be done at some point in time, and one reason to do it is because it makes us compatible with k8s 1.25 and higher among other things.

- Closes #1541 
- Enables resolving #1444 in a way like z2jh has done it

---

Since this may be relevant knowing about I pinged a lot of you who I guessed could be influenced.

## Draft status of PR because of ...

```
[D 2022-10-10 14:15:14.238 JupyterHub app:2035] Loading roles into database
[I 2022-10-10 14:15:14.249 JupyterHub roles:173] Role jupyterhub-idle-culler added to database
[I 2022-10-10 14:15:14.251 JupyterHub app:1934] Not using allowed_users. Any authenticated user will be allowed.
[W 2022-10-10 14:15:14.264 JupyterHub app:2312] Service binder sets `admin: True`, which is deprecated in JupyterHub 2.0. You can assign now assign roles via `JupyterHub.load_roles` configuration. If you specify services in the admin role configuration, the Service admin flag will be ignored.
[I 2022-10-10 14:15:14.265 JupyterHub roles:238] Adding role admin for Service: binder
[E 2022-10-10 14:15:14.268 JupyterHub app:3297]
    Traceback (most recent call last):
      File "/usr/local/lib/python3.9/site-packages/jupyterhub/app.py", line 3294, in launch_instance_async
        await self.initialize(argv)
      File "/usr/local/lib/python3.9/site-packages/jupyterhub/app.py", line 2824, in initialize
        self.init_services()
      File "/usr/local/lib/python3.9/site-packages/jupyterhub/app.py", line 2338, in init_services
        setattr(service, key, value)
      File "/usr/local/lib/python3.9/site-packages/traitlets/traitlets.py", line 715, in __set__
        self.set(obj, value)
      File "/usr/local/lib/python3.9/site-packages/traitlets/traitlets.py", line 689, in set
        new_value = self._validate(obj, value)
      File "/usr/local/lib/python3.9/site-packages/traitlets/traitlets.py", line 723, in _validate
        value = self._cross_validate(obj, value)
      File "/usr/local/lib/python3.9/site-packages/traitlets/traitlets.py", line 729, in _cross_validate
        value = obj._trait_validators[self.name](obj, proposal)
      File "/usr/local/lib/python3.9/site-packages/traitlets/traitlets.py", line 1132, in __call__
        return self.func(*args, **kwargs)
[...]

Issue #1541, opened 2022-10-10
CI system broken with k8s 1.25 (at least k3s 1.25)


consideRatio:
Without changes to this repo, our CI system has started failing, where a created build pod get stuck in pending state.

I'm quite confident it is because we started using k8s 1.25. Note though that the jupyterhub/action-k3s-helm action is providing us with some bridging tech called [cri-dockerd](https://github.com/Mirantis/cri-dockerd), so I thought maybe we wouldn't end up with issues related to that.

Oldest test failure: https://github.com/jupyterhub/binderhub/actions/runs/3210481910/jobs/5247954722#step:28:106
Most recent test success: https://github.com/jupyterhub/binderhub/actions/runs/2987255039/jobs/4789548250

consideRatio:
I conclude its z2jh as a dependency, with the user scheduler, that doesn't work with k8s 1.25. So the fix is to upgrade to z2jh 2.0.0

PR #1549, opened 2022-10-12
ci: extend test timeout

minrk:
tests are getting [cancelled](https://github.com/jupyterhub/binderhub/actions/runs/3214936169/jobs/5255678267) because 10 minutes is not enough.

time limits added in #1518

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
   # jobs or have "template" jobs, so we use `if` conditions on steps
   tests:
     runs-on: ubuntu-20.04
-    timeout-minutes: 10
+    timeout-minutes: 15
     strategy:
       # keep running so we can see if tests with other k3s/k8s/helm versions pass
       fail-fast: false


betatim:
Tests got cancelled after 15min :-/

From the logs of the helm tests:
```
============================= test session starts ==============================
platform linux -- Python 3.9.14, pytest-7.1.3, pluggy-1.0.0 -- /opt/hostedtoolcache/Python/3.9.14/x64/bin/python
cachedir: .pytest_cache
rootdir: /home/runner/work/binderhub/binderhub
plugins: cov-4.0.0, asyncio-0.19.0
asyncio: mode=strict
collecting ... collected 129 items / 118 deselected / 11 selected
```
and then it got cancelled. This seems weird no? I was expecting it to run some test or do something, not spend 12minutes finding tests to run.

consideRatio:
No, they are stuck.

User pods are stuck pending, because user scheduler isnt supporting k8s 1.25.

See #1544

betatim:
Ah ok. I had expected to see some output about a particular test having started (and then getting stuck).

Looking at the PR you linked, I think we don't need to extend the timeout for the tests. Instead we should migrate BinderHub to be compatible with JupyterHub 2.

manics:
> Ah ok. I had expected to see some output about a particular test having started (and then getting stuck).

Yes, that's why I originally added the timeout. I noticed that coding errors during development could lead to tests hanging due to an unexpected server response.... which meant they got stuck for the default timeout of 6(?) hours.

Is there a way to set a default timeout for each pytest test (e.g. maybe 1 or 2 minutes?), since that would at least give us a bit more information on how extensive the failures are.

As a short term fix, will pinning the CI tests to K8s 1.24 solve the failures?

minrk:
> As a short term fix, will pinning the CI tests to K8s 1.24 solve the failures?

Yes, I think that's the right change. k8s version should be pinned and upgraded explicitly in CI anyway.

minrk:
> Is there a way to set a default timeout for each pytest test (e.g. maybe 1 or 2 minutes?)

Yeah, [pytest-timeout](https://github.com/pytest-dev/pytest-timeout) should work.

minrk:
#1550 pins k3s to 1.24 to get tests working again (sorry @consideRatio for not seeing #1541 first). #1551 adds per-test timeouts so we shouldn't hang the whole test suite anymore when similar bugs crop up.

PR #1550, opened 2022-10-12
ci: test k3s 1.24

minrk:
gets tests passing, pending #1541.

Should probably stick with explicit versioning? I'm not sure - it's useful to have 'latest' to catch these issues, too.


manics (commented):
👍

PR #1551, opened 2022-10-12
add timeout to tests

minrk:
using pytest-timeout

default: 60s

can be overridden per test with `@pytest.mark.timeout`


minrk:
Because this doesn't fix the hang (#1550), we should see individual tests fail with timeouts instead of the whole thing hanging. If there are enough tests that use the full timeout, it'll still exhaust the limit and get cancelled.

PR #1555, opened 2022-10-13
helm chart: use modern chartpress and baseVersion 1.0.0-0.dev

consideRatio:


helm-chart/chartpress.yaml
```diff
@@ -5,12 +5,11 @@
 #       helm-chart folder.
 charts:
   - name: binderhub
+    baseVersion: 1.0.0-0.dev
```
manics, line 8: Does this mean the dev releases will come after 1.0.0 (which is confusing since we haven't had a major release yet), or before?

consideRatio: No if we release 1.0.0, it will be more recent just like 1.0.0-beta.1 is before 1.0.0 the 1.0.0-0.dev is before 1.0.0 (and 1.0.0-beta.1).

manics: Great!

manics (approved):

PR #1565, opened 2022-10-21
docs: add devenv and linkcheck to Makefile and misc details

consideRatio:
## Related
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2909
- https://github.com/jupyterhub/repo2docker/pull/1197


minrk:
Merging, since it's good to have these. But linkcheck's not run on CI, is it? Should that be a new workflow? Or is there a way to run linkcheck on RTD _only on PRs_ (because real docs builds shouldn't be prevented if a link breaks)?

consideRatio:
I think having linkcheck in the ci system for push and prs is too much noise overall, having a weekly/monthly run via schedule and a badge for failing or not in the project readme could make sense i think.

PR #1571, opened 2022-11-01
keep old default branch links working

### LINKED ISSUE #1570: Long-term fix for deleted default branches
### Proposed change

Redirect builds where  `unresolved_ref=master` and that ref doesn't resolve to the same spec with `unresolved_ref=HEAD`.

As discussed in #1145 (cc @bryevdv), there are likely a lot of old links that can't easily be updated after a default branch rename. We've fixed this for _new_ default URLs in #1186, but not in a way that keeps old links working.

### Alternative options

Let old links break.

Anyone with old Binder links in documentation, across the web who has renamed their default branch in the past few years, prior to the adoption of 'HEAD' as the default ref. We know folks are keeping an unused 'master' branch up-to-date just to avoid breaking links. This would allow them to stop doing that.

Two possible solutions:

1. (better, IMO, but more complex): I think the 'right' behavior would be for master builds to redirect the browser to HEAD, rather that keeping the resolution of dead links behind the scenes. That's a little tricky, since the resolution occurs in the event stream, not in the regular page handler, and event streams can't redirect in the middle. So we'd actually need to do the redirect in javascript on the page in response to a new event type.

2. Retrying with unresolved_ref=HEAD within the event stream handler would be a lot simpler to implement, but would keep broken links canonically working instead of redirecting. But since it's not a 'true' redirect (page load succeeds before the redirect is discovered), there's little difference, I think.

minrk:
If 'master' or 'main' fail to resolve, try again with 'HEAD' after a delay.

this keeps outdated links to renamed default branches working for the big master->main migration

closes #1570

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -2,6 +2,7 @@
 Handlers for working with version control services (i.e. GitHub) for builds.
 """
 
+import asyncio
 import hashlib
 import json
 import re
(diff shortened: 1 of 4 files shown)


minrk:
This is the simpler proposal in #1570, without the browser redirect, which I think has more moving parts (need to compute new URL somewhere, handle communicating updated spec from server to browser, etc.)

manics (commented):
Is the idea to only do this for GitHub but not GitLab because GitHub is where problems are most likely to occur?

minrk:
> Is the idea to only do this for GitHub but not GitLab because GitHub is where problems are most likely to occur?

That's what we've been doing, but we can probably expand it to other providers, too, if we want.

manics (approved):
Cool, just wanted to check it's not an oversight.

PR #1586, opened 2022-12-14
Add compatibility for fetching tokens from other Docker registries

manics:
The current `DockerRegistry` class only works with docker.io and gcr.io, as the token endpoints are hardcoded:
https://github.com/jupyterhub/binderhub/blob/fb773a0891d0ed84ad397ec4eeabec7b79e2947b/binderhub/registry.py#L126-L136

Support for other registries can be added be parsing a `WWW-Authentication: Bearer` header if there's one to obtain the token endpoint, e.g.
`WWW-Authenticate: Bearer realm="https://uk-london-1.ocir.io/12345678/docker/token",service="uk-london-1.ocir.io",scope=""`

This has been tested with [Oracle Cloud Infrastructure Container Registry](https://docs.oracle.com/en-us/iaas/Content/Registry/Concepts/registryoverview.htm)

### DIFF
--- a/binderhub/registry.py
+++ b/binderhub/registry.py
@@ -4,6 +4,7 @@ Interaction with the Docker Registry
 import base64
 import json
 import os
+import re
 from urllib.parse import urlparse
 
 from tornado import httpclient
@@ -131,8 +132,7 @@ class DockerRegistry(LoggingConfigurable):
         elif self.url.endswith(".docker.io"):
             return "https://auth.docker.io/token?service=registry.docker.io"
         else:
-            # is gcr.io's token url common? If so, it might be worth defaulting
-            # to https://registry.host/v2/token?service=registry.host
+            # If necessary we'll look for the WWW-Authenticate header
             return ""
 
     username = Unicode(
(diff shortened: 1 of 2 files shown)


binderhub/registry.py
```diff
@@ -185,30 +185,87 @@ def _default_password(self):
...
+    def _parse_www_authenticate_header(self, header):
+        # Header takes the form
+        # WWW-Authenticate: Bearer realm="https://uk-london-1.ocir.io/12345678/docker/token",service="uk-london-1.ocir.io",scope=""
+        self.log.debug(f"Parsing WWW-Authenticate {header}")
+
+        if not header.lower().startswith("bearer "):
+            raise ValueError(f"Only WWW-Authenticate Bearer type supported: {header}")
+        try:
+            realm = re.search(r'realm="([^"]+)"', header).group(1)
+            # Should service and scope parameters be optional instead of just empty?
```
manics, line 197: I've made `scope` and `service` required parameters (though they can be `""`), in future we can make them optional if other registries don't return them.

binderhub/registry.py
```diff
@@ -185,30 +185,87 @@ def _default_password(self):
...
+            service = re.search(r'service="([^"]*)"', header).group(1)
+            scope = re.search(r'scope="([^"]*)"', header).group(1)
+            return realm, service, scope
+        except AttributeError:
+            raise ValueError(
+                f"Expected WWW-Authenticate to include realm service scope: {header}"
+            ) from None
+
+    async def _get_token(self, client, token_url, service, scope):
```
manics, line 206: This new function is the same token handling code that used to be in `get_image_manifest()`, but moved into a standalone helper method.

consideRatio (approved):
Nice, LGTM!

Issue #1520, opened 2022-08-02
Should k8s-binder-k8s-hub/binderhub-chart-config-old.yaml be kept or deleted?


manics:
`k8s-binder-k8s-hub/binderhub-chart-config-old.yaml` says it should be deleted:
https://github.com/jupyterhub/binderhub/blob/7fa1d77f5d82ec60e05e2ef4ecd6e92950d2b87e/testing/k8s-binder-k8s-hub/binderhub-chart-config-old.yaml#L1-L2
However it's referenced in the CI workflows:
https://github.com/jupyterhub/binderhub/blob/7fa1d77f5d82ec60e05e2ef4ecd6e92950d2b87e/.github/workflows/diff.yml#L54-L59
https://github.com/jupyterhub/binderhub/blob/7fa1d77f5d82ec60e05e2ef4ecd6e92950d2b87e/.github/workflows/test.yml#L185-L190

Should it be kept and the comment removed, or should the file and all references to it be deleted?

consideRatio:
- #1472 is merged at this point. I presume we should remove all references to it. I recognize the need for having two sets of helm chart configuration because the CI system tests against an old version and then upgrades to a new version etc. I presume this is about that, and that we can now remove everything related to it.

PR #1587, opened 2022-12-14
binderhub-chart-config-old.yaml `imageBuilderType: "local"`

manics:
This should fix the CI failure resulting from
https://github.com/jupyterhub/binderhub/pull/1531#issuecomment-1352352454


manics:
Alternatively we could delete the file https://github.com/jupyterhub/binderhub/issues/1520 though this PR is a quick fix.

consideRatio:
#1520 seems fine as well, but I went for a merge so that choice becomes an optional followup and not required to fix the CI system

PR #1597, opened 2022-12-27
Fix the default `use_registry` value

manics:
`use_registry` defaults to try in BinderHub
https://github.com/jupyterhub/binderhub/blob/03545a1587f2d781ff4934beac8b8b8a19280754/binderhub/app.py#L261-L270

but the Helm chart implicitly defaults to `False`:
https://github.com/jupyterhub/binderhub/blob/03545a1587f2d781ff4934beac8b8b8a19280754/helm-chart/binderhub/templates/deployment.yaml#L68-L76

Which will lead to a broken deployment.

Also includes a commit to fix the CI upgrade/diff check.

### DIFF
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -40,7 +40,9 @@ service:
   loadBalancerIP:
 
 config:
-  BinderHub: {}
+  BinderHub:
+    # This must equal the default value of c.BinderHub.use_registry
+    use_registry: true
 
 extraConfig: {}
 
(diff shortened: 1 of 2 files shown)


consideRatio:
Ah, so `c.BinderHub.use_registry` is one of those software configurations that is required to be known also during chart template rendering. In z2jh `c.JupyterHub.base_url` is a configuration like that as well.

A big plus if we can make that explicit in comments going onwards as that is a very relevant piece of understanding for refactoring work in general.

PR #1598, opened 2022-12-28
Add `health_handler_class` to make `/health` handler is customisable

manics:
The original `/health` handler assumes everything is on Kubernetes, now it's configurable. The default behaviour is unchanged except for an additional field `"_ignored": true`:
```json
{
  "ok": true,
  "checks": [
    {
      "service": "Docker registry",
      "ok": true
    },
    {
      "service": "JupyterHub API",
      "ok": true
    },
    {
      "service": "Pod quota",
      "total_pods": 0,
      "build_pods": 0,
      "user_pods": 0,
      "quota": null,
      "ok": true,
      "_ignored": true
    }
  ]
}
```
I could filter it out, but I think it's worth keeping since it makes it clear that the status of the check is always ignored:
https://github.com/jupyterhub/binderhub/blob/a6f2701196427d11e44415f62dce26870e6cf511/binderhub/health.py#L201-L205

### DIFF
--- a/binderhub/health.py
+++ b/binderhub/health.py
@@ -159,7 +159,7 @@ class HealthHandler(BaseHandler):
                 results.append(dict({"service": service}, **result))
 
         # Some checks are for information but do not count as a health failure
-        overall = all(r["ok"] for r in results if not r.get("_ignored", False))
+        overall = all(r["ok"] for r in results if not r.get("_ignore_failure", False))
         if not overall:
             unhealthy = [r for r in results if not r["ok"]]
             app_log.warning(f"Unhealthy services: {unhealthy}")
@@ -229,6 +229,6 @@ class KubernetesHealthHandler(HealthHandler):
             "ok": total_pods <= quota if quota is not None else True,
             # The pod quota is treated as a soft quota
             # Being above quota doesn't mean the service is unhealthy
-            "_ignored": True,
+            "_ignore_failure": True,
         }
         return usage
--- a/binderhub/health.py
+++ b/binderhub/health.py
@@ -105,6 +105,81 @@ class HealthHandler(BaseHandler):
     def initialize(self, hub_url=None):
         self.hub_url = hub_url
 
+    @false_if_raises
+    @retry
+    async def check_jupyterhub_api(self, hub_url):
+        """Check JupyterHub API health"""
+        await AsyncHTTPClient().fetch(hub_url + "hub/api/health", request_timeout=3)
+        return True
+
+    @false_if_raises
+    @at_most_every(interval=15)
+    @retry
+    async def check_docker_registry(self):
+        """Check docker registry health"""
+        app_log.info("Checking registry status")
+        registry = self.settings["registry"]
+        # we are only interested in getting a response from the registry, we
+        # don't care if the image actually exists or not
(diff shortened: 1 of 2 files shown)


binderhub/health.py
```diff
@@ -155,58 +127,43 @@ async def check_docker_registry(self):
...
-        # quota doesn't mean the service is unhealthy
-        overall = all(
-            check["ok"] for check in checks if check["service"] != "Pod quota"
-        )
+        # Some checks are for information but do not count as a health failure
+        overall = all(r["ok"] for r in results if not r.get("_ignored", False))
         if not overall:
-            unhealthy = [check for check in checks if not check["ok"]]
+            unhealthy = [r for r in results if not r["ok"]]
             app_log.warning(f"Unhealthy services: {unhealthy}")
```
consideRatio, line 165: Here we will warn about the `_ignored` services as well and not hint that they are accepted to fail. But hmmm, we have the `_ignored` field as part of the `r` dictionary, so maybe its fine as the reader could deduce that.

I think about the naming `_ignore` though, maybe we can make it more descriptive? Its a "check" in a "health check" context, and we don't consider the health to be bad / not ok if this check is failing. What about `_accept_failure` or something like that? We can go with `_ignore` as well if this doesn't resonate with you!

consideRatio:
```python
# Some checks are for information but do not count as a health failure
```

Maybe even `_count_as_health_failure`?

consideRatio (approved):
This LGTM and I just commented about the name of the `_ignore` field that I consider if we should name more verbosively, such as `_count_as_health_failure`. We can go with `_ignore` as well but I've grown appreciative of the more descriptive names.

binderhub/health.py
```diff
@@ -218,3 +175,60 @@ async def head(self):
...
+        total_pods = n_user_pods + n_build_pods
+        usage = {
+            "total_pods": total_pods,
+            "build_pods": n_build_pods,
+            "user_pods": n_user_pods,
+            "quota": quota,
+            "ok": total_pods <= quota if quota is not None else True,
+            # The pod quota is treated as a soft quota
+            # Being above quota doesn't mean the service is unhealthy
+            "_ignored": True,
```
manics, line 232:
```suggestion
            "_ignore_failure": True,
```

binderhub/health.py
```diff
@@ -155,58 +127,43 @@ async def check_docker_registry(self):
...
-                check.update(result)
+                results.append(dict({"service": service}, **result))
 
-        # The pod quota is treated as a soft quota this means being above
-        # quota doesn't mean the service is unhealthy
-        overall = all(
-            check["ok"] for check in checks if check["service"] != "Pod quota"
-        )
+        # Some checks are for information but do not count as a health failure
+        overall = all(r["ok"] for r in results if not r.get("_ignored", False))
```
manics, line 162:
```suggestion
        overall = all(r["ok"] for r in results if not r.get("_ignore_failure", False))
```

manics (commented):
How about `_ignore_failure`? It's a bit shorter and describes the issue.

Ideally I think we shouldn't return `"ok": false` in the first place, since it's misleading and it clearly is OK!

consideRatio:
Nice `_ignore_failure` is great!

PR #1602, opened 2022-12-29
Test dind and pink

manics:
We currently test 3 Helm chart configurations. This means we can test all three `imageBuilderType` without increasing the number of matrix items.

Currently we only test `host`, but since docker is no longer part of Kubernetes most production deployments will use `dind` (or similar) so I think we should test it as standard even though it means deploying a registry.

When building images with chartpress we can use `localhost` since it's running outside K8s, but we need to use the hosts's external IP (`$REGISTRY_HOST`) when referring to images from inside K8s.

(extracted from https://github.com/jupyterhub/binderhub/pull/1521)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,8 @@ jobs:
           - main
           - auth
           - helm
+        test-variation:
+          - ""
         include:
           # Chart.yaml contains the chart's oldest supported k8s version, we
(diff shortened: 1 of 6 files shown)


consideRatio:
Wow it seems like a lot of effort went into this given the complexity of it. Thank you for your work @manics!!!!

manics:
Thanks for reviewing!

PR #1611, opened 2023-01-04
breaking: test and use Python 3.11 in Helm chart image, up from Python 3.9

consideRatio:
- ci: update from ubuntu-20.04 to ubuntu-22.04 in workflows
- breaking: test and use Python 3.11, up from Python 3.9
- Refreeze requirements.txt with Python 3.11


consideRatio:
Rebased

consideRatio:
Argh it seems like we still get github api rate limited sometimes.

manics (approved):

PR #1612, opened 2023-01-04
ci: set GITHUB_ACCESS_TOKEN to temporary read only token

### LINKED ISSUE #1608: CI tests frequently hit a GitHub API rate limit
### Bug description

`binderhub/tests/test_build.py` often fails with the error
```
 >       assert launch_events > 0
E       assert 0 > 0

binderhub/tests/test_build.py:87: AssertionError
----------------------------- Captured stdout call -----------------------------
failed: Error resolving ref for gh:binderhub-ci-repos/cached-minimal-dockerfile/HEAD: GitHub rate limit exceeded. Try again in 15 minutes.
_ test_build[gh/binderhub-ci-repos/cached-minimal-dockerfile/596b52f10efb0c9befc0c4ae850cc5175297d71c] _
```

I think this is related to how often PRs are opned and updated, the required timeout can easily reach 60 minutes

#### Expected behaviour

GitHub API limits are only occasionally reached for PR workflows

#### Actual behaviour

Limits are often hit, causing CI tests to incorrectly fail.

### How to reproduce

I think opening a PR that runs all tests, and updating if several times, is enough.

I think we should go through the tests and minimise the number of live GH API calls. Note we shouldn't use a GitHub API token since these are PRs.

consideRatio:
- Closes #1608.

We ran into rate limitations, but by setting the env var `GITHUB_ACCESS_TOKEN` in our GitHub Job's, which is the default value for `c.GitHubRepoProvider.access_token`, this is no longer an issue. 

We still have to mock tests with GistRepoProvider as the `secrets.github_token` is not provided rights to access api.github.com/gists.

### Remaining test failure is resolved by #1615


manics:
I started looking last night, saw a massive JSON blob.... and decided to leave it 😄. Everything else looked fine though, so happy to go with @betatim

consideRatio:
Ah sorry about the JSON blob. I wrote about it but then felt my PR description ended up so long that it was boring to read through and removed my notes about it ^^

The JSON changes is two keys representing api.github.com/gists URLs that are to be mocked that I relocated from one file to a standalone file. In other words, its just one file split in two.

Thanks for reviewing this, I'll go for a merge!

consideRatio:
I spotted a typo in this PR so the issue wasn't resolved. I added a commit fixing that typo in #1615 now though.

PR #1615, opened 2023-01-04
ci: fix changed resolved ref for dataverse test, fix recent github ratelimit PR

### LINKED ISSUE #1614: Dataverse test failure in main branch
Is this failure just tests being outdated, or indicative of a malfunction in the dataverse repoprovider?

Recent [main branch test failure](https://github.com/jupyterhub/binderhub/actions/runs/3837733452/jobs/6533372219#step:29:279):

image

consideRatio:
There was a new version of a DOI that was tested against, so "resolved_ref" transitioned from v3.0 to v4.0 it seems. The last update to this DOI was several years ago. I've made it into a regexp now though so if new releases happen, it should not break this again.

- Closes #1614

- Complements #1612 with a typo fix, and with that, our test suite should become reliable I think

### DIFF
--- a/binderhub/tests/test_repoproviders.py
+++ b/binderhub/tests/test_repoproviders.py
@@ -171,7 +171,7 @@ async def test_hydroshare_doi():
         [
             "10.7910/DVN/TJCLKP",
             "10.7910/DVN/TJCLKP",
-            "3035124.v3.0",
+            "3035124.v4.0",
             "https://doi.org/10.7910/DVN/TJCLKP",
             "dataverse-dvn-2ftjclkp",
         ],
--- a/binderhub/tests/test_repoproviders.py
+++ b/binderhub/tests/test_repoproviders.py
@@ -170,21 +170,21 @@ async def test_hydroshare_doi():
     # "10.7910/DVN/TJCLKP" is a DOI associated with all versions of the
     # dataverse dataset, including the latest version and previous versions (v3,
     # v2, etc). Dataverse doesn't mint DOIs for each version of a dataset, but
-    # that has been discussed in https://github.com/IQSS/dataverse/issues/4499,
+    # that has been discussed in https://github.com/IQSS/dataverse/issues/4499
     #
     "spec,resolved_spec,resolved_ref,resolved_ref_url,build_slug",
     [
         [
             "10.7910/DVN/TJCLKP",
             "10.7910/DVN/TJCLKP",
-            r"3035124.v\d+.\d+",
+            r"3035124\.v\d+\.\d+$",
             "https://doi.org/10.7910/DVN/TJCLKP",
             "dataverse-dvn-2ftjclkp",
         ],
         [
             "10.25346/S6/DE95RT",
             "10.25346/S6/DE95RT",
-            r"20460.v\d+.\d+",
+            r"20460\.v\d+\.\d+$",
             "https://doi.org/10.25346/S6/DE95RT",
             "dataverse-s6-2fde95rt",
         ],
--- a/binderhub/tests/test_repoproviders.py
+++ b/binderhub/tests/test_repoproviders.py
@@ -1,3 +1,4 @@
+import re
 from unittest import TestCase
 from urllib.parse import quote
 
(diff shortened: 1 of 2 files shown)


binderhub/tests/test_repoproviders.py
```diff
@@ -171,7 +171,7 @@ async def test_hydroshare_doi():
         [
             "10.7910/DVN/TJCLKP",
             "10.7910/DVN/TJCLKP",
-            "3035124.v3.0",
+            "3035124.v4.0",
```
manics, line 174: @Xarthisius @pdurbin You were involved in the PR that added support for Dataverse and this test in
https://github.com/jupyterhub/binderhub/pull/969

Is this change in the resolved reference expected? Is there a way to make this constant so it doesn't change again in future?

consideRatio: Is "10.7910/DVN/TJCLKP" associated with v3 or the latest version? Now we get the latest version from that as a reference given to binderhub, when we may or may not have expected we would get v3 even if there was a newer version.

pdurbin: @consideRatio I'm not sure I'd put it quite that way.

"10.7910/DVN/TJCLKP" is associated with all versions of the dataset, including the latest version and previous versions (v3, v2, etc). Dataverse doesn't mint DOIs for each version of a dataset (though this functionality has certainly been [requested](https://github.com/IQSS/dataverse/issues/4499) for quite a while).

I still think `:latest-published` (above) might help but there's also a Dataverse API to list each version of a dataset, if that's of interest: https://guides.dataverse.org/en/5.12.1/api/native-api.html#list-versions-of-a-dataset

Sorry if I'm not quite following. If you're getting the latest version from "10.7910/DVN/TJCLKP" I'm not sure why v3 is hard coded (and updated to v4 in this PR).

manics: https://github.com/IQSS/dataverse/issues/4499 explains it, I assumed Dataverse supported exact DOIs resolving to a particular version for reproducibility 😞 .

I can think of two options here (I'm happy with whatever you prefer):
- Change this to `v4` as in the PR
- Change this (and the test case below) to a regex match for `v\d+`

Either way, adding a comment to this test case explaining all this would be helpful for future readers!

consideRatio: Regex added and I've included a comment based on your comment here @pdurbin - thank you!!

consideRatio:
@manics I squeezed in a refactor of a function that could lead to a silent error and a complementary fix for #1612.

This is done now I think!

pdurbin (approved):
Look great! So much better than hard coding a version. Ship it! :shipit: 🚀

binderhub/tests/test_repoproviders.py
```diff
@@ -166,19 +167,24 @@ async def test_hydroshare_doi():
...
+    # v2, etc). Dataverse doesn't mint DOIs for each version of a dataset, but
+    # that has been discussed in https://github.com/IQSS/dataverse/issues/4499,
+    #
     "spec,resolved_spec,resolved_ref,resolved_ref_url,build_slug",
     [
         [
             "10.7910/DVN/TJCLKP",
             "10.7910/DVN/TJCLKP",
-            "3035124.v3.0",
+            r"3035124.v\d+.\d+",
```
manics, line 180:
```suggestion
            r"3035124\.v\d+\.\d+$",
```
`.` is a wildcard, `re.match` does a prefix match

binderhub/tests/test_repoproviders.py
```diff
@@ -166,19 +167,24 @@ async def test_hydroshare_doi():
...
             "https://doi.org/10.7910/DVN/TJCLKP",
             "dataverse-dvn-2ftjclkp",
         ],
         [
             "10.25346/S6/DE95RT",
             "10.25346/S6/DE95RT",
-            "20460.v1.0",
+            r"20460.v\d+.\d+",
```
manics, line 187:
```suggestion
            r"20460\.v\d+\.\d+$",
```

binderhub/tests/test_repoproviders.py
```diff
@@ -166,19 +167,24 @@ async def test_hydroshare_doi():
 
 
 @pytest.mark.parametrize(
+    # "10.7910/DVN/TJCLKP" is a DOI associated with all versions of the
+    # dataverse dataset, including the latest version and previous versions (v3,
+    # v2, etc). Dataverse doesn't mint DOIs for each version of a dataset, but
+    # that has been discussed in https://github.com/IQSS/dataverse/issues/4499,
```
manics, line 173:
```suggestion
    # that has been discussed in https://github.com/IQSS/dataverse/issues/4499
```

manics (approved):

Issue #1620, opened 2023-01-11
ACR support damaged by Gitlab registry change


wallyhall:
### Bug description

I believe ACR support has been damaged by the Gitlab registry change in this commit:

https://github.com/jupyterhub/binderhub/commit/f442f326320eaac68e08b663cac17055b615731a

#### Expected behaviour

When following a Binderhub link, if a pre-built/cached image already exists in the ACR, I expect Binderhub to pull it, not rebuild from scratch.

#### Actual behaviour

Binderhub runs `repo2docker` and builds a new image, pushing it to the ACR.

The logs contain the following errors:

```
[I 230111 09:52:26 launcher:257] Starting server for user ****** with image ******.azurecr.io/******/******
[I 230111 09:52:29 builder:744] Launched ****** in 3s
[I 230111 09:53:02 log:135] 200 GET /v2/git/******/2cd69cb92c9fa36877786c4672158f343e2d1a79?urlpath=******.ipynb (anonymous@******) 2.64ms
[E 230111 09:53:02 builder:394] Tornado HTTP Timeout error: Failed to get image manifest for ******.azurecr.io/******/******:2cd69cb92c9fa36877786c4672158f343e2d1a79
    Traceback (most recent call last):
      File "/usr/local/lib/python3.9/site-packages/binderhub/builder.py", line 388, in get
        image_manifest = await self.registry.get_image_manifest(
      File "/usr/local/lib/python3.9/site-packages/binderhub/registry.py", line 263, in get_image_manifest
        token = await self._get_token(
      File "/usr/local/lib/python3.9/site-packages/binderhub/registry.py", line 221, in _get_token
        auth_resp = await client.fetch(auth_req)
    tornado.httpclient.HTTPClientError: HTTP 400: Bad Request
```

I believe this is due to the `token_url` configured (shown below) being partially overwritten - specifically the `?service=` by the linked git commit above...

```
config:
  DockerRegistry:
    token_url: https://********.azurecr.io/oauth2/token?service=********.azurecr.io
```

(I suspect it is becoming `https://********.azurecr.io/oauth2/token?service=container_registry`.)

### How to reproduce

Follow the instructions in Binderhub's installation docs for an ACR, and attempt to launch the same repository multiple times.

Observe the logs (and behaviour).

### Your personal set up

Have installed via Helm:

```
config:
  BinderHub:
    hub_url: https://hub.binder.*********
    hub_url_local: http://10.100.100.122
[...]

PR #1628, opened 2023-01-23
Proposed fix for ACR caching while providing guidance for GitLab users.

wallyhall:
Proposed fix for https://github.com/jupyterhub/binderhub/issues/1620

This supersedes a single line change of PR https://github.com/jupyterhub/binderhub/pull/1283 ... and provides documentation (like already exists for ACR) to maintain GitLab compatibility (but the documented instructions have not been validated).

### DIFF
--- a/binderhub/registry.py
+++ b/binderhub/registry.py
@@ -264,7 +264,6 @@ class DockerRegistry(LoggingConfigurable):
                 client,
                 self.token_url,
                 scope=f"repository:{image}:pull",
-                service="container_registry",
             )
             req = httpclient.HTTPRequest(
                 url,
--- a/docs/source/zero-to-binderhub/setup-binderhub.rst
+++ b/docs/source/zero-to-binderhub/setup-binderhub.rst
@@ -174,6 +174,29 @@ where:
   If this is not provided, you may find BinderHub rebuilds images every launch instead of pulling them from the ACR.
   Suggestions for `<project-name>` could be `ACR_NAME` or the name you give your BinderHub.
 
+If you are using GitLab Container Registry 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you want your BinderHub to push and pull images from a GitLab Container Registry, then your `config.yaml` file will look as follows::
+
+    config:
+      BinderHub:
+        use_registry: true
+        image_prefix: registry.gitlab.com/<gitlab-user>/<project-name>/<prefix>-
+      DockerRegistry:
+        token_url: "https://gitlab.com/jwt/auth?service=container_registry"
+        url: "https://registry.gitlab.com"
+    registry:
+      url: https://registry.gitlab.com
+
+where:
+
+* `<gitlab-user>` is your gitlab username,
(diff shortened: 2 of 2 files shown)


wallyhall:
Has anyone advice for testing this PR?  The updated tests are largely all integration ... and I'm not entirely clear on the consequences of what I've changed with regards to other container registries (and I think that probably is why the original PR this seeks to "patch" probably wasn't caught for the ACR scenario I'm trying to repair here).

I've added an attempt at a regression test too, but I'm no Python wizard and I've been unable to get the tests running locally (I've relied on CI).

leviplj (approved):

leviplj:
Can any maintainer review/merge this? We've stuck with the same problem using ACR.

Thanks!

binderhub/tests/test_registry.py
```diff
@@ -169,7 +169,15 @@ async def test_get_token():
...
-            (r"/token", MockTokenHandler, {"test_handle": test_handle}),
+            (
+                r"/token",
+                MockTokenHandler,
+                {
+                    "test_handle": test_handle,
+                    "service": "service.1",
+                    "scope": "scope.2",
+                },
+            ),
```
manics, line 180: Is this change needed?

binderhub/tests/test_registry.py
```diff
@@ -196,6 +204,49 @@ async def test_get_token():
     assert token == test_handle["token"]
 
 
+async def test_get_token_with_config_supplied_service():
```
manics, line 207: It looks like this is a copy of `test_get_token` apart from the arguments to `registry._get_token()`. Can you use `@pytest.mark.parametrize` instead to reduce duplication?

binderhub/tests/test_registry.py
```diff
@@ -230,7 +281,7 @@ async def test_get_image_manifest(tmpdir, token_url_known):
             f,
         )
     if token_url_known:
-        token_url = url + "/token"
+        token_url = url + "/token?service=container_registry"
```
manics, line 284: Is this needed? If it's testing another scenario can you use `@pytest.mark.parametrize` to test with and without this extra parameter?

manics (changes requested):
Sorry for the delay in reviewing. Over time the registry class has been incrementally modified to support other registries but unfortunately the documentation of how each registry works hasn't always been included, hence the difficulty in maintaining this code.

I think the best we can do is to ensure the existing tests pass unmodified (unless it's clear why the test needs to be changed), and to add more tests for new functionality or fixes as you've done.

bilke:
I have setup BinderHub with ACR and in principal it works until the built image is downloaded where it fails with:

```
Failed to pull image "myacr.azurecr.io/...": failed to pull and unpack image: failed to resolve reference ... :
failed to authorize: failed to fetch anonymous token:
unexpected status from GET request to https://myacr.azurecr.io/oauth2/token?scope=repository%3Amyacracr%2Fmy-bilke-2dbinder-2dogs-2drequirements-d16279%3Apull&service=mybenacr.azurecr.io: 401 Unauthorized
```

I guess it is because of the issue described in #1620.

Is there a way to test this patch in my instance? I am a k8s / helm noobie so it would awesome could point out the required steps to test this, thanks a lot!

PR #1635, opened 2023-02-15
switch default server to jupyter-server-based jupyter-lab

### LINKED ISSUE #1634: Jupyter Server support
### Proposed change

Recent JupyterLab, and a growing number of extensions require the newer Jupyter Server, and will no longer run on the legacy `jupyter-notebook` server. But BinderHub (without auth) hardcodes the launch command as `jupyter-notebook`, and there's no way for repos to opt-in to the current standard Jupyter server (except by [redefining `jupyter-notebook`](https://github.com/jorgensd/dolfinx-tutorial/pull/113)).

At this point, I think we should switch to launching jupyter-server by default, which I think will be more compatible more often, but is a backward-incompatible change for _some_ repos, especially those that may have old server extensions. In a _binder_ context, I think there are far more lab extensions than classic nb extensions, so I think at this point it will result in _fewer_ broken repos, not more. But it's still a breaking change since which repos are broken will _change_.

xref https://github.com/jupyter-server/jupyter_server/issues/1038

### Alternative options

Make this an option somehow, and possibly opt-in (eventually, it has to be opt-out since we're currently defaulting to a deprecated package that already doesn't work with the current version of our default UI). I don't know where to express that., though, in a way that's going to work.

Everyone who wants to use recent JupyterLab or any other Jupyter frontends other than `notebook<7`.

Add to our launch command `jupyter-server` _by default_ with a fallback on `jupyter-notebook`. We could gate this on the presence of Jupyter Server 2.0, because Jupyter Server 1.0 _tends_ to be compatible with the legacy notebook server. This decision isn't really made by the user though, it's made for most repos by the repo2docker base env. And I think it would be pretty weird to say 'if you want to launch the old jupyter-notebook, install jupyter-server<2.'

minrk:
This will launch jupyter-server, and be far more likely to be compatible with current repos than launching the old jupyter-notebook server.

It also makes the behavior of unauthenticated binderhub more similar to authenticated binderhub, which already does this same thing via `jupyterhub-singleuser`.

more detail in #1634 

This is a breaking change, in that anything that _relied_ on running the old server (mostly legacy extensions, which _often_, but do not _always_ work) may break. But it's also a big fix, because now anything that relies on the _current_ server is not broken (i.e. JupyterLab itself). I think this will result in a _reduction_ of broken repos, but since it's a _change_, not a pure improvement, it's still a backward-incompatibility issue. It's also not great that repos cannot pick this, really, other than uninstalling jupyterlab, given how this works.

I can't seem to find it now, but I feel like we had a discussion and decided on switching the UI first (which we did a long time ago), and then switching the server at some later point, which is what's happening here. That was a long time ago, and I can't find it.

closes #1634 

xref https://github.com/jupyter-server/jupyter_server/issues/1038 - this should fix issues with jupyter-server 2.

### DIFF
--- a/binderhub/binderspawner_mixin.py
+++ b/binderhub/binderspawner_mixin.py
@@ -84,6 +84,10 @@ class BinderSpawnerMixin(Configurable):
             if self.cors_allow_origin == "*":
                 args.append("--NotebookApp.allow_origin_pat=.*")
             args += self.args
+            # ServerApp compatibility: duplicate NotebookApp args
+            for arg in list(args):
+                if arg.startswith("--NotebookApp."):
+                    args.append(arg.replace("--NotebookApp.", "--ServerApp."))
         return args
 
     def start(self):
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -169,6 +169,10 @@ jupyterhub:
                     if self.cors_allow_origin == "*":
                         args.append("--NotebookApp.allow_origin_pat=.*")
                     args += self.args
+                    # ServerApp compatibility: duplicate NotebookApp args
+                    for arg in list(args):
+                        if arg.startswith("--NotebookApp."):
+                            args.append(arg.replace("--NotebookApp.", "--ServerApp."))
                 return args
 
             def start(self):
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -224,12 +224,14 @@ jupyterhub:
         else:
             have_lab = major >= 3
 
-        if have_lab and "NotebookApp.default_url" not in " ".join(sys.argv):
-            # if recent-enough lab is available, make it the default UI
-            sys.argv.insert(1, "--NotebookApp.default_url=/lab/")
+        if have_lab:
+          # technically, we could accept another jupyter-server-based frontend
+          exe = "jupyter-lab"
+        else:
+          exe = "jupyter-notebook"
 
         # launch the notebook server
-        os.execvp("jupyter-notebook", sys.argv)
+        os.execvp(exe, sys.argv)
     events: true
     storage:
       type: none


consideRatio (approved):
Looks good to me, and makes sense to me

manics (approved):

minrk:
I think we should do this, and pretty soon. I think _few_ users will be affected, but not none. I opened https://discourse.jupyter.org/t/binder-switching-to-jupyter-server-by-default/18011 to solicit feedback, since I want to do a better job communicating the change than we did with the UI switch.

manics:
What are your thoughts on deploying this and other potentially breaking changes in repo2docker (e.g. default python 3.10) to mybinder.org- one change at a time, or combine them all into a single deployment and announcement?

minrk:
How long do folks think is appropriate to wait for feedback on this one? The Python 3.10 switch has been announced for almost 6 weeks, with no objections, so I feel like that one's ready to deploy.

I'm not sure if it's better to deploy several breaking changes spread out over a week or two, or one big deploy that changes all 3.

consideRatio:
I think one big. Less communication, and less states to consider if someone reports an issue - then one can know that either it was the big patch or not, as compared to before any patch, after patch 1, 2, or 3 etc.

minrk:
Shall we do this update this week? We've received no feedback other than some hearts in 12 days so far.

consideRatio:
@minrk I think that would be great! If I can help you with review or similar please do ping me and I'll try assist in such capacity

manics:
Can we merge this now?

Then open a combined mybinder.org-deploy PR to bump both BinderHub and repo2docker, along with a prepared announcement to post to Discourse when the deployment PR is merged.

consideRatio:
Merging to help offload with misc followups if issues surface in the CI system or similar!

minrk:
https://github.com/jupyterhub/mybinder.org-deploy/pull/2509 is the combined-update PR

minrk:
@manics thanks for the review on the post! Going to publish and deploy.

minrk:
Actually, I realized the blog post should have been an Issue on team-compass to make sure folks can review it. I'll do that and give it a bit more time.

minrk:
This has been deployed.

Issue #1636, opened 2023-02-17
Container builds take 4x as long following base image switch to Python 3.11


manics:
### Bug description

https://github.com/jupyterhub/binderhub/actions/workflows/publish.yml?query=branch%3Amain

Container images based on Python 3.9 used to take around 30 minutes total for chartpress to publish the chart.

Following the [switch to Python 3.11](https://github.com/jupyterhub/binderhub/actions/runs/3877664894) this job takes around 2 hours.

consideRatio:
I bet it relates to compiling things that have pre-built wheels for py39 but not for py311

consideRatio:
The single build step that takes notable time is [this one](https://github.com/jupyterhub/binderhub/actions/runs/4121415327/jobs/7117080666#step:10:1742), but not form amd64 but for arm64.

```
#29 [linux/arm64 build-stage 6/6] RUN --mount=type=cache,target=/tmp/pip-cache     pip install build  && pip wheel         .         pycurl         -r helm-chart/images/binderhub/requirements.txt
```

The slowness stems from this dependency:

https://github.com/jupyterhub/binderhub/blob/c71a2034aed6deb86935dbe9ce76304ad4f871e3/helm-chart/images/binderhub/requirements.txt#L59-L64

minrk:
Maybe best to switch the image to 3.10 for now?

consideRatio:
It seems that this is to be fixed quite soon, I suggest we wait. See https://github.com/grpc/grpc/issues/32454 and note that their current release candidate has aarch64 wheels for py311

consideRatio:
- Closed by #1654 that bumps to a grpc version that has an aarch64 build for py311

PR #1638, opened 2023-02-23
helm-chart: refactor Dockerfile to make it obvious we don't fail to pass built wheels

consideRatio:
I think we actually built the wheel for grpcio correctly in the build-stage, and that was transferred, but I hesitated on that when I saw changes compared to a Dockerfile in z2jh.

With this, I'm making the Dockerfiles align, but in practice the reason for #1636 is that it takes ages to build `grpcio` on arm64 where they don't have a wheel published.

### DIFF
--- a/helm-chart/images/binderhub/Dockerfile
+++ b/helm-chart/images/binderhub/Dockerfile
@@ -3,11 +3,12 @@
 
 # The build stage
 # ---------------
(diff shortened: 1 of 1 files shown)


consideRatio:
Self merging as this now conflicts with #1648 so I can rebase that on this

PR #1640, opened 2023-03-01
Improve detection of jupyterlab server

minrk:
some prior repo2docker builds have broken environments where labapp fails to start, even though the `jupyterlab` package is present

tested in https://github.com/jupyterhub/mybinder.org-deploy/pull/2517


minrk:
https://dataverse.ucla.edu/ appears to be entirely down, causing the one test failure.

PR #1641, opened 2023-03-06
Ensure hub_url is configured before use

berghaus:
This PR follows the discussion in discourse post [18215](https://discourse.jupyter.org/t/podman-inside-kubernetes-pink-or-docker-inside-docker-dind-to-build-repositories/18215).

Problem: the configuration of BinderHub fails if hub_url is not set.

Solution: set a default value for the `hub_url` in the helm chart. Check that `hub_url` is set before use.

I tested the helm chart (as per the documentation), but am unsure how to test the change to `binderhub_config.py` properly.

### DIFF
--- a/helm-chart/binderhub/files/binderhub_config.py
+++ b/helm-chart/binderhub/files/binderhub_config.py
@@ -55,6 +55,8 @@ if imageBuilderType in ["dind", "pink"]:
         c.BinderHub.build_docker_host = f"unix://{hostSocketDir}/{socketname}.sock"
 
 if c.BinderHub.auth_enabled:
+    if "hub_url" not in c.BinderHub:
+        c.BinderHub.hub_url = ""
     hub_url = urlparse(c.BinderHub.hub_url)
     c.HubOAuth.hub_host = f"{hub_url.scheme}://{hub_url.netloc}"
     if "base_url" in c.BinderHub:
--- a/helm-chart/binderhub/values.yaml
+++ b/helm-chart/binderhub/values.yaml
@@ -45,7 +45,7 @@ config:
     # auth_enabled:
     base_url: /
     build_node_selector: {}
-    hub_url: ""
+    # hub_url:
     # hub_url_local:
     use_registry: true
   KubernetesBuildExecutor: {}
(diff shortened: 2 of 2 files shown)


manics:
You can just set the default to `""` in
https://github.com/jupyterhub/binderhub/blob/6b16f8d3888263be8b3a1ccbc8f63734b55627ec/binderhub/app.py#L512-L513
e.g.
https://github.com/jupyterhub/binderhub/blob/6b16f8d3888263be8b3a1ccbc8f63734b55627ec/binderhub/app.py#L404-L417

berghaus:
> You can just set the default to `""` in
> 
> https://github.com/jupyterhub/binderhub/blob/6b16f8d3888263be8b3a1ccbc8f63734b55627ec/binderhub/app.py#L512-L513


Unfortunately, that won't address the issue.  The error is raised because an unset configuration value is retrieved, not the BinderHub parameter that the configuration value is meant to set:

https://github.com/jupyterhub/binderhub/blob/ddba2322d9a3f8d8d1a91e405f4ceec488071740/helm-chart/binderhub/files/binderhub_config.py#L58

I proved this to myself like so:
```python
# Sample configurable:
from traitlets.config.configurable import Configurable
from traitlets.config import get_config
from traitlets import Int, Unicode

class MyClass(Configurable):
    name = Unicode(config=True)
    number = Int(config=True)

# Create a configuration
c = get_config()
c.MyClass.number = 10
try:
    urlparse(c.MyClass.name)
except AttributeError:
    print("using config item produced error")

# construct class instance:
m = MyClass()
try:
    urlparse(m.name)
except AttributeError:
    print("using instane property produced error")

print("test done")
```
which produces:
```bash
using config item produced error
test done
```

helm-chart/binderhub/files/binderhub_config.py
```diff
@@ -55,6 +55,8 @@ def get_value(key, default=None):
         c.BinderHub.build_docker_host = f"unix://{hostSocketDir}/{socketname}.sock"
 
 if c.BinderHub.auth_enabled:
+    if "hub_url" not in c.BinderHub:
```
berghaus, line 58: The more idiomatic to catch the exception

manics (approved):
Thanks! Sorry for the delay

PR #1647, opened 2023-03-24
Allow building the image without needing to launch it

consideRatio:
(PR written and made by @GeorgianaElena, it was just initialized by @consideRatio to get a branch to reference)

This PR adds a feature that offers the option of an API only mode, and in this mode, it allows opting-out of launching directly after build.

It adds:
- [x] a traitlet called `enable_api_only_node` that  defaults to False to preserve existing default behavior. When enabled, there will be no UI, with the only registered endpoints being:
   - `/metrics`
   - `/versions`
   - `/build`
   - `/health`
   - `/_config`
   - `/*` -> shows a custom 404 page
- [x] a query parameter called `build_only` to be passed to individual requests to enable the build only **only** when `enable_api_only_node` is True

```mermaid
flowchart LR
  subgraph API only mode enabled
     direction TB
       id1{{false}}
       id2{{false}}
       id3{{false}}
       id4{{true}}
       id5{{true}}
       id6{{true}}
  end

  subgraph `build_only` query param
     direction TB
     id1 ---|and|id11{{missing}}
     id2 ---|and|id12{{false}}
     id3 ---|and|id13{{true}}
     id4 ---|and|id14{{missing}}
     id5 ---|and|id15{{false}}
     id6 ---|and|id16{{true}}
  end

   subgraph Outcome
     direction TB
     id11 ==> id21{{OK, image will be launched after build}}
     id12 ==> id22{{OK, image will be launched after build}}
     id13 ==> id23{{ERROR, building but not launching is not permitted when API only mode is not enabled}}
     id14 ==> id24{{OK, image will be launched after build}}
     id15 ==> id25{{OK, image will be launched after build}}
     id16 ==> id26{{OK, image will NOT be launched after build}}
  end
```

- [x] tests
   - [x] for the configuration corner cases that generate errors
   - [x] for the actual build only config

It also updates the `auth` tests for consistency to expect a more descriptive pytest request parameter value in order for the app fixture to load the extra auth configuration. This is for consistency with the build only extra config that was added in this PR.

### Todo
- [ ] Document this once there's agreement on naming an evaluation of the the traitlet and query parameter values to come to an outcome.

### Background

This is meant as a building block for the https://github.com/2i2c-org/binderhub-service project. In brief, its positioning itself to be relevant specifically for situations when a binderhub UI is relevant together with a JupyterHub behind authentication and that also provides user home folder storage.

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -993,35 +993,59 @@ class BinderHub(Application):
                 (
                     r"/(badge\.svg)",
                     tornado.web.StaticFileHandler,
-                    {"path": os.path.join(self.tornado_settings["static_path"], "images")},
+                    {
+                        "path": os.path.join(
+                            self.tornado_settings["static_path"], "images"
+                        )
+                    },
                 ),
                 # /badge_logo.svg
                 (
                     r"/(badge\_logo\.svg)",
                     tornado.web.StaticFileHandler,
-                    {"path": os.path.join(self.tornado_settings["static_path"], "images")},
+                    {
+                        "path": os.path.join(
+                            self.tornado_settings["static_path"], "images"
+                        )
+                    },
                 ),
                 # /logo_social.png
                 (
                     r"/(logo\_social\.png)",
                     tornado.web.StaticFileHandler,
-                    {"path": os.path.join(self.tornado_settings["static_path"], "images")},
+                    {
+                        "path": os.path.join(
+                            self.tornado_settings["static_path"], "images"
+                        )
+                    },
                 ),
                 # /favicon_XXX.ico
                 (
                     r"/(favicon\_fail\.ico)",
                     tornado.web.StaticFileHandler,
-                    {"path": os.path.join(self.tornado_settings["static_path"], "images")},
+                    {
+                        "path": os.path.join(
+                            self.tornado_settings["static_path"], "images"
+                        )
+                    },
                 ),
                 (
                     r"/(favicon\_success\.ico)",
                     tornado.web.StaticFileHandler,
-                    {"path": os.path.join(self.tornado_settings["static_path"], "images")},
+                    {
+                        "path": os.path.join(
+                            self.tornado_settings["static_path"], "images"
+                        )
+                    },
                 ),
                 (
                     r"/(favicon\_building\.ico)",
                     tornado.web.StaticFileHandler,
-                    {"path": os.path.join(self.tornado_settings["static_path"], "images")},
+                    {
+                        "path": os.path.join(
+                            self.tornado_settings["static_path"], "images"
+                        )
+                    },
                 ),
             ]
         # This needs to be the last handler in the list, because it needs to match "everything else"
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -789,6 +789,7 @@ class BinderHub(Application):
     no_launch = Bool(
         False,
         help="When enabled, the hub will no longer launch the image after the build",
+        config=True,
     )
 
     _build_config_deprecated_map = {
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -786,6 +786,11 @@ class BinderHub(Application):
         help="Origin to use when emitting events. Defaults to hostname of request when empty",
     )
 
+    no_launch = Bool(
+        False,
+        help="When enabled, the hub will no longer launch the image after the build"
+    )
+
     _build_config_deprecated_map = {
         "appendix": ("BuildExecutor", "appendix"),
         "push_secret": ("BuildExecutor", "push_secret"),
@@ -943,6 +948,7 @@ class BinderHub(Application):
                 "auth_enabled": self.auth_enabled,
                 "event_log": self.event_log,
                 "normalized_origin": self.normalized_origin,
+                "no_launch": self.no_launch,
             }
         )
         if self.auth_enabled:
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -967,13 +967,18 @@ class BinderHub(Application):
                 "Access-Control-Allow-Origin"
             ] = self.cors_allow_origin
 
-        if self.enable_api_only_mode:
-            # Deactivate the UI's for the `/` and `/v2` endpoints
-            # as they won't make sense in a build only scenario
-            # when a REST API usage is expected
-            handlers = []
-        else:
-            handlers = [
+        handlers = [
+            (r"/metrics", MetricsHandler),
+            (r"/versions", VersionHandler),
+            (r"/build/([^/]+)/(.+)", BuildHandler),
+            (r"/health", self.health_handler_class, {"hub_url": self.hub_url_local}),
+            (r"/_config", ConfigHandler),
+        ]
+        if not self.enable_api_only_mode:
+            # In API only mode the endpoints in the list below
+            # are unregistered as they don't make sense in a API only scenario
+            handlers += [
+                (r"/about", AboutHandler),
                 (r"/v2/([^/]+)/(.+)", ParameterizedMainHandler),
                 (r"/", MainHandler),
                 (r"/repo/([^/]+)/([^/]+)(/.*)?", LegacyRedirectHandler),
@@ -1018,16 +1023,9 @@ class BinderHub(Application):
                     tornado.web.StaticFileHandler,
                     {"path": os.path.join(self.tornado_settings["static_path"], "images")},
                 ),
-                (r"/about", AboutHandler),
-                (r".*", Custom404),
             ]
-        handlers += [
-            (r"/metrics", MetricsHandler),
-            (r"/versions", VersionHandler),
-            (r"/build/([^/]+)/(.+)", BuildHandler),
-            (r"/health", self.health_handler_class, {"hub_url": self.hub_url_local}),
-            (r"/_config", ConfigHandler),
-        ]
+        # This needs to be the last handler in the list, because it needs to match "everything else"
+        handlers.append((r".*", Custom404))
         handlers = self.add_url_prefix(self.base_url, handlers)
         if self.extra_static_path:
             handlers.insert(
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -786,34 +786,16 @@ class BinderHub(Application):
         help="Origin to use when emitting events. Defaults to hostname of request when empty",
     )
 
-    require_build_only = Bool(
+    enable_api_only_mode = Bool(
         False,
-        help="""When enabled, the hub will no longer launch the image after the build when combined with the appropriate
-        `build_only` request query parameter.
-
-        Below it's the table for evaluating whether or not the image will be launched after build,
-        based on the values of the `require_build_only` traitlet and the `build_only` query parameter.
-
-        +----------------------+--------------+---------------------------------------------------------------------------------------+
-        | `require_build_only` | `build_only` | Outcome                                                                               |
-        +======================+==============+=======================================================================================+
-        | false                | missing      | OK, image will be launched after build                                                |
-        +----------------------+--------------+---------------------------------------------------------------------------------------+
-        | false                | false        | OK, image will be launched after build                                                |
-        +----------------------+--------------+---------------------------------------------------------------------------------------+
-        | false                | true         | ERROR, building but not launching is not permitted when require_build_only == False   |
-        +----------------------+--------------+---------------------------------------------------------------------------------------+
-        | true                 | missing      | ERROR, query parameter must be explicitly set to true when require_build_only == True |
-        +----------------------+--------------+---------------------------------------------------------------------------------------+
-        | true                 | false        | ERROR, query parameter must be explicitly set to true when require_build_only == True |
-        +----------------------+--------------+---------------------------------------------------------------------------------------+
-        | true                 | true         | OK, image won't be launched after build                                               |
-        +----------------------+--------------+---------------------------------------------------------------------------------------+
-
-        .. note::
-            When this feature is enabled the only supported UX is the REST API based UX, as it's out of scope to support
-            a new UI based UX around not launching.
-
+        help="""
+        When enabled, BinderHub will operate in an API only mode,
+        without a UI, and with the only registered endpoints being:
+            - /metrics
+            - /versions
+            - /build/([^/]+)/(.+)
+            - /health
+            - /_config
         """,
         config=True,
     )
@@ -975,7 +957,7 @@ class BinderHub(Application):
                 "auth_enabled": self.auth_enabled,
                 "event_log": self.event_log,
                 "normalized_origin": self.normalized_origin,
-                "require_build_only": self.require_build_only,
+                "enable_api_only_mode": self.enable_api_only_mode,
             }
         )
         if self.auth_enabled:
@@ -985,7 +967,7 @@ class BinderHub(Application):
                 "Access-Control-Allow-Origin"
             ] = self.cors_allow_origin
 
-        if self.require_build_only:
+        if self.enable_api_only_mode:
             # Deactivate the UI's for the `/` and `/v2` endpoints
             # as they won't make sense in a build only scenario
             # when a REST API usage is expected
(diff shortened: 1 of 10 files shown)


yuvipanda (changes requested):
Thank you so much for working on this, @GeorgianaElena!

In thinking about config, I was originally thinking about two traitlets and a query param, and made this table:


| Scenario | `allow_no_launch` traitlet | `require_no_launch` traitlet | `launch=false` query parameter |
| - | - | - | - |
| mybinder.org / public binders | false | false | error is returned, saying 'building but not launching is not permitted' |
| binderhub used as a service to build images for a JupyterHub | true | true | works as intended |
| Mysterious scenario where a single binderhub is used both to launch into a preconfigured JupyterHub but *also* as an API for just building images | true | false | works as intended |

 However, upon looking at this table now, I think we don't really want to support row 3 - I don't see a good use case for that. So instead, I think the table should be:


| Scenario | `require_no_launch` traitlet | `launch=false` query parameter |
| - | - | - | 
| mybinder.org / public binders |  false | error is returned, saying 'building but not launching is not permitted' |
| binderhub used as a service to build images for a JupyterHub |  true | works as intended. If `launch=false` is not specified, return an error |

So in effect, to deploy a binderhub that will not try to launch into an existing JupyterHub automatically, the following things must both happen:

1. The binderhub Admin must set this traitlet config to True, stating their intention that they do not want to launch anywhere. This also allows them to skip the config for connecting a JupyterHub to this.
2. The person building the frontend code that makes this API call must *also* set a query parameter, stating their intention to just get back a built image spec - and not want an automatic launch.

I think the current combination of traitlets and query parameters is close, but doesn't quite implement this. Currently, the query parameter just *overrides* the traitlet - so the image gets built and not launched if **either** the traitlet or the query param is true. For public facing binders like mybinder.org, I don't think we want to allow this behavior - we want to allow image building but not launching if **both** the traitlet and the query parameter is true. Switch from an **or** to an **and**. 

I hope this makes sense! Let me know if it does not, and I can try clarify.

GeorgianaElena:
I think I understand @yuvipanda. Thanks for the explanation. Can you double check the following points to make sure I got it right, please?

### Double check list

1. When `require_no_launch = False`,  then a `mybinder` scenario  is considered, meaning that the query can be:
 - missing, in which case everything remains unchanged
 - `launch=true`, in which case everything remains unchanged
 - `launch=false`, in which case error is returned, saying 'building but not launching is not permitted'

2. When `require_no_launch = True`, then a service-like scenario is considered, meaning that the query can be:
 - missing, in which case  return an error
 - `launch=true`, in which case we launch?? (don't sure about this one. Do we still want an `and` situation herre also?)
 - `launch=false`, in which case we don't launch

### Prior discussions
> Currently, the query parameter just overrides the traitlet - so the image gets built and not launched if either the traitlet or the query param is true.

I believe @consideRatio made a point about why this (overriding through the query the traitlet) might be useful, but don't remember what it was 😬 @consideRatio wdyt?

### Naming
Thinking more about the naming, I realize that having a traitlet with a negation in its name, and then the query without, it's kind of confusion. Also, checking `if no require_no_launch` is also confusing I believe.

What do you think about a:
- traitlet called `allow_build_only` if we want to allow image building only if both the traitlet and the query parameter are try true 
- query param called `build_only`

yuvipanda:
> `launch=true`, in which case we launch?? (don't sure about this one. Do we still want an `and` situation herre also?)

I think in this case, we return an error - `launch=false` is *required*. This allows the binderhub to exist without requiring a JupyterHub it has launch rights on to be configured.

I like your proposed `build` based naming scheme, we can go with that. Thank you for thinking that through!

binderhub/builder.py
```diff
@@ -408,31 +408,69 @@ async def get(self, provider_prefix, _unescaped_spec):
...
 
+        require_build_only = self.settings.get("require_build_only", False)
+        build_only = False
+        if not require_build_only:
+            build_only_query_parameter = self.get_query_argument(
+                name="build_only", default=""
+            )
+            if build_only_query_parameter.lower() == "true":
+                raise HTTPError(
+                    log_message="Building but not launching is not permitted!"
```
consideRatio, line 419: The error stems from not being passed a `build_only=true` query parameter when BinderHub is configured with `require_build_only`. I think it would be good to surface that more explicitly as well in this message to help anyone seeing this directly or indirectly via a issue report figure out what is going on.

The error message can be lengthy without issues, so we can include both a non-technical description as well.

binderhub/builder.py
```diff
@@ -408,31 +408,69 @@ async def get(self, provider_prefix, _unescaped_spec):
...
+                )
+            # If we're here, it means a build only deployment is required
+            build_only = True
+
+        if build_only:
+            await self.emit(
+                {
+                    "phase": "info",
+                    "imageName": image_name,
+                    "message": "Build only was enabled. Image will not be launched after build.\n",
```
consideRatio, line 437:
```suggestion
                    "message": "build_only was passed, built image will not be launched.\n",
```

binderhub/builder.py
```diff
@@ -408,31 +408,69 @@ async def get(self, provider_prefix, _unescaped_spec):
...
-            )
-            return
+                if build_only:
+                    await self.emit(
+                        {
+                            "phase": "ready",
+                            "imageName": image_name,
+                            "message": "Image won't be launched\n",
+                        }
+                    )
```
consideRatio, line 456: https://github.com/jupyterhub/binderhub/blob/554549c5ff83d87c57e7bb88b05c0120f1b17afe/binderhub/builder.py#L55-L58

It seems `LAUNCHES_INPROGRESS` is about a prometheus metric tracking launch attempts. I figure we should not run code associated with build_only under it as it may briefly then increment a counter (to later quickly decrement it).

binderhub/builder.py
```diff
@@ -408,31 +408,69 @@ async def get(self, provider_prefix, _unescaped_spec):
...
+                            "build_token": self._have_build_token,
+                            "origin": self.settings["normalized_origin"]
+                            if self.settings["normalized_origin"]
+                            else self.request.host,
+                        },
+                    )
+                    return
 
         # Prepare to build
         q = Queue()
```
consideRatio, line 476: This return statement is only provided when launching and having found a built image, but probably makes sense no matter what as the following logic goes on with a `# Prepare to build` comment

binderhub/builder.py
```diff
@@ -500,7 +538,7 @@ def _check_result(future):
                     elif progress.payload == ProgressEvent.BuildStatus.BUILT:
                         event = {
                             "phase": phase,
-                            "message": "Built image, launching...\n",
+                            "message": "Done! Image built.\n",
```
consideRatio, line 541: I'm a bit worried that we change to much by changing this frequently seen log messages. It feels like a part of the binderhub experience at this point. I'm so used to `Built image, launching...` and would be a bit uncertain on what was going on if I just saw `Done! Image built.` instead, especially without any further messages about waiting for a launch.

I consider the options of:
- making the messages exactly like they were before this feature was around
- breaking apart the message `Built image, launching...` into two parts, one about the image being built, and one about launching or being done.

binderhub/builder.py
```diff
@@ -544,21 +582,30 @@ def _check_result(future):
...
+                            "provider": provider.name,
+                            "spec": spec,
+                            "ref": ref,
+                            "status": "success",
+                            "build_token": self._have_build_token,
+                            "origin": self.settings["normalized_origin"]
+                            if self.settings["normalized_origin"]
+                            else self.request.host,
+                        },
+                    )
```
consideRatio, line 608: I read about event logging in https://binderhub.readthedocs.io/en/latest/eventlogging.html. I figure it could be relevant to track the amount of builds we do, even though binderhub isn't launching.

I think that could be a related but separate PR though, independent on this feature.

consideRatio (commented):
Thank you for working on this @GeorgianaElena and @yuvipanda!!

GeorgianaElena:
Thanks for the feedback @consideRatio ✨ it helped me simplify and correct the logic a bit more while also keeping the old log messages. I've also managed to get an understanding of the current tests and added some for the new functionality.

This is why I marked this as ready for review.

Note that it still lacks documentation but I plan to add that once there's agreement on the implementation.

consideRatio:
Thank you for working this @GeorgianaElena!!! I'll look at this tomorrow morning!

consideRatio (commented):
Great work on this @GeorgianaElena!!!

binderhub/tests/test_build.py
```diff
@@ -96,9 +96,61 @@ async def test_build(app, needs_build, needs_launch, always_build, slug, pytestc
     assert r.url.startswith(final["url"])
 
 
+@pytest.mark.asyncio(timeout=900)
+@pytest.mark.parametrize(
+    "app,build_only",
+    [
+        ("app_with_require_build_only", True),
+        ("app_with_require_build_only", "true"),
+        ("app_with_require_build_only", "True"),
```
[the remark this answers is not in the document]
GeorgianaElena: Sure!

binderhub/tests/conftest.py
```diff
@@ -252,10 +253,25 @@ def app(request, io_loop, _binderhub_config):
...
+            _binderhub_config.merge(cfg)
+            build_only = True
+
+    if not build_only:
+        # load conf that sets BinderHub.require_build_only = False
+        # otherwise because _binderhub_config has a session scope,
+        # any previous set of require_build_only to True will stick around
+        cfg = Config({"BinderHub": {"require_build_only": False}})
         _binderhub_config.merge(cfg)
+
```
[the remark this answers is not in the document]
GeorgianaElena: Me too :D

binderhub/builder.py
```diff
@@ -408,30 +408,89 @@ async def get(self, provider_prefix, _unescaped_spec):
...
+                )
+        else:
+            # Raise an error if the `build_only` query parameter is anything but `(T)true`
+            if build_only_query_parameter.lower() != "true":
+                raise HTTPError(
+                    log_message="The `build_only=true` query parameter is required when traitlet `require_build_only` is false! "
+                    "See the table below for more info.\n\n" + log_message_eval_table
+                )
+            # If we're here, it means a build only deployment is required
+            build_only_outcome = True
```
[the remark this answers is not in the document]
GeorgianaElena: Very good suggestions @consideRatio. Thank you! Done in https://github.com/jupyterhub/binderhub/pull/1647/commits/a95acae2f5dc8bc9caa9e31a7e4a340280cc1199

binderhub/app.py
```diff
@@ -767,6 +767,12 @@ def _template_path_default(self):
         help="Origin to use when emitting events. Defaults to hostname of request when empty",
     )
 
+    require_build_only = Bool(
+        False,
+        help="When enabled, the hub will no longer launch the image after the build",
+        config=True,
+    )
```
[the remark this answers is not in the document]
GeorgianaElena: Good point! Added it in https://github.com/jupyterhub/binderhub/pull/1647/commits/cbb5d0b19dbf574fd6bf583c5570a9c0d96c3635

consideRatio (commented):
This looks great to me @GeorgianaElena!

With this well documented traitlet for an opt-in feature only to be used in certain circumstances, I consider this mergable as it is.

GeorgianaElena:
After testing this out with @consideRatio yesterday, we concluded that even though we decided that it is out of scope of this PR to support a new UI for the build only option, it is confusing to still have the classical build + launch available.

When clicking the `launch` button, that will imediatlly fail with:
Screenshot 2023-04-28 at 11 08 02

With the last commit, https://github.com/jupyterhub/binderhub/pull/1647/commits/b82758ee00ace584ab53607806ba01127297fa44, when `require_build_only` is set, the classical UI is deactivated and instead a 204 No content page is rendered using the error template. It looks like below:
Screenshot 2023-04-28 at 10 41 01


Does something like this makes sense?

GeorgianaElena:
Failures seem to be happening on the main branch as well https://github.com/jupyterhub/binderhub/actions/runs/4829259002/jobs/8604077929 so not sure which if any are related to the PR

consideRatio:
@GeorgianaElena I've been debugging this for quite some time now, I'm drawing blank.

Apparently the hub pods launching are stuck pending without k8s Events associated with them. That indicates to me they aren't getting scheduled, not even failing to get scheduled to a node I think - but why?

consideRatio (commented):
I'm still brainstorming not confident on what I think is desirable

- Do we want to de-register `/about` as well, and possibly `/_config` that I don't know what it does? And along with that perhaps deregistering everything about favicons and badges etc?
- Should Custom404 respond with JSON instead if require_build_only is set?
- If require_build_only pretty much is "api only mode", should we make api_only_mode a standalone configuration or similar? Hmmm, this is a jupyterhub thing as well i recall... Yes https://jupyterhub.readthedocs.io/en/latest/howto/api-only.html writes about an API Only mode.

binderhub/builder.py
```diff
@@ -542,31 +590,33 @@ def _check_result(future):
...
-                    "status": "success",
-                    "build_token": self._have_build_token,
-                    "origin": self.settings["normalized_origin"]
-                    if self.settings["normalized_origin"]
-                    else self.request.host,
-                },
-            )
+        if build_only_outcome:
+            return
+        else:
```
consideRatio, line 597: A code style detail, feel free to mark as resolved without further ado if you think its nicer as it is!

I've grown to appreciate the pattern of returning and avoiding nesting inside an `else` section. In this case it would reduce the lines changed by indentation only, which provides a small upside for reviewers as well.

Current:

```python
if x:
    return
else:
    print("hi")
```

The considered option:

```python
if x:
    return

print("hi")
```

binderhub/builder.py
```diff
@@ -542,31 +590,33 @@ def _check_result(future):
...
+            # Launch after building an image
+            if not failed:
+                BUILD_TIME.labels(status="success").observe(
+                    time.perf_counter() - build_starttime
+                )
+                BUILD_COUNT.labels(status="success", **self.repo_metric_labels).inc()
```
consideRatio, line 603: I think we should have BUILD_TIME and BUILD_COUNT tracked before returning independently `if build_only_outcome`.

Let's relocate this part to be under `elif progress.payload == ProgressEvent.BuildStatus.BUILT:` (similarly to how BUILD_TIME is updated if failure is observed on line 587).

consideRatio (commented):
This looks great to me!

I think that the prometheus metrics `BUILD_TIME` and `BUILD_COUNT` should be tracked still, let's resolve that and ask someone else to review this after that.

binderhub/app.py
```diff
@@ -952,7 +985,17 @@ def initialize(self, *args, **kwargs):
...
+            # Deactivate the UI's for the `/` and `/v2` endpoints
+            # as they won't make sense in a build only scenario
+            # when a REST API usage is expected
+            handlers = []
+        else:
+            handlers = [
+                (r"/v2/([^/]+)/(.+)", ParameterizedMainHandler),
+                (r"/", MainHandler),
+            ]
+        handlers += [
```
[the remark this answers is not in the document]
GeorgianaElena: I just checked and `/metrics` and `/about` render the correct pages

GeorgianaElena:
@consideRatio, I believe I've addressed everything in your latest feedback. Thank you!

binderhub/builder.py
```diff
@@ -542,31 +596,29 @@ def _check_result(future):
...
+                        "provider": provider.name,
+                        "spec": spec,
+                        "ref": ref,
+                        "status": "success",
+                        "build_token": self._have_build_token,
+                        "origin": self.settings["normalized_origin"]
+                        if self.settings["normalized_origin"]
+                        else self.request.host,
+                    },
+                )
```
consideRatio, line 621: The indentation here seems unrelated to the PR. The indentation change could make sense, but I'm not sure. It would influence the time of the prometheus metric a bit, decrementing LAUNCHES_INPROGRESS after event_log.emit() has been called also.

Since the BUILD_TIME tracking is resolved before emitting the event, I suggest we revert this indentation for consistency and to make this PR scoped tighter

```
            with LAUNCHES_INPROGRESS.track_inprogress():
                await self.launch(provider)
```

GeorgianaElena: Oh, yes. I think the inndentation is like this by mistake. Thanks for spotting it

consideRatio:
@minrk and @manics I've requested your review for this PR which is done by @GeorgianaElena (in my fork because I wanted a placeholder branch), and reviewed by me and @yuvipanda so far.

We want to request your review to ensure that this is considered reasonable from jupyterhub team-members outside 2i2c. The key use of this feature is to enable binderhub to run in a mode where it is decoupled from jupyterhub, as deployed via https://github.com/2i2c-org/binderhub-service.

minrk (commented):
The implementation looks solid to me. My only questions are about how the behavior is expressed via configuration. I would expect the `require_build_only` flag to imply that the build_only flag is _required_, but may be True or False. But that's not what we have. What we have is a single flag that makes at least 3 changes:

1. disables UI, so it becomes an “api-only” service
2. disables launch entirely
3. requires an extra arg to be specified for all requests that is only allowed to have a single value

I don’t think `require_build_only` communicates the lack of UI, and requiring an argument be specified when there’s no possibility of it being any other value doesn’t feel right to me.

I’d propose only these two small (separate) changes:

1. give the config a name that indicates how much is changing, e.g. `build_only_api_only` or `disable_launch_and_ui`
2. accept the only allowed value when unspecified in both cases, rather than requiring opt-in to the only accepted value (I’m less sure about this, especially in more detailed proposal below)

I know build-only UI is out of scope for this PR, but if we think it might be in scope for a _later_ change, it may be worth exploring what that config would look like. Same goes for _allowing_ build-only on a deployment that also allows launch.

That would suggest to me these options:

- enable_ui (default: true)
- allow_build_only (default: false)
- allow_launch (default: true)

with these relationships:

- (for now) allow_build_only must be combined with enable_ui = false and allow_launch = false. These _may_ be temporary restrictions that can be relaxed over time without any changes in config options.
- allow_launch must not (for now) be combined with allow_build_only
- allow_build_only and allow_launch must not both be False (binderhub can’t do anything!)
- UI can be disabled independently of the build_only flag (new!)


I realize that makes the config required for build-only more complex, because it must be:

```python
allow_build_only = True
allow_launch = False
enable_ui = False
```


but there are some advantages to this, because it’s explicit that 3 separate behaviors are changing. It also leaves room for future combinations of values to be supported without breaking any deployments or configurations, because we don’t _imply_ the other changes when opting in to build-only/api-only, each change is explicit.

I also think if we go this route, then _requiring_ the build_only=true query param starts to make more sense, because then in the API it seems clearer to me that build_only has a default value of true in the API.

I know naming discussions are tedious, but I think it's important to try to be clear, especially if we are thinking of of supporting more possible combinations in the future.

consideRatio:
Thank you for reviewing @minrk!!! :heart: :tada:

> I realize that makes the config required for build-only more complex, [...], but there are some advantages to this, because it’s explicit that 3 separate behaviors are changing.

I've come to appreciate these more explicit config choices greatly, I think overall they are easier to implement, document, understand, maintain, and finally build other functional upon.

yuvipanda:
Here's a demo of this in action, in a `profile_list` page!

https://github.com/jupyterhub/binderhub/assets/30430/ede5e8a5-b705-4b81-8279-1ad913b0a128

GeorgianaElena:
@minrk, I've added a few commits based on your feedback above (thank you),. I've implemented a combination of your suggestions above! What do you think about:

> 
>    - give the config a name that indicates how much is changing, e.g. build_only_api_only or disable_launch_and_ui
>    - accept the only allowed value when unspecified in both cases, rather than requiring opt-in to the only accepted value (I’m less sure about this, especially in more detailed proposal below)


- `enable_api_only_mode` config
    This introduces an API mode that only keeps the `/build` endpoint and a few other that are informative, like metrics, versions, health etc.

   _Without additional config, this mode does not assume anything about a build only behavior_

- `build_only` query parameter

   This flag enables build only  when it's set to True. In all other cases, the `/build` endpoint will do what it does by default, which is to build + launch.
   
   ⬆️ This is what I am unsure of. Does an api only mode only makes sense in a build only scenario?

binderhub/builder.py
```diff
@@ -228,6 +228,24 @@ def set_default_headers(self):
...
+        # Get the value of the `enable_api_only_mode` traitlet
+        enable_api_only_mode = self.settings.get("enable_api_only_mode", False)
+        # Get the value of the `build_only` query parameter if present
+        build_only_query_parameter = str(
+            self.get_query_argument(name="build_only", default="")
+        )
+        build_only_outcome = False
+        if build_only_query_parameter.lower() == "true":
+            if not enable_api_only_mode:
+                raise HTTPError(
```
minrk, line 241: This lacks a status code. Presumably it should be `400`?

binderhub/builder.py
```diff
@@ -228,6 +228,24 @@ def set_default_headers(self):
         self.set_header("content-type", "text/event-stream")
         self.set_header("cache-control", "no-cache")
 
+    def _get_build_only_outcome(self):
```
minrk, line 231: I'd call this `_get_build_only()`, since it's the dynamic resolution of the value of `build_only` with checks, etc., not a separate concept called 'outcome', which confused me a bit.

binderhub/builder.py
```diff
@@ -408,33 +426,52 @@ async def get(self, provider_prefix, _unescaped_spec):
             else:
                 image_found = True
 
-        if image_found:
+        build_only_outcome = self._get_build_only_outcome()
```
minrk, line 429: Same here, let's call it `build_only`, which I think is clearer than adding `outcome`

minrk (approved):
Sorry for leaving you waiting so long, it's a busy Fall! I left some minor comments to address in naming and clarity, but I think this is solid as is, and feel free to merge and iterate on refinements in smaller later PRs.

> Does an api only mode only makes sense in a build only scenario?

I don't think so. A separate service could use an api-only BinderHub to build and launch and then redirect to the running server. An example would be a service that uses BinderHub, but presents its own UI to the user before redirecting to a running Jupyter server. Much like BinderHub itself uses JupyterHub in api-only mode.

GeorgianaElena:
Thank you very much @minrk for the review and suggestions. I've added a commit that addresses them.

Some tests seem to be failing now though, but the same are failing for the main branch as well, so I don't think they are related to this PR?

minrk:
Yes, failures appear to be helm related, not related to this PR (one image pull failure and one memory limit issue). Going to merge this one!

PR #1648, opened 2023-03-24
maint: refactoring around pycurl as an optional dependency

consideRatio:
I saw an opportunity to reduce the complexity a bit by relying more on the optional dependency declared in setup.py on `pycurl` over specifying it separately. This would be especially relevant if we would enforce a version constraint on pycurl in setup.py.

We couldn't reduce the complexity that much since binderhub is installed locally. This is always tricky since requirements.in freezes into requirements.txt using `pip-compile` via the `ci/refreeze` script, and when it does we get an absolute path to install binderhub. And with that, I think we can't avoid issues when we are to install pre-built wheels in the Dockerfile.

### DIFF
--- a/docs/source/contribute.md
+++ b/docs/source/contribute.md
@@ -123,7 +123,7 @@ continue.
    requirements locally.
 
    ```bash
-   python3 -m pip install -e . -r dev-requirements.txt
+   python3 -m pip install -e ".[pycurl]" -r dev-requirements.txt
    ```
 
 1. Install the JupyterHub Helm chart by itself into your Kubernetes cluster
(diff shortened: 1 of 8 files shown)


docs/source/contribute.md
```diff
@@ -123,7 +123,7 @@ continue.
    requirements locally.
 
    ```bash
-   python3 -m pip install -e . -r dev-requirements.txt
+   python3 -m pip install -e ".[pycurl]" -r dev-requirements.txt
```
consideRatio, line 126: I think maybe we don't need pycurl here, but I wasn't sure so I retained it to match current setup.

consideRatio: I now know we need pycurl here and to run the test suite due to conftest.py imports MockAsyncHTTPClient which depends on a tornado Curl client that depends on pycurl.

consideRatio:
I've done a delayed self-review, and think it makes sense still as it is. I'll go for a self-merge in the intrest of minimizing use of our limited maintenance capacity.

PR #1659, opened 2023-04-18
maint: Prompt check of public repo during ref resolution

### LINKED ISSUE #1658: logic for error message for private repo
### Bug description

If you provide a private repo in GitHub, the error message is not helpful or what is intended. It seems there is a logic error in the code.

#### Expected behaviour

I'd expect it to say something like `Check to make sure this repository is public. ` 
The code expects to say `"Is your repo public?"`

https://github.com/jupyterhub/binderhub/blob/e27f63f8597a84a8cf079de284e6486fabbece88/binderhub/builder.py#L338

#### Actual behaviour

The actual message is 

`Could not resolve ref for gh:<username>/<repo>/HEAD. Double check your URL. GitHub recently changed default branches from "master" to "main".`

here's the line of code. 
https://github.com/jupyterhub/binderhub/blob/e27f63f8597a84a8cf079de284e6486fabbece88/binderhub/builder.py#L300

The desired message isn't produced because
`    if provider.name == "GitHub":` is True.

### How to reproduce

Use any private repository on GitHub in the mybinder interface.

matthewfeickert:
Resolves #1658

Instead of prompting the user to check if the URL for their repo is public only if it is not on GitHub and if all other resolution has failed, escalate prompting the user to check if the repo is public as soon as there are ref resolution problems.

### DIFF
--- a/binderhub/builder.py
+++ b/binderhub/builder.py
@@ -297,7 +297,9 @@ class BuildHandler(BaseHandler):
             return
 
         if ref is None:
-            error_message = [f"Could not resolve ref for {key}. Double check your URL."]
+            error_message = [
+                f"Could not resolve ref for {key}. Double check your URL and that your repo is public."
+            ]
 
             if provider.name == "GitHub":
                 error_message.append(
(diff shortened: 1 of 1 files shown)


matthewfeickert:
(Rebased to bring in the pre-commit fix as a `fixup` to avoid extra commit noise. Sorry for any noise.)

matthewfeickert:
@consideRatio as it [looks like you're the current maintainer](https://github.com/jupyterhub/binderhub/pulse/monthly) this is passing CI and ready for review. :+1:

consideRatio (approved):
Thank you @matthewfeickert!! This makes sense to me!

PR #1665, opened 2023-04-26
ci: fix watch-dependencies workflow so it bumps docker and podman images

consideRatio:
I figure I'd fix the non-functional automation to bump these instead.

ci-binderhub

### DIFF
--- a/.github/workflows/watch-dependencies.yaml
+++ b/.github/workflows/watch-dependencies.yaml
@@ -26,13 +26,17 @@ jobs:
       matrix:
         include:
           - name: docker
-            registry: registry.hub.docker.com
+            registry: docker.io
             repository: library/docker
             values_path: dind.daemonset.image.tag
+            tag_prefix: ""
+            tag_suffix: -dind
           - name: podman
             registry: quay.io
             repository: podman/stable
             values_path: pink.daemonset.image.tag
+            tag_prefix: v
+            tag_suffix: ""
 
           # FIXME: After docker-image-cleaner 1.0.0 is released, we can enable
           #        this. So far, there isn't any available stable release, and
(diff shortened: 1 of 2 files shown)


helm-chart/binderhub/values.yaml
```diff
@@ -272,8 +272,8 @@ dind:
   initContainers: []
   daemonset:
     image:
-      name: docker
-      tag: 20.10.12-dind
+      repository: docker.io/library/docker
```
manics, line 275:
```suggestion
      image: docker.io/library/docker
```
`repository` is a better name for this property though, and matches other Helm charts

consideRatio: Wooops renaming the values property was a mistake

.github/workflows/watch-dependencies.yaml
```diff
@@ -71,17 +71,15 @@ jobs:
...
         run: |
           sed --in-place 's/tag: "${{ steps.local.outputs.tag }}"/tag: "${{ steps.latest.outputs.tag }}"/g' helm-chart/binderhub/values.yaml
 
       - name: git diff
-        if: steps.local.outputs.tag != steps.latest.outputs.tag
         run: git --no-pager diff --color=always
 
       # ref: https://github.com/peter-evans/create-pull-request
       - name: Create a PR
-        if: steps.local.outputs.tag != steps.latest.outputs.tag
```
consideRatio, line 81: A PR won't be created if there isn't a change anyhow.

consideRatio:
@manics I reworked this to be fixing the CI workflow watching these dependencies instead.

consideRatio:
Going for a merge as this is CI only and I've tested it!

PR #1669, opened 2023-04-27
log duration of health checks

minrk:
OVH is having trouble, in large part due to very slow health checks

but there's currently no way to tell what's being so slow

I have a _hunch_ that it's queued requests in the shared executor used to talk to k8s (and ultimately caused by the k8s API being slow), but that's just a guess. Could be the registry as well.

This should help us identify what's wrong, at least.

### DIFF
--- a/binderhub/health.py
+++ b/binderhub/health.py
@@ -30,6 +30,9 @@ def retry(_f=None, *, delay=1, attempts=3):
                         raise
                     else:
                         attempts -= 1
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
Tests are caused by k3s latest isn't so stable, because they upload artifacts a while after updating the release channel "latest".

PR #1679, opened 2023-04-28
build binderhub wheel outside docker

minrk:
with `python3 -m build .`

Avoids rebuilding the binderhub wheel in each Docker build stage, which takes ~10 minutes for each arm build on GHA.

emulated webpack takes ~10 minutes on GHA, so this should save quite a bit of time in the publish action, which is currently taking more than 20 minutes.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -39,9 +39,15 @@ jobs:
         with:
           # chartpress requires the full history
           fetch-depth: 0
+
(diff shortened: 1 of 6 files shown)


minrk:
docker build test is down to 8 minutes from 23.

minrk:
Self-merging to unblock mybinder.org deploys, which need to be finished for the GKE shutdown today.

PR #1680, opened 2023-04-28
Limit stacking jupyterhub API requests

minrk:
Without this, several API requests can be outstanding at the same time if things are getting slow

### DIFF
--- a/binderhub/health.py
+++ b/binderhub/health.py
@@ -25,13 +25,13 @@ def retry(_f=None, *, delay=1, attempts=3):
             while attempts > 0:
                 try:
                     return await f(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


consideRatio:
A cache is used otherwise? Makes sense

consideRatio (approved):

minrk:
Yes, the other health checks are already cached in this way. The at_most_every decorator does two things:

1. cache results for `interval` seconds, and
2. (more directly relevant) ensure only one check is outstanding at a time, returning the outstanding Future if there is already an outstanding call

I'm going to rewrite the health handlers to poll in the background instead of on request, so the health handler should always be responsive.

PR #1684, opened 2023-05-01
k3s latest: specify version (v1.26)

manics:
Using latest means things are liable to break when there's a new release. Although this happens infrequently when it does happen it either leads to wasted time, or PRs being merged without tests being completed.

If we want to automate the bump it could be added to the https://github.com/jupyterhub/binderhub/blob/main/.github/workflows/watch-dependencies.yaml workflow

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,6 +49,8 @@ jobs:
       contents: read
     env:
       GITHUB_ACCESS_TOKEN: "${{ secrets.github_token }}"
+      # Don't use latest, instead bump it using a PR so we know when a new version breaks BinderHub
+      K3S_LATEST: v1.26
 
     strategy:
       # keep running so we can see if tests with other k3s/k8s/helm versions pass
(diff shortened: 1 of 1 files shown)


manics:
The
```
on:
  pull_request:
    paths-ignore:
      ...
      - "!.github/workflows/test.yml"
```
exclusion obviously isn't working since this workflow hasn't run on this PR

manics:
Maybe exclusions can only be used with `paths` and not `paths-ignore`?
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

Though it's still not working

consideRatio (approved):
Yeah lets do automated bumps of the k3s version we test against.

consideRatio (changes requested):
The workflow was invalid, thats why its not running.

Screenshot_20230501-225311

consideRatio (approved):

manics:
Yeah... I realised that eventually. I'm going to merge this so we can try and fix the other issues. Incidentally I noticed some of the tests are incorrectly passing because `curl http://localhost:30901/health --max-time 5 --retry 5 --retry-delay 1 --retry-connrefused` doesn't return a non-zero exit code when `/heath` returns a HTTP error- something to be fixed separately.

PR #1683, opened 2023-05-01
launch_quota.total_quota

manics:
Fix for https://github.com/jupyterhub/binderhub/pull/1682#issuecomment-1530106672
(and therefore https://github.com/jupyterhub/mybinder.org-deploy/issues/2617)

This should be `total_quota` since `LaunchQuota` is platform agnostic
https://github.com/jupyterhub/binderhub/blob/0c3d8b23d396f1d0d276605356c8c329b2cc1376/binderhub/quota.py#L43-L54

PR #1686, opened 2023-05-01
test.yml: check curl status for HTTP error

manics:
By default curl doesn't return a non-zero exit code when it receives a HTTP error, e.g. even if `http://binderhub/health` returns a 500 error curl will exit with 0.

`--fail-with-body` should ensure it exits with an error code, whilst still showing the retrieved content (`-f` exits with an error code but doesn't display the content).


manics:
It makes more sense to bundle this with https://github.com/jupyterhub/binderhub/pull/1683

PR #1689, opened 2023-05-03
JS: Move the binderhub API client to its own npm package

yuvipanda:
Many projects want / need to connect to the BinderHub API:

- [thebe](https://thebe.readthedocs.io)
- [prototype UI for connecting to a private binderhub from profile_list](https://github.com/yuvipanda/prototype-kubespawner-dynamic-building-ui)

These all need to recreate a JS client interface to the BinderHub API.

This PR splits out the *existing* client API we have in the BinderHub JS codebase into its own package, and refers to that from the existing codebase. This has several advantages:

- Multiple projects can re-use this client code easily, without having to copy-paste.
- Separation of concerns allows us to eventually do wonderful things like **add unit tests** for the JS code.
- By using the npm [workspaces](https://docs.npmjs.com/cli/v9/using-npm/workspaces?v=true) feature, extra complexity overhead is very minimal. Note how small this PR actually is.

As added incentive, I've added doc strings for the methods in `BinderImage`. I will work on refactoring the JS code in there as well, but as a separate PR.

### DIFF
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -11,12 +11,14 @@ on:
       - ".eslintrc.js"
       - "package.json"
       - "binderhub/static/js/**"
+      - "js/**"
   push:
     paths:
       - ".github/workflows/eslint.yml"
       - ".eslintrc.js"
       - "package.json"
       - "binderhub/static/js/**"
+      - "js/**"
     branches-ignore:
       - "dependabot/**"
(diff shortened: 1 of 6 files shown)


manics (approved):

yuvipanda:
AAAAND published! https://www.npmjs.com/package/@jupyterhub/binderhub-client

PR #1691, opened 2023-05-03
Get rid of eslint GH workflow

yuvipanda:
This is handled by prettier in the pre-commit workflow instead


manics:
I thought prettier only formats code, whereas eslint also detects some likely errors?

PR #1694, opened 2023-05-03
JS: Rename BinderImage to BinderRepository

yuvipanda:
- Makes more sense, as we're representing a repository and then building it.
- Moves away from this being a default export, so we could export other functions and classes in the future. This is a 'breaking' change but nobody else uses this for now.

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -15,7 +15,7 @@ import { FitAddon } from 'xterm-addon-fit';
 import ClipboardJS from 'clipboard';
 import 'event-source-polyfill';
 
-import BinderImage from '@jupyterhub/binderhub-client';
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
Makes sense to me!

I didnt review this technically and isnt very familiar with the use of this code, so i'll let Simon go for a merge

consideRatio:
We have BuildExecutor Python classes doing work in the backend, is this class maybe representing an inidividual BuildExecution?

yuvipanda:
@consideRatioI don't think the name here should be based on the internal details of the python backend. This class also doesn't necessarily represent a single BuildExecution as EventSource by default will retry and can make multiple connections.  I think BinderRepository is the correct name here conceptually.

minrk (approved):

PR #1695, opened 2023-05-04
use python for local hook

### LINKED ISSUE #1690: pre-commit does not work for this repo if you do not use conda
### Bug description

When trying to use `pre-commit` on a machine with no `conda`, it does not work. The following exception occurs:

```
➜ g commit
[INFO] Installing environment for local.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('conda', 'env', 'create', '-p', '/Users/yuvipanda/.cache/pre-commit/repo2xt7fjkj/conda-default', '--file', 'environment.yml')
return code: 1
stdout:
    Executable `conda` not found
stderr: (none)
Check the log at /Users/yuvipanda/.cache/pre-commit/pre-commit.log
```

The script being run doesn't actually use anything conda specific, should just use whatever python pre-commit is already using.

minrk:
shouldn't need to use conda for a single Python dependency

closes #1690

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,7 +39,7 @@ repos:
     hooks:
       - id: update-values-based-on-binderspawner-mixin
         name: Update helm-chart/binderhub/values.yaml based on binderhub/binderspawner_mixin.py
-        language: conda
+        language: python
         additional_dependencies: ["ruamel.yaml"]
         entry: python ci/check_embedded_chart_code.py
         args:


consideRatio:
Does pre-commit run without erroring in pre-commit.ci if the relevant files are changed, where this hook is doing autoformatting?

yuvipanda:
YAY!

PR #1699, opened 2023-05-16
[docs] help with creating namespace

dwinston:


consideRatio:
Thank you @dwinston! Note that this could also be done with `--create-namespace` flag of the `helm install` command.

PR #1702, opened 2023-05-22
maint: fix broken test by upgrading to non-deprecated ruamel.yaml syntax

### LINKED ISSUE #1701: ci: indentation failure - why!?
We error in the k8s 1.26 / main test, because the built pod can fail to startup because of an indentation error in `singleuser.cmd`.

image

But we don't have an indentation error in the provided config:

https://github.com/jupyterhub/binderhub/blob/9cdae11b6671d87b4355534d3edeedfac0a26774/helm-chart/binderhub/values.yaml#L213-L231

Yet we have it [in the pod based on k8s namespace report](https://github.com/jupyterhub/binderhub/actions/runs/5042805566/jobs/9043834434?pr=1697#step:34:184):
image

---

Is the singleuser.cmd processed in any way by us or `helm` making this happen? Is variable expansion involved adding a space or similar?

consideRatio:
- Closes #1701

### DIFF
--- a/testing/local-binder-k8s-hub/install-jupyterhub-chart
+++ b/testing/local-binder-k8s-hub/install-jupyterhub-chart
@@ -10,7 +10,9 @@ import sys
 from subprocess import check_call
 from tempfile import NamedTemporaryFile
 
-from ruamel import yaml
+from ruamel.yaml import YAML
+
+yaml = YAML(typ="safe")
 
 here = os.path.abspath(os.path.dirname(__file__))
 helm_chart = os.path.join(here, os.pardir, os.pardir, "helm-chart")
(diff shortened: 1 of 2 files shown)


testing/local-binder-k8s-hub/install-jupyterhub-chart
```diff
@@ -10,7 +10,9 @@ import sys
 from subprocess import check_call
 from tempfile import NamedTemporaryFile
 
-from ruamel import yaml
```
consideRatio, line 13: This was deprecated syntax on how to use ruamel.yaml, and transitioning away from it like done elsewhere solved the issue.

PR #1711, opened 2023-05-30
maint: fix ruamel.yaml issue, flush needed now but not before

consideRatio:
Not sure why, but it seems now ruamel.yaml require us to flush content written to a file while we didn't before. Or? I'm quite confused, but our [test suite started failing](https://github.com/jupyterhub/binderhub/actions/workflows/test.yml?query=branch%3Amain), and now its resolved with this PR.

### DIFF
--- a/ci/check_embedded_chart_code.py
+++ b/ci/check_embedded_chart_code.py
@@ -17,7 +17,7 @@ import sys
 
 from ruamel.yaml import YAML
 
-yaml = YAML(typ="rt")
+yaml = YAML()
 yaml.preserve_quotes = True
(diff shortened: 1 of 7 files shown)


ci/check_embedded_chart_code.py
```diff
@@ -17,7 +17,7 @@
 
 from ruamel.yaml import YAML
 
-yaml = YAML(typ="rt")
+yaml = YAML()
```
consideRatio, line 20: The default, at least in a new version, is the `typ=rt` (round trip), and that hasn't caused issues to use.

I don't know what caused things to start failing, but I hope to not need to dig in and learn the details.

PR #1637, opened 2023-02-21
Support dynamic registry tokens, add wrapper for working with registry microservice

manics:
This is a companion for https://github.com/jupyterhub/repo2docker/pull/1245 that allows dynamic credentials t be used for pushing to a registry. This is required for e.g. ECR where the token used to login to the reigistry expires.

This also implements the non-vendor specific parts of
https://github.com/jupyterhub/binderhub/issues/1623

Closes https://github.com/jupyterhub/binderhub/issues/705

You can see an example implementation of the microservice side in
https://github.com/manics/binderhub-container-registry-helper
This is working with AWS ECR using an IRSA role for the microservice.

It's written in Go as I wanted to learn it, but if we choose to pursue this rewriting it in Python (or Javascript, or any other language) shouldn't be too difficult.

PR #1724, opened 2023-06-23
Support dynamic push credentials obtained from registry

manics:
Extracted from https://github.com/jupyterhub/binderhub/pull/1637
This is required to support time-limited registry credentials, as required by AWS ECR.

Required for https://github.com/jupyterhub/mybinder.org-deploy/issues/2629

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -87,18 +87,20 @@ class BuildExecutor(LoggingConfigurable):
 
     push_secret = Unicode(
         "",
-        help="Implementation dependent secret for pushing image to a registry.",
+        help="Implementation dependent static secret for pushing image to a registry.",
         config=True,
     )
 
-    push_secret_content = Unicode(
+    registry_credentials = Unicode(
         "",
         help=(
-            "Content of an implementation dependent secret for pushing image to a registry. "
-            "For example, if push tokens are temporary this can be used to pass the token "
-            "as an environment variable CONTAINER_ENGINE_REGISTRY_CREDENTIALS to "
-            "repo2docker."
-            "If provided this will be used instead of push_secret."
+            "Implementation dependent credentials for pushing image to a registry. "
+            "For example, if push tokens are temporary this could be used to pass "
+            "dynamically created credentials as an encoded JSON blob "
+            '`{"registry": "docker.io", "username":"user", "password":"password"}` '
+            "in the environment variable `CONTAINER_ENGINE_REGISTRY_CREDENTIALS` to "
+            "repo2docker. "
+            "If provided this will be used instead of push_secret. "
         ),
         config=True,
     )
@@ -243,7 +245,26 @@ class KubernetesBuildExecutor(BuildExecutor):
     # Overrides the default for BuildExecutor
     push_secret = Unicode(
         "binder-build-docker-config",
-        help="Implementation dependent secret for pushing image to a registry.",
+        help=(
+            "Name of a Kubernetes secret containing static credentials for pushing "
+            "an image to a registry."
+        ),
+        config=True,
+    )
+
+    registry_credentials = Unicode(
+        "",
+        help=(
+            "Implementation dependent credentials for pushing image to a registry. "
+            "For example, if push tokens are temporary this could be used to pass "
+            "dynamically created credentials as an encoded JSON blob "
+            '`{"registry": "docker.io", "username":"user", "password":"password"}` '
+            "in the environment variable `CONTAINER_ENGINE_REGISTRY_CREDENTIALS` to "
+            "repo2docker. "
+            "If provided this will be used instead of push_secret. "
+            "Currently this is passed to the build pod as a plan text environment "
+            "variable, though future implementations may use a Kubernetes secret."
+        ),
         config=True,
     )
 
@@ -415,11 +436,11 @@ class KubernetesBuildExecutor(BuildExecutor):
                 client.V1EnvVar(name="GIT_CREDENTIAL_ENV", value=self.git_credentials)
             )
 
-        if self.push_secret_content:
+        if self.registry_credentials:
             env.append(
                 client.V1EnvVar(
                     name="CONTAINER_ENGINE_REGISTRY_CREDENTIALS",
-                    value=self.push_secret_content,
+                    value=self.registry_credentials,
                 )
             )
         elif self.push_secret:
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -91,16 +91,16 @@ class BuildExecutor(LoggingConfigurable):
         config=True,
     )
 
-    registry_credentials = Unicode(
-        "",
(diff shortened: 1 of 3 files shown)


binderhub/build.py
```diff
@@ -394,7 +406,23 @@ def submit(self):
...
+        if self.git_credentials:
+            env.append(
+                client.V1EnvVar(name="GIT_CREDENTIAL_ENV", value=self.git_credentials)
+            )
+
+        if self.push_secret_content:
+            env.append(
+                client.V1EnvVar(
+                    name="CONTAINER_ENGINE_REGISTRY_CREDENTIALS",
+                    value=self.push_secret_content,
```
consideRatio, line 422: This is less secure than the `push_secret` option I think, because `kubectl get pod` can reveal content while when using a dedicated k8s Secret to house the content only `kubectl get secret` can reveal the content.

KubeSpawner has ended up dynamically creating also a k8s Secret for user pods to house misc information, for example info relevant for internal_ssl. It certainly adds complexity to do that, and this is far less complicated.

I think it may be acceptable to do this if we describe it clearly in the help string of the config.

binderhub/build.py
```diff
@@ -91,6 +91,18 @@ class BuildExecutor(LoggingConfigurable):
...
+        help=(
+            "Content of an implementation dependent secret for pushing image to a registry. "
+            "For example, if push tokens are temporary this can be used to pass the token "
+            "as an environment variable CONTAINER_ENGINE_REGISTRY_CREDENTIALS to "
+            "repo2docker."
+            "If provided this will be used instead of push_secret."
+        ),
+        config=True,
+    )
+
```
consideRatio, line 105: Overall, the `push_secret` config has been challenging for me to understand well. Here are some notes on what I think can help.

- Its not obvious that its a `push_secret` references a name (k8s Secret resource for example) as compared to content to put inside one.
- Its not obvious that its going to be created or needs to be created manually
- Its not obvious what kind of content to put in the referenced secret created manually

Thinking about what I'd appreciate from docs to understand this good enough, I'd say something like the following, where its not obvious whats in scope and not for this PR:

1. I think "implementation dependant" could be written more verbosively to clarify that: its dependant on the BuildExecutor implementation, and no base implemention exist.
2. `push_secret` and `push_secret_content`'s help strings are both are updated to reflect that there are separate ways of providing container registry credentials to the builder, where `push_secret` references something existing by name, and `push_secret_content` is content.
3. `push_secret` and `push_secret_content`'s help strings are both are updated on how to know what kind of resource to reference and what content to provide. This may be implementation dependent and not something BuildExecutor can know, but I think we can declare here where to find that kind of details.

binderhub/build.py
manics: I agree it's not as good as using a secret, however we had a long discussion in https://github.com/jupyterhub/binderhub/pull/1577 over the pros and cons of dynamically creating secrets attached to pods, as the same problem arises with dynamic Git credentials.

binderhub/build.py
manics: I'll update the docs. AFAIK it should be fine to duplicate the traitlet in the subclass to have more specialist documentation?

If it's easier (I'm not sure whether it is), we could switch `push_secret` to be of type `(Unicode, Dict)`. where Dict would for example take the form `{content: <secret>}` or `{secret_name: <k8s-secret-name>}`. `Unicode` would support the current default behaviour.

binderhub/build.py
```diff
@@ -87,7 +87,21 @@ class BuildExecutor(LoggingConfigurable):
...
+    registry_credentials = Dict(
+        {},
+        help=(
+            "Implementation dependent credentials for pushing image to a registry. "
+            "For example, if push tokens are temporary this could be used to pass "
+            "dynamically created credentials "
+            '`{"registry": "docker.io", "username":"user", "password":"password"}`. '
+            "This will be JSON encoded and passed in the environment variable "
+            "CONTAINER_ENGINE_REGISTRY_CREDENTIALS` to repo2docker. "
```
consideRatio, line 102:
```suggestion
            "BINDERHUB_REGISTRY_CREDENTIALS` to repo2docker. "
```

binderhub/build.py
```diff
@@ -231,7 +245,26 @@ def _default_api(self):
...
```
consideRatio, line 263:
```suggestion
            "BINDERHUB_REGISTRY_CREDENTIALS` to repo2docker. "
```

binderhub/build.py
```diff
@@ -394,7 +427,23 @@ def submit(self):
...
+        ]
+        if self.git_credentials:
+            env.append(
+                client.V1EnvVar(name="GIT_CREDENTIAL_ENV", value=self.git_credentials)
+            )
+
+        if self.registry_credentials:
+            env.append(
+                client.V1EnvVar(
+                    name="CONTAINER_ENGINE_REGISTRY_CREDENTIALS",
```
consideRatio, line 442:
```suggestion
                    name="BINDERHUB_REGISTRY_CREDENTIALS",
```

consideRatio (approved):
I've googled on `CONTAINER_ENGINE_REGISTRY_CREDENTIALS` and understand the name to be defined by us and not an env var name respected by another project already, right?

What do you think about transition to prefixing it with `BINDERHUB_` as this is the software project that defines them and sets them?

I've grown opinionated about this after seeing env vars respected by jupyter/docker-stacks images show up in z2jh and other places, where its very hard to understand they won't work for other images.

Reviewing this PR, I've googled the internet to find if `CONTAINER_ENGINE_REGISTRY_CREDENTIALS` and `GIT_CREDENTIAL_ENV` was used elsewhere, for example is `GIT_CREDENTAL_ENV` an env variable respected by git? I suspected it was because of its name, but [it probably isn't as its not listed in git documentation](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables). If it was named `BINDERHUB_GIT_CREDENTIALS` it would be obvious to me `git` wouldn't respect them, so then I could understand that it had to be consumed by something else. Searching in repo2docker, it seems that it isn't respected by that either but that there was some legacy notes about it leaving me confused if it will be respected by repo2docker.

I'd like to see us not lock into defining another env variable name that isn't obviously defined by this project by not prefixing it, but besides that I think we should merge this.

manics:
Both of these environment variables are used by repo2docker:
- https://github.com/jupyterhub/repo2docker/blob/247e9535b167112cabf69eed59a6947e4af1ee34/docker/git-credential-env
- https://github.com/jupyterhub/repo2docker/blob/247e9535b167112cabf69eed59a6947e4af1ee34/repo2docker/engine.py#L148-L177

In hindsight prefixing them with `R2D_` would've been a good idea and makes sense for future additions. We could add new `R2D_` env vars and check for both old and new, but that risks adding to the confusion.

consideRatio (approved):

PR #1740, opened 2023-09-14
Bump and publish @jupyterhub/binderhub-client

yuvipanda:
Brings in https://github.com/jupyterhub/binderhub/pull/1694

There are currently no real external users of this package, except for
github.com/yuvipanda/prototype-kubespawner-dynamic-building-ui (which
I'm working on again now). I think the API will change around a bit as this gets
worked on.


yuvipanda:
Let's wait a bit actually.

PR #1741, opened 2023-09-15
Use URLSearchParams() to construct query strings

yuvipanda:
This is cleaner, and also allows us to support more query parameters in the future (when
https://github.com/jupyterhub/binderhub/pull/1647
lands)


GeorgianaElena (approved):
Thanks @yuvipanda!

PR #1748, opened 2023-09-24
local-binder-local-hub: Set min versions in requirements

manics:
NullAuthenticator is replaced by the built-in `null`.

I've set minimum versions of the requirements to more closely reflect what's we're likely to test locally.

### DIFF
--- a/testing/local-binder-local-hub/requirements.txt
+++ b/testing/local-binder-local-hub/requirements.txt
@@ -1,4 +1,3 @@
-dockerspawner
-jupyter-repo2docker
-jupyterhub
-nullauthenticator
+dockerspawner>=12
+jupyter-repo2docker>=2023.06.0
+jupyterhub>=3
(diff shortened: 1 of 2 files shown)


testing/local-binder-local-hub/requirements.txt
```diff
@@ -1,4 +1,3 @@
-dockerspawner
-jupyter-repo2docker
-jupyterhub
-nullauthenticator
+dockerspawner>=12
```
manics, line 1: The first version that's greater than 0.x.x

testing/local-binder-local-hub/requirements.txt
```diff
@@ -1,4 +1,3 @@
...
+jupyter-repo2docker>=2023.06.0
```
manics, line 2: Older versions will fail on some R repos since MRAN no longer works,

consideRatio (approved):

Issue #1747, opened 2023-09-24
Can't launch repos in mybinder.org UI: `Uncaught TypeError: URL constructor: /build/gh/binder-examples/conda/HEAD is not a valid URL.`


manics:
### Bug description

I had to revert https://github.com/jupyterhub/mybinder.org-deploy/pull/2746 (which includes https://github.com/jupyterhub/binderhub/pull/1741) as it causes a Javascript error on the main mybinder.org page. This seems to prevent launching any repository.

### How to reproduce

The occurred after entering `https://github.com/binder-examples/conda` into mybinder.org (deployment has since been reverted and is working again). It failed to launch, and this error was in the browser console:
```
Uncaught TypeError: URL constructor: /build/gh/binder-examples/conda/HEAD is not a valid URL.
```
The browser debugger indicated
https://github.com/jupyterhub/binderhub/blob/527ac45a377d2470dfa101823fbd8d3a00193bc5/js/packages/binderhub-client/lib/index.js#L28
is the source

![image](https://github.com/jupyterhub/mybinder.org-deploy/assets/1644105/535fceb6-58f5-44fc-885c-ed4ef8ebbab9)

Note this is not reproducible using the test configurations in https://github.com/jupyterhub/binderhub/tree/main/testing

yuvipanda:
Sorry about missing this @manics. I've now opened https://github.com/jupyterhub/binderhub/pull/1758, which starts adding JS unit tests here as well. There's going to be a bit of work needed before the unit tests can 'fully' catch things, but it's a start.

yuvipanda:
This has since been fixed

PR #1755, opened 2023-10-02
Revert "Use URLSearchParams() to construct query strings"

manics:
Reverts jupyterhub/binderhub#1741

This code isn't tested by our CI, and resulted in a failure on mybinder.org: https://github.com/jupyterhub/binderhub/issues/1747 which now blocks further BinderHub updates.

This syntax can be seen to be invalid in a browser javascript console:
```
>> new URL("/" + "build/")
Uncaught TypeError: URL constructor: /build/ is not a valid URL.
```

### DIFF
--- a/js/packages/binderhub-client/lib/index.js
+++ b/js/packages/binderhub-client/lib/index.js
@@ -25,13 +25,10 @@ export class BinderRepository {
    * Call the BinderHub API
    */
   fetch() {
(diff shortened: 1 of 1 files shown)


yuvipanda:
Hmm, not sure how this worked when I tested this locally?!

yuvipanda:
Ugh, I can see how this worked when I tested this locally but failed on mybinder.org :( I'll put up another change.

PR #1756, opened 2023-10-02
Z2JH 3.1.0

manics:
It's about time we updated this :smile:


consideRatio (approved):
Test failures are unrelated i think, thanks @manics - no breaking change expected, but an important kubespawner bugfix!

Issue #1754, opened 2023-10-01
Test failures in main branch


consideRatio:
I've not looked into this further, just going through every repo in jupyterhub merging dependabot updates atm.

manics:
```
2023-10-18T07:54:57.4452270Z ----------------------------- Captured stdout call -----------------------------
2023-10-18T07:54:57.4452943Z built: Found built image, launching...
2023-10-18T07:54:57.4453471Z launching: Launching server...
2023-10-18T07:54:57.4453927Z launching: Server requested
2023-10-18T07:54:57.4455703Z launching: 2023-10-18T07:53:34.021076Z [Warning] 0/1 nodes are available: 1 Insufficient memory. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod..
```
Maybe we aren't cleaning up properly?

PR #1758, opened 2023-10-03
JS: Add unit tests (with coverage) & use `URL` object instead of strings to construct some URLs

yuvipanda:
- Re-adds https://github.com/jupyterhub/binderhub/pull/1741, but with [Jest](https://jestjs.io/) based unit tests. Jest seems fairly popular and is also what JupyterLab & JupyterHub admin uses, so it feels appropriate here.
- Adds a small unit test for the binderhub frontend too (for `updateFavicon`), so we can slowly add more
  tests there. This actually helped catch a bug in `updateFavicon` - we were unconditionally appending
  `<link>` tags to `<head>`, so the number of tags grew each time the icon changed. There is now a unit
   test to catch this.
- Use typed URL objects to construct build URLs, rather than just doing string concatenation. I'll try to slowly move all URL manipulation to using URL object
- Use typed URL objects to construct the URL we eventually redirect users to, and add unit tests. Some cleanup here to make this actually unit testable rather than test side effects (`window.location`)
- Uploads JS test coverage stats to codecov as well, so we can track and improve our JS coverage
- Remove implicit dependency of `binderhub-client` on jquery for URL parameter manipulation.
- This still isn't enough to actually fully catch the issues of 1741 yet, so here's  a screen recording of me testing it via the  UI (this includes the changes in #1761 as well)

https://github.com/jupyterhub/binderhub/assets/30430/2f0c7423-13cc-4f12-9d74-7e186c42628e

### DIFF
--- a/.github/workflows/jest.yml
+++ b/.github/workflows/jest.yml
@@ -6,6 +6,7 @@ on:
     paths: &paths
       - "binderhub/static/js/**"
       - "js/packages/binderhub-client/**"
+      - ".github/workflows/jest.yml"
   push:
     paths: *paths
     branches-ignore:
--- a/.github/workflows/jest.yml
+++ b/.github/workflows/jest.yml
@@ -1,5 +1,5 @@
 # Runs jest based unit tests for the binderhub-client JS package
-name: eslint
+name: jest
 
 on:
   pull_request:
--- a/.github/workflows/jest.yml
+++ b/.github/workflows/jest.yml
@@ -1,0 +1,28 @@
+# Runs jest based unit tests for the binderhub-client JS package
+name: eslint
+
+on:
+  pull_request:
+    paths:
+      - "js/**"
+  push:
+    paths:
+      - "js/**"
+    branches-ignore:
+      - "dependabot/**"
+      - "pre-commit-ci-update-config"
+      - "update-*"
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - run: |
+        cd js/packages/binderhub-client
+        npm install
+
+      - run: |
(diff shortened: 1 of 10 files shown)


yuvipanda:
The test failures seem unrelated

.github/workflows/jest.yml
```diff
@@ -0,0 +1,32 @@
+# Runs jest based unit tests for frontend javascript and @jupyterhub/binderhub-client
+name: "JS Unit tests"
+
+on:
+  pull_request:
+    paths: &paths
```
manics, line 6: This workflow hasn't been run in this PR yet.
```suggestion
    paths: &paths
      - .github/workflows/jest.yml
```

yuvipanda: Hmm, I added this but it doesn't seem to be running still. It was running before (https://github.com/jupyterhub/binderhub/actions/runs/6387299049/job/17335421212)

yuvipanda: Ah because github actions doesn't support YAML anchors https://github.com/actions/runner/issues/1182

yuvipanda:
JS tests are running here again, and looks like the failures are unrelated

manics (approved):
I'm pretty sure the test failures are unrelated....we'll need to fix them at some point though
https://github.com/jupyterhub/binderhub/issues/1754

yuvipanda:
Yep, given the same tests are failing for https://github.com/jupyterhub/binderhub/pull/1766 as well, I'm pretty sure they're unrelated.

consideRatio:
+1 to disregard the test failure atm.

I think its good if the title reflects that its more than added tests in this PR though, and a label is added as well.

yuvipanda:
@consideRatio good call! I've updated the title and added some labels.

PR #1759, opened 2023-10-03
Enable prettier on JS & CSS as well

yuvipanda:
While the pre-existing comment suggested using 4-space indents, prettier ships by default with 2 space indents for JS & CSS. I looked at the JS code shipped in JupyterHub as well as JupyterLab, and found they are also using 2 space indents. So I would suggest we just use two space indents here.


yuvipanda:
Nah, I'll come back to this later.

Issue #777, opened 2019-01-25
Refactor index.js megafile into mutliple files


captainsafia:
To aid with our conversion to React and to ease front-end development on Binder (https://github.com/jupyterhub/binderhub/issues/774), we'll need to refactor the `index.js` mega file into multiple files with discrete functions. The goal here is to move all the functions into files that will roughly map to React components or library functions in the future.

### Steps

1. Pick one of the tasks in the checklist below.
2. Move the function/functions into a new file within the `src` directory.
3. Add any helper functions or adjust function definitions/body to make it easy for the relocated functions to work independently of jQuery.
4. Export the necessary and functions and use them in `index.js`.

### Tips
- Where possible, opt for passing a value into a function as a parameter instead of extracting it within the function body using jQuery selectors.
- When possible, use functions that are part of standard JS instead of jQuery.
- If possible, address some of the FIXMEs left in the code if they are small and isolated.
- Make sure that all functions and variables are in camelCase, the JavaScript function naming convention.

### Checklist

Here's a checklist of the functions/components that need to be moved out into separate files. This list isn't complete yet but is enough to get the ball rolling.

- [x] Move Image class into `image.js` (https://github.com/jupyterhub/binderhub/pull/776)
- [ ] Move `setUpLog` into `log.js`
- [ ] Move `updateUrls` and `v2url` to `url.js`
- [ ] Move `getBuildFormValues` to `form.js`. The function should take the jQuery form element as a parameter and extract the individual form fields from the children within.
- [ ] Move constants in constants.js, export each individually, and import them where needed
- [x] Move `markdownBade` and `rstBadge` functions into badge.js (https://github.com/jupyterhub/binderhub/pull/778)
- [x] Move `getPathType` and `updatePathText` to `path.js`. (https://github.com/jupyterhub/binderhub/pull/778)
- [ ] Move `updateRepoText` to `repo.js`

cc: @yuvipanda  @choldgraf  @betatim

captainsafia:
> To be clear - this is before any reactification (reacting? reaction?) happens, right?

Yep. No React-ification. Just moving and refactoring things so it is easier to Reactify  down the line.

yuvipanda:
https://github.com/jupyterhub/binderhub/pull/1761 fixes this, following the plan @captainsafia laid out :)

yuvipanda:
It only took about 4+ years, but this is finally done :) Thanks for the guidance, @captainsafia!

PR #1761, opened 2023-10-03
Split big index.js file into multiple smaller files

yuvipanda:
- Follows the plan laid out in
https://github.com/jupyterhub/binderhub/issues/777
- Only non-trivial move is to cleanup the `_config` call to
  use the more modern 'fetch' API rather than the older
  XMLHttpRequest API

Fixes https://github.com/jupyterhub/binderhub/issues/777

A screen recording of me testing this is available in
https://github.com/jupyterhub/binderhub/pull/1758

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -10,14 +10,11 @@
   pushing -> built
   pushing -> failed
 */
-import { Terminal } from 'xterm';
-import { FitAddon } from 'xterm-addon-fit';
 import ClipboardJS from 'clipboard';
 import 'event-source-polyfill';
 
 import { BinderRepository } from '@jupyterhub/binderhub-client';
-import { makeBadgeMarkup } from './src/badge';
-import { getPathType, updatePathText } from './src/path';
+import { updatePathText } from './src/path';
 import { nextHelpText } from './src/loading';
 import { updateFavicon } from './src/favicon';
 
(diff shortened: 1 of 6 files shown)


choldgraf:
omg is it finally happening?!

binderhub/static/js/src/repo.js
```diff
@@ -0,0 +1,43 @@
...
+    const configUrl = BASE_URL + "_config";
+    fetch(configUrl).then(resp => {
+      resp.json().then(data => {
+        configDict = data
+        setLabels();
+      });
+    })
+  } else {
+    setLabels();
+
```
consideRatio, line 41:
```suggestion
```

binderhub/static/js/src/urls.js
```diff
@@ -0,0 +1,53 @@
+import { makeBadgeMarkup } from './badge';
+import { getBuildFormValues } from './form';
+import { BADGE_BASE_URL, BASE_URL } from './constants';
+
+ function v2url(providerPrefix, repository, ref, path, pathType) {
```
consideRatio, line 5:
```suggestion
function v2url(providerPrefix, repository, ref, path, pathType) {
```

binderhub/static/js/index.js
```diff
@@ -29,121 +26,12 @@ import 'bootstrap/dist/css/bootstrap.min.css';
...
-  $("label[for=repository]").text(text);
-}
-
-function updateRepoText() {
-  if (Object.keys(config_dict).length === 0){
-    loadConfig(function(res) {
-      config_dict = JSON.parse(res);
-      setLabels();
-    });
-  } else {
```
consideRatio, line 87: This section previously looked like:

```js
  if (Object.keys(configDict).length === 0) {
    const configUrl = BASE_URL + "_config";
    fetch(configUrl).then(resp => {
      resp.json().then(data => {
        configDict = data
        setLabels();
      });
    })
  }
```

The loadConfig function now in repo.js is unchanged, looking like below with XMLHttpRequest being used still. I understand it as you've inlined the function into updateRepoText, and that updateRepoText was the only place where loadConfig was used, so I figure we should delete it.

```js
function loadConfig(callback) {
  const req = new XMLHttpRequest();
  req.onreadystatechange = function() {
    if (req.readyState == 4 && req.status == 200)
      callback(req.responseText)
  };
  req.open('GET', BASE_URL + "_config", true);
  req.send(null);
}
```

binderhub/static/js/src/repo.js
```diff
@@ -0,0 +1,43 @@
...
+  const req = new XMLHttpRequest();
+  req.onreadystatechange = function() {
+    if (req.readyState == 4 && req.status == 200)
+      callback(req.responseText)
+  };
+  req.open('GET', BASE_URL + "_config", true);
+  req.send(null);
+}
+
+
```
consideRatio, line 15:
```suggestion
```

I think this is removable at this point, motivated in another comment

consideRatio (commented):
Thank you for working this @yuvipanda!!!

LGTM except a few (I think) easy-to-resolve comments!

binderhub/static/js/index.js
yuvipanda: Good catch. done

yuvipanda:
Thanks @consideRatio. I've applied your suggested changes.

I think after this gets merged, I'll enable prettier for JS in this repo as well, so that should handle all other formatting inconsistencies.

binderhub/static/js/src/repo.js
```diff
@@ -1,18 +1,5 @@
...
-      callback(req.responseText)
-  };
-  req.open('GET', BASE_URL + "_config", true);
-  req.send(null);
-}
-
-
 function setLabels() {
   const provider = $("#provider_prefix").val();
   const text = configDict[provider]["text"];
```
consideRatio, line 5: This commit removed `configDict = {}`, but that is seen referenced via `setLabels`. I think something is going to go wrong

yuvipanda: Ah yes, fixed. I thought I had gotten rid of it, but forgotten I hadn't actually gotten rid of all the places it's used :) coming in a different PR!

yuvipanda: @consideRatio also I'm not sure if you caught it or the linter did, but eslint would've caught this one as well. So yay for more linting and testing.

consideRatio:
Wieeee thank you @yuvipanda for taking the time to this!!! :heart: :tada:

Issue #1765, opened 2023-10-04
Allow passing additional args to repo2docker


yuvipanda:
There's no way for us to configure binderhub to allow passing additional args to repo2docker. This is helpful when used with https://github.com/jupyterhub/binderhub/pull/1647 for example - I wanna pass `repo_dir` to be `/srv/repo`, as when used with a persistent home directory $HOME is overriden with persistent directory. Instead of adding each possible option, we should just allow a passthrough.

This is only configured by admins setting up binderhub, not end users.

PR #1766, opened 2023-10-06
Add `BuildExecutor.repo2docker_extra_args` config

yuvipanda:
Very helpful for a specific binderhub instance to customize how images are built. Off the top of my head, two useful things would be to pass `--repo-dir` (so you can have this work with a persistent home directory but still get access to the files in the image), and `--base-image` (so an install can use a different base image than rep2docker default).

As it's a simple passthrough, I've added a unit test only (no integration test).

Fixes https://github.com/jupyterhub/binderhub/issues/1765

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -16,7 +16,7 @@ import kubernetes.config
 from kubernetes import client, watch
 from tornado.ioloop import IOLoop
 from tornado.log import app_log
-from traitlets import Any, Bool, Dict, Integer, Unicode, default
+from traitlets import Any, Bool, Dict, Integer, List, Unicode, default
 from traitlets.config import LoggingConfigurable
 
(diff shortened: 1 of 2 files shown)


yuvipanda:
Failures are unrelated I think

manics (approved):
My original thinking behind the  `get_r2d_cmd_options()` function was that you could easily override it by subclassing `BuildExecutor`.

Adding a new parameter seems fine though, especially if this means we can say we won't add traitlets to pass other repo2docker arguments, and instead refer people to this new property instead.

consideRatio:
Lgtm!

I tweaked the title to "Add `repo2docker_extra_args` config" as it by itself clarifies the mechanism to allow passing args and that then enables the changelog to provide a better overview. I also added the "new" label for this PR!

PR #1768, opened 2023-10-09
Make eslint understand modern .js with babel's parser

consideRatio:
With this, we can use prettier in pre-commit without getting errors stemming from the formatting changes. Previously we did because the eslint's .js parser was not modern enough to support the formatting changes.

By switching to use the `@babel/eslint-parser` we are fine for now at least, and I think in the future as well, because we end up using a parser that is provided by babel, which is also what we use to transform the .js into something else.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,16 +2,14 @@ module.exports = {
     "env": {
         "browser": true,
         "jquery": true,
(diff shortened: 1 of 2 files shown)


yuvipanda:
Thanks for sorting this out, @consideRatio

PR #1769, opened 2023-10-09
pre-commit: let prettier format .js files

consideRatio:
A one line change in pre-commit-config -> autoformatting made by pre-commit.ci.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,15 +1,11 @@
 module.exports = {
-    "env": {
-        "browser": true,
-        "jquery": true,
-        "jest/globals": true,
-    },
-    "extends": [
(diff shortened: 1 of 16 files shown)


yuvipanda:
Yep, let's wait for the one I currently have brewing and then we can merge this. Will definitely help me avoid merge conflicts.

Thanks @consideRatio

consideRatio:
> let's wait for __the one__ I currently have brewing and then we can merge this

This was #1771 that is now merged I think

yuvipanda:
@consideRatio nope, there's one more coming (#1771 was a part of it).

yuvipanda:
Actually let's just do this, I can rebase. Thanks @consideRatio

Issue #774, opened 2019-01-23
Long term planning for the Frontend


yuvipanda:
Currently, we have one JS file + 1 HTML file (rendered with Jinja2) that does all the frontend work. This includes:

1. Client for the BinderHub API
2. Main page (when people see mybinder.org)
3. Loading page (when people go to a mybinder.org/v2/gh/user/repo launch URL)

(2) and (3) are particularly intertwined - they're restyled versions of the same page (HTML + JS). 

In the future, I'd like us to improve on the following aspects:

1. JS Tests (we have none)
2. Export a BinderHub client API library that others can re-implement. Right now, everyone does their own implementation, and particularly run into problems against eventsource retries.
3. Implement both v2 (current) and newer API (whenever #358 lands)
4. Split the main & loading pages
5. Make main page a lot more dynamic - ideally showcasing useful / featured repos & other forms of discovery
6. Better (privacy conscious) frontend instrumentation to see how people use the main page & how we can improve it.

Doing a refactor of the current JS in some form or way while it's still smallish (~400 lines) seems like a good idea to me!

## Factors

Some factors to consider:

1. How much is it used inside the Jupyter community? We want to make it as easy as possible for other Jupyter contributors to work on this.
2. Does it help bring new people with different skillsets into the project? We should balance this against (1) - we wanna diversify who comes in and contributes to the project, and specific tech might help / hinder that.
3. How much is it used in the frontend dev world? We wanna use an approach that's likely to be well debugged & with a lot of resources out there on how to use it.
4. Is it appropriately scoped for our size? We shouldn't overcomplicate our lives by choosing tools meant for 100k lines projects while ours is much smaller.
5. Does it make working on the frontend a happy & joyful thing for many people in the long run? Often projects end up with convoluted untested frontend code that is hard to touch since it might break.

## Options

Here are some options

### Option 1: React + TypeScript

We can switch to using TypeScript as a language & use React for frontend components + tests. 

### Option 2: React + JS

Similar to above, but with just plain JS (JSX) instead of TypeScript.

### Option 3: Next.js

Recommended by @betatim: https://nextjs.org/docs

### Option 4: No React

We'll refactor and move away from monolithic single JS file, but keep using our current setup of pure DOM manipulation rather than use a library.

There are probably more options than this, but this should be a starting point to discuss!

yuvipanda:
Fun fact - when I first started writing BinderHub, the very first line of JS I wrote was:

```
[...]

PR #1771, opened 2023-10-09
JS: Remove dead state transition validation code

yuvipanda:
Soooo, because `this.state` was initialised to `null` and not 'start', this code was always actually dead.
`validateStateTransition` had no branch for `oldState` being null, so always returned `false`. This meant `this.state` was *never* set to current state, and so in all callbacks, `oldState` was always `null`!

I validated this again by console.logging `oldState` and `newState` inside `validateStateTransition`. Here's the output for when a full build is performed:

```
null -> waiting
null -> fetching
null -> unknown
null -> building
null -> failed
null -> built
null -> launching
```

And for when launching a previously built image
```
null -> launching
null -> ready
```

I also put one just above where `this.state` is set, and that code was never called.

I was hoping to remove all this code anyway, as it doesn't actually validate nor test anything - that's really done by unit tests and integration tests. I thought this would hide bugs - but turns out, it was a bug itself, and hid nothing.

This also changes the signature of the callback for `onChangeState`, dropping `oldState` and `newState`. `oldState` was always `null`, and in our code, we actually don't use `newState` *at all*. The one place that was conditional on `oldState` being null is now unconditional, because `oldState` was always `null`.

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -1,14 +1,4 @@
 /* If this file gets over 200 lines of code long (not counting docs / comments), start using a framework
-  State transitions that are valid are:
-  start -> waiting
-  start -> built
-  start -> failed
-  waiting -> building
-  waiting -> failed
-  building -> pushing
-  building -> failed
-  pushing -> built
-  pushing -> failed
 */
 import ClipboardJS from 'clipboard';
 import 'event-source-polyfill';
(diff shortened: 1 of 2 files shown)


yuvipanda:
This is one way to get to 100% unit test coverage - just find code that was never used and get rid of it :)

consideRatio (approved):

consideRatio:
Wieee thank you for working this @yuvipanda!

binderhub/static/js/index.js
```diff
@@ -92,16 +82,14 @@ function build(providerSpec, log, fitAddon, path, pathType) {
...
 
-  image.onStateChange('built', function(oldState) {
-    if (oldState === null) {
-      $('#phase-already-built').removeClass('hidden');
-      $('#phase-launching').removeClass('hidden');
-    }
+  image.onStateChange('built', function() {
+    $('#phase-already-built').removeClass('hidden');
+    $('#phase-launching').removeClass('hidden');
     $('#phase-launching').removeClass('hidden');
```
manics, line 88: This line is a duplicate

consideRatio: Thanks for revieiwng @manics! I opened https://github.com/jupyterhub/binderhub/pull/1772 as I was too quick on the ball

PR #1773, opened 2023-10-10
JS: Use async generator instead of callbacks + add tests

yuvipanda:
JS now also supports writing async code with [Asnc Generators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator) which work very much like how async code would in Python. This allows us to write code that flows better, is more familiar to python folks, and avoids messy callback hell.

This code is a *straight* port of the existing callback based code to use an Async Generator instead. The EventIterator library is a very helpful convenience function here, allowing us to translate the callbacks from EventSource into an async generator without having to write a queue of promises ourselves. The iterator is stopped exactly the same way the callbacks were earlier stopped (whenever .stop() is called).

All the callbacks have been switched over to a switch / case statement instead. We can get cleaner and use exceptions to handle failures, instead of the 'failure' case, but that can come in a later PR so this is a purely mechanical refactoring from one async method to another.

An appropriate test for this is also added. To fully test the EventSource flow, we create a temporary web server that will serve a real captured eventsource response, with 1ms gaps between lines to fully simulate how this would work in reality. This gives us much more confidence about how this would work than mocking.

With this, @jupyterhub/binder-client now has 100% unit test coverage!

Here is a video of me testing this. I tested both something that worked, and an image that is known to fail (RStudio on ARM).

https://github.com/jupyterhub/binderhub/assets/30430/2ce22011-ee53-4c27-b9b8-1ee27ba13c63

### DIFF
--- a/binderhub/static/js/index.js
+++ b/binderhub/static/js/index.js
@@ -22,7 +22,7 @@ import { BASE_URL } from "./src/constants";
 import { getBuildFormValues } from "./src/form";
 import { updateRepoText } from "./src/repo";
 
-function build(providerSpec, log, fitAddon, path, pathType) {
+async function build(providerSpec, log, fitAddon, path, pathType) {
   updateFavicon(BASE_URL + "favicon_building.ico");
   // split provider prefix off of providerSpec
   const spec = providerSpec.slice(providerSpec.indexOf("/") + 1);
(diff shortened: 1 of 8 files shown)


binderhub/static/js/index.js
```diff
@@ -52,67 +52,67 @@ function build(providerSpec, log, fitAddon, path, pathType) {
...
-    // If we fail for any reason, show an error message and logs
-    updateFavicon(BASE_URL + "favicon_fail.ico");
-    log.show();
-    if ($("div#loader-text").length > 0) {
-      $("#loader").addClass("error");
-      $("div#loader-text p.launching").html(
-        "Error loading " + spec + "!<br /> See logs below for details.",
-      );
+    switch (data.phase) {
+      case "waiting":
```
consideRatio, line 65: Apparently the content under case should be put in a curly bracket

https://docs.w3cub.com/eslint/rules/no-case-declarations

yuvipanda: oooh, fun lexical scope thingies. I'll add.

yuvipanda: Fixed

yuvipanda:
I added a screen recording of me testing this as well.

binderhub/static/js/index.js
```diff
@@ -170,21 +177,21 @@ function indexMain() {
     return false;
   });
 
-  $("#build-form").submit(function () {
+  $("#build-form").submit(async function (e) {
+    e.preventDefault();
```
yuvipanda, line 181: This was necessary as `return false` doesn't work from async functions. Same effect

consideRatio: I figure no - that the event to submit this form isnt considered by parent elements, but I'm not confident reviewing this in general.

yuvipanda: I don't think it's relevant.

consideRatio (approved):

consideRatio:
I'm not soo comfortable reviewing the JS code, but I think this makes sense and I didn't spot any issues are carefully going through this PR.

Thank you @yuvipanda for working this! This seems great!!

PR #1774, opened 2023-10-11
Do not escape query included in url launch parameter

manics:
I'm not sure if this test is correct, but I think it illustrates the bug in
https://github.com/jupyterhub/nbgitpuller/issues/329
https://discourse.jupyter.org/t/nb-on-binder-with-parameters-now-resulting-in-404/21860
```
 FAIL  js/packages/binderhub-client/lib/index.test.js
  ● Get full redirect URL and deal with query and encoded query (with pathType=url)

    expect(received).toBe(expected) // Object.is equality

    Expected: "https://hub.test-binder.org/user/something/endpoint?a=1/2&b=3%3F%2F&token=token"
    Received: "https://hub.test-binder.org/user/something/endpoint%3Fa=1/2&b=3%3F%2F?token=token"

      179 |       )
      180 |       .toString(),
    > 181 |   ).toBe(
          |     ^
      182 |     "https://hub.test-binder.org/user/something/endpoint?a=1/2&b=3%3F%2F&token=token",
      183 |   );
      184 | });

      at Object.toBe (js/packages/binderhub-client/lib/index.test.js:181:5)
```

### DIFF
--- a/js/packages/binderhub-client/lib/index.js
+++ b/js/packages/binderhub-client/lib/index.js
@@ -118,30 +118,30 @@ export class BinderRepository {
     // Make a copy of the URL so we don't mangle the original
     let url = new URL(serverUrl);
     if (path) {
-      // strip trailing / from URL
-      url.pathname = url.pathname.replace(/\/$/, "");
-
+      // Ensure there is a trailing / in serverUrl
+      if (!url.pathname.endsWith("/")) {
+        url.pathname += "/";
+      }
       // trim leading '/' from path to launch users into
       path = path.replace(/(^\/)/g, "");
 
       if (pathType === "lab") {
         // The path is a specific *file* we should open with JupyterLab
-
         // trim trailing / on file paths
         path = path.replace(/(\/$)/g, "");
 
         // /doc/tree is safe because it allows redirect to files
-        url.pathname = url.pathname + "/doc/tree/" + encodeURI(path);
+        url = new URL("doc/tree/" + encodeURI(path), url);
       } else if (pathType === "file") {
         // The path is a specific file we should open with *classic notebook*
 
         // trim trailing / on file paths
         path = path.replace(/(\/$)/g, "");
-        // /tree is safe because it allows redirect to files
-        url.pathname = url.pathname + "/tree/" + encodeURI(path);
+
+        url = new URL("tree/" + encodeURI(path), url);
       } else {
         // pathType is 'url' and we should just pass it on
-        url.pathname = url.pathname + "/" + path;
+        url = new URL(path, url);
       }
     }
 
--- a/js/packages/binderhub-client/tests/index.test.js
+++ b/js/packages/binderhub-client/tests/index.test.js
@@ -233,3 +233,22 @@ describe(
   },
   10 * 1000,
 );
+
+test("Get full redirect URL and deal with query (with pathType=url)", () => {
+  const br = new BinderRepository(
+    "gh/test/test",
+    new URL("https://test-binder.org/build"),
+  );
+  expect(
+    br
+      .getFullRedirectURL(
+        "https://hub.test-binder.org/user/something/",
+        "token",
+        "endpoint?a=1/2&b=3",
+        "url",
+      )
+      .toString(),
+  ).toBe(
+    "https://hub.test-binder.org/user/something/endpoint?a=1/2&b=3&token=token",
+  );
+});
--- a/js/packages/binderhub-client/tests/index.test.js
+++ b/js/packages/binderhub-client/tests/index.test.js
@@ -234,7 +234,7 @@ describe(
   10 * 1000,
 );
 
-test("Get full redirect URL and deal with query (with pathType=url)", () => {
+test("Get full redirect URL and deal with query and encoded query (with pathType=url)", () => {
   const br = new BinderRepository(
     "gh/test/test",
     new URL("https://test-binder.org/build"),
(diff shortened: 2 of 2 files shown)


js/packages/binderhub-client/lib/index.test.js
```diff
@@ -163,3 +163,22 @@ test("Get full redirect URL and deal with excessive slashes (with pathType=file)
...
+  const br = new BinderRepository(
+    "gh/test/test",
+    new URL("https://test-binder.org/build"),
+  );
+  expect(
+    br
+      .getFullRedirectURL(
+        "https://hub.test-binder.org/user/something/",
+        "token",
+        "endpoint?a=1/2&b=3%3F%2F",
```
manics, line 177:
```suggestion
        "endpoint?a=1%2F2&b=3%3F%2F",
```

manics:
First of all was my test definitely correct? I wasn't completely sure whether the `path` (including query) argument of `getFullRedirectURL` was meant to be encoded or not. If `/` is meant to be encoded in a query then I think let's assume it is for now since browsers often auto encode/decode some things even something doesn't quite follow the spec.

The main problem is the leading `?` being encoded which it looks like you've fixed!

manics:
Rebased

consideRatio:
@manics could you update the title of this PR as well to reflect its more than just test related now?

manics:
Done!

yuvipanda:
I added an additional test specifically using nbgitpuller, and showing that URL encoding seems to be done now *just the right amount*, rather than double or none.

js/packages/binderhub-client/lib/index.js
```diff
@@ -118,30 +118,30 @@ export class BinderRepository {
     // Make a copy of the URL so we don't mangle the original
     let url = new URL(serverUrl);
     if (path) {
-      // strip trailing / from URL
-      url.pathname = url.pathname.replace(/\/$/, "");
-
+      // Ensure there is a trailing / in serverUrl
+      if (!url.pathname.endsWith("/")) {
+        url.pathname += "/";
+      }
```
consideRatio, line 124: I think this is redudant when using the URL objects, but I'm not 100%.

```javascript
let baseUrl = new URL("https://developer.mozilla.org");
console.log(new URL("doc/tree/?test=ok", baseUrl))
```

image

Do we delete it or retain it?

```suggestion
```

consideRatio: Hmm okay so pathname may sometimes not endWith `/`, for example...

image

But it seems that the path is ignored when using this URL to construct a new URL via the URL constructor that includes a baseUrl argument.

image

consideRatio: Ah!

I think there is a breaking change here.

```diff
-url.pathname = url.pathname + "/doc/tree/" + encodeURI(path);
+url = new URL("doc/tree/" + encodeURI(path), url);
```

These are different I think, to verify that:

```
console.log("old new behavior presented below")

let nonRoot1 = new URL("https://developer.mozilla.org/test")
nonRoot1.pathname = nonRoot1.pathname.replace(/\/$/, "");
nonRoot1.pathname = nonRoot1.pathname + "/doc/tree/"
console.log(nonRoot1.href)

let root1 = new URL("https://developer.mozilla.org")
root1.pathname = root1.pathname.replace(/\/$/, "");
root1.pathname = root1.pathname + "/doc/tree/"
console.log(root1.href)

console.log("new behavior comes below")

let nonRoot2 = new URL("https://developer.mozilla.org/test")
let newNonRoot2 = new URL("doc/tree", nonRoot2)
console.log(newNonRoot2.href)

let root2 = new URL("https://developer.mozilla.org")
let newRoot2 = new URL("doc/tree", root2)
console.log(newRoot2.href)
```

image

consideRatio (changes requested):
See https://github.com/jupyterhub/binderhub/pull/1774/files#r1358088687, I think this isn't correct yet because there is a behavior change unrelated to escaping.

js/packages/binderhub-client/lib/index.js
manics: The trailing `/` in the pathname is required (`/test/` not `/test`), it affects how the parts are joined.

yuvipanda: Yeah, preventing this behavior is why we enforce path ends with `/` even if it is not passed in. So baseUrl will never not have trailing slash. As @manics says, it affects how parts are joined.

yuvipanda: @consideRatio I added another test case to demonstrate why I think this is fine.

consideRatio: Hmmm but what if baseUrl is a jupyterhub with a prefix?

Note that now, but not before, the nonRoot examples doesnt have the /test prefix in the path?

consideRatio: If you both think there is no loss of path issue, this is good even if i dont understand the details.

consideRatio: Ah, so the href can be the same for a URL object, but end up resulting in a different URL when used to construct another. I think i got it and this LGTM

consideRatio (approved):

yuvipanda:
ok, I tested this locally. You can first see me try an nbgitpuller link on the main branch, and it fails. Then I switch to this branch, and it succeeds (once if hard refresh the page to get new JS).

https://github.com/jupyterhub/binderhub/assets/30430/925ff703-92e7-4f15-bd31-6a5cb565c6cc

Thank you for working on this @manics and thanks for the review, @consideRatio. I am hoping that my cleanup + automated testing efforts will allow us to build more of these things without breaking.

PR #1776, opened 2023-10-14
Bump docker image to use python 3.12

yuvipanda:


consideRatio:
Real test failure, lets not merge

yuvipanda:
I don't think I'll be able to focus on fixing up these anytime soon, so I'm just going to close this so this isn't 'cookie licked'!

PR #1783, opened 2023-10-20
JS: Fix some JSDoc types, and misc details in tests

yuvipanda:
- Optional args get [] around the param name, not the type!
- Remove extra increased timeout for some jest tests, as they are not needed
- Pass actual URL objects to some of the methods, rather than strings. These accidentally worked but shouldn't be relied upon.
- Import just the function used during tests from node modules, rather than the entire module.
- Fix type of `fetch` to indicate that's actually just an *iterator*, not a *generator*.


consideRatio (approved):

consideRatio:
Amazing work on this @yuvipanda !!!!

PR #1784, opened 2023-10-20
Kill singleuser test servers to reduce k3s resource usage

manics:
Closes https://github.com/jupyterhub/binderhub/issues/1754

We build and launch several servers, but we don't shut them down.
https://github.com/jupyterhub/binderhub/blob/af81d7fd53b8fcab78b2f4276bee9828693e677c/binderhub/tests/test_build.py#L52-L96


consideRatio (approved):
Wieeeee thanks for resolving this @manics!!!!!!!!!

PR #1792, opened 2023-11-04
Add extraPodSpec

manics:
Similar to Z2JH https://z2jh.jupyter.org/en/stable/resources/reference.html#hub-extrapodspec


consideRatio (approved):

consideRatio:
Went for "new" label for this, but this looks great to me!

PR #1801, opened 2023-12-07
Update z2jh version

yuvipanda:
I checked the changelog, nothing really needed here.

Helpful for me to bring in newer oauthenticator :)


manics (approved):
Failing test should be fixed by https://github.com/jupyterhub/binderhub/pull/1802

PR #1807, opened 2023-12-15
Enable and fix eslint for all packages

manics:
It was previously hardcoded to only run on a subdirectory


yuvipanda:
yay awesome ty @manics

PR #1813, opened 2024-01-17
Remove ingress.https with unused kube-lego references, rely on ingress.tls and ingress.annotations

yuvipanda:
[kube-lego](https://github.com/jetstack/kube-lego) was the precursor to cert-manager, and has been dead for years and years now

### DIFF
--- a/helm-chart/binderhub/templates/ingress.yaml
+++ b/helm-chart/binderhub/templates/ingress.yaml
@@ -6,7 +6,6 @@ metadata:
   {{- with .Values.ingress.annotations }}
   annotations:
     {{- . | toYaml | nindent 4 }}
-    {{- end }}
   {{- end }}
 spec:
   {{- with .Values.ingress.ingressClassName }}
--- a/helm-chart/binderhub/templates/ingress.yaml
+++ b/helm-chart/binderhub/templates/ingress.yaml
@@ -3,12 +3,8 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: binderhub
-  {{- if or (and .Values.ingress.https.enabled (eq .Values.ingress.https.type "kube-lego")) .Values.ingress.annotations }}
+  {{- with .Values.ingress.annotations }}
   annotations:
-    {{- if and .Values.ingress.https.enabled (eq .Values.ingress.https.type "kube-lego") }}
-    kubernetes.io/tls-acme: "true"
-    {{- end }}
-    {{- with .Values.ingress.annotations }}
     {{- . | toYaml | nindent 4 }}
     {{- end }}
   {{- end }}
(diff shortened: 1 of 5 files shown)


helm-chart/binderhub/templates/ingress.yaml
```diff
@@ -3,12 +3,8 @@ apiVersion: networking.k8s.io/v1
...
   name: binderhub
-  {{- if or (and .Values.ingress.https.enabled (eq .Values.ingress.https.type "kube-lego")) .Values.ingress.annotations }}
+  {{- with .Values.ingress.annotations }}
   annotations:
-    {{- if and .Values.ingress.https.enabled (eq .Values.ingress.https.type "kube-lego") }}
-    kubernetes.io/tls-acme: "true"
-    {{- end }}
-    {{- with .Values.ingress.annotations }}
     {{- . | toYaml | nindent 4 }}
     {{- end }}
```
manics, line 9:
```suggestion
```

consideRatio (approved):
## Cleanup of `ingress.https`

`ingress.https` as a whole is soleley used for `ingress.https.type=kube-lego`. If we remove `type` from the schema, it will error as much as if we remove all of `ingress.https`. Due to this, **I've added a commit cleaning it up fully**.

Note that mybinder.org-deploy [makes use of this configuration](https://github.com/jupyterhub/mybinder.org-deploy/blob/c4efa315da72a36b886aec7c435d159714287d31/mybinder/values.yaml#L306-L312). In practice the annotation `kubernetes.io/tls-acme` is also respected by cert-manager as a signal that it should be active, but it requires some configuration to specify default issuer etc alongside it - see https://cert-manager.io/docs/usage/ingress/.

I think removing this is a fine call, but it requires adjustments to deployments having it explicitly enabled (the default is not).

## Config migration

Deployments with `ingress.https.enabled=true` need to set `ingress.annotations` and `ingress.tls -> .hosts / .secretName` going onwards instead I think.

In practice for mybinder.org-deploy, examplifying with the prod config, I think the following changes would need to be made.

1. Let the mybinder chart remove the `binderhub.ingress.https` section
2. Let the mybinder chart have `binderhub.ingress.annotations` including `kubernetes.io/tls-acme=true`
3. Let `binderhub.ingress.tls` look like `binderhub.jupyterhub.ingress.tls` [configured here](https://github.com/jupyterhub/mybinder.org-deploy/blob/c4efa315da72a36b886aec7c435d159714287d31/config/prod.yaml#L43-L72), declaring a secret name `kubelego-tls-binder-prod` to avoid making a change of name.

Like that, I think mybinder.org will render like it did before this change as well.

### Config migration verified

I did some chart rendering and concluded that the before / after config below led to the same content (besides key ordering and indentation).

```yaml
# possible use before this config
# helm template . --values before-config.yaml --show-only templates/ingress.yaml > before-ingress.yaml
ingress:
  enabled: true
  hosts:
    - example.com
  https:
    enabled: true
    type: kube-lego

# equivalent config after this PR
# helm template . --values after-config.yaml --show-only templates/ingress.yaml > after-ingress.yaml
ingress:
  enabled: true
  hosts:
    - example.com
  annotations:
    kubernetes.io/tls-acme: "true"
  tls:
    - secretName: binderhub-tls-binder-release-name
      hosts:
        - example.com
```

```shell
git diff --no-index -- before-ingress.yaml after-ingress.yaml
```

image

yuvipanda:
Thanks @consideRatio.

I am going to wait to make sure I have access to the curvenote member as well (https://github.com/jupyterhub/mybinder.org-deploy/issues/2920) before merging this and deploying it on mybinder.org

yuvipanda:
Closing in favor of plan in https://github.com/jupyterhub/binderhub/pull/1862#issuecomment-2197822914

PR #1815, opened 2024-01-21
Use setup-helm action to install helm

yuvipanda:
This reduces overall custom bash code we maintain, which is great!


manics (approved):
🎸

PR #1828, opened 2024-02-23
Remove incorrect docker socket if found

manics:
BinderHub has a well known problem where the docker.socket is created as a directory, which then causes dind to crash:
https://github.com/jupyterhub/mybinder.org-deploy/blob/61c0cbedddee92dbcd252e02d9bd02b5a5b94d58/docs/source/operation_guide/common_problems.md#the-docker-in-docker-socket

Given that this problem has existed for years, occurs on multiple deployments, and can result in a prolonged outage until a manual fix is made (most recently today- both Curvenote nodes were hit by this bug simultaneously blocking all builds), I think we should automate the fix.

I originally made this controllable with parameter `{dind,pink}.cleanupIncorrectSocketDir` but I've removed it in commit d6c66045a0168c8e00f685e1fbf8f21498171392 to reduce the added logic, I'm happy to add it back though.

### DIFF
--- a/helm-chart/binderhub/schema.yaml
+++ b/helm-chart/binderhub/schema.yaml
@@ -414,11 +414,6 @@ properties:
             type: array
             description: |
               Additional volume mounts for the Container. Use a k8s native syntax.
-      cleanupIncorrectSocketDir: &cleanupIncorrectSocketDir-spec
-        type: boolean
-        description: |
-          If the container socket has been incorrectly created as a directory
-          remove the directory
       storageDriver:
         type: string
         description: |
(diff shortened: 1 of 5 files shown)


helm-chart/binderhub/schema.yaml
```diff
@@ -417,16 +417,20 @@ properties:
...
+          Docker storage driver
       resources: *resources-spec
-      hostSocketDir:
+      hostSocketDir: &hostSocketDir-spec
         type: string
         description: |
-          TODO
-      hostLibDir:
+          Host directory where the container socket will be located
+      hostLibDir: &hostStorageDir-spec
```
manics, line 426: We could change this to the more generic `hostStorageDir` at some point to match the pink parameter names, but that's a breaking change.

minrk (approved):
absolutely _thrilled_ to have an automated fix for this

manics:
Merging, the failing test is unrelated and the docker socket directory problem has appeared again on mybinder

PR #1868, opened 2024-07-08
Update message about repo not allowed to launch

consideRatio:
We are checking `is_banned` (really a check if the repo is allowed or not), and if its not allowed we write a message saying "has been temporarily disabled" - but I think its incorrect to include "temporarily" here as we may work against a check of allowed repositories and not just bans (temporary?).

image


yuvipanda:
Thanks @consideRatio

PR #1884, opened 2024-11-07
Update Z2JH to 4.0.0

manics:
Z2JH changelog: https://z2jh.jupyter.org/en/4.0.0/changelog.html

### DIFF
--- a/binderhub/tests/test_build.py
+++ b/binderhub/tests/test_build.py
@@ -21,7 +21,7 @@ from .utils import async_requests
 
 # We have optimized this slow test, for more information, see the README of
 # https://github.com/binderhub-ci-repos/minimal-dockerfile.
-@pytest.mark.asyncio(timeout=900)
+@pytest.mark.asyncio
 @pytest.mark.parametrize(
     "slug",
     [
@@ -102,7 +102,7 @@ async def test_build(app, needs_build, needs_launch, always_build, slug, pytestc
     stop.raise_for_status()
 
 
-@pytest.mark.asyncio(timeout=900)
+@pytest.mark.asyncio
 @pytest.mark.parametrize(
     "app,build_only_query_param",
     [
@@ -152,7 +152,7 @@ async def test_build_only(app, build_only_query_param, needs_build):
     assert final["phase"] == "ready"
 
 
-@pytest.mark.asyncio(timeout=120)
+@pytest.mark.asyncio
 @pytest.mark.remote
 async def test_build_fail(app, needs_build, needs_launch, always_build):
     """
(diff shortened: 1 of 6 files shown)


binderhub/tests/test_build.py
```diff
@@ -21,7 +21,7 @@
 
 # We have optimized this slow test, for more information, see the README of
 # https://github.com/binderhub-ci-repos/minimal-dockerfile.
-@pytest.mark.asyncio(timeout=900)
+@pytest.mark.asyncio
```
manics, line 24: According to https://github.com/pytest-dev/pytest-asyncio/issues/215 the timeout argument was only a proposal.
This means the argument would've been silently ignored until additional error checking was added in https://github.com/pytest-dev/pytest-asyncio/pull/886

manics:
The test failure is due to https://github.com/jupyterhub/binderhub/issues/1886
(though we're only showing a max of two failures so there may be more)

consideRatio (approved):

consideRatio:
Ckan resolved, next error is now in binderhub/tests/test_eventlog.py::test_emit_event

manics:
https://github.com/jupyterhub/binderhub/pull/1884/commits/ba490a751ab9511ea4b69b10e367a96c2233f3cd
deletes the `taskName` attribute.
This is a new logging attribute in Python 3.13:
https://docs.python.org/3.13/library/logging.html#logrecord-attributes

I don't understand how the EventLog works though. It only removes the `message` attribute
https://github.com/jupyterhub/binderhub/blob/58bb7dba89baf739c4015a6afc59d6ec3c8d59d7/binderhub/events.py#L15-L104
but https://docs.python.org/3.13/library/logging.html#logrecord-attributes lists many more attributes.

consideRatio (approved):

consideRatio:
Makes sense to me given what i get about this, which isnt much, but i dont see how kt would be breaking or similar to do this change now that python 3.12 is used. (Note it was in 3.12 the taskname thing got introduced)

binderhub/tests/test_build.py
```diff
@@ -176,7 +179,8 @@ async def test_build_fail(app, needs_build, needs_launch, always_build):
     assert failed_events > 0, "Should have seen phase 'failed'"
 
 
-@pytest.mark.asyncio(timeout=120)
+@pytest.mark.timeout(120)
+@pytest.mark.asyncio
```
consideRatio, line 183: Was pytest.mark.asyncio needed? We have a lot of jupyterhub repos with an automatic detection mode, making all async test functions assumed this decorator

consideRatio: I think this may make it not needed, but also not a problem to have explicit

https://github.com/jupyterhub/binderhub/blob/58bb7dba89baf739c4015a6afc59d6ec3c8d59d7/pyproject.toml#L34

manics: I'll take it out

consideRatio:
Thank you @manics!!!

PR #1885, opened 2024-11-07
Update python, node, helm

manics:
Update GH workflows to use
- Python 3.12
- NodeJS 22
- Helm 3.16.2

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -42,12 +42,12 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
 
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -42,12 +42,12 @@ jobs:
...
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
 
       - uses: actions/setup-node@v4
         # node required to build wheel
         with:
-          node-version: "16"
+          node-version: "22"
```
consideRatio, line 50: I've made this "lts/*" in various other projects to reduce the amount of manual bumping needs - fine to bump manually also

PR #1889, opened 2024-11-27
Better ipv6 support when checking network bans

yuvipanda:
- Rely on using `in` between IP addresses and IP Networks (https://docs.python.org/3/library/ipaddress.html#networks-as-containers-of-addresses) rather than re-implement that ip address / network matching with CIDRs ourselves.
- Brings in *much* better ipv6 support
- Adds a few test cases for ipv6 support as well
- Remove `ban_networks_min_prefix_len` which is no longer necessary

Discovered while investigating the test failures in https://github.com/jupyterhub/binderhub/pull/1856

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -736,19 +736,6 @@ class BinderHub(Application):
 
         return networks
 
-    ban_networks_min_prefix_len = Integer(
-        1,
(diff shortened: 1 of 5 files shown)


manics (approved):
This looks a lot nicer!

PR #1891, opened 2024-12-07
Add playwright based UI integration tests for existing UI

yuvipanda:
Primarily written to make UI testing easier in https://github.com/jupyterhub/binderhub/pull/1856, and
steals code from there

- Add a new 'integration-tests' directory that does white box
  UI testing
- Install dockerspawner in dev-requirements, as the playwright
  integration tests now need it
- Move loading and about page tests to use playwright
- Add some tests for the home page
- Use a single instance of local-binder-local-hub for doing thes
  integration tests
- Upload playwright traces (https://playwright.dev/python/docs/trace-viewer)
  on failure as a github artifact so we can debug things better 
- Split some more tests out of `test_main.py`, so we can get rid of it as part of #1856

### DIFF
--- a/integration-tests/conftest.py
+++ b/integration-tests/conftest.py
@@ -1,0 +1,6 @@
+import nest_asyncio
+
+
+def pytest_configure(config):
+    # Required for playwright to be run from within pytest
+    nest_asyncio.apply()
--- a/.github/workflows/playwright.yaml
+++ b/.github/workflows/playwright.yaml
@@ -1,0 +1,97 @@
+name: Playwright Tests
+
+on:
+  pull_request:
+    paths-ignore:
+      - "**.md"
+      - "**.rst"
+      - "docs/**"
+      - "examples/**"
+      - ".github/workflows/**"
+      - "!.github/workflows/playwright.yaml"
+  push:
+    paths-ignore:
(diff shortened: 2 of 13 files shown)


integration-tests/conftest.py
```diff
@@ -0,0 +1,6 @@
+import nest_asyncio
+
+
+def pytest_configure(config):
+    # Required for playwright to be run from within pytest
```
minrk, line 5: The alternative to this is to use async tests with the playwright async API. That's what we use in the jupyterhub tests.

minrk (approved):
Nice! I won't merge myself because I don't know how this will interact with #1856, but feel free to go ahead.

integration-tests/conftest.py
yuvipanda: @minrk oooh, this led me down a hole and I ended up opening https://github.com/jupyterhub/binderhub/issues/1894 to not derail this PR.

yuvipanda:
I opened https://github.com/jupyterhub/binderhub/issues/1894 to handle follow-up task around asyncio.

Ty for the review @minrk! I'm going to merge this now, and then rebase #1856 on top!

PR #1892, opened 2024-12-07
Don't stop when 2 tests fail

yuvipanda:
This gives the false impression that only two tests are failing, and is very surprising behavior to have as default. Users who want this can pass the commandline arg to `py.test` when they want it.


yuvipanda:
Took me *entirely* by surprise in #1856

manics (approved):

Issue #1903, opened 2025-01-06
Support different container registry to push and pull


rgaiacs:
### Proposed change

In some advanced use cases, the operator might want to push to one container registry and pull from another container registry. For example, the operator want to push to a public container registry like https://hub.docker.com/ and pull from a private [proxy cache](https://goharbor.io/docs/2.12.0/administration/configure-proxy-cache/).

![binderhub](https://github.com/user-attachments/assets/bddd4d1d-de0b-4c63-b336-3124ebf53798)

This will require

- [ ] Add extra option in Helm chart

    ```yaml
    config:
      BinderHub:
        use_registry: true
        image_prefix_to_push: "your-push-registry.io/<prefix>-"
        image_prefix_to_pull: "your-pull0registry.io/<prefix>-"
    ```

- [ ] Support new option when building container
- [ ] Support new option when contacting JupyterHub

### Who would use this feature?

@arnim and @rgaiacs want to use it at the mybinder.org server operated by GESIS.

rgaiacs:
I'm closing this because GESIS is replacing the server with one from Hetzner Online GmbH as announced in https://github.com/jupyterhub/mybinder.org-deploy/issues/3264 and we will not pursue this idea.

PR #1904, opened 2025-01-06
Support different container registry to push and pull

rgaiacs:
This covers https://github.com/jupyterhub/binderhub/issues/1903.

### DIFF
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -445,7 +445,7 @@ class BinderHub(Application):
 
     image_prefix_push = Unicode(
         help="""
-        Prefix for built docker images push to container registry.
+        Prefix for built docker images being pushed to the to container registry.
 
         If you are pushing to gcr.io, this would start with:
             gcr.io/<your-project-name>/
@@ -463,7 +463,7 @@ class BinderHub(Application):
 
     image_prefix_pull = Unicode(
         help="""
-        Prefix for built docker images pull from container registry.
+        Prefix for built docker images being pulled from container registry.
 
         If you are pushing to gcr.io, this would start with:
             gcr.io/<your-project-name>/
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -479,7 +479,6 @@ class BinderHub(Application):
     def _image_prefix_pull_default(self):
         return self.image_prefix
 
-
     build_memory_request = ByteSpecification(
         0,
         help="""
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -444,7 +444,6 @@ class BinderHub(Application):
     )
 
     image_prefix_push = Unicode(
-        "",
         help="""
         Prefix for built docker images push to container registry.
 
@@ -458,8 +457,11 @@ class BinderHub(Application):
         config=True,
     )
 
+    @default("image_prefix_push")
+    def _image_prefix_push_default(self):
+        return self.image_prefix
+
     image_prefix_pull = Unicode(
-        "",
         help="""
         Prefix for built docker images pull from container registry.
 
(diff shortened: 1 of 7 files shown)


binderhub/app.py
```diff
@@ -443,6 +443,42 @@ def _pod_quota_deprecated(self, change):
         config=True,
     )
 
+    image_prefix_push = Unicode(
+        help="""
+        Prefix for built docker images push to container registry.
```
manics, line 448:
```suggestion
        Prefix for built docker images being pushed to the container registry.
```

manics (commented):
My initial thoughts are that most BinderHubs will use the same prefix for push and pull, so how about just have `image_prefix` as the default property as at present, and maybe add just one new property, `image_prefix_pull`?

I think it's fine to keep using the `*_push` and `*_pull` variables in builder.py if you want though, for clarity.

rgaiacs:
> My initial thoughts are that most BinderHubs will use the same prefix for push and pull

You are absolutely right. I think that 99% of BinderHubs will use a single registry. But there will be the 1% like GESIS that want to push to Docker Hub or GitHub Packages Registry and pull from a local Harbor that operates as [proxy cache](https://goharbor.io/docs/2.12.0/administration/configure-proxy-cache/).

>  so how about just have `image_prefix` as the default property as at present, and maybe add just one new property, `image_prefix_pull`?

`image_prefix` is use as the default property, see https://github.com/rgaiacs/binderhub/blob/c322eabfdd78574e6b2ea03863f04ad4593060d7/binderhub/app.py#L460-L462 and https://github.com/rgaiacs/binderhub/blob/c322eabfdd78574e6b2ea03863f04ad4593060d7/binderhub/app.py#L478-L480. **No change required to existing deployments.**

rgaiacs:
I forgot that the credentials to the registries might be different and must also be provided.

@manics @minrk @yuvipanda can I get a bit of help on how the registry credentials are handle? Thanks!

rgaiacs:
@manics @minrk @yuvipanda my understanding is that in https://github.com/jupyterhub/binderhub/blob/2ac912493f3d06e21146048f6240c017ef4fc126/binderhub/builder.py#L408-L421 BinderHub will contact the registry to check if the image exists. For a private registry, BinderHub will use the credentials available for the Docker client configured at https://github.com/jupyterhub/binderhub/blob/2ac912493f3d06e21146048f6240c017ef4fc126/binderhub/registry.py#L139-L146 and https://github.com/jupyterhub/binderhub/blob/2ac912493f3d06e21146048f6240c017ef4fc126/binderhub/registry.py#L164-L171

For this PR to move forward, the user will need to pass two credentials. One for the registry to push and another for the registry to pull. For this, backward compatibility will be broken. Can you advise me what is your preferable way to move this forward? Duplicate the credentials as done for the image name? Or use another format? Thanks!

rgaiacs:
09051c43517d0ddaf1ccf5b61ed72562a8ab0e4d has a backward compatibility change in the Helm chart. `registry` changed to be a list.

rgaiacs:
BinderHub is missing the existing image in the container registry because the [`DockerRegistry` class](https://github.com/jupyterhub/binderhub/blob/2ac912493f3d06e21146048f6240c017ef4fc126/binderhub/registry.py#L20C7-L20C21) is implemented to support only one registry as mentioned in https://github.com/jupyterhub/binderhub/blob/2ac912493f3d06e21146048f6240c017ef4fc126/binderhub/registry.py#L42 And the Helm chart implementation of `.docker/config.json` in 09051c43517d0ddaf1ccf5b61ed72562a8ab0e4d does **not** preserve the order from the Helm value.

PR #1915, opened 2025-01-25
Set btn-secondary on dropdown selectors

manics:
This ensures you receive visual feedback when you use the keyboard TAB to move between form fields. At present there is no visual indication when you get to a dropdown.

https://github.com/user-attachments/assets/194d7ec2-018c-4dc3-8245-327d0ce8df01


rgaiacs:
Looks good to me.

Issue #1907, opened 2025-01-09
Document how to customise index / root page


rgaiacs:
The [documentation](https://binderhub.readthedocs.io) includes a section named "[About page customization](https://binderhub.readthedocs.io/en/latest/customizing.html#about-page-customization)" but it misses a section "Index page customization".

After https://github.com/jupyterhub/binderhub/pull/545, it was possible to customise the index page. GESIS has used it, see screenshot from https://notebooks.gesis.org.

![Screenshot 2025-01-09 at 16-45-16 Binder](https://github.com/user-attachments/assets/c5d71cba-5c0b-4191-b0e6-05f48a8d3182)

After the frontend rewrite in https://github.com/jupyterhub/binderhub/pull/1856, `binderhub` moved from use [Jinja](https://jinja.palletsprojects.com/en/stable/) to use [React](https://react.dev/).

The React app [looks for a HTML DOM element with ID equals to `root`](https://github.com/jupyterhub/binderhub/blob/3b3743b65a7cc7016a2823a4ea1ca3f1aacd03f3/binderhub/static/js/index.jsx#L4) to construct the page.

Given the component nature of React, should be possible for GESIS to reuse the new [`<LinkGenerator>`](https://github.com/jupyterhub/binderhub/blob/3b3743b65a7cc7016a2823a4ea1ca3f1aacd03f3/binderhub/static/js/pages/HomePage.jsx#L62-L76) and [`<BuilderLauncher>`](https://github.com/jupyterhub/binderhub/blob/3b3743b65a7cc7016a2823a4ea1ca3f1aacd03f3/binderhub/static/js/pages/HomePage.jsx#L77-L85) components.

I will start testing by overwriting https://github.com/jupyterhub/binderhub/blob/3b3743b65a7cc7016a2823a4ea1ca3f1aacd03f3/binderhub/templates/page.html with

```html
<body>
<div id="link-generator"></div>
</body>

<script src="{{static_url("dist/bundle.js")}}"></script>
<script>
    document.addEventListener('DOMContentLoaded', function () {
        const root = ReactDOM.createRoot(document.getElementById('link-generator'));
        root.render(React.createElement(LinkGenerator));
    });
</script>
```

@yuvipanda because I have not used React in the last couple of years, am I missing something? Thanks!

cc @arnim

yuvipanda:
hi @rgaiacs!

We'll have to build in whatever we need. I think the process we should take is:

1. Figure out what you wanna actually change
[...]

PR #1923, opened 2025-01-30
Add template_file config and document page customization

rgaiacs:
Closes https://github.com/jupyterhub/binderhub/issues/1907

We might want to make add a new `traitlets` for users to provide their own Jinja template that will be used in https://github.com/jupyterhub/binderhub/blob/4f0a91ac7a851238d0857e68691edf2ab70c4b3d/binderhub/main.py#L42 from https://github.com/jupyterhub/binderhub/blob/4f0a91ac7a851238d0857e68691edf2ab70c4b3d/binderhub/main.py#L41-L46


docs/source/customizing.rst
```diff
@@ -1,6 +1,63 @@
...
 =====================================
 
+Frontend
+--------
+
+Header and Footer customization
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+BinderHub uses `Jinja <https://jinja.palletsprojects.com/en/stable/>`_ as template engine
+to process the page template `binderhub/templates/page.html <https://github.com/jupyterhub/binderhub/blob/main/binderhub/templates/page.html>`.
```
manics, line 11: There's something wrong with the link syntax- it's not linked:
https://binderhub--1923.org.readthedocs.build/en/1923/customizing.html#header-and-footer-customization

manics: I think it'd be helpful to also highlight the important parts of the template- is `<div id="root"></div>` the critical bit?

docs/source/customizing.rst
```diff
@@ -1,6 +1,63 @@
...
+    metadata:
+      name: binderhub-template-custom
+      labels:
+        app: binder
+        component: etc-binderhub
+        heritage: {{ .Release.Service }}
+        release: {{ .Release.Name }}
+    data:
+      {{- (.Files.Glob "files/*").AsConfig | nindent 2 }}
+
```
manics, line 41: Is additional Helm configuration needed to mount the file in the ConfigMap into the BinderHub pod?

rgaiacs:
I will address the comments in a couple of weeks.

rgaiacs:
Sorry for the delay to return to this.

I address the comments.

Do I need to add some tests? If yes, can I have some help? Thanks!

agahkarakuzu:
@rgaiacs thanks for this PR! 

How complicated would it be to customize components like `HowItWorks.jsx`, and have the template page use the built dist to control what gets rendered inside `<div id="root"></div>`?

For example, I can easily add header/footer

image

but I cannot figure out a way to "replace" the binder logo with this design. 

Also, I think `extra_static_path` no longer takes effect, or at least I could not figure out how to link assets I used to mount at `/etc/binderhub/custom/static`.

rgaiacs:
Hi @agahkarakuzu,

the challenge to customise any file in [`binderhub/static/js`](https://github.com/jupyterhub/binderhub/tree/837169dfebf9640c4d3b4f4247240b15fe2278a0/binderhub/static/js) is due to how the JavaScript works.

We are using [webpack](https://webpack.js.org/) to merge all JavaScript files into a single `bundle.js` file that is we reference in the HTML that we send to the user, see https://github.com/jupyterhub/binderhub/blob/837169dfebf9640c4d3b4f4247240b15fe2278a0/binderhub/templates/page.html#L27

> How complicated would it be to customize components like HowItWorks.jsx

If you want to change any of the JavaScript files, for example [`HowItWorks.jsx`](https://github.com/jupyterhub/binderhub/blob/837169dfebf9640c4d3b4f4247240b15fe2278a0/binderhub/static/js/components/HowItWorks.jsx), you will have to

1. re-run webpack to create a new `bundle.js`.
2. create a new volume that overwrite the default `bundle.js` with the one that you created.

agahkarakuzu:
Thanks @rgaiacs ! I actually created a new bundle.js. Would simply mounting the new distribution override the existing one?

rgaiacs:
> Would simply mounting the new distribution override the existing one?

It should.

rgaiacs:
One of the tests are with 

```
FAILED binderhub/tests/test_registry.py::test_get_image_manifest[False] - OSError: [Errno 98] Address already in use
```

I believe the failure not related with this pull request.

@yuvipanda and @minrk this is blocking https://github.com/jupyterhub/mybinder.org-deploy/pull/3186 and I would like to merge this earlier next week.

manics:
Do we need the new `template_file` property? Can't we already override the template using `template_path`?

https://github.com/jupyterhub/binderhub/blob/43ecb3a8677ad4a5e0a0a70e1a0ab240f67d52dd/binderhub/app.py#L791-L798

PR #1929, opened 2025-02-13
Only mount ~/.docker/config.json, not ~/.docker

yuvipanda:
docker buildx wants to write to ~/.docker/buildx, but since we mount the secret *entirely* (to ~/.docker/buildx), it can not. I think mounting *just* the subpath should work.

Follow-up to https://github.com/jupyterhub/repo2docker/pull/1402

### DIFF
--- a/binderhub/build.py
+++ b/binderhub/build.py
@@ -448,7 +448,11 @@ class KubernetesBuildExecutor(BuildExecutor):
 
         if not self.registry_credentials and self.push_secret:
             volume_mounts.append(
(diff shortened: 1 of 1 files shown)


minrk (approved):

consideRatio:
There is a slight behavior change that i don't think is a problem, but I'm not 100% confident on it - it could even be that this is a fix to an edge case.

With subPath is used, it wont refresh the mounted config file over time if the source Secret/ConfigMap changes, but that seems fine as long as the pod isn't staying around for multiple builds etc - which it won't, right?

PR #1930, opened 2025-02-13
Bump base image versions used

yuvipanda:
There's a segfault in the helm chart publish workflow that isn't being resolved by simply rebuilding (https://github.com/jupyterhub/binderhub/actions/runs/13316495693/job/37191967146).

So this bumps up the base versions of everything to something more modern, and hopefully when merged will cause a helm chart to actually publish.


consideRatio (approved):
Reasonable changes, but the bump to py 3.13 from 3.11 in the binder pod is worth highlighting in the title and a "breaking" label (as it could impact extraConfig or added dependencies not compatible with py 3.13)

consideRatio:
If you just want to refreeze in this/z2jh/binderhub-service repos you can also trigger the workflow https://github.com/jupyterhub/binderhub/actions/workflows/watch-dependencies.yaml btw!

Issue #1934, opened 2025-02-27
tornado.httpclient.HTTPClientError: HTTP 401: Unauthorized for GCP service account which works locally


paololazzari:
### Bug description

I am trying to setup binderhub on gcloud with gcp artifact registry. I created a service account with a bunch of permissions and deployed binderhub with the following:

`secret.yaml`:
```
registry:
  url: us-central1-docker.pkg.dev

  # below is the content of the JSON file downloaded earlier for the container registry from Service Accounts
  # it will look something like the following (with actual values instead of empty strings)
  # paste the content after `password: |` below
  password: |
    {
      censored
    }
```

`config.yaml`:
```
config:
  BinderHub: 
    debug: true
    use_registry: true
    image_prefix: us-central1-docker.pkg.dev/my-project-name/binderhub/binder
    ...
```

I tried to launch this repository https://github.com/yunabe/lgo-binder. I built that image locally and pushed it to the registry. When I try to launch the repo from binder I see the following errors in the `binder` pod:

```
[D 250227 10:53:49 registry:302] Getting image manifest from https://us-central1-docker.pkg.dev/v2/my-project-name/binderhub/binderyunabe-2dlgo-2dbinder-798bce/manifests/cc3ac7aa13ddf6602d70f2d8bbfb228e403fb419
[E 250227 10:53:49 builder:417] Failed to get image manifest for us-central1-docker.pkg.dev/my-project-name/binderhub/binderyunabe-2dlgo-2dbinder-798bce:cc3ac7aa13ddf6602d70f2d8bbfb228e403fb419
    Traceback (most recent call last):
      File "/usr/local/lib/python3.11/site-packages/binderhub/builder.py", line 411, in get
        image_manifest = await self.registry.get_image_manifest(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/binderhub/registry.py", line 304, in get_image_manifest
        resp = await client.fetch(req)
               ^^^^^^^^^^^^^^^^^^^^^^^
    tornado.httpclient.HTTPClientError: HTTP 401: Unauthorized
```

I thought perhaps the service account didn't have enough permissions, but if I try to hit the same endpoint using the service account credentials locally it works fine:

```
$ curl -H "Authorization: Bearer $(gcloud auth print-access-token binderhubsa@my-project-name.iam.gserviceaccount.com)" \
[...]

PR #1936, opened 2025-02-28
Add support for Google Artifact Registry

paololazzari:
Fixes https://github.com/jupyterhub/binderhub/issues/1934

### DIFF
--- a/binderhub/registry.py
+++ b/binderhub/registry.py
@@ -225,7 +225,10 @@ class DockerRegistry(LoggingConfigurable):
             ) from None
 
     async def _get_token(self, client, token_url, service, scope):
-        if token_url == "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token":
+        if (
+            token_url
+            == "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token"
+        ):
             auth_req = httpclient.HTTPRequest(
                 token_url, headers={"Metadata-Flavor": "Google"}
             )
--- a/binderhub/registry.py
+++ b/binderhub/registry.py
@@ -229,8 +229,7 @@ class DockerRegistry(LoggingConfigurable):
     async def _get_token(self, client, token_url, service, scope):
         if self.url.endswith("docker.pkg.dev"):
             auth_req = httpclient.HTTPRequest(
-                token_url,
-                headers={"Metadata-Flavor":"Google"}
+                token_url, headers={"Metadata-Flavor": "Google"}
             )
         else:
             auth_req = httpclient.HTTPRequest(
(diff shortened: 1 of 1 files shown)


paololazzari:
Looks like this change isn't enough for pushing a new image to the registry:

```
{"message": "unauthorized: authentication failed", "phase": "failed"}
Exception ignored in: <function Application.__del__ at 0x7c7322d5c7c0>
  File "/opt/venv/lib/python3.11/site-packages/traitlets/config/application.py", line 1065, in __del__
  File "/opt/venv/lib/python3.11/site-packages/traitlets/config/application.py", line 1054, in close_handlers
  File "/opt/venv/lib/python3.11/site-packages/traitlets/traitlets.py", line 687, in __get__
  File "/opt/venv/lib/python3.11/site-packages/traitlets/traitlets.py", line 666, in get
TypeError: 'NoneType' object is not callable
```

paololazzari:
This pull requests allows for the GCP's Google Metadata Server to be used as a `token_url`:

```
config:
  BinderHub: 
    DockerRegistry:
      url: "https://us-central1-docker.pkg.dev"
      token_url: http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token
      username: "oauth2accesstoken"
```

paololazzari:
@consideRatio could you take a look please? this change is the only way I can get binderhub to work with artifact registry

paololazzari:
@rgaiacs could you take a look please?

consideRatio:
@paololazzari Sorry I am too low on time to help at the moment =/

rgaiacs:
> could you take a look please?

I never used Google Compute Engine. I cannot help. Sorry.

paololazzari:
Is there anyone who can help? it's a very small change.

manics (commented):
Sorry for the delay in reviewing. Based on your comments in https://github.com/jupyterhub/binderhub/issues/1934 this seems reasonable. However rather than adding a conditional here I think it'd be cleaner to create a subclass, especially if we need to add more conditionals for other registries in future.

e.g.
```python
class GoogleArtifactRegistry(DockerRegistry):
    @default("token_url")
    def _default_token_url(self):
        ....

    async def _get_token(self, client, token_url, service, scope):
        ....
```

This also makes it easier to use- you can just specify the registry class instead of having to set the token_url since that'll be embedded in the class.

paololazzari:
@manics Thanks for the feedback, I've made those changes.

binderhub/registry.py
```diff
@@ -339,6 +339,39 @@ async def get_credentials(self, image, tag):
...
+    @default("token_url")
+    def _default_token_url(self):
+        return "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token"
+
+    async def _get_token(self, client, token_url, service, scope):
+        auth_req = httpclient.HTTPRequest(
+            token_url, headers={"Metadata-Flavor": "Google"}
+        )
+        self.log.debug(
+            f"Getting registry token from {token_url} service={service} scope={scope}"
```
manics, line 361: Does the metadata server support the service/scope argument? If not we should omit it from the log.

manics (approved):

Issue #1940, opened 2025-03-07
Move away from using a daemonset to serve the docker socket


yuvipanda:
When I first set it up, I made the docker socket for builds a unix socket for the following reasons:

1. It allowed us to re-use the existing docker socket used by k8s on the machine, allowing a no-registry approach
2. The only real way to secure a tcp docker socket is via a CA, which was... complicated.
3. It allowed for rudimentary load balancing of the build pods

(1) is no longer the case for like years now, as docker is no longer used by kubernetes.

(2) is easier now (see https://helm.sh/docs/chart_template_guide/function_list/#genprivatekey and friends).

Doing this would allow us to simplify our operational model - we can have a single docker daemon that's doing builds instead of a large number. And we can do load balancing by switching to https://docs.docker.com/build/builders/drivers/kubernetes/.

So let's move away from the daemonset to a deployment + service for dind!

PR #1862, opened 2024-06-28
WIP: Allow BinderHub to work with an existing JupyterHub

manics:
I thought this used to be possible, but it might be my imagination.

https://discourse.jupyter.org/t/installing-binderhub-to-existing-jupyterhub/26561

PR #1941, opened 2025-03-07
Make binderhub helm chart not depend on jupyterhub

yuvipanda:
This is a WIP implementation of the proposal that I spoke about with @minrk a while back, and outlined in https://github.com/jupyterhub/binderhub/pull/1862#issuecomment-2197822914.

It basically gets rid of the existing helm chart in this repo which embeds jupyterhub *into* it, causing all sorts of issues due to the inheritance. We instead want the charts to be composable, so it can be deployed more easily in many different places. This binderhub-service chart was developed thanks to contributions from GESIS (ty @arnim), and I've merged it in here while preserving all the git history. 

I've worked on migrating one of the mybinder.org federation members to using this chart to make sure it works, and 2i2c-bare.mybinder.org now successfully runs on this chart!

I wanted to work on https://github.com/jupyterhub/binderhub/issues/1940 as well as using kubernetes buildx support for distributed building, and wanted to unify the helm charts before I do so.

## TODO

- [ ] Rename chart from `binderhub-service` to `binderhub`
- [ ] Update all the existing deployment documentation to take into account how this should work now
- [ ] Figure out a plan forward for `BinderHubSpawnerMixin`
- [ ] Announce this change, along with the last version of the old chart


yuvipanda:
Opened a topic to discuss paths forward for the spawner mixin: [#jupyterhub > Future of binderhub_mixin.py in binderhub](https://jupyter.zulipchat.com/#narrow/channel/469744-jupyterhub/topic/Future.20of.20binderhub_mixin.2Epy.20in.20binderhub)

Issue #1943, opened 2025-03-11
Update authentication with docker cli


rgaiacs:
As reported in https://github.com/jupyterhub/mybinder.org-deploy/issues/3253, BinderHub instances using Docker Hub as container registry will encounter authentication issues because **base64** `user:password` and `user:token` is not accepted by Docker Hub.

We need to use

```
cat ~/my_password.txt | docker login --username foo --password-stdin
```

This is how the Docker's login GitHub Action operates, see https://github.com/docker/login-action/blob/327cd5a69de6c009b9ce71bce8395f28e651bf99/src/docker.ts#L35.

This change is limited to Docker Hub. For example, [Project Quay](https://docs.projectquay.io/api_quay.html#creating-v2-oauth-access-token) continues to mention **base64** `user:token` in their documentation.

PR #1945, opened 2025-03-11
Append / to Docker Hub auth URL

rgaiacs:
Closes https://github.com/jupyterhub/binderhub/issues/1943

On a Linux bare metal machine, when I run

```
docker login
```

the `~/.docker/config.json` is

```json
{
  "auths": {
    "https://index.docker.io/v1/": {
      "auth": "****"
    }
  }
}
```

This is very similar to the file being generated except for the trailing slash.

This might be a regression on Docker client.

### DIFF
--- a/binderhub/registry.py
+++ b/binderhub/registry.py
@@ -46,7 +46,7 @@ class DockerRegistry(LoggingConfigurable):
             # which assumes https
             auth_config_url = "https://" + auth_config_url
 
-        if auth_config_url.rstrip("/") == DEFAULT_DOCKER_AUTH_URL:
+        if auth_config_url == DEFAULT_DOCKER_AUTH_URL:
             # default docker config key is the v1 registry,
             # but we will talk to the v2 api
             return DEFAULT_DOCKER_REGISTRY_URL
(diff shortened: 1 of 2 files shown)


minrk:
Looks like tests need updating, but 👍 . Some code might need to be updated to handle the trailing slash, but I'm not sure.

rgaiacs:
I patched the test that was comparing the strings.

rgaiacs:
I don't understand why https://github.com/jupyterhub/binderhub/blob/579dd78430422bae896c90d69ffad637d9fcfc26/binderhub/tests/test_registry.py#L42 is failing now.

rgaiacs:
I tried to understand why we were using `https://index.docker.io/v1` (no trailing slash) but the code was introduced 7 years ago at https://github.com/jupyterhub/binderhub/commit/db06622fef0076052e6dd2939248af1a40e140d7. In all places that I look regarding the `.docker/config.json`, I always see `https://index.docker.io/v1/` (with trailing slash).

All tests are passing now.

yuvipanda:
ty @rgaiacs!

Issue #1949, opened 2025-03-13
Provide way to inject custom css in page


paololazzari:
I'm looking to customize the banner message and to do that I need to be able to provide my own css. 

Currently it's possible to inject custom scripts with `extra_footer_scripts`, but it is not possible to inject custom css.

PR #1950, opened 2025-03-13
Add extra_css for custom css injection

paololazzari:
Fixes https://github.com/jupyterhub/binderhub/issues/1949.

I tested this locally by adding the following to `testing/local-binder-mocked-hub/binderhub_config.py`:

```
c.BinderHub.banner_message = 'This is headline <a href="#">news.</a>'
c.BinderHub.extra_css = """
.p-3.bg-light.shadow-sm.text-center {
    color: red;
}
"""
```

the style was loaded and the headline color changed.

### DIFF
--- a/binderhub/templates/page.html
+++ b/binderhub/templates/page.html
@@ -16,6 +16,9 @@
   <link id="favicon" rel="shortcut icon" type="image/png" href="{{static_url("favicon.ico")}}" />
   <link href="{{static_url("dist/styles.css")}}" rel="stylesheet"></link>
 
+  <style>
+    {{extra_css}};
+  </style>
   <script>
     window.pageConfig = {{page_config|tojson}};
   </script>
--- a/binderhub/app.py
+++ b/binderhub/app.py
@@ -188,6 +188,16 @@ class BinderHub(Application):
         config=True,
     )
 
+    extra_css = Unicode(
+        {},
+        help="""
+        Extra bits of CSS that should be loaded in each page.
+
+        Omit the <style> tag.
+        """,
+        config=True,
+    )
+
     base_url = Unicode("/", help="The base URL of the entire application", config=True)
 
     @validate("base_url")
@@ -959,6 +969,7 @@ class BinderHub(Application):
                 "about_message": self.about_message,
                 "banner_message": self.banner_message,
                 "extra_footer_scripts": self.extra_footer_scripts,
+                "extra_css": self.extra_css,
                 "jinja2_env": jinja_env,
                 "build_docker_config": self.build_docker_config,
                 "base_url": self.base_url,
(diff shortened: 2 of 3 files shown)


binderhub/templates/page.html
```diff
@@ -16,6 +16,9 @@
   <link id="favicon" rel="shortcut icon" type="image/png" href="{{static_url("favicon.ico")}}" />
   <link href="{{static_url("dist/styles.css")}}" rel="stylesheet"></link>
 
+  <style>
```
rgaiacs, line 19: Jinja has [control structures](https://jinja.palletsprojects.com/en/stable/templates/#list-of-control-structures). Can we add a [if-clause](https://jinja.palletsprojects.com/en/stable/templates/#if) here?

```html
{% if extra_css %}
<style>
  {{ extra_css }}
</style>
```
```

**Review by rgaiacs — CHANGES_REQUESTED** [review]
It looks good to me. Before merge, can we

- use if-clause in the HTML template
- document it for the Helm chart

File: binderhub/templates/page.html (hunk 1 above)
Comment (paololazzari) [reply binderhub/templates/page.html:20]: Good idea, I've fixed this

**Comment by paololazzari** [comment]
@rgaiacs Thanks! Where should I document this for the helm chart?

**Comment by rgaiacs** [comment]
> Where should I document this for the helm chart?

I believe that the traitlets are ["magic"](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1396) connected to the helm chart. But it is nice to add to https://github.com/jupyterhub/binderhub/blob/c07d252a98f7906ec36d3c152e60f7bbd2ae3319/helm-chart/binderhub/values.yaml#L42-L51 like

```yaml
config:
  # These c.BinderHub properties are referenced by the Helm chart
  BinderHub:
    ...
    # extra_css: |
    #  .p-3.bg-light.shadow-sm.text-center {
    #      color: red;
    #  }
```

Also mention it in [the documentation](https://github.com/jupyterhub/binderhub/blob/31e5d099132787ee2fd5915311317bc7c52c7fbd/docs/source/customizing.rst?plain=1#L45-L52)

I would like to merge https://github.com/jupyterhub/binderhub/pull/1923 before this one.

paololazzari:
@rgaiacs I did think about documenting `extra_css` in those two files, but I did not do it because I found that `extra_footer_scripts` was not documented there. I assumed whatever reasoning was behind not documenting `extra_footer_scripts` would hold true in this case too.

rgaiacs:
Hi @paololazzari, thanks again for the PR. I noticed that this PR will only work in some scenarios. To enable custom CSS as CSS was originally design for, we also need https://github.com/jupyterhub/binderhub/issues/1952.

The scenario that the solution that you proposed will not work is when you try to change some style from https://github.com/jupyterhub/binderhub/blob/main/binderhub/static/js/index.scss. For example, I was trying to change the font used in the page but using your solution resulted in

grafik

paololazzari:
My proposal was to add CSS, not replace the existing one

PR #1956, opened 2025-03-31
Draft: Move the frontend into its own package

rgaiacs:


yuvipanda:
I think I just published it from my personal npm token!

Thanks for taking this on, @rgaiacs! I'll try find some reviewers.

rgaiacs:
I was trying to build a example of how to use the package, https://github.com/rgaiacs/binderhub-frontend-example, but I end up with some JavaScript compilation challenges.

rgaiacs:
Close as replaced by https://github.com/jupyterhub/binderhub/pull/1978.

PR #1961, opened 2025-05-14
Add comment to values.yaml to show how extraPodSpec can be used

paololazzari:


rgaiacs:
I think that we already fixed the reason why the test was failing. @paololazzari could you rebase this pull request for us to verify the test?

consideRatio:
I think by close/re-open the PR, the tests will re-run against on the PR merged into the main branch I think - lets see if it resolves @rgaiacs!

rgaiacs (approved):
I'm not familiar with the `securityContext` but this pull request is about improve documentation. It looks good to me.

Issue #1962, opened 2025-05-14
Expose additional tolerations for image cleaner


paololazzari:
I have deployed binderhub in a tainted pool. I can't add a toleration to the image cleaner daemon set because tolerations are not exposed https://github.com/jupyterhub/binderhub/blob/main/helm-chart/binderhub/templates/image-cleaner.yaml

PR #1963, opened 2025-05-14
Add binder's tolerations to image-cleaner and container-builder

paololazzari:
Fixes https://github.com/jupyterhub/binderhub/issues/1962

### DIFF
--- a/helm-chart/binderhub/templates/image-cleaner.yaml
+++ b/helm-chart/binderhub/templates/image-cleaner.yaml
@@ -33,14 +33,8 @@ spec:
         key: hub.jupyter.org_dedicated
         operator: Equal
         value: user
-      {{- range .Values.tolerations }}
-      - effect: {{ .effect }}
-        key: {{ .key }}
-        operator: {{ .operator }}
-        value: {{ .value }}
-        {{- if .tolerationSeconds }}
-        tolerationSeconds: {{ .tolerationSeconds }}
-        {{- end }}
+      {{- with .Values.tolerations }}
+      {{- . | toYaml | nindent 6 }}
       {{- end }}
       nodeSelector: {{ .Values.config.BinderHub.build_node_selector | toJson }}
       {{- if .Values.rbac.enabled }}
--- a/helm-chart/binderhub/templates/image-cleaner.yaml
+++ b/helm-chart/binderhub/templates/image-cleaner.yaml
@@ -33,6 +33,13 @@ spec:
         key: hub.jupyter.org_dedicated
         operator: Equal
         value: user
+      {{- range .Values.tolerations }}
+      - effect: {{ .effect }}
+        key: {{ .key }}
+        operator: {{ .operator }}
+        value: {{ .value }}
+        {{- if .timeLimit }}timeLimit: {{ .timeLimit }}{{ end }}
+      {{- end }}
       nodeSelector: {{ .Values.config.BinderHub.build_node_selector | toJson }}
       {{- if .Values.rbac.enabled }}
       serviceAccountName: {{ .Release.Name }}-image-cleaner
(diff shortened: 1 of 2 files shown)


helm-chart/binderhub/templates/image-cleaner.yaml
```diff
@@ -33,6 +33,13 @@ spec:
         key: hub.jupyter.org_dedicated
         operator: Equal
         value: user
+      {{- range .Values.tolerations }}
+      - effect: {{ .effect }}
+        key: {{ .key }}
+        operator: {{ .operator }}
+        value: {{ .value }}
+        {{- if .timeLimit }}timeLimit: {{ .timeLimit }}{{ end }}
```
manics, line 41: Looks like a pasting error

helm-chart/binderhub/templates/image-cleaner.yaml
```diff
@@ -33,6 +33,13 @@ spec:
         key: hub.jupyter.org_dedicated
         operator: Equal
         value: user
+      {{- range .Values.tolerations }}
```
manics, line 36: Using `{{- with .Values.tolerations }} ... {{.}} ...` is a nicer way to do this instead of repeating all keys:
https://github.com/jupyterhub/binderhub/blob/ed61e2fb73142a0c3ccf01877c3ad1de83ce6d59/helm-chart/binderhub/templates/deployment.yaml#L44-L47

manics (commented):
It's also missing from the builder daemonset:
https://github.com/jupyterhub/binderhub/blob/ed61e2fb73142a0c3ccf01877c3ad1de83ce6d59/helm-chart/binderhub/templates/container-builder/daemonset.yaml#L29-L37

helm-chart/binderhub/templates/image-cleaner.yaml
paololazzari: That does not append to the existing tolerations already defined in image-cleaner.yaml

helm-chart/binderhub/templates/image-cleaner.yaml
paololazzari: There was a typo here as well, fixed.

helm-chart/binderhub/templates/image-cleaner.yaml
consideRatio: You can do it still but without wrapping the `tolerations:` part in the with clause in such situations.

I appreciate that this method decouples us from needing to know the k8s spec of tolerations. One could for example omit one of the fields i think, but they may as the template is become  now be required.

consideRatio: You want to conditionally render the tolerations as yaml if they are non-empty, so include a if or with clause for this as well.

The with makes you able to not repeat .Values.tolerations and instead write just 
 and is what we often use

paololazzari: Does that work for you? Because it doesn't for me.

With:

```
      tolerations:
      - effect: NoSchedule
        key: hub.jupyter.org/dedicated
        operator: Equal
        value: user
      - effect: NoSchedule
        key: hub.jupyter.org_dedicated
        operator: Equal
        value: user
      {{- with .Values.tolerations }}
      {{- . | toYaml | nindent 8 }}
      {{- end }}
```

```sh
$ helm template . -f values.yaml
Error: YAML parse error on binderhub/templates/image-cleaner.yaml: error converting YAML to JSON: yaml: line 28: did not find expected key
```

manics: You might need a different indentation level, it'll need to match the indentation of the existing tolerations. E.g. try `6`?

paololazzari: Spot on 👍🏻

consideRatio:
A single toleration for all may or may not make sense, I am not sure. Is this the right config to peovide, or is it more suitable to provide tolerations under imagecleaner specific config etc?

Writing from mobile with little time, so have not looked into it myself

manics:
I think it's fine, if necessary we can add `.Values.<component>.tolerations` in future and have them override `.Values.tolerations`

paololazzari:
Can we get this merged then?

manics (approved):

PR #1965, opened 2025-05-21
Fix f-string syntax in test_ui.py for python<3.12

douardda:
The `expected_url` f-string was incorrectly using strong quotes inside the brackets, while using strong quotes itself (which can only be done since python 3.12 IIRC).


minrk:
Thanks!

PR #1980, opened 2025-07-24
Changes introduced by pre-commit run --all-files

rgaiacs:
02768abc10722db3c16cda302273388ec9886f9b fails the tests.

image

One of the tests if fails is the pre-commit.

image


consideRatio:
Is prettier v4 involved anywhere? Downgrading to v3 is needed i think. Ghere is a team compass issue around it

Issue #1998, opened 2025-08-01
test_loading_page is failing


rgaiacs:
https://github.com/jupyterhub/binderhub/blob/5ce05acd5a7b229f8fd058b640fcaabd502ccfae/integration-tests/test_ui.py#L52-L113 is failing in a few pull requests with

```
=========================== short test summary info ============================
FAILED integration-tests/test_ui.py::test_loading_page[firefox-gh-binderhub-ci-repos/requirements-master---200] - AssertionError: 404 http://127.0.0.1:51433/services/binder/v2/gh/binderhub-ci-repos/requirements/master?
assert 404 == 200
 +  where 404 = <Response url='http://127.0.0.1:51433/services/binder/v2/gh/binderhub-ci-repos/requirements/master?' request=<Request url='http://127.0.0.1:51433/services/binder/v2/gh/binderhub-ci-repos/requirements/master?' method='GET'>>.status
======================== 1 failed, 11 passed in 40.34s =========================
```

The know pull requests to be effected are

- [ ] https://github.com/jupyterhub/binderhub/pull/1978
- [ ] https://github.com/jupyterhub/binderhub/pull/1993

rgaiacs:
It is interesting that the URL without slash (`/`) at the end

```
http://127.0.0.1:51433/services/binder/v2/gh/binderhub-ci-repos/requirements/master?
```

fails but

```
http://127.0.0.1:51433/services/binder/v2/gh/binderhub-ci-repos/requirements/master/?
```

passes.

I cannot replicate the error locally following https://github.com/jupyterhub/binderhub/blob/5ce05acd5a7b229f8fd058b640fcaabd502ccfae/integration-tests/test_ui.py.

```bash
py.test --cov=binderhub -s integration-tests/ -k 'test_loading_page[firefox-gh-binderhub-ci-repos/requirements-master---200]'
```

```
============================================================= test session starts =============================================================
platform linux -- Python 3.13.5, pytest-8.4.1, pluggy-1.6.0 -- /home/raniere/github.com/jupyterhub/binderhub/.pixi/envs/default/bin/python3
cachedir: .pytest_cache
rootdir: /home/raniere/github.com/jupyterhub/binderhub
configfile: pyproject.toml
plugins: base-url-2.1.0, anyio-4.9.0, playwright-0.7.0, cov-6.2.1, asyncio-0.25.0, timeout-2.4.0
[...]

PR #2003, opened 2025-08-08
Fix race condition in test_loading_page

rgaiacs:
Closes https://github.com/jupyterhub/binderhub/issues/1998

This is based on https://github.com/rgaiacs/binderhub/actions/runs/16830429222/job/47676539438.

The call https://github.com/jupyterhub/binderhub/blob/457ce23a0b6c7f7da56bd15810cc12032e565b32/.github/workflows/playwright.yaml#L88 starts the execution of https://github.com/jupyterhub/binderhub/blob/457ce23a0b6c7f7da56bd15810cc12032e565b32/integration-tests/test_ui.py#L92 that has as dependency the fixture https://github.com/jupyterhub/binderhub/blob/457ce23a0b6c7f7da56bd15810cc12032e565b32/integration-tests/test_ui.py#L19 responsible to start the JupyterHub and BinderHub server as a subprocess https://github.com/jupyterhub/binderhub/blob/457ce23a0b6c7f7da56bd15810cc12032e565b32/integration-tests/test_ui.py#L26 A for-loop is used to force the test to wait for the JupyterHub and BinderHub to be running https://github.com/jupyterhub/binderhub/blob/457ce23a0b6c7f7da56bd15810cc12032e565b32/integration-tests/test_ui.py#L37-L45 The for-loop assumes that the JupyterHub and BinderHub will be running in less than 10 seconds. This assumption is valid as usually takes less than 5 seconds for JupyterHub and BinderHub to be running.

```
2025-08-08T12:30:43.9693623Z [I 2025-08-08 12:30:43.968 JupyterHub app:3354] Running JupyterHub version 5.3.0
2025-08-08T12:30:44.7843121Z [I 2025-08-08 12:30:44.784 JupyterHub service:423] Starting service 'binder': ['python', '-mbinderhub', '--config=/home/runner/work/binderhub/binderhub/testing/local-binder-local-hub/binderhub_config.py', '--port=50087']
2025-08-08T12:30:47.6670578Z [I 250808 12:30:47 app:1120] BinderHub starting on port 50087
2025-08-08T12:30:47.8608506Z [W 2025-08-08 12:30:47.860 JupyterHub proxy:445] Adding missing route for binder (Server(url=http://localhost:50087/services/binder/, bind_url=http://localhost:50087/services/binder/))
2025-08-08T12:30:47.8674210Z [I 2025-08-08 12:30:47.867 JupyterHub app:3778] JupyterHub is now running at http://:55483/
2025-08-08T12:30:47.8694761Z [D 2025-08-08 12:30:47.868 JupyterHub app:3347] It took 3.945 seconds for the Hub to start
```

 Unfortunately, the tests to check if JupyterHub and BinderHub are running happens 10 seconds before the server is ready.

```
2025-08-08T12:30:36.1494173Z [35mDEBUG   [0m test_ui:test_ui.py:43 Debugging 1998. Making request #0 to http://127.0.0.1:55483/services/binder/
2025-08-08T12:30:37.1557788Z [35mDEBUG   [0m test_ui:test_ui.py:43 Debugging 1998. Making request #1 to http://127.0.0.1:55483/services/binder/
2025-08-08T12:30:38.1586443Z [35mDEBUG   [0m test_ui:test_ui.py:43 Debugging 1998. Making request #2 to http://127.0.0.1:55483/services/binder/
2025-08-08T12:30:39.1624992Z [35mDEBUG   [0m test_ui:test_ui.py:43 Debugging 1998. Making request #3 to http://127.0.0.1:55483/services/binder/
2025-08-08T12:30:40.1651928Z [35mDEBUG   [0m test_ui:test_ui.py:43 Debugging 1998. Making request #4 to http://127.0.0.1:55483/services/binder/
2025-08-08T12:30:41.1686785Z [35mDEBUG   [0m test_ui:test_ui.py:43 Debugging 1998. Making request #5 to http://127.0.0.1:55483/services/binder/
2025-08-08T12:30:42.1713366Z [35mDEBUG   [0m test_ui:test_ui.py:43 Debugging 1998. Making request #6 to http://127.0.0.1:55483/services/binder/
2025-08-08T12:30:43.1739505Z [35mDEBUG   [0m test_ui:test_ui.py:43 Debugging 1998. Making request #7 to http://127.0.0.1:55483/services/binder/
```

This pull request changes the `time.sleep()` strategy to a increased time instead of a linear one. This will make the tests to take longer to run but will reduce the changes of having problems with race condition.

A `print()` statement was included to aid similar problems in the future.

### DIFF
--- a/integration-tests/test_ui.py
+++ b/integration-tests/test_ui.py
@@ -42,7 +42,12 @@ async def local_hub_local_binder(request):
                 break
         except requests.exceptions.ConnectionError:
             pass
-        time.sleep(1)
+        # We use a increased linear time between requests
+        # because BinderHub server might require more than 10s to start.
+        time.sleep(i)
+    else:
+        print("BinderHub server is down! Tests will probably fail.")
+    # PyTest requires the yield instruction here!
     yield url
 
     proc.terminate()
--- a/integration-tests/test_ui.py
+++ b/integration-tests/test_ui.py
@@ -2,9 +2,9 @@
 Integration tests using playwright
 """
 
-import asyncio
 import subprocess
 import sys
+import time
 
 import pytest
 import requests
@@ -42,7 +42,7 @@ async def local_hub_local_binder(request):
                 break
         except requests.exceptions.ConnectionError:
             pass
-        await asyncio.sleep(1)
+        time.sleep(1)
     yield url
 
     proc.terminate()


integration-tests/test_ui.py
```diff
@@ -42,7 +42,12 @@ async def local_hub_local_binder(request):
                 break
         except requests.exceptions.ConnectionError:
             pass
-        await asyncio.sleep(1)
+        # We use a increased linear time between requests
```
manics, line 45: Rather than using a linear increase we can just do `for i in range(60)`, (summing the linear increases means this takes up to 45s, so we can round it off to 1 minute). I'd keep `await asyncio.sleep(1)` since although it's not the cause of the bug this is an async function, and using async.sleep will avoid future problems.

integration-tests/test_ui.py
```diff
@@ -42,7 +42,12 @@ async def local_hub_local_binder(request):
...
+        # because BinderHub server might require more than 10s to start.
+        time.sleep(i)
```
manics, line 47: We should move the `sleep` call to the start of the loop before the `try`, otherwise we end up with a delay at the end without a check.

integration-tests/test_ui.py
```diff
@@ -42,7 +42,12 @@ async def local_hub_local_binder(request):
...
+    else:
+        print("BinderHub server is down! Tests will probably fail.")
```
manics, line 49: If we move the `sleep` to the start then we're certain the server is returning 404, so we can fail here for clarity.

manics (approved):
Looks great!

PR #2008, opened 2025-08-25
Update the commands of the helm post install notes to map helm chart fields correctly:

EmmanuelKasper:
When binder is installed with the helm chart, and is only exposed via a ClusterIP Kubernetes Service, the `kubectl` commands of the post install notes are erroring in subtle ways:

~~~
kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "fullname" . }}" -o jsonpath="{.items[0].metadata.name}"
 ~~~

can fail because `{{ template "fullname" . }}` does not always match the hardcoded labels we set in https://github.com/jupyterhub/binderhub/blob/33bcc37c2fd18573d19a13c5a41b1d8d563c3663/helm-chart/binderhub/templates/deployment.yaml#L20

and

~~~
kubectl port-forward $POD_NAME 8080:{{ .Values.service.externalPort }} 
~~~

always fails because we don't expose set or expose an  externalPort field in https://github.com/jupyterhub/binderhub/blob/33bcc37c2fd18573d19a13c5a41b1d8d563c3663/helm-chart/binderhub/values.yaml#L34

This commit fixes the two errors by using immutables, non-overridables values from the helm templates.

### DIFF
--- a/helm-chart/binderhub/templates/NOTES.txt
+++ b/helm-chart/binderhub/templates/NOTES.txt
@@ -159,9 +159,10 @@ config:
   export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ template "fullname" . }} -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
   echo http://$SERVICE_IP:{{ .Values.service.externalPort }}
 {{- else if contains "ClusterIP"  .Values.service.type }}
-  export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "fullname" . }}" -o jsonpath="{.items[0].metadata.name}")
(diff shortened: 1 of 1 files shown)


EmmanuelKasper:
To reproduce the issue, and verify the fix, it should be enough to change in 
 testing/k8s-binder-k8s-hub/binderhub-chart-config.yaml the nodePort service to a ClusterIP service in this way:

~~~
 service:
  type: ClusterIP
~~~

and install the helm chart following https://binderhub.readthedocs.io/en/latest/contribute.html#develop-helm-chart

rgaiacs (approved):

rgaiacs:
> is there anything I need to do to get this merged ?

I would like a review of another contributor as I'm a beginner regarding Kubernetes.

minrk:
Thanks!

EmmanuelKasper:
Thanks for the review !

Issue #2019, opened 2025-09-27
launch-url field is not reset when when repo field is cleared


manics:
By default the binder launch-url field says `Fill in the fields to see a URL for sharing your Binder.` If you start typing a repo name it dynamically updates to display a URL, for example entering `x` in the repo field results in a displayed launch url of `https://mybinder.org/v2/gh/x/HEAD`

However if you then delete the text in the repo field the launch URL gets stuck at `https://mybinder.org/v2/gh/x/HEAD` instead of returning to the default text `Fill in the fields to see a URL for sharing your Binder.`:

<img width="1860" height="468" alt="Image" src="https://github.com/user-attachments/assets/5589bf0d-e367-46bb-91a5-d2244656626c" />

I can't reproduce this locally in the [BinderHub main branch](https://github.com/jupyterhub/binderhub), I only see this on mybinder.org.

PR #2020, opened 2025-09-27
LinkGenerator: set repo to empty if invalid

manics:
https://github.com/jupyterhub/binderhub/commit/3d29702496b8d879a5a541c06a3dc4f4914bd16c fixes https://github.com/jupyterhub/binderhub/issues/2019

The other commits are for CI testing

### DIFF
--- a/js/packages/binderhub-react-components/babel.config.js
+++ b/js/packages/binderhub-react-components/babel.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  presets: [
+    ["@babel/preset-env", { targets: { node: "current" } }],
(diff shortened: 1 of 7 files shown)


js/packages/binderhub-react-components/babel.config.js
rgaiacs, line 1: Sorry my ignorance here. Is `preset` required only for the test? If it is required for the test, I think it should be included in `js/packages/binderhub-react-components/package.json`?

manics: You're right, I've added it

consideRatio:
@manics, should we go for a merge here, or do you want additional review?

PR #2044, opened 2025-12-16
feat: support Node.js for `@jupyterhub/binderhub-client`

### LINKED ISSUE #2111: Allow `binderhub-client` to run in Node.js
The `@jupyterhub/binderhub-client` package can only run in a browser, not in `node`.
This makes it hard to drive a BinderHub from outside the browser, for example from a command-line tool or a CI job.

The problem is [`@microsoft/fetch-event-source`](https://www.npmjs.com/package/@microsoft/fetch-event-source), which is browser-only (and doesn't seem to be maintained anymore, the last release was like 5 years ago).
The client uses it to stream build logs from the BinderHub `/build` endpoint, so the package fails when imported in a `node` runtime.

### Why this would be useful

We're working on a command-line interface to manage BinderHub sessions over here: https://github.com/2i2c-org/clinder in collaboration with [Project Pythia](https://2i2c.org/collaborators/pythia). We'd like to request Binder sessions for notebook execution as part of a CI/CD pipeline, and this improvement would let us do this!

### Implementation

We already have a vendored+modified version of binderhub with this fix working over here:

- https://github.com/agoose77/binderhub/tree/ef429843ed048d96ffae1c816aa53559f346a19a/js/packages/binderhub-client
- Which is what's behind the PR here: https://github.com/jupyterhub/binderhub/pull/2044

agoose77:
This PR switches out [`@microsoft/fetch-event-source`](https://www.npmjs.com/package/@microsoft/fetch-event-source) in favour of [`eventsource`](https://www.npmjs.com/package/eventsource). This package is both newer (and maintained) and explicitly supports the Node.js runtime.

I'm consuming a vendored version of these changes in https://github.com/2i2c-org/clinder/tree/main/packages/binderhub-client-next, which is Node.js only.

---

- closes https://github.com/jupyterhub/binderhub/issues/2111

### DIFF
--- a/js/packages/binderhub-client/lib/client.js
+++ b/js/packages/binderhub-client/lib/client.js
@@ -1,4 +1,4 @@
-import { fetchEventSource } from "@microsoft/fetch-event-source";
+import { EventSource } from "eventsource";
 import { EventIterator } from "event-iterator";
 
 function _getXSRFToken() {
@@ -21,8 +21,6 @@ function _getXSRFToken() {
   return null;
 }
 
-/* throw this to close the event stream */
-class EventStreamClose extends Error {}
 /* throw this to close the event stream */
 class EventStreamRetry extends Error {}
 
@@ -68,8 +66,7 @@ export class BinderRepository {
     }
     this.apiToken = apiToken;
 
-    this.eventIteratorQueue = null;
-    this.abortSignal = null;
+    this.stopQueue = null;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


js/packages/binderhub-client/lib/client.js
```diff
@@ -106,106 +102,73 @@ export class BinderRepository {
...
       }
     }
-    // setTimeout(() => this.close(), 1000);
+
+    const es = new EventSource(this.buildUrl, {
+      fetch: async (input, init) => {
+        const response = await fetch(input, {
+          ...init,
+          headers: { ...init.headers, ...headers },
+        });
```
agoose77: Ensure we pass through the headers

choldgraf:
I've added a bit more explanation for why this feature is useful and what the problem is in this tracking issue, and linked this PR to it: https://github.com/jupyterhub/binderhub/issues/2111

agoose77:
@rgaiacs what is the best pathway to try and get this PR merged, in your view?

PR #2092, opened 2026-04-10
fix figshare article version resolution, update http-record caches for tests

minrk:
created by running `pytest binderhub/tests` locally and updating http-record files

my guess is CI is getting blocked from one or more dataverse requests

the same tests pass fine locally, which generated these caches


consideRatio (approved):

minrk:
I realized the figshare test failing was rightly a bug. Figshare appeared to change their URL scheme recently.

Rather than parsing figshare URLs, which are subject to change, I updated to parse the figshare _doi_, which has a clearer (hopefully more stable) scheme.

Also a fix: when a doi lacks a version, get the _latest_ version from the figshare API, which is how figshare resolves unversioned dois, not assume v1, which isn't correct unless there is only one version.

